PR #26, opened 2016-01-16
Adding full chain to LE

lucasRolff:
- Added functionality to create a fullchain cert file, I didn't add it as a 'RealPath' option, because it would either have to be in the last argument, or we would shift other arguments, which would not be backward compatible.

@Neilpang I don't know if you require more places for this to be changed, I think I covered the most.

Code is tested, and it actually creates the fullchain.pem file as well!

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -279,6 +279,9 @@ _initpath() {
   if [ -z "$CA_CERT_PATH" ] ; then
     CA_CERT_PATH="$WORKING_DIR/$domain/ca.cer"
   fi
+  if [ -z "$CERT_CHAIN_PATH" ] ; then
+    CERT_CHAIN_PATH="$WORKING_DIR/$domain/fullchain.pem"
+  fi
 
   
 }
@@ -619,6 +622,12 @@ issue() {
     echo -----END CERTIFICATE-----  >> "$CA_CERT_PATH"
     _info "The intermediate CA cert is in $CA_CERT_PATH"
   fi
+
+  if [[ -f $CERT_PATH && -f $CA_CERT_PATH ]] ; then
+    cat $CERT_PATH >> "$CERT_CHAIN_PATH"
+    cat $CA_CERT_PATH >> "$CERT_CHAIN_PATH"
+    _info "The full certificate chain is in $CERT_CHAIN_PATH"
+  fi
   
   Le_CertCreateTime=$(date -u "+%s")
   _setopt "$DOMAIN_CONF"  "Le_CertCreateTime"     "="  "$Le_CertCreateTime"
@@ -713,6 +722,7 @@ renewAll() {
     CERT_KEY_PATH=""
     CERT_PATH=""
     CA_CERT_PATH=""
+    CERT_CHAIN_PATH=""
     ACCOUNT_KEY_PATH=""
     
     wellknown_path=""


samueltardieu:
You can squash your commits together with `git rebase -i` and force push them to avoid getting the intermediate buggy version in the tree.

samueltardieu:
Another alternative would have been to:
- remove (or move to `.bak`) the previous certificate and CA files;
- write the certificate to its new location (`>`);
- append the CA to its new location (`>>`).

This way, you can get the full chain by specifying the same file for the certificate and the CA, and you get one file instead of two containing the full chain.

le.sh
```diff
@@ -279,6 +279,9 @@ _initpath() {
   if [ -z "$CA_CERT_PATH" ] ; then
     CA_CERT_PATH="$WORKING_DIR/$domain/ca.cer"
   fi
+  if [ -z "$CERT_CHAIN_PATH" ] ; then
```
samueltardieu: What is the point of defining this variable as empty and then testing if it is still empty ?I don't like the way it's not configurable, hence the proposed alternative.

lucasRolff: You're right on this one.

Also you can specify the path twice, or can be changed to do so. With that said, I somehow like that both files are separate, but you also have a full chain of it (Let's say you automate this for multiple systems), some servers needs a chain (haproxy e.g) - where others you specify the certificate and the CA separately.

samueltardieu: Well, going this way, why not just concatenate the files externally to this script? For example, my reload command is a script which concatenates both files to make a full cert chain then restarts my servers. I'm not sure this is worth the added complexity when this can be automated already using the reload command argument.

lucasRolff: Sure - I mainly this this pull request because of https://github.com/Neilpang/le/issues/24#issuecomment-172228674

Neilpang:
Yes, I agree with @samueltardieu .   If the user wants a fullchain.pem, he should concatenate the files in the "reload command" script.

The most I concern here is that: in some circumstances the "fullchain.pem" may be different.  As you know, the cert is not only used by Webserver, but also mail server, or even VPN server.  Some case may require single separated CA.cer,  some other cases may require  Ca.cer + domain.cer. While in some rare cases, it may requires Root.cer + ca.cer + domain cer.

We can not handle all those cases automatically, so, we just make it simplest.  I think the administrator must know how to deal with his server.

markogresak:
What is the problem with creating all 3 files, so just adding another line in the script which will do something like `cat [domain].cer ca.cer > fullchain.pem`? You avoid the problem where one would need only one certificate and avoid having to add concatenation to each domain we issue.

samueltardieu:
@markogresak The problem has been explained by @Neilpang in https://github.com/Neilpang/le/pull/26#issuecomment-172298782. Moreover, you would need to add yet another argument for this new file path.

markogresak:
@samueltardieu I've read that, but I still can't understand what's the big problem and why would it have to be an argument? In cases where ca.cer should be separate, there is no problem at all. Cases where one would need domain.cer + ca.cer would be fulfilled by fullchain.pem containing what I've proposed in my previous comment. Lastly, the rare cases requiring domain.cer + ca.cer + root.cer are probably done by more experienced users who know what they are doing and putting a note what fullchain.pem contains would be enough to let them know what they have to do. Moreover, there is no root.cer created by le.sh at the moment, so one would have to take care of this first anyway. I believe that some opinionated design that fulfills most users' requirements would be better than making everyone figure it out for themselves and put in additional commands for each certificate they issue.

Neilpang:
I just fixed it:  If the real cert path equals to the ca cert path, we will append the ca cert to the domain cert, which makes it a fullchain.

agarzon:
@Neilpang 

> I just fixed it: If the real cert path equals to the ca cert path, we will append the ca cert to the domain cert, which makes it a fullchain.

Can you please explain this?

I was having the same issue (version 1.1.8) and nginx. The solution was creating manually the chained certificate dom.cer + ca.cer as proposed by @markogresak 

Shouldn't be the chained version created automatically?

Neilpang:
example:

```
le  installcert   aa.com     /etc/nginx/ssl/aa.com/aa.cer   /etc/nginx/ssl/aa.com/aa.key   /etc/nginx/ssl/aa.com/aa.cer
```

Take a look at the arguments above.

PR #28, opened 2016-01-17
Chown directories to parent

lucasRolff:
- You might configure web-servers to not allow reading files owned by root (or user you execute as), modified script to try chowning the .well-known recursively
- If you do not have chown rights it will work anyway

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -538,9 +538,9 @@ issue() {
       mkdir -p "$wellknown_path"
       echo -n "$keyauthorization" > "$wellknown_path/$token"
 
-      webroot_owner=$(stat -c '%U' $Le_Webroot)
-      _debug "Changing owner of .well-known to $webroot_owner"
-      chown -R $webroot_owner. "$Le_Webroot/.well-known"
+      webroot_owner=$(stat -c '%U:%G' $Le_Webroot)
+      _debug "Changing owner/group of .well-known to $webroot_owner"
+      chown -R $webroot_owner "$Le_Webroot/.well-known"
 
     fi
     wellknown_url="http://$d/.well-known/acme-challenge/$token"
--- a/le.sh
+++ b/le.sh
@@ -537,6 +537,11 @@ issue() {
       
       mkdir -p "$wellknown_path"
       echo -n "$keyauthorization" > "$wellknown_path/$token"
+
+      webroot_owner=$(stat -c '%U' $Le_Webroot)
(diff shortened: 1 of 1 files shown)


lucasRolff:
I don't know if it's a feature that should actually be included, but I use a modified version of the script to support web-servers that prevents reading files from other users but itself.

If you don't want it, feel free to decline

Neilpang:
I have an idea:   we should add "read" permission  instead of "chown".
Something like:
```
chmod -R  +r   "$Le_Webroot/.well-known"
``` 

Which may requires less permissions, while not dependent on "stat".

It's better to keep simple, and less dependencies.

And we also should delete the  ".well-known" folder after verification.

What do you think?

lucasRolff:
Giving read recursively doesn't always fix the problem, e.g. directories with 755 permissions and files with 644 permissions - do have `read` permission, but a web-server that is locked down (only allowing the files to be owned by the user the web-server is running as), or even a server running with SELinux, will still not be able to actually read the file.

Sure if you do not run SELinux (People should IMO), or if you do not lock down your web-server it wouldn't be a problem just giving read permissions on a file.

But in the case that you actually do either lock down your web-server or run SELinux we need to actually change the ownership of the file.

Now I used stat because it's simple, I could also use `ls` with some `grep` to solve it - either way we either add a dependency of `ls`, `chown`, `chmod` or `stat`.

I know it needs to be simple, I fully agree, but we also have to keep in mind that Let's Encrypt's goal is to be able to give SSL to everyone, to make it easy, and to automate it across single and multi tenant systems - if you're working with multi tenant systems you do often have security measures in place that requires changing permissions of files. So sure either people can do it by building their own let's encrypt 'client' or they can e.g. use a script like this which would already try to change the permissions. Either way, if there's not a chown functionality in the script people would have to do it in a different way. (Which was what I personally ended up doing with the official letsencrypt client before I found yours) - I basically had to start a sub-process to chown a directory for the validation to actually work because security measures that prevented viewing files owned by in this case `root`

I do think that the .well-known directory should get removed mainly for security purposes as well.

Neilpang:
Ok,
I agree, but why not change the group at the same time.
Some thing looks like:

```
stat -c  '%U:%G'
```

Is there any other concern ?

lucasRolff:
@Neilpang - can do that indeed! - wait :dancer:

Neilpang:
Thanks

PR #29, opened 2016-01-20
Update le.sh

tombii:
If we regenerate the key by using FORCE=1 then we also need to regenerate the CSR, otherwise the key will not match the certificate.


Neilpang:
Thanks.

PR #32, opened 2016-01-23
Script should create world-readable verification files

samueltardieu:
The script should use a permissive umask (022) in order to create world-readable verification files. 

Some would find this debatable, as the key file might be made readable for everyone as well. In this case, a better solution could be to change the mode of `.well-known` and `.well-known/acme-challenge` (0755 when they are created by the script), and on the verification file in any case (mode 0644).

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -384,29 +384,44 @@ _clearup () {
   _restoreApache
 }
 
-# webroot  removelevel tokenfile
+# webroot removelevel tokenfile
+_tokenlevel() {
+  __path="$1/.well-known"
+  if [ "$2" != '1' ]; then
+    __path="$__path/acme-challenge"
+    if [ "$2" == '3' ]; then
+      __path="$__path/$3"
+    elif [ "$2" != '2' ]; then
+      _err "removelevel invalid: $2"
+      return 1
+    fi
+  fi
+  echo "$__path"
+}
+
+# webroot removelevel tokenfile
+_fixtokenperms() {
+  __path=$(_tokenlevel "$@")
+  _debug "Setting world-readable permissions on $__path"
+  chmod -R og=u-w "$__path"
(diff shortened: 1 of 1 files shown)


samueltardieu:
(don't ask me how I could convert an issue into a pull request, I don't know it myself, this is a side effect of using the hub command-line script to create the pull-request, and this option looks obsolete anyway) :-)

kevinoid:
I just ran into this issue for the reason @samueltardieu stated (I run `acme.sh` as a non-root user different from the owner of the webroot directory, which results in `chown: Operation not permitted` as described in #163).

I agree that writing files to `.well-known` with umask 022 should be sufficient.  If world-readability is an issue, using umask 027 (and chown/chgrp to webroot owner if running as root) should also work.

I'm a fan of the approach in this patch, although I'd call `umask` before creating the files/dirs rather than `chmod` after.  Obviously the patch is now a bit out of date.  Is it worth updating?  Is there disagreement about this approach or its implementation?  If so, what's the alternative?

kevinoid:
@Neilpang Does this mean running as a user other than the webroot owner (and root) is an unsupported use case?

Neilpang:
@kevinoid  
Please create a new issue `Script should create world-readable verification files`,  I will fix it soon.

kevinoid:
@Neilpang I'm sorry, but I don't understand.  You want me to create a new issue with the same name as this issue?  If so, why not reopen this one until it is fixed?

Alternatively, I could send a PR of [the fix that I am using](https://github.com/kevinoid/acme.sh/commit/6cebfe5c0940076ccaab11d148ebc38a39f273a7) if you prefer.

Neilpang:
@kevinoid 
This is a PR, not an issue.  And the PR is outdated,  so I want a new Issue.

Yes, if you have PR, please send me.

PR #37, opened 2016-01-24
Skip verification if the account is already verified

NITEMAN:
There is no need to make a new-authz call if the account is already verified for the domain.

Skipping it may be quite helpful when using dns.


Neilpang:
The second one looks better,  I will review/test it soon.  Thanks.

NITEMAN:
As far as I understand both commits are needed to achieve the goal.

It might be easier if verification were performed in it's own functions

Shuro:
So this would mean that already verified domains wouldn't be checked on renew? This would be good for some projects.

PR #73, opened 2016-02-17
change "openssl sha" to "openssl dgst"

travislee8964:
LibreSSL doesn't support "openssl sha" command, use "openssl dgst" instead of "openssl sha".


Neilpang:
Thanks.

PR #81, opened 2016-03-02
simplify call by using $@

phanimahesh:
$@ refers to all arguments, which is a nice way of saying $1 $2 ..., plus it doesn't need updating if we need more arguments.


BlueAnanas:
Ah, now I see which commit introduced the problem. Don't worry I have made the #82 for that.

phanimahesh:
Oops, I should have checked that. Thanks for finding and fixing this. :)

PR #82, opened 2016-03-02
Replace $@ by quoted "$@" in function selector

BlueAnanas:
An $@ without quotes is not putting quotes around expanded parameters its basically expanding to $1 $2 ... while "$@" is expanding to "$1" "$2" . The quotes around the server reload command get lost in the process and any parameters of the command are not transferred to installcert


Neilpang:
Thanks.

PR #84, opened 2016-03-03
Removed redundant backup in installcert 

BlueAnanas:
when RealCACertPath == RealCertPath and certs are concatenated. In this case the Cert file is copied twice to backup resulting in loss of the original file being there before installcert

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -1020,13 +1020,13 @@ installcert() {
   fi
   
   if [ "$Le_RealCACertPath" ] ; then
-    if [ -f "$Le_RealCACertPath" ] ; then
-      cp -p "$Le_RealCACertPath" "$Le_RealCACertPath".bak
-    fi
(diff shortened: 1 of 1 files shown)


Neilpang:
I don't think this is necessary and is worth to make the change.    I'd like to keep it simple, we only do the critical things.

Less is more.

BlueAnanas:
I think I didn't express myself well: This PR is not an optimization but **fixing a real bug**: If cert concatenation is used, install will trash the previous cert file. **The .bak file** which should hold the saved previous cert concatenation **will hold the new nonconcatenated cert**, the same one which is in the newly written cert file.
A simple rollback is not possible anymore as the old cert file is lost - the .bak files are useless.
Complexity isn't increased - the tree code lines are simply moved into the right conditional block

Maybe you could reconsider the PR rejection in light of these facts.

PR #87, opened 2016-03-07
Add post-renew hook

theit8514:
Allows for running commands when the certificate is renewed. The command is only run if a) the file exists and is executable by the current user and b) the certificate renewal was a success.

I needed a way to send the key to ESXi when the cert was renewed. The following is an example of a post-renew.sh file:

    #!/bin/bash

    scp esx.example.com.cer root@esx:/etc/vmware/ssl/rui.crt
    scp esx.example.com.key root@esx:/etc/vmware/ssl/rui.key
    ssh root@esx 'services.sh restart' > restart.log

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -412,9 +412,6 @@ _initpath() {
   if [ -z "$CA_CERT_PATH" ] ; then
     CA_CERT_PATH="$domainhome/ca.cer"
   fi
-  if [ -z "$POST_RENEW_PATH" ] ; then
-    POST_RENEW_PATH="$domainhome/post-renew.sh"
-  fi
 
 }
 
@@ -962,9 +959,6 @@ renew() {
   local res=$?
   IS_RENEW=""
 
-  if [ -x "$POST_RENEW_PATH" -a "$res" -eq "0" ] ; then
-    (cd $DOMAIN_PATH && exec $POST_RENEW_PATH)
-  fi
   return $res
 }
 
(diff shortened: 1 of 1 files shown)


Neilpang:
Thanks for the code, but we have a "Le_ReloadCmd" config there.  You can write any script there, which will be executed after the cert is renewed.

Open the "~/.le/domain.com/domain.com.conf"
It make look like:
```
Le_ReloadCmd="chmod +x post-renew.sh && post-renew.sh"
```


Let me know whether it fits for you.


Thanks.

theit8514:
It does look like Le_ReloadCmd would suffice except for the fact that it runs from the current directory of the shell, and not from the LE root directory or the directory of the certificate.

> /home/jack/.le/test.example.com/test.example.com.conf:8:Le_ReloadCmd="chmod +x post-renew.sh && ./post-renew.sh"
> Run Le_ReloadCmd: chmod +x post-renew.sh && ./post-renew.sh
> chmod: cannot access 'post-renew.sh': No such file or directory

If I try to use a variable in the Le_ReloadCmd (i.e. $domainhome from _initpath), the command parses the variable correctly, but the command is rewritten back with the variable expanded.

Would you be opposed to using the `(cd $DOMAIN_PATH && eval $Le_ReloadCmd)` to change the directory before evaluating any commands? With this change, ./post-renew.sh is executed in the correct directory.

Neilpang:
Thanks.

PR #90, opened 2016-03-09
Update initconf

Fabian-Gruenbichler:
Making it obvious that ACCOUNT_KEY_PATH can be set in the account config file seems like a good idea to me - the cron/renewAll function resets the ACCOUNT_KEY_PATH set via the environment, so the config file is the only way to set this for cron/renewAll.

This behaviour is needed for setups where the account config and key files are shared between hosts, but the LE_WORKING_DIR is not.

The second commit just fixes non escaped quotes in an echo statement.

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -1150,22 +1150,22 @@ _initconf() {
 #######################
 #Cloudflare:
 #api key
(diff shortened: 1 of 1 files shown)


Neilpang:
thanks.

PR #97, opened 2016-03-11
Fix dns error message

kaklakariada:
Fixed error message when dns api function is not found in dos api file.


kaklakariada:
Sorry, the history of my branch is not clean, i will send another pull request with only one commit.

PR #98, opened 2016-03-11
Fix missing dns function message

kaklakariada:
Fix error message for missing dns api function


kaklakariada:
Good point, I did not see this.

Neilpang:
Thanks, 
BTW, I suggest you rebase the code.  it seems too complicated.

PR #99, opened 2016-03-12
concatenate cert parts during issue

avonderluft:
accomodates needs of various web servers


Neilpang:
Thanks.   I just added fullchain support.  please try again.

PR #120, opened 2016-04-08
Save ACCOUNT_CONF_PATH in crontab

Fabian-Gruenbichler:
Setting ACCOUNT_CONF_PATH to a non-default location when
calling le.sh install or le.sh installcronjob should have an
effect on the generated crontab entry, otherwise the cron
job runs with unexpected (default) settings.

For example, installing with ACCOUNT_CONF_PATH=/some/path/account.conf , then setting STAGE=1 in that config file and getting some test certificates could lead to a cron job that would run with STAGE=0 in 2.5 months and issue non-stage certificates! 

A similar issue applies to the DNS API related settings, but those probably just mean that the cron job will fail? Haven't tested those as I don't use any of the configured DNS providers.

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -1449,12 +1449,12 @@ installcronjob() {
       _err "Can not install cronjob, le.sh not found."
       return 1
     fi
-    crontab -l | { cat; echo "0 0 * * * LE_WORKING_DIR=\"$LE_WORKING_DIR\" $lesh cron > /dev/null"; } | crontab -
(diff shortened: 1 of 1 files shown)


Neilpang:
fixed.  d53289d707d72ceec0282c265f345228ef333f90

If you set " ACCOUNT_CONF_PATH=/some/path/account.conf " when install:

```
ACCOUNT_CONF_PATH=/some/path/account.conf   ./le.sh  --install
``` 

This value will be saved to the default `~/.le/accoung.conf` file.   The default file only contains one line:
```
 ACCOUNT_CONF_PATH=/some/path/account.conf 
```

Which points to the real config file.

Any further configs will be saved in the real config file.

The cron job is not changed.

In this fix,  the customized account conf can also be used in interactive command line.

Please try again,   you'd better remove your existing `~/.le/accoung.conf` file first, and try rre-install again.

Fabian-Gruenbichler:
Yes, this solution works as well!

PR #121, opened 2016-04-08
Set ACCOUNT_KEY_PATH and ACCOUNT_EMAIL in _initconf()

Fabian-Gruenbichler:
this enables the following one liner for initial configuration at install time, which would mean that most users would never need to touch their account.conf manually:

    ACCOUNT_CONF_PATH=/path/to/account.conf ACCOUNT_KEY_PATH=/path/to/account.key ACCOUNT_EMAIL=some@valid.email ./le.sh install

### DIFF
--- a/le.sh
+++ b/le.sh
@@ -1609,6 +1609,13 @@ USER_AGENT=\"le.sh client: $PROJECT\"
 
     " > $ACCOUNT_CONF_PATH
   fi
+  if [ -n "$ACCOUNT_KEY_PATH" ]; then
+    _sed_i "s/^#ACCOUNT_KEY_PATH=.*$/ACCOUNT_KEY_PATH=\"${ACCOUNT_KEY_PATH//\//\\\/}\"/g" "$ACCOUNT_CONF_PATH"
+  fi
+
+  if [ -n "$ACCOUNT_EMAIL" ]; then
+    _sed_i "s/^#ACCOUNT_EMAIL=.*$/ACCOUNT_EMAIL=\"${ACCOUNT_EMAIL//\//\\\/}\"/g" "$ACCOUNT_CONF_PATH"
+  fi
 }
 
 _precheck() {


Neilpang:
It's good idea to save the conf path and  email.

Please use `_saveaccountconf()`  function to save conf values,  do not use `sed`.

please fix and re-pull

Neilpang:
It's not necessary to define it in the command line.

The user can save it in the `account.conf` file,  yes,  the  saved value will override the default one.

See the `_initpath()`  function:
```
_initpath() {

  if [ -z "$LE_WORKING_DIR" ]; then
    LE_WORKING_DIR=$HOME/.le
  fi
  
  if [ -z "$ACCOUNT_CONF_PATH" ] ; then
    ACCOUNT_CONF_PATH="$LE_WORKING_DIR/account.conf"
  fi
  
  if [ -f "$ACCOUNT_CONF_PATH" ] ; then
    source "$ACCOUNT_CONF_PATH"
  fi

```

Neilpang:
BTW,  and it's not normal case and not recommended to re-define `ACCOUNT_CONF_PATH`.

Fabian-Gruenbichler:
I know that it is possible to set those in the `account.conf` itself, but I figure that setting them when calling `le.sh install` probably means that the user wants them to be set in the generated config file as well. Would you be more happy if instead of sed-ing them into the file, they are put into the the `account.conf` template itself (with the default values set to the current, commented out settings)? If so, I would refresh this PR with a sed-less version like this (still only affecting the initally generated default configuration file!).

Some background on why we need to set `ACCOUNT_CONF_PATH` (and `ACCOUNT_KEY_PATH`) to non-standard locations, while keeping `LE_WORKING_DIR` standard might shed some light onto our motivation:
In Proxmox VE, we use a cluster wide synchronized file system mounted at the same path on every node. which is used to manage configuration files for the cluster (storage, cluster network, firewall) and for the VMs and containers running on the various nodes. When enabling LE certificates, each node needs to issue and renew the certificate for that one node (so we need per-node `LE_WORKING_DIR`s), but we want to use a single shared `account.conf` and `account.key` for all nodes in a cluster (i.e., `ACCOUNT_CONF_PATH` and the contained `ACCOUNT_KEY_PATH` both point to a location on our cluster filesystem). 

We'd like to make this setup as easy as possible for our users, which is why a single `le.sh install` and a single `le.issue` command per node - without manual fiddling with config files or crontab entries - would be perfect. Unfortunately it is technically not possible to use  symlinks from `LE_WORKING_DIR` to our cluster fs, so setting `ACCOUNT_CONF_PATH` is the only way to achieve this without complicated workarounds.

Neilpang:
Ok, adding it.

PR #127, opened 2016-04-12
incorrect parameter for keypath

roozbehk:
after issuing certificate 
```bash
le --issue   -d domain.tld  -w  /var/www/domain.tld 
```
--keypath is set to "no" and is not copied to ssl directories. 
```bash
le --installcert  -d domain.tld \
--certpath /etc/nginx/ssl/domain.tld.cer  \
--keypath  /etc/nginx/ssl/domain.tld.key \
--capath   /etc/nginx/ssl/ca.cer   \
--reloadcmd  "service nginx reload" \
--fullchainpath /etc/nginx/ssl/fullchain.cer 
```


Neilpang:
Thanks.

PR #158, opened 2016-04-18
Fixes hidden curl errors, adds exit level checking to curl calls

tsaavik:
Took me a really long time to troubleshoot a curl failure (certs), so I added this fixes to help others out

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -492,12 +492,12 @@ _post() {
     if [ "$needbase64" ] ; then
       response="$($CURL -A "User-Agent: $USER_AGENT" -X POST --data "$body" $url | _base64)"
       if [[ $? -gt 0 ]] ;then
-         _err "Curl failed with exit code $?, trying running script in DEBUG=1 mode"
+         _err "Curl failed with exit code $?, trying running script in DEBUG=2 mode"
       fi
     else
       response="$($CURL -A "User-Agent: $USER_AGENT" -X POST --data "$body" $url)"
       if [[ $? -gt 0 ]] ;then
-         _err "Curl failed with exit code $?, trying running script in DEBUG=1 mode"
+         _err "Curl failed with exit code $?, trying running script in DEBUG=2 mode"
       fi
     fi
   else
@@ -521,12 +521,12 @@ _get() {
     if [ "$onlyheader" ] ; then
       $CURL -I -A "User-Agent: $USER_AGENT" $url
       if [[ $? -gt 0 ]] ;then
-       _err "Curl failed with exit code $?, trying running script in DEBUG=1 mode"
+       _err "Curl failed with exit code $?, trying running script in DEBUG=2 mode"
       fi
     else
       $CURL -A "User-Agent: $USER_AGENT" $url
       if [[ $? -gt 0 ]] ;then
-       _err "Curl failed with exit code $?, trying running script in DEBUG=1 mode"
+       _err "Curl failed with exit code $?, trying running script in DEBUG=2 mode"
       fi
     fi
   else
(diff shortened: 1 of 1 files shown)


acme.sh
```diff
@@ -491,8 +491,14 @@ _post() {
     CURL="$CURL --dump-header $HTTP_HEADER "
     if [ "$needbase64" ] ; then
       response="$($CURL -A "User-Agent: $USER_AGENT" -X POST --data "$body" $url | _base64)"
+      if [[ $? -gt 0 ]] ;then
+         _err "Curl failed with exit code $?, trying running script in DEBUG=1 mode"
```
Neilpang: please use  new  `--debug`

acme.sh
```diff
@@ -491,8 +491,14 @@ _post() {
     CURL="$CURL --dump-header $HTTP_HEADER "
     if [ "$needbase64" ] ; then
       response="$($CURL -A "User-Agent: $USER_AGENT" -X POST --data "$body" $url | _base64)"
+      if [[ $? -gt 0 ]] ;then
```
Neilpang: do not use  '[[',   just use '[' instead.

```
if [ "$?" = "0" ] ;then

```

Since, '[['  is only in bash.

acme.sh
```diff
@@ -520,13 +520,13 @@ _get() {
   if _exists "curl" ; then
     if [ "$onlyheader" ] ; then
       $CURL -I -A "User-Agent: $USER_AGENT" $url
-      if [[ $? -gt 0 ]] ;then
-       _err "Curl failed with exit code $?, trying running script in DEBUG=2 mode"
+      if [ $? -gt 0 ] ;then
```
Neilpang: In the '_get()' function,  please move the change to the end of the function, so it covers both the curl and wget.  There is a variable `ret` for use.

And I think just using equal sign is ok,  instead of  `-gt`

acme.sh
```diff
@@ -491,13 +491,13 @@ _post() {
     CURL="$CURL --dump-header $HTTP_HEADER "
     if [ "$needbase64" ] ; then
       response="$($CURL -A "User-Agent: $USER_AGENT" -X POST --data "$body" $url | _base64)"
-      if [[ $? -gt 0 ]] ;then
-         _err "Curl failed with exit code $?, trying running script in DEBUG=2 mode"
+      if [ $? -gt 0 ] ;then
```
Neilpang: In this `post()` function, please add a `ret` variable to use, just as we did in the `_get()` function.
Thanks.

acme.sh
tsaavik: wget can return error codes from 0-8.
curl can return different error levels from 1 to 90

I can flip the logic to succeed when == 0, but then the final 'good' exit point will no longer be at the bottom, which seemed to be a convention that was being followed? (maybe not?)

acme.sh
tsaavik: done!

acme.sh
```diff
@@ -491,8 +491,14 @@ _post() {
     CURL="$CURL --dump-header $HTTP_HEADER "
     if [ "$needbase64" ] ; then
       response="$($CURL -A "User-Agent: $USER_AGENT" -X POST --data "$body" $url | _base64)"
+      if [ $? -gt 0 ] ;then
```
Neilpang: so,  why not add a variable "ret" here,  and move this block to the end of this "_post()" function?

tsaavik: Isn't that what I did in the last patch?

Neilpang:
Sorry, please  pull  the latest code, and create a new pull request.

Thanks.

Neilpang:
As far as I just tested,  the `wget` may return non-zero code even if it successed.   Please double check.

tsaavik:
will check, although it shouldn't according to docs
https://www.gnu.org/software/wget/manual/html_node/Exit-Status.html

tsaavik:
This was done in a later commit

PR #176, opened 2016-04-27
Added standalonePort option

akaoj:
If you can't be root on the machine, or already have something running on port 80, you can now specify which port for the HTTP server to bind to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Parameters:
 
   --webroot, -w  /path/to/webroot   Specifies the web root folder for web root mode.
   --standalone                      Use standalone mode.
+  --standalonePort                  The port to bind the HTTP server to, if in standalone mode.
   --apache                          Use apache mode.
   --dns [dns-cf|dns-dp|dns-cx|/path/to/api/file]   Use dns mode or dns api.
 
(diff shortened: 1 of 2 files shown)


Shuro:
This would be a good thing :+1:

arrowcircle:
Thats awesome!!!

heldchen:
let's encrypt will check port 80 tho, so you'd have to make sure to redirect the request somehow...

arrowcircle:
@heldchen How to deal with nginx running on 80 port? Only with webroot? Running verification on different port will easy the life.

Neilpang:
As acme protocol requires,  the public port for validation must be `80` or `443`.     

This `Le_HTTPPort` is only for use in case that the server is behind a load balancer or reverse proxy.

Do not expect it to work on non `80` port.

akaoj:
Well I should have made my commit more explicit.
This is only useful if you have, like me, a reverse proxy / load balancer running in front of your Web servers, so that you can forward the LE requests to another port (maybe on another machine).
In my personal case, I can't use this script without my patch.

Neilpang:
added.  --httpport     39c8f79f3ef8b1e16c21198fbc8f2936d806a1d8

PR #178, opened 2016-05-03
fix Le_Webroot check

defanor:
Go for other options if it is equal to "no", not if it merely contains "no".

Not sure if fixed properly, just had to do that in order to use a webroot path that contains "no".

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1057,7 +1057,7 @@ issue() {
     Le_Keylength=""
   fi
   
-  if _contains "$Le_Webroot" "no" ; then
+  if [ "$Le_Webroot" = "no" ] ; then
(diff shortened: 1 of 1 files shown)


Neilpang:
Thanks for you code, but it's not working.  We support hybrid mode. For example, the webroot string may look like:
```
no
/home/a.com,no
/home/a.com,no,no,/home/b.com
/home/a.com,no,apache,/home/wwwroot/b.com

```

So, just use equal sign is not enough.

Anyway, I just fixed it :  c53da1ef7264f96eddefcc37c8e3361591f3b266

Please try again.

Thanks.

defanor:
Works now, thanks.

PR #205, opened 2016-05-31
Add Nginx reverse proxy option

Thulium-Drake:
This adds documentation for the configuration required to make a Nginx reverse proxy
handle the ACME client's requests for proof of domain ownership.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,30 @@ If your DNS provider supports API access, we can use API to automatically issue 
 
 You don't have do anything manually!
 
+# Run on Nginx Reverse Proxy
+
+**(requires you be root/sudoer, since it is required to interact with nginx server)**
+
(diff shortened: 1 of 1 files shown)


tinproject:
What I do in nginx is:
```nginx
        location /.well-known/acme-challenge/ {
                alias /var/www/letsencrypt_webroot/.well-known/acme-challenge/;
        }
```
The `/var/www/letsencrypt_webroot` must be writable by the user that excutes acme.sh and readable by nginx user, in my case acme:www-data.

heldchen:
@Thulium-Drake careful, the new section you added broke up the existing DNS section. 

also, I think this should rather go into a wiki page as it does *not* describe how to use acme.sh but rather shows one of many, many possible webserver configuration methods.

PR #207, opened 2016-06-03
Added new parameter dnssleep

rajcz:
Added new parametr dnssleep for better handling sleep time after succesful insert of records via dns api

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1371,8 +1371,8 @@ issue() {
   fi
   
   if [ "$dnsadded" = '1' ] ; then
-    _info "Sleep 60 seconds for the txt records to take effect"
-    sleep 60
+    _info "Sleep $dnssleep seconds for the txt records to take effect"
+    sleep $dnssleep
   fi
   
   _debug "ok, let's start to verify"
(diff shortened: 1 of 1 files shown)


Neilpang:
It's not enough.  It's also needed to save such sleep time to the domain conf file, so that the auto-renewal can get use it.

rajcz:
OK, this I fix 1) save to conf 2) default 60s sleep. But If I understand right - I must add parametr to issue function with sleep too? Or can you help me with renew in code, thanks.

rajcz:
Hi Neilpang, today I completed all for dnssleep parametr. Tested and all is ok, from renew action too. Pavel

acme.sh
```diff
@@ -1105,7 +1105,8 @@ issue() {
   Le_RealCACertPath="$7"
   Le_ReloadCmd="$8"
   Le_RealFullChainPath="$9"
-  
+  dnssleep="${10}"
```
Neilpang: why not name it as something like:  "Le_Dns_Sleep"  ?

acme.sh
```diff
@@ -2235,6 +2236,7 @@ _process() {
   _accountkey=""
   _certhome=""
   _httpport=""
+  dnssleep="60"
```
Neilpang: why not  _dnssleep  ?

acme.sh
```diff
@@ -2361,13 +2363,13 @@ _process() {
           _webroot="$_webroot,$wvalue"
         fi
         ;;
-    --dnssleep 
+    --dnssleep)          
         re='^[0-9]+$'
-        if ! [[ $2 =~ $re ]] ; then
```
Neilpang: Do not use "[[",  it's bash only.   We need keep sh compatible.
Use "["  instead.

acme.sh
```diff
@@ -2361,13 +2363,13 @@ _process() {
...
-           dnssleep="60"
+        if ! [[ "$2" =~ $re ]] ; then
```
Neilpang: ??

Neilpang:
fixed  0e38c60da35620059ca3aa99a0982276260fc3c4

So, close.

PR #236, opened 2016-07-08
Add missing escape in regular expression.

tohojo:
There was a missing escape of a { character in one of the regular
expressions passed to grep. This adds that.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1725,7 +1725,7 @@ issue() {
       fi
       
       if [ "$status" = "invalid" ] ; then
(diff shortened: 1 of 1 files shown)


Neilpang:
What  is your OS version and grep version for this issue ?

tohojo:
An oldish version of openwrt; grep from busybox. Can find exact versions when I'm near that box again. 

Also, I noticed that the other place that had a similar grep pattern already escaped the { - so figured just adding the escape was a no-brainer. Certainly seems to have fixed it :)

PR #243, opened 2016-07-22
Support for shells without 'stat' (busybox ash)

t413:
I successfully got acme.sh to work on OpenWRT using the internal busybox ash shell!
It required installing curl, ca-certificates, and adding `export SSL_CERT_DIR=/etc/ssl/certs` to /etc/profile. The only problem was when using the webroot verification option it tried to chown the created folder but couldn't use the `stat` command to get the correct permissions– that command doesn't exist in busybox.

My fix is simple:
- the _stat command can now return an error
- the issue() command chown command isn't run if _stat fails

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -549,6 +549,7 @@ _stat() {
   if stat -f  '%Su:%Sg' "$1" 2>/dev/null ; then
     return
   fi
+  return 3; #error, 'stat' not found
 }
 
 #keyfile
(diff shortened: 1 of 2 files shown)


acme.sh
```diff
@@ -549,6 +549,7 @@ _stat() {
   if stat -f  '%Su:%Sg' "$1" 2>/dev/null ; then
     return
   fi
+  return 3; #error, 'stat' not found
```
Neilpang: Please just return 1  here.

Neilpang:
BTW,   can you please add a wiki page here:　https://github.com/Neilpang/acme.sh/wiki

Add a new page like:  How to use on busybox.

Thanks.

t413:
Alright, how's that look? Added the line return, changed the exit code, added a wiki page howto article, and linked to it from the README.md.

Neilpang:
Thanks.

PR #245, opened 2016-07-29
Create private key files not world-readable

vmmello:
One thing that I find annoying is that the private key files are create world-readable (e.g. when the user uses umask 022, that is the default mostly everywhere). On most cases this is not a problem, but on shared systems of universities, shared hosting companies, etc, it's a problem.

This pull request changes the _createKey() function to create private key files readable only by the user executing the script.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -337,6 +337,18 @@ _createkey() {
     _info "Using ec name: $eccname"
   fi
 
+  # to prevent the key file from being world-readable
+  # create an empty file and chmod 600 before saving the key contents
+  if ! touch "$f"; then
+    _err "unable to create empty file '$f' for private key"
+    return 1
+  fi
+
+  if ! chmod 600 "$f"; then
+    _err "unable to chmod 600 key file $f"
+    return 1
+  fi
+
   #generate account key
   if [ "$isec" ] ; then
     openssl ecparam  -name $eccname -genkey 2>/dev/null > "$f"


Neilpang:
At line 2363,  the acme home dir is set the mode  700
https://github.com/Neilpang/acme.sh/blob/master/acme.sh#L2363

So, nobody else can read your keys.

What the actual problem/case  did you see ?

vmmello:
I did see that line meanwhile I was writing the patch. Though on the Ubuntu 14.04 server that I'm using the directory ~/.acme.sh/ was still 755, not 700. I had used the standalone options to create my certs (...what likely created the ~/.acme.sh/ directory).

This fix is for correctness, as private keys should never be world-readable.

heldchen:
In our setup, the group also needs read wrights as there are multiple users for the same cert (mail + web for example). I guess everyone has different requirements, so in my opinion, a sane default for acme.sh is good enough, further hardening if required could be up to the user...

vmmello:
Right.

A possibility is to make the key protection optional and create the option --keyChmod NNN

I knew that this change was intrusive and could break some setups, but I wanted the first PR to be as simple as possible to get feedback.

I'll think about the options and see if I re-submit the proposal with a less intrusive change.

Thanks guys for the feedback so far.

Neilpang:
I just tried to install acme.sh on a fresh new machine:
```
root@adm:~# ll
total 168K
drwx------ 16 root root 4.0K Jul 30 01:19 .
drwxr-xr-x 22 root root 4.0K Aug  4  2014 ..
drwx------  3 root root 4.0K Jul 30 01:19 .acme.sh

```

The `~/.acme.sh`  is mode "700".  

I don't want to add more arguments for file mode,  since it's too trivial.


If you want to do extra operations after the cert is issued,  please use `--reloadcmd`.

```
cat /root/.acme.sh/myreload.sh

echo "this will be called after the cert is issued/renewed."
```

Install reload shell
```
acme.sh --issue -d aa.com .................

acme.sh  --installcert  --reloadcmd ".  /root/.acme.sh/myreload.sh"
```


You can do anything in `myreload.sh`, it will be called when cert is issued/renewed automatically.

Neilpang:
We had some discussion about the file mode bofore.
for example:  https://github.com/Neilpang/acme.sh/issues/174

PR #247, opened 2016-07-30
Added TTL to dns_lexicon.sh

geekuality:
Realized missing `--ttl` parameter while debugging and ultimately discovering that there is a bug (or actually completely missing functionality) in lexicon that causes the parameter currently not to be processed at all, but I submitted an issue for it too. (https://github.com/AnalogJ/lexicon/issues/49)

### DIFF
--- a/dnsapi/dns_lexicon.sh
+++ b/dnsapi/dns_lexicon.sh
@@ -61,7 +61,7 @@ dns_lexicon_add() {
     _saveaccountconf $Lx_domaintoken "$Lx_domaintoken_v"
   fi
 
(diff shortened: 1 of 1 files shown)


Neilpang:
Thanks for your code.
But I just saw that  lexicon was fixed,  and the default ttl is  300 now.

So, I think this PR is not necessary now. 
And I'm a little worried that some of the dns providers may not support very small ttl, such like 120.
I think 300 is a good value.

What do you think?

Neilpang:
BTW,  I see you are using DNS made easy API. 
Can you please add an example on this wiki page:

https://github.com/Neilpang/acme.sh/wiki/How-to-use-lexicon-dns-api

Just like what we have for cloudflare and namesilo there.

Neilpang:
@geekuality

PR #249, opened 2016-08-01
Make `--install' respect $DESTDIR if set.

archiecobbs:
This allows `--install` to work for example during RPM builds, where installation needs to go into a staging area before packaging.


Neilpang:
you can use `--home` or `--config-home` instead:
```
acme.sh --install   --home "$DEST"
```

or
```
acme.sh --install   --config-home "$DEST"
```

PR #250, opened 2016-08-01
Use correct syntax for test(1) non-empty string.

archiecobbs:


Neilpang:
Is there any problems for you with the old code ?

archiecobbs:
I thought it was causing a problem but actually it is working fine. So there's no bug. However it might make the code a little more clear.

Neilpang:
Ok.
We are always using this style to check empty string.
I'd like to keep it.

Would you mind if I close this PR ?
Thanks.

archiecobbs:
Sounds good. Closing now.

PR #255, opened 2016-08-05
dns_cx.sh prefix and suffix domain bug fixed

jim3mar:
If we have multi domain in Cloudxns, such as:  "acme.sh", "**acme.sh**.sh", "acme.**acme.sh**".
When we issue a cert for "**acme.sh**", all of the domains are matched!
We must exactly determine the domain.
The Cloudxns rest api return json data, so we can match the domain use double quotation marks in grep command.

### DIFF
--- a/dnsapi/dns_cx.sh
+++ b/dnsapi/dns_cx.sh
@@ -145,7 +145,7 @@ _get_root() {
     fi
 
     if printf "$response" | grep "$h." >/dev/null ; then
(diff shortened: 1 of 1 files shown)


Neilpang:
Looks good.  thanks.

jim3mar:
Woo, merged quick as a flash

Neilpang:
I'm just doing the test too.

PR #262, opened 2016-08-13
Support RSA and ECDSA dual certs

Neilpang:
The ecc cert forlder will have a `_ecc` suffix.
```
~/.acme.sh/domain.com_ecc
```

All the existing certs(rsa or ecc) will keep renewed in the old dir.  The new issued ecc certs will use this ne domain folder.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env sh
 
-VER=2.3.6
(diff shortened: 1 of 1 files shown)


Neilpang:
https://github.com/Neilpang/acme.sh/issues/74

PR #274, opened 2016-08-25
Add option for a custom ca-bundle file.

sodre:
Sometimes an OS comes without a default ca-cert bundle file. This option allows one to point curl or wget to a CA certificate bundle file.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -743,6 +743,10 @@ _inithttp() {
       CURL="$CURL --trace-ascii $_CURL_DUMP "
     fi
 
+    if [ "$CACERT_BUNDLE" ] ; then
+      CURL="$CURL --cacert $CACERT_BUNDLE "
+    fi
+
     if [ "$HTTPS_INSECURE" ] ; then
       CURL="$CURL --insecure  "
     fi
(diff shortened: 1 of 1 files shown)


Neilpang:
We have a `--insecure `  param,  which indicates not to check the  ca cert.  Is that ok for you ?

sodre:
I saw that option, the concern was with a possible MITM attack.

Neilpang:
I prefer the param name  "--ca-bundle" to the name "--cacert".
Please change the name, and re-pull.  I will review it soon.

Thanks.

sodre:
No problem, I actually prefer it as well after you mentioned it.
I am retesting the code after the rename and will push it soon.

sodre:
It is ready for you to review.

Neilpang:
We need to save the ca bundle  to the account.conf,  so that, when we are renew the certs automatically, the ca bundle can also work for us.

Please search for "HTTPS_INSECURE" macro,  and see what it does.

sodre:
That is done as well. 
Is there an easy way that I can force a renew in my certificates? Right now all the ones I have are brand new.

Neilpang:
add  `--force` to force to renew.

sodre:
👍 It works with --renew as well.

PR #345, opened 2016-10-27
add retry for temp authz request error

Neilpang:
https://github.com/Neilpang/acme.sh/issues/344


guestisp:
Cool, with this merge, I can remove my retry login (that was repeating the whole procedure from start)

PR #354, opened 2016-10-31
support --reloadcmd on renew

pdbogen:


Neilpang:
if you want to set/change the reloadcmd,  please use `--installcert` command.  Do not use `--renew`.

PR #372, opened 2016-11-07
adding DNSMadeEasy API

justmwa:
No need for lexicon either.

### DIFF
--- a/dnsapi/dns_me.sh
+++ b/dnsapi/dns_me.sh
@@ -123,8 +123,8 @@ _me_rest() {
   data="$3"
   _debug $ep
 
-  cdate=$(date -Ru)
-  hmac=$(printf "$cdate" | openssl dgst -sha1 -hmac $ME_Secret | cut -d = -f 2 | tr -d ' ')
+  cdate=$(date -u +"%a, %d %b %Y %T %Z")
+  hmac=$(printf "$cdate"  | _hmac  sha1  "$ME_Secret"   1)
 
   _H1="x-dnsme-apiKey: $ME_Key"
   _H2="x-dnsme-requestDate: $cdate"
--- a/dnsapi/dns_me.sh
+++ b/dnsapi/dns_me.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 # bug reports to dev@1e.ca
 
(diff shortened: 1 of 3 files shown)


dnsapi/dns_me.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/sh
```
Neilpang: use `#!/usr/bin/env sh`  instead.

dnsapi/dns_me.sh
```diff
@@ -0,0 +1,148 @@
...
+}
+
+_me_rest() {
+  m=$1
+  ep="$2"
+  data="$3"
+  _debug $ep
+
+  cdate=$(date -Ru)
+  hmac=$(printf "$cdate" | openssl dgst -sha1 -hmac $ME_Secret | cut -d = -f 2 | tr -d ' ')
```
Neilpang: I just added a new function "_hmac()" for you. at b001840dee20d31423a7fb29e18ac9177b7f9a4c

Please use `_hmac()` instead of `openssl`.

```
printf "$cdate"  | _hmac  sha1  "$ME_Secret"   1
```

In future, we will have other implementations instead of openssl.
do not use openssl in the dns apis.

dnsapi/dns_me.sh
```diff
@@ -0,0 +1,148 @@
...
+  return 1
+}
+
+_me_rest() {
+  m=$1
+  ep="$2"
+  data="$3"
+  _debug $ep
+
+  cdate=$(date -Ru)
```
Neilpang: Is it a must to use `-R` option here ?
As in my tests,  the `-R` option is not supported in `sun solaris` and `openbsd`.

justmwa: Since custom strings are localized the only way I found is
LANG=C date +"%a, %d %b %Y %T %Z" -u
can't tell if it works on solaris / openbsd though.

Neilpang: good.
Please use `date -u +"%a, %d %b %Y %T %Z"`  instead.
It's working on all the platforms I tested.

justmwa:
Just tested, it's OK.

Neilpang:
merged to https://github.com/Neilpang/acme.sh/pull/376

Thanks.

Neilpang:
Hi @justmwa 

Would you mind implementing the `rm` function:  https://github.com/Neilpang/acme.sh/issues/681

Thanks.

PR #414, opened 2016-11-21
Cyon.ch DNS API

noplanman:
Add DNS API support for cyon.ch

### DIFF
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -305,6 +305,24 @@ Note that you cannot use acme.sh automatic DNS validation for FreeDNS public dom
 you create under a FreeDNS public domain.  You must own the top level domain in order to automaitcally
 validate with acme.sh at FreeDNS.
 
+## 16. Use cyon.ch
+
+You only need to set your cyon.ch login credentials.
+If you also have 2 Factor Authentication (OTP) enabled, you need to set your secret token too and have `oathtool` installed.
+
+```
+export cyon_username="your_cyon_username"
+export cyon_password="your_cyon_password"
+export cyon_otp_secret="your_otp_secret" # Only required if using 2FA
+```
+
+To issue a cert:
+```
+acme.sh --issue --dns dns_cyon -d example.com -d www.example.com
+```
+
+The `cyon_username`, `cyon_password` and `cyon_otp_secret` will be saved in `~/.acme.sh/account.conf` and will be reused when needed.
+
 # Use custom API
 
 If your API is not supported yet, you can write your own DNS API.
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -311,9 +311,9 @@ You only need to set your cyon.ch login credentials.
 If you also have 2 Factor Authentication (OTP) enabled, you need to set your secret token too and have `oathtool` installed.
 
 ```
-export cyon_username="your_cyon_username"
-export cyon_password="your_cyon_password"
-export cyon_otp_secret="your_otp_secret" # Only required if using 2FA
+export CY_Username="your_cyon_username"
+export CY_Password="your_cyon_password"
+export CY_OTP_Secret="your_otp_secret" # Only required if using 2FA
 ```
 
 To issue a cert:
@@ -321,7 +321,7 @@ To issue a cert:
 acme.sh --issue --dns dns_cyon -d example.com -d www.example.com
 ```
 
-The `cyon_username`, `cyon_password` and `cyon_otp_secret` will be saved in `~/.acme.sh/account.conf` and will be reused when needed.
+The `CY_Username`, `CY_Password` and `CY_OTP_Secret` will be saved in `~/.acme.sh/account.conf` and will be reused when needed.
 
 # Use custom API
 
--- a/dnsapi/dns_cyon.sh
+++ b/dnsapi/dns_cyon.sh
@@ -7,7 +7,6 @@
 #
 # Dependencies:
 # -------------
-# - jq (get it here: https://stedolan.github.io/jq/download)
 # - oathtool (When using 2 Factor Authentication)
 #
 # Author: Armando Lüscher <armando@noplanman.ch>
@@ -34,10 +33,6 @@
 ########
 
 dns_cyon_add() {
-  if ! _exists jq; then
-    _fail "Please install jq to use cyon.ch DNS API."
-  fi
-
   _load_credentials
   _load_parameters "$@"
 
@@ -165,8 +160,8 @@ _login() {
   _debug login_response "${login_response}"
 
   # Bail if login fails.
-  if [ "$(echo "${login_response}" | jq -r '.onSuccess')" != "success" ]; then
-    _fail "    $(echo "${login_response}" | jq -r '.message')"
+  if [ "$(echo "${login_response}" | _get_response_success)" != "success" ]; then
+    _fail "    $(echo "${login_response}" | _get_response_message)"
   fi
 
   _info "    success"
@@ -199,8 +194,8 @@ _login() {
     _debug otp_response "${otp_response}"
 
     # Bail if OTP authentication fails.
-    if [ "$(echo "${otp_response}" | jq -r '.onSuccess')" != "success" ]; then
-      _fail "    $(echo "${otp_response}" | jq -r '.message')"
+    if [ "$(echo "${otp_response}" | _get_response_success)" != "success" ]; then
+      _fail "    $(echo "${otp_response}" | _get_response_message)"
     fi
 
     _info "    success"
@@ -227,11 +222,11 @@ _domain_env() {
 
   _check_2fa_miss "${domain_env_response}"
 
-  domain_env_success=$(echo "${domain_env_response}" | jq -r '.authenticated')
+  domain_env_success=$(echo "${domain_env_response}" | _egrep_o '"authenticated":\w*' | cut -d : -f 2)
 
   # Bail if domain environment change fails.
   if [ "${domain_env_success}" != "true" ]; then
-    _fail "    $(echo "${domain_env_response}" | jq -r '.message')"
+    _fail "    $(echo "${domain_env_response}" | _get_response_message)"
   fi
 
   _info "    success"
@@ -252,14 +247,11 @@ _add_txt() {
 
   _check_2fa_miss "${addtxt_response}"
 
-  addtxt_message=$(echo "${addtxt_response}" | jq -r '.message')
-  addtxt_status=$(echo "${addtxt_response}" | jq -r '.status')
+  addtxt_message=$(echo "${addtxt_response}" | _get_response_message)
+  addtxt_status=$(echo "${addtxt_response}" | _get_response_status)
 
   # Bail if adding TXT entry fails.
   if [ "${addtxt_status}" != "true" ]; then
-    if [ "${addtxt_status}" = "null" ]; then
-      addtxt_message=$(echo "${addtxt_response}" | jq -r '.error.message')
-    fi
     _fail "    ${addtxt_message}"
   fi
 
@@ -306,14 +298,11 @@ _delete_txt() {
 
     _check_2fa_miss "${delete_txt_response}"
 
-    delete_txt_message=$(echo "${delete_txt_response}" | jq -r '.message')
-    delete_txt_status=$(echo "${delete_txt_response}" | jq -r '.status')
+    delete_txt_message=$(echo "${delete_txt_response}" | _get_response_message)
+    delete_txt_status=$(echo "${delete_txt_response}" | _get_response_status)
 
     # Skip if deleting TXT entry fails.
     if [ "${delete_txt_status}" != "true" ]; then
-      if [ "${delete_txt_status}" = "null" ]; then
-        delete_txt_message=$(echo "${delete_txt_response}" | jq -r '.error.message')
-      fi
       _err "    ${delete_txt_message} (${_identifier})"
     else
       _info "    success (${_identifier})"
@@ -322,6 +311,18 @@ _delete_txt() {
 
   _info "    done"
   _info ""
+}
+
+_get_response_message() {
+  _egrep_o '"message":"[^"]*"' | cut -d : -f 2 | tr -d '"'
+}
+
+_get_response_status() {
+  _egrep_o '"status":\w*' | cut -d : -f 2
+}
+
+_get_response_success() {
+  _egrep_o '"onSuccess":"[^"]*"' | cut -d : -f 2 | tr -d '"'
 }
 
 _check_2fa_miss() {
--- a/dnsapi/dns_cyon.sh
+++ b/dnsapi/dns_cyon.sh
@@ -9,27 +9,12 @@
 # -------------
 # - oathtool (When using 2 Factor Authentication)
 #
+# Issues:
+# -------
+# Any issues / questions / suggestions can be posted here:
+# https://github.com/noplanman/cyon-api/issues
+#
 # Author: Armando Lüscher <armando@noplanman.ch>
-########
-
-########
-# Define cyon.ch login credentials:
-#
-# Either set them here: (uncomment these lines)
-#
-# cyon_username='your_cyon_username'
-# cyon_password='your_cyon_password'
-# cyon_otp_secret='your_otp_secret' # Only required if using 2FA
-#
-# ...or export them as environment variables in your shell:
-#
-# $ export cyon_username='your_cyon_username'
-# $ export cyon_password='your_cyon_password'
-# $ export cyon_otp_secret='your_otp_secret' # Only required if using 2FA
-#
-# *Note:*
-# After the first run, the credentials are saved in the "account.conf"
-# file, so any hard-coded or environment variables can then be removed.
 ########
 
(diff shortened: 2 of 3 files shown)


noplanman:
@Neilpang I'd appreciate your input on the last issue that is making the script fail.
I don't need double quotes there, as I would probably have to escape everything, making it quite unreadable.

Thanks!

Neilpang:
@noplanman 

Please do not use `jq` here.

Use `_normalizejson()` and `_egrep_o()` combination instead.  If you need any help there,  please paste your difficulty here, I can help you.

Thanks.

noplanman:
@Neilpang Thanks for offering your help, so very much appreciated!

Wow, what a learning curve, phew. I've managed to change everything to not require `jq` 🎉 

The test is still failing though, but it gives me no output to go by, so I have no idea how to fix it.
Maybe an error in the travis configuration?

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,347 @@
...
+  if test "${1#*multi_factor_form}" != "$1"; then
+    _fail "    Missed OTP authentication!"
+  fi
+}
+
+_fail() {
+  _err "$1"
+  _err ""
+  _cleanup
+  exit 1
```
Neilpang: Do not call `exit 1`,  you should just `return 1` to the main shell.

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,347 @@
...
+  # Save the login credentials to the account.conf file.
+  _debug "Save credentials to account.conf"
+  _saveaccountconf cyon_username "${cyon_username}"
+  _saveaccountconf cyon_password_b64 "$cyon_password_b64"
+  if [ ! -z "${cyon_otp_secret}" ]; then
+    _saveaccountconf cyon_otp_secret "$cyon_otp_secret"
+  fi
+}
+
+_is_idn() {
```
Neilpang: Hnnnnnnnnnnnn.......

_is_idn .....

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,347 @@
...
+    _info ""
+    _info "  * Full Domain: ${fulldomain}"
+    _info "  * Cookie Jar:  ${cookiejar}"
+    _info ""
+  fi
+}
+
+_login() {
+  _info "  - Logging in..."
+  login_response=$(curl \
```
Neilpang: do not use `curl`,   please use `_get()` or `_post()` function instead.

If you need cookies,  please check the `responseHeaders`.

If yo want to set cookies headers,  please set `_H1` ,`_H2` ,`_H3` , or `_H4`.

If there is any problem,  please give me the  debug2 log, I can help you.

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,347 @@
...
+  # 2FA authentication with OTP?
+  if [ ! -z "${cyon_otp_secret}" ]; then
+    _info "  - Authorising with OTP code..."
+
+    if ! _exists oathtool; then
+      _fail "Please install oathtool to use 2 Factor Authentication."
+    fi
+
+    # Get OTP code with the defined secret.
+    otp_code=$(oathtool --base32 --totp "${cyon_otp_secret}" 2>/dev/null)
```
Neilpang: what is  oathtool ?

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,347 @@
...
+  fi
+
+  _info ""
+}
+
+_domain_env() {
+  _info "  - Changing domain environment..."
+
+  # Get the "example.com" part of the full domain name.
+  domain_env=$(printf "%s" "${fulldomain}" | sed -E -e 's/.*\.(.*\..*)$/\1/')
```
Neilpang: You can not detect the root domain like that.
For example,  you will get wrong result for domain:  `www.mydomain.co.uk`

See the `_get_root()` function:  
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_cf.sh#L98

dnsapi/dns_cyon.sh
noplanman: Shall I rename it to something own?
Also, does this function overwrite the acme.sh one outside of my script?

dnsapi/dns_cyon.sh
noplanman: oathtool is used to generate TOTP codes for the 2 Factor Authentication.
Since cyon offer 2FA, having oathtool and setting the OTP secret as an environment variable, the script can successfully log into the account.

dnsapi/dns_cyon.sh
sjau: @noplanman 

What I did for ISPConfig was looping through the string given by acme.sh and check if such a (sub.)domain exists.

```
  curZone="${fulldomain}"
  while [ "${zoneEnd}" = false ]; do
    # we can strip the first part of the fulldomain, since it's just the _acme-challenge string
    curZone="${curZone#*.}"
    # suffix . needed for zone -> domain.tld.
    curData="{\"session_id\":\"${sessionID}\",\"primary_id\":{\"origin\":\"${curZone}.\"}}"
    curResult="$(_post "${curData}" "${ISPC_Api}?dns_zone_get")"
    _debug "Calling _ISPC_getZoneInfo: '${curData}' '${ISPC_Api}?login'"
    _debug "Result of _ISPC_getZoneInfo: '$curResult'"
    if _contains "${curResult}" '"id":"'; then
      zoneFound=true
      zoneEnd=true
      _info "Retrieved zone data."
      _debug "Zone data: '${curResult}'"
    fi
    if [ "${curZone#*.}" != "$curZone" ]; then
      _debug2 "$curZone still contains a '.' - so we can check next higher level"
    else
      zoneEnd=true
      _err "Couldn't retrieve zone data."
      return 1
    fi
  done
```
Not very pretty but it works...

It's just a loop that strips text from the front of a string to the first "." and then I check if there's a zone entry for that in the db so that I get according infos.

first, fulldomain is like:   _acme-challenge.sub.domain.tld. It strips away the _acme-challenge. part and checks if there's an entry for sub.domain.tld. If not, the script notices there are still "." in the string so it strips the next part from the front. It will look up then domain.tld. 

Maybe that is of use to you.

noplanman: @sjau Thanks for posting your solution.

I guess I could do something similar to that, I just don't think it makes sense though.
I'll think about it and see if I do something along that line.

sjau: @noplanman 

good luck :)

dnsapi/dns_cyon.sh
noplanman: Has been renamed.

dnsapi/dns_cyon.sh
noplanman: Finally got it working 👍 

I still need curl though, as a helper to correctly urlencode the post data, so just locally.
There is a `_urlencode` function in the `acme.sh` script, but that doesn't do it.

noplanman:
Right, after an epic delay, I've finally updated the script to the latest version.

@Neilpang You may scrutinise 😃

noplanman:
@Neilpang Any feedback would be much appreciated 😊

Neilpang:
Sorry,  I was just too busy recently.

Please update the usage in the readme file:
https://github.com/Neilpang/acme.sh/blob/master/README.md
https://github.com/Neilpang/acme.sh/tree/master/dnsapi

noplanman:
Done. Is it ok like that?

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,326 @@
...
+}
+
+# comment on https://stackoverflow.com/a/10797966
+_cyon_urlencode() {
+  curl -Gso /dev/null -w "%{url_effective}" --data-urlencode @- "" | cut -c 3-
+}
+
+_cyon_load_parameters() {
+  # Read the required parameters to add the TXT entry.
+  fulldomain="$(printf "%s" "${1}" | tr '[:upper:]' '[:lower:]')"
```
Neilpang: Do not use `[:upper:]` or `[:lower:]`, in some old version of tr, it's not supported.
please use this instead:
```
tr '[A-Z]' '[a-z]'
```

see: https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_lexicon.sh#L33

dnsapi/README.md
```diff
@@ -240,6 +240,24 @@ acme.sh --issue --dns dns_ispconfig -d example.com -d www.example.com
 
 The `ISPC_User`, `ISPC_Password`, `ISPC_Api`and `ISPC_Api_Insecure` will be saved in `~/.acme.sh/account.conf` and will be reused when needed.
 
+## 13. Use cyon.ch
+
+You only need to set your cyon.ch login credentials.
+If you also have 2 Factor Authentication (OTP) enabled, you need to set your secret token too and have `oathtool` installed.
+
+```
+export cyon_username="your_cyon_username"
```
Neilpang: Would you mind following the name rule ?
Something like: `CY_username` instead of `cyon_username`

And `CY_password`  `CY_otp_secret`

Neilpang: Sorry,  should be `CY_Username`

Neilpang: `CY_OTP_Secret`

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,326 @@
...
+
+_cyon_is_idn() {
+  _idn_temp="$(printf "%s" "${1}" | tr -d "[0-9a-zA-Z.,-_]")"
+  _idn_temp2="$(printf "%s" "${1}" | grep -o "xn--")"
+  [ "$_idn_temp" ] || [ "$_idn_temp2" ]
+}
+
+# comment on https://stackoverflow.com/a/10797966
+_cyon_urlencode() {
+  curl -Gso /dev/null -w "%{url_effective}" --data-urlencode @- "" | cut -c 3-
```
Neilpang: Is it possible to implement it without `curl` ?  Or is there a way to do that with `wget` too?

It will be better if we can work without `curl`.

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,326 @@
...
+    _info "| Deleting DNS TXT entry from your cyon.ch domain |"
+    _info "+-------------------------------------------------+"
+    _info ""
+    _info "  * Full Domain: ${fulldomain}"
+    _info ""
+  fi
+}
+
+_cyon_get_cookie_header() {
+  printf "%s" "$(sed -n 's/Set-\(Cookie:.*cyon=[^;]*\).*/\1/p' "$HTTP_HEADER" | _tail_n 1)"
```
Neilpang: If you want to get the cookie header.  Please use `_egrep_o` or `grep`,   `sed` has problems when crossing platforms.

dnsapi/dns_cyon.sh
noplanman: I got that solution from [here](https://stackoverflow.com/questions/296536/how-to-urlencode-data-for-curl-command)
There's another page with some [other solutions](https://stackoverflow.com/questions/6250698/how-to-decode-url-encoded-string-in-shell), but most use sed.

As I'm new to bash and don't really know about the compatibility and cross-platform functionality, do you have any input on this?

dnsapi/dns_cyon.sh
noplanman: I saw that `_egrep_o` also has `sed` in it, which is why I used it. But the `sed` code that is used didn't work for my use case.

How can I do this line with `_egrep_o`? I'm lost when it comes to `grep`, as it doesn't seem to do what I want...

This is what this line does:
Input (from `http.header`):
```
HTTP/1.1 200 OK
Server: nginx
Set-Cookie: cyon=IWANTTHIS; something-else=anything;
Set-Cookie: other=values;
...
```
output (the value I want from this):
```
Cookie: cyon=IWANTTHIS
```

Neilpang:
```bash
cat "$HTTP_HEADER" | grep "cyon=" | grep "^Set-Cookie:" |  _head_n 1 |　_egrep_o  'cyon=[^;]*;'  | tr -d ';'
```

dnsapi/dns_cyon.sh
Neilpang: Tell me what you want here.
What is the input and what is the output.

dnsapi/dns_cyon.sh
noplanman: Ok, have updated this, thanks!

noplanman:
Ok, latest changes are rebased and shellcheck is now happy too :blush:

dnsapi/dns_cyon.sh
Neilpang: How is going on?

noplanman: @Neilpang I still haven't found a good / better solution to this than using curl directly, which encodes everything properly 😕

Neilpang: @noplanman 
I was finding too, but I was too busy recently.

If it's just username or password,  I think we can do a simple replacement.  replace the char with it's hex code.

noplanman: Is there an easy way to do that for all characters?
The problem I was running into, is that the password can obviously have all sorts of characters and we'd have to decide which ones are "valid".

I guess just sticking to the western alphabet and adding most special characters should do the trick.

Neilpang: see the function:  `_is_idn`.  I think `printf` and `tr` can deal with any English char or no-English char.

Neilpang: Hi @noplanman 

I just implemented the `_url_encode()` function.   https://github.com/Neilpang/acme.sh/commit/542d7977db1479f9383561ae1657dcfb2ecfcb2c


Please rebase and try again.

noplanman: @Neilpang Done, nice work!

noplanman:
@Neilpang Please don't merge just yet, there still seems to be a cyon-specific issue that has just been reported by a user, so that needs to get solved first.

noplanman:
@Neilpang Could you maybe assist with the build/shellcheck issues? How do I fix them?

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,321 @@
...
+
+_cyon_is_idn() {
+  _idn_temp="$(printf "%s" "${1}" | tr -d "[0-9a-zA-Z.,-_]")"
+  _idn_temp2="$(printf "%s" "${1}" | grep -o "xn--")"
+  [ "$_idn_temp" ] || [ "$_idn_temp2" ]
+}
+
+_cyon_load_parameters() {
+  # Read the required parameters to add the TXT entry.
+  fulldomain="$(printf "%s" "${1}" | tr '[A-Z]' '[a-z]')"
```
Neilpang: Don't use [] around ranges in tr,
Use:
```
tr "A-Z" "a-z"
```

dnsapi/dns_cyon.sh
```diff
@@ -0,0 +1,321 @@
...
+  fi
+
+  _debug fulldomain "${fulldomain}"
+  _debug fulldomain_idn "${fulldomain_idn}"
+
+  txtvalue="${2}"
+  _debug txtvalue "${txtvalue}"
+
+  # This header is required for curl calls.
+  _H1="X-Requested-With: XMLHttpRequest"
```
Neilpang: use `export`

```
export  _H1="X-Requested-With: XMLHttpRequest"
```

dnsapi/dns_cyon.sh
noplanman: Have done this now, but shellcheck now tells me I should use what I had to start with:
https://travis-ci.org/noplanman/acme.sh/jobs/196962127#L196

Neilpang:
@noplanman 

Please rebase and fix the errors.

noplanman:
@Neilpang Have rebased and added a fix to a problem some users were facing. It's not a 100% solution yet, having trouble finding the issue.

Also, regarding the [failing travis test](https://travis-ci.org/Neilpang/acme.sh/jobs/200843683), it's moaning about something you suggested I change: `Use '[:upper:]' to support accents and foreign alphabets.`

Unless it's failing because of something else?

Neilpang:
@noplanman 
Please suppress the warnings,  see: https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_lexicon.sh#L34

noplanman:
@Neilpang Ok, all good now 👍

Neilpang:
@noplanman 
What is `oathtool` used for ? I'm not familiar with it at all.   Is it possible not to use `oathtool` in the script ?

noplanman:
`oathtool` is used to generate one time passwords (OTP).

This is required, because cyon offers 2 factor authentication. So if a user has 2FA activated, the script need to be able to log in with the 2FA token, which is generated by `oathtool` using the secret key that gets passed with the `CY_OTP_Secret`.

Because cyon doesn't yet have an official API with bearer/authentication tokens, the only way to use it is to simulate a user actually logging in and adding the domain TXT entry manually.

So it's only required if a user has 2FA activated.

PR #454, opened 2016-12-06
Support of temporary access keys from AWS IAM Roles

AntonTimiskov:
Ref: https://github.com/Neilpang/acme.sh/issues/453

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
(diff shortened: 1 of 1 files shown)


dnsapi/dns_aws.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
Neilpang: NO, we can not use bash.

AntonTimiskov:
Closed until implementation of "Support IAM Roles"

PR #474, opened 2016-12-18
Fix typo.

noplanman:


Neilpang:
good catch.

But we must deal with the compatibility too.   when renewing the cert, we must read the old name first.

noplanman:
Ok, have added the check of the old typo, removing it from the config at the same time.

Neilpang:
Thank you.

PR #500, opened 2017-01-02
Add support for HTTP DELETE method.

wpk-:
Hi Neil,

My host's API requires a HTTP DELETE to delete the DNS records. I've added the `_delete` function starting from the `_get` function and making a few small changes. Hope you're happy with this. Please let me know if you desire any modifications.

Cheers,
Paul


Neilpang:
I believe we have already supported `DELETE` method.
See: 
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_cf.sh#L114
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_cf.sh#L172

wpk-:
Ah, I see. My bad. Only checked `_get` and not `_post`.

PR #513, opened 2017-01-06
Fall back to python if OpenSSL does not support '-macopt'

ksperling:
This allows the script to work with the tools shipped with OS X.

The AWS DNS plugin in particular requires HMAC operations with binary keys, because the output form one HMAC becomes the key for the next round.

(The uses of HMAC in dns_ali and dns_me could be rewritten to instead pass the key as plain text rather than hex, and thus work with OpenSSL 0.9, but this wouldn't work for dns_aws).

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -461,10 +461,19 @@ _hmac() {
   fi
 
   if [ "$alg" = "sha256" ] || [ "$alg" = "sha1" ]; then
-    if [ "$outputhex" ]; then
-      $OPENSSL_BIN dgst -"$alg" -mac HMAC -macopt "hexkey:$secret_hex" | cut -d = -f 2 | tr -d ' '
+    # OpenSSL only support -macopt from version 1.0. OS X ships version 0.9
+    if echo -n "" | $OPENSSL_BIN dgst -sha1 -mac HMAC -macopt hexkey:00 >/dev/null 2>&1; then
+      if [ "$outputhex" ]; then
+        $OPENSSL_BIN dgst -"$alg" -mac HMAC -macopt "hexkey:$secret_hex" | cut -d = -f 2 | tr -d ' '
+      else
+        $OPENSSL_BIN dgst -"$alg" -mac HMAC -macopt "hexkey:$secret_hex" -binary
+      fi
     else
-      $OPENSSL_BIN dgst -"$alg" -mac HMAC -macopt "hexkey:$secret_hex" -binary
+      # Try to fall back to python's built-in hmac/hashlib modules
+       [ -n "$outputhex" ] && outputhex=hex
+      python -u -c \
+          'import sys,binascii,hmac,hashlib;sys.stdout.write(getattr(hmac.new(binascii.unhexlify(sys.argv[3]),sys.stdin.read(),getattr(hashlib,sys.argv[1])),sys.argv[2])())' \
+          "$alg" "${outputhex}digest" "$secret_hex"
     fi
   else
     _err "$alg is not supported yet"
--- a/acme.sh
+++ b/acme.sh
@@ -470,10 +470,10 @@ _hmac() {
       fi
     else
       # Try to fall back to python's built-in hmac/hashlib modules
-       [ -n "$outputhex" ] && outputhex=hex
+      [ -n "$outputhex" ] && outputhex=hex
       python -u -c \
-          'import sys,binascii,hmac,hashlib;sys.stdout.write(getattr(hmac.new(binascii.unhexlify(sys.argv[3]),sys.stdin.read(),getattr(hashlib,sys.argv[1])),sys.argv[2])())' \
-          "$alg" "${outputhex}digest" "$secret_hex"
+        'import sys,binascii,hmac,hashlib;sys.stdout.write(getattr(hmac.new(binascii.unhexlify(sys.argv[3]),sys.stdin.read(),getattr(hashlib,sys.argv[1])),sys.argv[2])())' \
+        "$alg" "${outputhex}digest" "$secret_hex"
     fi
   else
     _err "$alg is not supported yet"


Neilpang:
Sorry,  it's never an option to use python in `acme.sh`.
Please upgrade your openssl to the latest version, or  define `OPENSSL_BIN` pointing to the new openssl executablefile.

```
OPENSSL_BIN=/usr/local/bin/openssl   acme.sh --issue -d xxxx.com ......
```

`acme.sh` can remember `OPENSSL_BIN` position automatically for future use.

ksperling:
Thanks for your response. I understand that you don't want to have a hard dependency on python, but for OS X specifically (which is pretty widely used among developers) being able to work with the tools shipped with the operating system is a major plus in my mind.

In fact that's one of the primary things that made me pick acme.sh over a bunch of other clients, as I can just have a copy or submodule of it in our configuration repository, and any developer can just run it in-place without having to bother with installing Ruby gems, python modules, a Go runtime, or indeed a custom OpenSSL.

The code specifically tries to use the modern OpenSSL options first, and only falls back to python as a last resort. I'm happy to update the code to only try this fallback on OS X if that would make the workaround acceptable to be merged.

Neilpang:
Hi @ksperling ,
Sorry, but I really want to keep the code pure in sh.

I knew that the default openssl on MacOS is v0.9,  which doesn't support some modern features.
But developers should know how to update openssl to the latest version,  which brings much more benefits than just making acme.sh working.
  
If you don't want to upgrade the system-wide openssl version, you can just use `OPENSSL_BIN` to specify a custom/private  openssl executable to work. You need only specify it once, `acme.sh` can remember it automatically . And it will use it in any future use without specifying anymore.

I think that's is enough.  Users should not keep using the old version/buggy openssl anymore.

ksperling:
Thanks for your detailed response.

I've put my workaround code into an `openssl-compat` script now that I'm passing to `acme.sh` as `OPENSSL_BIN`. Just posting it here in case someone runs across this thread looking for a similar solution.

```
#!/bin/bash
# Wrapper for OpenSSL that implements the 'dgst' command with a binary key
# (specified in hex) on OpenSSL v0.9 where it is not natively available.

if [ "$1" = dgst ]; then
  # Parse a 'dgst' command like 'dgst -$ALG -mac HMAC -macopt hexkey:$KEY [-binary]'
  args=("$@") compat=1 alg= mac= hexkey= output=hexdigest
  shift
  while [[ "$compat" -gt 0 && $# -gt 0 ]]; do
    case "$1" in
      -binary) output=digest;;
      -sha1|-sha224|-sha256|-sha384|-sha512) alg="${1:1}";;
      -mac) mac="$2"; shift;;
      -macopt) [[ "$2" =~ ^hexkey:([0-9a-fA-F]+)$ ]] && hexkey="${BASH_REMATCH[1]}" || compat=0; shift;;
      *) compat=0;;
    esac
    shift
  done
  set -- "${args[@]}" # restore

  # If we can emulate check for native OpenSSL support first anyway
  if [[ "$compat" -gt 0 && -n "$alg" && -n "$hexkey" && "$mac" = HMAC ]]; then
    if ! openssl dgst -"$alg" -mac HMAC -macopt hexkey:00 </dev/null >/dev/null 2>&1; then
      exec python -u -c 'import sys,binascii,hmac,hashlib;(alg,key,output)=sys.argv[1:];sys.stdout.write(
        ("(stdin)= " if output=="hexdigest" else "") +
        getattr(hmac.new(binascii.unhexlify(key),sys.stdin.read(),getattr(hashlib,alg)),output)() +
        ("\n" if output=="hexdigest" else ""))' \
        "$alg" "$hexkey" "$output"
    fi
  fi
fi

exec openssl "$@"
```

Neilpang:
@ksperling 

Thank you.

Neilpang:
Hi @ksperling 
Sorry, but this is fixed  at c70432996a70c82bdb4af526496bbc9b33408ab3
It falls back to the old 0.9 way to calc hmac.  Please try again.

But,  `OPENSSL_BIN` is renamed to `ACME_OPENSSL_BIN`  at 851fedf7512216c1926f3f64e899167ed6f06f56

Please take care.

ksperling:
Hi Neil,

I tried a workaround like that previously, before I went down the Python route, but the problem is that it doesn't work correctly when any byte in the HMAC secret is a 0. Those bytes just get silently dropped by the shell when it's doing the command substitution (reading back the output of the `_h2b`)

Especially in the AWS case where HMAC is applied something like 4 times (using the output from the previous step as the key for the next) this is quite likely to happen in practice, and will just end up generating incorrect signatures in that case.

```
$ echo -n hello | ./openssl-compat dgst -"sha256" -mac HMAC -macopt "hexkey:00112233" | cut -d = -f 2 | tr -d ' '
1e686202515f109116d87cbb392c3510190848aa308b509e30691859160ec8aa

$ echo -n hello | openssl dgst -"sha256" -hmac "$(printf "%s" "00112233" | _h2b)" | cut -d = -f 2 | tr -d ' '
76d62566ef589706fdc6bcce3e84d68134d3385645b2bed86a8f0d71d9835a05

$ echo -n hello | openssl dgst -"sha256" -hmac "$(printf "%s" "112233" | _h2b)" | cut -d = -f 2 | tr -d ' '
76d62566ef589706fdc6bcce3e84d68134d3385645b2bed86a8f0d71d9835a05

$ echo -n hello | ./openssl-compat dgst -"sha256" -mac HMAC -macopt "hexkey:112233" | cut -d = -f 2 | tr -d ' '
76d62566ef589706fdc6bcce3e84d68134d3385645b2bed86a8f0d71d9835a05
```

Neilpang:
@ksperling 
Yes, it was a known issue. if the secret contains '\0'.

It's  a shell limitation.  The string can not contain "\0".

Please update your openssl version,  or  use `--openssl-bin` to specify a custom new version openssl to use.

ksperling:
Just ran into the issue again... I guess I hadn't remembered to change OPENSSL_BIN to ACME_OPENSSL_BIN, and everything was working fine for a while, until suddenly I was getting signature errors, and just spent a couple of hours debugging and then remembered this issue.

It's really problematic if the `_hmac` function silently produces incorrect results some of the time (when the key contains a binary 0). It's fine if you don't want to support old OpenSSL etc, but at least there should be an actual error reported, rather than just silently producing invalid hashes.

The simplest fix would be just removing the fallback to `openssl -hmac` if `openssl -mac HMAC -macopt` is unsupported. The other alternative would be to keep the fallback, but actually check for null bytes and fail, rather than producing an invalid hmac. That way the AWS signatures would still only sometimes work, but at least it would be obvious why.

PR #520, opened 2017-01-09
dnsapi/dns_lexicon.sh: shellcheck: fix 4 occurences of SC2021:

bittorf:
"Don't use [] around ranges in tr, it replaces literal square brackets."

this introduces another warning:
"Use '[:lower:]' to support accents and foreign alphabets."

This is more a style thingy because we really want to only catch A-Z.
work around this by using a shellcheck-directive and a comment
that the [:lower:] will not work with e.g. busybox-ash.

if we later really want to use [:lower:], we should use 'sed' for that.


Neilpang:
Thank you.
Please send the pull request to `dev` branch, instead of `master` branch.

bittorf:
@Neilpang - sorry, will resend to 'dev

bittorf:
@Neilpang - looks strange: does it work? https://github.com/Neilpang/acme.sh/pull/521

PR #522, opened 2017-01-09
dnsapi/dns_dp.sh: shellcheck: fix 1 occurence of SC2126

bittorf:
message:
"Consider using grep -c instead of grep | wc"

### DIFF
--- a/dnsapi/dns_dp.sh
+++ b/dnsapi/dns_dp.sh
@@ -199,15 +199,16 @@ _get_root() {
 
 #Usage: method  URI  data
 _rest() {
-  m=$1
+  m="$1"	# method e.g. GET|POST|PUT|DELETE
   ep="$2"
   data="$3"
+
   _debug "$ep"
   url="$REST_API/$ep"
 
   _debug url "$url"
 
-  if [ "$data" ]; then
+  if [ "$m" != 'GET' -a "$data" ]; then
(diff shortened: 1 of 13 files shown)


dnsapi/dns_dp.sh
```diff
@@ -199,7 +199,7 @@ _get_root() {
 
 #Usage: method  URI  data
 _rest() {
-  m=$1
+  export m=$1		# e.g. GET|POST
```
Neilpang: this is local variable, do not use export

bittorf:
i do not understand the build/check-error: it bails out in line 223:

<code>
The command "if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then git diff --exit-code && echo "shfmt OK" ; fi" exited with 1.
</code>

dnsapi/dns_dp.sh
bittorf: this is not true: check it again: $m is just not used in this function, but in others...

bittorf: ok, i see:

if [ "$data" ]; then ...

maybe we should do "if [ "$m" = 'POST' ]; then ...
otherwise the var is really not used. you are right, this var should not be exportet

bittorf:
now i understand - i have to rebase - sorry about that

PR #526, opened 2017-01-09
shellcheck: fix several occurences of SC2034

bittorf:
message:
SC2034: $VARNAME appears unused. Verify it or export it.

most of these are related to the style:
we generate global vars, which are used in other functions.

the var "lexicon_url" was really unused (left it as comment)

the travis-check now enforces a POSIX-shell (dash).
this makes it now possible to use 'local var=foo' in functions.

Signed-off-by: Bastian Bittorf <bb@npl.de>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ~/shfmt -l -w -i 2 . ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then git diff --exit-code && echo "shfmt OK" ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck -V ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck -e SC2034 **/*.sh && echo "shellcheck OK" ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck **/*.sh && echo "shellcheck OK" ; fi
   - cd ..
   - git clone https://github.com/Neilpang/acmetest.git && cp -r acme.sh acmetest/ && cd acmetest
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]] && [[ "$NGROK_TOKEN" ]]; then sudo NGROK_TOKEN="$NGROK_TOKEN" ./letest.sh ; fi
--- a/dnsapi/dns_ad.sh
+++ b/dnsapi/dns_ad.sh
@@ -127,8 +127,8 @@ _ad_rest() {
   _debug mtd "$mtd"
   _debug ep "$ep"
 
-  _H1="Accept: application/json"
-  _H2="Content-Type: application/json"
+  export _H1="Accept: application/json"
+  export _H2="Content-Type: application/json"
 
   if [ "$mtd" != "GET" ]; then
     # both POST and DELETE.
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -137,13 +137,13 @@ aws_rest() {
 
   #RequestDate="20161120T141056Z" ##############
 
-  _H1="x-amz-date: $RequestDate"
+  export _H1="x-amz-date: $RequestDate"
 
   aws_host="$AWS_HOST"
   CanonicalHeaders="host:$aws_host\nx-amz-date:$RequestDate\n"
   SignedHeaders="host;x-amz-date"
   if [ -n "$AWS_SESSION_TOKEN" ]; then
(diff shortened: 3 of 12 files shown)


bittorf:
i'am sorry - why do the tests fail?

bittorf:
anything I can help with? still questions? 8-)

.travis.yml
```diff
@@ -41,7 +41,7 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ~/shfmt -l -w -i 2 . ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then git diff --exit-code && echo "shfmt OK" ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck -V ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck -e SC2034 **/*.sh && echo "shellcheck OK" ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck --shell=dash **/*.sh && echo "shellcheck OK" ; fi
```
Neilpang: why is this change for ?

bittorf: this ensures, that we test a *POSIX* shell (so the same like .sh) but with the addition, that the keyword 'local' for e.g. 'local var=value' is allowed. This will make it possible in later commits to actually *use* that keyword without shellcheck-erros if we want it (at least I want it). fun fact: there seems to be no shell in the wild which aims to be POSIX but can not work with 'local'.

Neilpang: I remember it can check the `local` error without `-shell=dash`

bittorf: it is the other way around: if you use 'local' without this patch, shellcheck emits a warning, that 'local' is not allowed in POSIX shells.

Neilpang: NO, we can not use `local` keyword.  That is why I never used one `local` in the code.

`sh` on openBSD doesn't support `local`

Neilpang: we must keep  POSIX  compatible

bittorf: i strongly support beeing POSIX compatible.
when reading https://www.freebsd.org/doc/en/articles/linux-users/shells.html - are you talking about "tcsh" or "sh" ?

bittorf: i just checked: even openBSD1.0 supports 'local' in shells (with ash)

bittorf: and 'tcsh' does not even support functions 8-)

bittorf: i just asked in the 'openbsd' irc channel and they report that the default shell is 'ksh' and it does indeed support 'local'...

bittorf: one guy explained to me, that openbsd 3.8 used another shell ("csh"), but that was 2005 or earlier...also they say we should better use the shebang #!/bin/sh

Neilpang: I just tested on latest openBSD 5.6, yes, `local` is supported.  
I also tested in latest solaris,   but the `local` keyword is not supported in the default `sh`.

Anyway,  I won't add anything that is not in POSIX standard.

bittorf: thanks for you feedback...i changed the pullrequest, so that we use now a strict posix shell again. we really need a better testinfrastructure for automated tests on ALL these targets 8-)

.travis.yml
```diff
@@ -41,7 +41,7 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ~/shfmt -l -w -i 2 . ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then git diff --exit-code && echo "shfmt OK" ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck -V ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck -e SC2034 **/*.sh && echo "shellcheck OK" ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck --shell=sh **/*.sh && echo "shellcheck OK" ; fi
```
Neilpang: do not use `--shell` parameter here.

.travis.yml
Neilpang: I have set up all the VMs, but it's for my private use only.
https://github.com/Neilpang/acmetest

.travis.yml
bittorf: is'nt it a good idea to _enforce_ testing a POSIX shell instead of relying parsing the shebang? think about e.g. UTF8 things and just mistakes during checkin!

Neilpang: I think by the shebang is enough.
I don't like any extra parameter in the command line if not necessary.

Neilpang: I even don't think `SC2034 ` is not necessary to fix.
But whatever.

bittorf: ok, updateted the pullrequest and make the travis-call without any specials.

PR #571, opened 2017-02-01
Add FreeDNS plugin support to acme.sh

dkerr64:
New pull request to replace pull request 557 with a clean branch without all the commits to clean up the code

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ You don't have to do anything manually!
 1. ISPConfig 3.1 API
 1. Alwaysdata.com API
 1. Linode.com API
-1. FreeDNS
+1. FreeDNS (https://freedns.afraid.org/)
 
 **More APIs coming soon...**
 
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ You don't have to do anything manually!
 1. ISPConfig 3.1 API
 1. Alwaysdata.com API
 1. Linode.com API
+1. FreeDNS
 
 **More APIs coming soon...**
 
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -301,6 +301,10 @@ Now you can issue a certificate.
 acme.sh --issue --dns dns_freedns -d example.com -d www.example.com
 ```
 
+Note that you cannot use acme.sh automatic DNS validation for FreeDNS public domains or for a subdomain that
+you create under a FreeDNS public domain.  You must own the top level domain in order to automaitcally
+validate with acme.sh at FreeDNS.
+
 # Use custom API
 
 If your API is not supported yet, you can write your own DNS API.
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -278,6 +278,32 @@ acme.sh --issue --dns dns_linode --dnssleep 900 -d example.com -d www.example.co
 
 The `LINODE_API_KEY` will be saved in `~/.acme.sh/account.conf` and will be reused when needed.
 
+# 15.  Use FreeDNS
+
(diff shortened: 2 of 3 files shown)


Neilpang:
Please update the readme.md files with the usage of freedns.

Neilpang:
Please fix the error:  https://travis-ci.org/Neilpang/acme.sh/jobs/197461974

README.md
```diff
@@ -270,6 +270,7 @@ You don't have to do anything manually!
 1. ISPConfig 3.1 API
 1. Alwaysdata.com API
 1. Linode.com API
+1. FreeDNS
```
Neilpang: change to:
```
FreeDNS(https://freedns.afraid.org/)
```

dnsapi/README.md
```diff
@@ -278,6 +278,32 @@ acme.sh --issue --dns dns_linode --dnssleep 900 -d example.com -d www.example.co
...
+reuses that when needed.
+
+Now you can issue a certificate.
+
+```sh
+acme.sh --issue --dns dns_freedns --dnssleep 30 -d example.com -d www.example.com
+```
+
+FreeDNS updates records quite quickly so it is possible to reduce the dnssleep time, in the above example
+to 30 seconds.
```
Neilpang: I don't think explicitly using `30` seconds is a good idea.
I would prefer to use the default value.

README.md
dkerr64: Done

dnsapi/README.md
dkerr64: Done

Neilpang:
I just tried it, but It failed to add txt record.
can you please try again ?

Neilpang:
Another question,  why you save the cookies, instead of saving the username and password ?
How long will the cookies be valid ?

dkerr64:
It is working for me.... just tried twice now.  Was any error message issued?  if you run with --debug 2 then all HTML returned will be printed out too.

As for cookie... according to Joshua Anderson (admin for FreeDNS whom I consulted when creating this) this value will change only when password changes.  He recommended storing that and not the userid/password.... it is not necessary to login and obtain the cookie unless the password changes.

David

dkerr64:
Thoughts on why it might fail for you... I'm guessing that you are in china.  It is possible that the locale / language may be detected by the freeDNS website and different HTML returned.  It may be necessary to force to english/usa in the HTTP header sent.  Just a thought... would be helpful if you could determine if this is right.

David

Neilpang:
Please use a new domain to test. I guest you have cache.

dkerr64:
Are you testing with one of FreeDNS public domains or your own private domain?  FreeDNS public domains will not work for various reasons... 1) rate limiting at LetsEncrypt (there is much discussion of this at LetsEncrypt community pages) and 2) FreeDNS uses a captcha image for public domains to prevent bots creating lots of subdomains under free public domains.

I have added a test for failures to add TXT record because of the captcha/security code and issue an error message advising that public domains will not work.  And I have updated the README.md

If you are failing with a public domain please run with --debug 2 and send me the output. 

And yes, I am testing with new domains, no caching going on.

I have also added Accept-Language:en-US to the HTTP headers as a precaution agains future multi-lingual FreeDNS pages.

David

Neilpang:
Yes,  I was testing with the public free domains.

MukilanP:
@Neilpang

How did you skip the captcha issue while testing with public free domains. 

hu Dec 22 13:48:37 STD 2022] _ret='0'
[Thu Dec 22 13:48:37 STD 2022] FreeDNS failed to add TXT record for _acme-challenge.acmetest as FreeDNS requested security code
[Thu Dec 22 13:48:37 STD 2022] Note that you cannot use automatic DNS validation for FreeDNS public domains

PR #577, opened 2017-02-04
Add --syslog feature to enable logging to system logs.

dkerr64:
adds --syslog [level]

level may be one of error, info, debug, debug2 or debug3
All debug levels are merged into syslog "debug" level
If level not provided, defaults to error.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -136,7 +136,7 @@ _log() {
 _info() {
   _log "$@"
   if [ "$SYSLOG_LEVEL" -ge "$SYSLOG_INFO" ]; then
-    logger -i -t "acme.sh" -p "user.info"  "$@" >/dev/null 2>&1
+    logger -i -t "acme.sh" -p "user.info" "$@" >/dev/null 2>&1
   fi
   _printargs "$@"
 }
@@ -144,7 +144,7 @@ _info() {
 _err() {
   _log "$@"
   if [ "$SYSLOG_LEVEL" -ge "$SYSLOG_ERROR" ]; then
-    logger -i -t "acme.sh" -p "user.error"  "$@" >/dev/null 2>&1
+    logger -i -t "acme.sh" -p "user.error" "$@" >/dev/null 2>&1
   fi
   if [ -z "$NO_TIMESTAMP" ] || [ "$NO_TIMESTAMP" = "0" ]; then
     printf -- "%s" "[$(date)] " >&2
@@ -168,7 +168,7 @@ _debug() {
     _log "$@"
   fi
   if [ "$SYSLOG_LEVEL" -ge "$SYSLOG_DEBUG" ]; then
-    logger -i -t "acme.sh" -p "user.debug"  "$@" >/dev/null 2>&1
+    logger -i -t "acme.sh" -p "user.debug" "$@" >/dev/null 2>&1
   fi
   if [ -z "$DEBUG" ]; then
     return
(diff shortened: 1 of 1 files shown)


dkerr64:
I implemented this using the "logger" command.  I don't know if this is portable across all platforms you support.

dkerr64:
I do not save the syslog level in account.conf.  I considered this but decided that it should not be saved.  If saved then we would need a mechanism to "un-save" it or to change the level.  I think it is best to explicitly require it every time acme.sh is run somewhat like debug.

Maybe it could be saved only if done at the same time as --install or --installcronjob that might make sense.  But with cron, it could be specified in the crontab file rather than in .conf.

Thoughts?

Neilpang:
I just added a new branch `syslog`:  https://github.com/Neilpang/acme.sh/tree/syslog

At revision:  https://github.com/Neilpang/acme.sh/commit/62b6bda424c0f8b5aabb650f698bdbb8673b4e3a

Usage:
```
#enable syslog
--syslog

#or
--syslog 1

#disable syslog
--syslog 0
```

Please have a try, if there is no problem, I will merge it soon.

Neilpang:
we must save the `--syslog` value in account conf for future use and cron job use.

dkerr64:
Please study the version I proposed.  Your implementation is all or nothing... that is if syslog=1 it will log errors, info and debug.  This is bad... debug is very very verbose and should not be entered into syslog by default.  The default should be errors only.

Please provide a way to indicate to what level syslog should be done...  that is permit user to specify error, info or debug (and maybe debug2/3).

Also as I said above I don't think syslog level should be stored in account.conf every time it is used on command line.  I do think it appropriate to store on --install or --installcronjob.

David

Neilpang:
the debug info is not logged into syslog by default.
It only loggs  info and error.

dkerr64:
but if I run acme.sh with --debug flag then all debug messages go to syslog.  This is not right, what gets logged to syslog should be managed independently of the other --log or --debug levels. 

I strongly believe that you have implemented this wrong, syslog level should be controlled on its own as my original pull request proposed.  System admins hate clutter in syslog and want the flexibility to control what goes in there.

Strongly urge that you reconsider.

Neilpang:
`--debug` is just a temp state,  not a persist state.
I can remove the debug message from the syslog.

We can just record the info and error messages in the syslog.

System admins can change the config of syslog.conf to determine whether they want to hear the "user.info" message or "user.error" message, or both of them.

But from the internal point of view, we should just honestly send out any messages we have, to report the internal status.

 If there is any problems happen, the sys admin can review the info message and/or the error message.

Just imagine that:  if we just send error message out,  the sys admin would only get one error message, something like: "issue cert error ... bla bla bla.",  that doesn't help at all for checking/fixing the real problem.

On the other hand, if we record the whole process(both the info messages and error message, or even with debug message),   we can know what was really happening when the problem was happening.

That's what the word "log" means.  otherwise, it should be called as "sysalert/syserror", instead of "syslog".

dkerr64:
Debug should not go to syslog by default.... even with --debug enabled.  As you said it is temporary state, but when debugging you can be repeating app runs many times and debug can be very noisy... dumping a lot of text which I don't think should go to syslog.  Besides, when debugging you don't really bother to look at syslog and it is not necessary to maintain a persistent log.

If you don't want to provide granularity then just do error and Info.

David

PR #586, opened 2017-02-05
Ssh deploy

dkerr64:
Script to deploy certificates to a remote server by SSH.  Note that SSH must be able to login to remote host without a password... SSH Keys must have been exchanged with the remote host.  Validate and test that you can login to USER@URL from the host running acme.sh before using this script.

Validate and test that the USER@URL  at the remote server has permissions to execute any commands (e.g. to stop/start services) and write access to any certificate files required.

The following variables exported from environment will be used.  If not set then values previously saved in domain.conf file are used.  The server URL may also be obtained from parameter passed into this script if not overridden by the .conf file or exported variable.  SSH port defaults to 22 unless another value passed in.

Only a username is required.  All others are optional.... but in practice something must be set otherwise nothing will get deployed !

The following examples are for QNAP NAS running QTS 4.2.  Change to match your target server / service.  These are listed in the order in which they are used (command executed / file copied).

export ACME_DEPLOY_SSH_USER="admin"
export ACME_DEPLOY_SSH_SERVER="qnap"
export ACME_DEPLOY_SSH_PORT="22"
export ACME_DEPLOY_SSH_SERVICE_STOP=""
export ACME_DEPLOY_SSH_KEYFILE="/etc/stunnel/stunnel.pem"
export ACME_DEPLOY_SSH_CERTFILE="/etc/stunnel/stunnel.pem"
export ACME_DEPLOY_SSH_CAFILE="/etc/stunnel/uca.pem"
export ACME_DEPLOY_SSH_FULLCHAIN=""
export ACME_DEPLOY_SSH_REMOTE_CMD="/etc/init.d/stunnel.sh restart"
export ACME_DEPLOY_SSH_SERVICE_START=""

Commands, directories, paths, all refer to those on the remote server to which certificates are to be deployed.  Not for the host on which acme.sh is running.


deploy/sshdeploy.sh
```diff
@@ -0,0 +1,196 @@
...
+# export ACME_DEPLOY_SSH_CERTFILE="/etc/stunnel/stunnel.pem"
+# export ACME_DEPLOY_SSH_CAFILE="/etc/stunnel/uca.pem"
+# export ACME_DEPLOY_SSH_FULLCHAIN=""
+# export ACME_DEPLOY_SSH_REMOTE_CMD="/etc/init.d/stunnel.sh restart"
+# export ACME_DEPLOY_SSH_SERVICE_START=""
+
+########  Public functions #####################
+
+#domain keyfile certfile cafile fullchain
+sshdeploy_deploy() {
```
Neilpang: rename this hook to `ssh`,  instead of `sshdeloy`

deploy/sshdeploy.sh
```diff
@@ -0,0 +1,196 @@
...
+# using this script.
+#
+# The following variables exported from environment will be used.
+# If not set then values previously saved in domain.conf file are used.
+#
+# Only a username is required.  All others are optional.
+#
+# The following examples are for QNAP NAS running QTS 4.2 
+# export ACME_DEPLOY_SSH_USER="admin"
+# export ACME_DEPLOY_SSH_URL="qnap"
```
Neilpang: what is a "ACME_DEPLOY_SSH_URL" ?

I think we need a "ssh server" and a "ssh port" .

deploy/sshdeploy.sh
```diff
@@ -0,0 +1,196 @@
...
+sshdeploy_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+  _cmdstr=""
+  _homedir='~'
+  _backupprefix="$_homedir/.acme_ssh_deploy/certs-backup"
+  _backupdir="$_backupprefix-$(date +%Y%m%d%H%M%S)"
```
Neilpang: I just added a new function: `_utc_date`
https://github.com/Neilpang/acme.sh/commit/5d2c5b01a80c2cdc555673926e7b865e33297ddd

Please use that to compose your suffix.

deploy/sshdeploy.sh
```diff
@@ -0,0 +1,196 @@
...
+
+  if [ -f "$DOMAIN_CONF" ]; then
+    # shellcheck disable=SC1090
+    . "$DOMAIN_CONF"
```
Neilpang: why do you need to source the conf here ?

You should not source the conf here.

deploy/sshdeploy.sh
```diff
@@ -0,0 +1,196 @@
...
+    _cmdstr="mkdir -p $_backupdir ; $_cmdstr"
+    _info "Backup of old certificate files will be placed in remote directory $_backupdir"
+    _info "Backup directories erased after 180 days."
+  fi
+
+  _debug "Remote commands to execute: $_cmdstr"
+  _info "Submitting sequence of commands to remote server by ssh"
+  # quotations in bash cmd below intended.  Squash travis spellcheck error
+  # shellcheck disable=SC2029
+  ssh -T "$Le_Deploy_ssh_user@$Le_Deploy_ssh_url" bash -c "'$_cmdstr'"
```
Neilpang: please use `sh -c` instead of `bash -c`.  we can not assume that the target machine has bash installed.

deploy/sshdeploy.sh
```diff
@@ -0,0 +1,196 @@
...
+
+  return 0
```
Neilpang: remove this `return 0`,  we should just use the return status of the ssh command.

deploy/sshdeploy.sh
dkerr64: I'm not sure.  The reason I did not name it ssh is because the script is not deploying certificates for use by ssh, rather ssh is the tool we use to deploy the certificates to a remote server.  Hence why I called it sshdeploy.   I'll change it if you want, but I'm concerned that it would be misleading.

deploy/sshdeploy.sh
dkerr64: Yup, you are right.... I should have a server and port.  I'll change.

deploy/sshdeploy.sh
dkerr64: I am not able to access values stored in the domain.conf file unless I do this.  The only alternative I could see would be to call _readdomainconf() function for each value... doable but much less efficient.  Please tell me which you would like me to do.

deploy/sshdeploy.sh
dkerr64: Done...  but your _utc_date() function returns a space between the date and the time which is not ideal for a filename... not supported by all file systems.  I had to pipe through tr to convert the space to a hyphen.

deploy/sshdeploy.sh
dkerr64: Done.  But we are also dependent on other commands being available at the target machine.  hopefully all should be okay....  cp echo sleep find xargs mkdir
If not then I think that just becomes a restriction... or eliminate the certificate backup feature, that would get rid of requirement for cp find xargs and mkdir.

deploy/sshdeploy.sh
dkerr64: Done

deploy/sshdeploy.sh
Neilpang: yes, please change it.
when the user uses this hook, it should be look like:
```
acme.sh --deploy --deploy-hook  ssh  -d ......
```

This is obvious enough that it's for deploy only.

dkerr64: Done

deploy/ssh.sh
```diff
@@ -0,0 +1,205 @@
...
+    _info "Will start remote service with command $Le_Deploy_ssh_remote_cmd"
+  fi
+
+  if [ -z "$_cmdstr" ]; then
+    _err "No remote commands to excute. Failed to deploy certificates to remote server"
+    return 1
+  else
+    # something to execute.
+    # run cleanup on the backup directory, erase all older than 180 days.
+    _cmdstr="find $_backupprefix* -type d -mtime +180 2>/dev/null | xargs rm -rf ; $_cmdstr"
```
Neilpang: do not use `find` and `xargs` here.
please use for:
```
for file  in $_backupprefix*  ; do
    #get the datetime from the filename
    #check and remove the file if it's too old.
done
```

It's better if we as less command as we can.

deploy/ssh.sh
```diff
@@ -0,0 +1,205 @@
...
+# export ACME_DEPLOY_SSH_USER="admin"
+# export ACME_DEPLOY_SSH_SERVER="qnap"
+# export ACME_DEPLOY_SSH_PORT="22"
+# export ACME_DEPLOY_SSH_SERVICE_STOP=""
+# export ACME_DEPLOY_SSH_KEYFILE="/etc/stunnel/stunnel.pem"
+# export ACME_DEPLOY_SSH_CERTFILE="/etc/stunnel/stunnel.pem"
+# export ACME_DEPLOY_SSH_CAFILE="/etc/stunnel/uca.pem"
+# export ACME_DEPLOY_SSH_FULLCHAIN=""
+# export ACME_DEPLOY_SSH_REMOTE_CMD="/etc/init.d/stunnel.sh restart"
+# export ACME_DEPLOY_SSH_SERVICE_START=""
```
Neilpang: what is this "ACME_DEPLOY_SSH_SERVICE_START" for ?

I think only "ACME_DEPLOY_SSH_REMOTE_CMD" is enough to restart service is enough.

dkerr64: There are some applications/services that require two command to execute when certificates change.  For example for the Unifi Controller you must first import the certificates into a keystore and only then restart the service.  I therefore thought it easier to provide for two separate commands that could be executed after the certificates were in place.

deploy/ssh.sh
dkerr64: Remember that find/xargs/rm are commands that execute on the remote sever, not the acme.sh host.   I think we should keep it as simple as possible... packing in for loops makes it more complicated.
If we think there is a risk that find/xargs might not be available at the remote server then I would rather document that as a restriction and provide the option to disable all backup so that the rest of the script can still be used.

Neilpang: Yes, I knew it's run on the remote server.  but it would be better if we depend on less commands.

dkerr64: I have investigated and I still don't like it.  All we do is substitute two other commands in place of find and xargs.... date and expr and it becomes much much more complex.  This is what I came up with...
```
now=$(date -u +%s);
for fn in $_backupprefix*; do
  if [ "$(expr $(expr $now - $(date -ur $fn +%s)) / 86400)" -ge "180" ]; then
    rm -rf $fn
  fi
done
```
I don't know how portable this is.  And I am not testing for directories only.  And I am not handling the case where there are no files matching $_backupprefix* (in which case fn is set to $_backupprefix* which is not a valid filename / directory) so that would add additional complication.  I could simplify the math by doing -ge "15552000" (180 days * 86000 seconds) but that is crystal clear, right!

I much prefer the find/xargs version and feel like it is perfectly okay to document the requirements for the remote host.

Neilpang: No,  we should not use any more command if it's not really necessary.
You never know what the remove system is, and what command is missing from the remote system.

Let me show you an example.

I recently add a `_hex_dump()` function, first at https://github.com/Neilpang/acme.sh/commit/5f8daeeb6d987687ac896c128c327c2096b02009

It was firstly implemented by `od` command.   Since `od` is one the core utils of GNU, and I tested all the platforms we have here: https://github.com/Neilpang/acmetest,  including solaris, freebsd and Mac.

I thought it should be safe.
But soon, it gave me a bug:
https://github.com/Neilpang/acme.sh/issues/579

It was in a router, the `od` command is missing. unbelievable.

And soon again, another bug:
https://github.com/Neilpang/acme.sh/issues/595

It was in busybox, the `od` command is also missing.

I fixed a lot, and now the current version is:  https://github.com/Neilpang/acme.sh/pull/600/commits/4e4a6d83973fa0580ef8b2f1d7d0312fdd671543

I never thought the result when I first started function.


So, how about `xargs/find` ?   do you think it's more portable than `od` ?  I don't believe.

So, never use external command if not necessary.  That would make our code more safe and portable.


As for you script, please see: https://github.com/Neilpang/acme.sh/blob/master/acme.sh#L3343

we already have such `for/basename/_math/[ -ge ]` combination.

deploy/ssh.sh
Neilpang: I think one parameter something like: `ACME_DEPLOY_SSH_SERVICE_RELOAD` is enough.

IMOP, the current ssh deploy hook has too many parameters already,  which would make us too complicated for the user to use.
 
I was thinking about refactoring the parameters to make it easier to learn and use.

Neilpang:
It's not final,  but I have some thoughts about the parameters.

Basic parameters:
```sh
ssh_user
ssh_server

ssh_fullchain_path
ssh_key_path

ssh_reload_cmd
```

Advanced parameters:
```sh
ssh_cmd

ssh_cert_path
ssh_ca_path
```

The above `ssh_cmd` parameter is a way the user can customize the ssh port or ssh identity file.
For example:

customize the ssh port
```sh
export  ssh_cmd="ssh -p 1022"
acme.sh --deploy --deploy-hook ssh   -d .......
``

customize the ssh identity key file:
```sh
export  ssh_cmd="ssh -i /path/to/id_rsa"
acme.sh --deploy --deploy-hook ssh   -d .......
```

It's even possible for the user to use `sshpass`
```sh
export  ssh_cmd="sshpass -p mysshpassword  ssh"
acme.sh --deploy --deploy-hook ssh   -d .......
```

The above names "ssh_user" , "ssh_server" or "ssh_cmd" etc are just examples to illustrate,  not the real names.

what do you think ?

deploy/ssh.sh
dkerr64: I am open to suggestions on how to simplify.  It is possible to chain multiple remote commands into a single variable.... I have tested with cmd1 && cmd2 and it worked.  In the case of Unifi Controller I would need to openssl && keytool && service start so three commands would be chained.

The openssl btw is to create a pkcs12, something that acme.sh can do but then that is a binary file which is not easy to send over to the remote server by ssh, so I generate the pkcs12 at the remote end.

deploy/ssh.sh
dkerr64: No I don't know if find/xargs is any more portable.  I'm just saying that it is not unreasonable to make it a requirement that they exist on the remote host for the purpose of backing up the prior certificates on the remote host. 

I see two options...
1. Keep find/xargs and add a option to disable backup if it fails.
2. Implement something like the script I came up with that uses date/expr.  And maybe also implement an option to disable backup if even this causes problems

dkerr64:
Adding a ssh_cmd like you describe certainly greatly expands the flexibility of how ssh can be invoked. 

ssh_server can be optional... will use -d domain.com as provided on command line.

deploy/ssh.sh
Neilpang: acme.sh  has a command `--toPkcs` to convert cert to pkcs12 format.

deploy/ssh.sh
Neilpang: It's better if we require less but support more.
Please do not use xargs/find if we can use other simpler code there.

deploy/ssh.sh
dkerr64: yes, but how would I transfer that pkcs12 file over to a remote server by ssh... it is binary not ascii like the key/cert files.  So, simpler I thought to generate the pcks12 at the far end.

Neilpang: Here is how to transfer the binary file to remote server:

```sh
cat  cert.p12 | ssh  user@server "cat -  > cert.p12"
```

deploy/ssh.sh
dkerr64: I will change to use date/expr.

Neilpang:
I just added  "DOMAIN_BACKUP_PATH" for you.  at  https://github.com/Neilpang/acme.sh/commit/fd72cced13e37c3a5f877e14825bac2d6c455a2a

Please backup the cert and key to local path, install of on the remote server.  that could simplify your code.

dkerr64:
I cannot backup certificates on a remote sever into a location on the local host (where acme.sh is running).  That would require copying the old certificates back from remote server to local host and I don't want to tackle that.

dkerr64:
I think that this is ready to merge.  I do not have any further updates planned.

dkerr64:
It would be good to get this merged soon as it looks like you are about to start a lot of work on deploy scripts.... I'd rather not have to do a lot of manual merging.   Thanks.
David

deploy/README.md
```diff
@@ -24,7 +24,144 @@ Before you can deploy your cert, you must [issue the cert first](https://github.
...
+environment variables.  This is not required for subsequent runs as the
+values are stored by acme.sh in the domain configuration files.
+
+Required...
+```
+export ACME_DEPLOY_SSH_USER=username
+```
+Optional...
+```
+export ACME_DEPLOY_SSH_CMD=custom ssh command
```
Neilpang: Please remove all the prefix `ACME_*` : 
```
ACME_DEPLOY_SSH_CMD  =>  DEPLOY_SSH_CMD
```

dkerr64: Okay, I can do that.  Will work on it as soon as I can.

dkerr64: Done.

dkerr64:
Would be good to get this merged.... once again I had to resolve a conflict when merging upstream (albeit a simple one in readme).

jasonkeller:
Is this dead?  Really looking forward to this functionality (have redundant load balancers in my lab)

dkerr64:
I'd like this merged too.  It has been stable now for a couple of months and my own LetsEncrypt certificates happily auto renewed by cron job and deployed to two separate servers successfully with this.  Would really like @Neilpang to merge this so we can pull from master into [AstLinux](http://www.astlinux-project.org/)

Neilpang:
@jasonkeller  @dkerr64 

Sorry for the delay,  but there needs to be much more design in this hook.
The variable name and how to save/change the variable value, etc.

I was too busy recently,  but I'm thinking about some refactoring in this hook.

I will update on this PR soon.
Thanks.

dkerr64:
@Neilpang happy to work with you on modifications to the design.  Changes to variable values should just be as simple as export XXX a new value, that would then get picked up and saved into the domain conf file.  Names of the variables can easily be changed to conform to naming standards for other hooks.

The bigger issue I have identified is what if you want to deploy the same certificate to multiple servers... say you are load balancing across servers, you need to get the same certificate for the same domain name deployed to multiple places.  There is no way to do this today -- but solving that I think would be at a higher level than this deploy hook.

FernandoMiguel:
@dkerr64 technically, servers can each have their own cert. nothing requires all of them to have the same cert.

Neilpang:
@FernandoMiguel 

I think what @dkerr64  means is to deploy the same cert to different servers via ssh.

All the config of the hook is global now,  so it's impossible to use ssh deploy hook with 2 different target servers.

This is not a problem of ssh hook,  but is a common problem in all of our hooks/apis.

I was thinking about adding a new parameter to solve it, but have not a clear design yet.

dkerr64:
Nudge Nudge.  Just wondering what the status of this pull request is.  @Neilpang Any prospect of it being merged soon?

emestee:
Soooo.. it's 2018... any chance of this being merged?

emestee:
<3

PR #647, opened 2017-02-20
Weechat deploy

phil21:
A basic weechat deploy script, should work on the majority of common weechat installations.


phil21:
whoops, wrong branch

PR #682, opened 2017-02-28
DNSAPI for DNSEver and deploy plugin for strongswan 

hiskang:
I made below two DNSAPI and deploy plugin:
- DNSAPI for DNSEver (https://www.dnsever.com) which is very famous DNS service in Korea.
- deploy plugin for strongswan IPSec server(https://www.strongswan.org/).


deploy/strongswan.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
Neilpang: please use `#!/usr/bin/env/sh`,  we must be posix

deploy/strongswan.sh
```diff
@@ -0,0 +1,38 @@
...
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+#  _err "Not implemented yet"
+
+#  return 1
+
+  cp -pv $_ckey       /etc/ipsec.d/private/$(basename $_ckey)
```
Neilpang: do not use `cp -pv`.
Please use `cat "$keys" > targetfile.key`

dnsapi/dns_dnsever.sh
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bash
```
Neilpang: use `sh`

dnsapi/dns_dnsever.sh
```diff
@@ -0,0 +1,121 @@
...
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+  dnsever_delete_txt "$DNSEVER_ID" "$DNSEVER_PW" "$fulldomain" "$txtvalue"
+  return $?
+}
+
+####################  Private functions below ##################################
+
+dnsever_add_txt(){
+  local login_id="$1"; local login_password="$2"; local fulldomain="$3"; local new_txt="$4"
```
Neilpang: do not use `local` keyword.  it's not POSIX

dnsapi/dns_dnsever.sh
```diff
@@ -0,0 +1,121 @@
...
+  local n=$(echo $fulldomain |  grep -o '\.'  | wc -l)
+  local f=$(seq 1 $(($n - 1)) | paste -d, -s -)
+  local subname=$(echo $fulldomain | cut -f $f -d .)
+  local user_domain=$(echo $fulldomain | cut -f $n,$(( $n + 1 )) -d .)
+
+  curl -sS -k -c /tmp/dnsever.txt -d "login_id=$login_id" -d "login_password=$login_password" https://kr.dnsever.com/index.html > /tmp/dnsever.html
```
Neilpang: do not use `curl`,
Please use the `_get()` or `_post()` function instead.

deploy/strongswan.sh
hiskang: I didn't touch the first line from dns_myapi.sh, and I also see other files in dnsapi directory are having same line.
Do you really want me to use posix bourne shell instead of bash?

Neilpang: Yes, we must use `sh`.
You can check the original file on github, it's `sh`, not `bash`.
When installing on your system, we detect if `bash` is found, if found we will change the shebang to us `bash`,  since bash has a better performance in some of the operations.

hiskang:
I updated the files. please check.

PR #706, opened 2017-03-06
Shell script for deploying changes to a routeros server.

palhaland:
Shell script for deploying certificate and key to a routeros server


deploy/routeros.sh
```diff
@@ -0,0 +1,70 @@
...
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  if [ -z "$ROUTER_OS_HOST" ]; then
```
aklowther: if this doesn't exist, couldn't you use `_cdomain`?

palhaland: that is a good point, could add that addition

deploy/README.md
```diff
@@ -255,3 +255,20 @@ acme.sh --deploy -d fritzbox.example.com --deploy-hook fritzbox
...
 ```
+
+## 10. Deploy the cert to remote routeros
+
+```sh
+acme.sh --deploy -d ftp.example.com --deploy-hook routeros
+```
+
+Before you can deploy the certificate to router os, you need to add the id_rsa.pub key to the routeros and assign a user to that key.
+The user need to have access to ssh, ftp, read and write.
```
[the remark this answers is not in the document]
palhaland: Good point, removing that from the doc

deploy/routeros.sh
```diff
@@ -0,0 +1,70 @@
...
+    _err "Need to set the env variable ROUTER_OS_USERNAME"
+    return 1
+  fi
+
+  _info "Trying to push key '$_ckey' to router"
+  scp "$_ckey" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.key"
+  _info "Trying to push cert '$_ccert' to router"
+  scp "$_ccert" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.cer"
+  _info "Trying to push ca cert '$_cca' to router"
+  scp "$_cca" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.ca"
```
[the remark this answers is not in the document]
palhaland: I guess that is true as long as the client accessing the router has access to internet, if using a hotspot with https this might not be the case. Can still remove this from the script

deploy/routeros.sh
```diff
@@ -0,0 +1,70 @@
...
+
+/file remove $_cdomain.cer
+
+/file remove $_cdomain.key
+
+/file remove $_cdomain.ca
+
+delay 2
+
+/ip service set www-ssl certificate=$_cdomain.cer_0
```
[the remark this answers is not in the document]
palhaland: Will add additional comment to the README.md, was also thinking about making an script, that could be executed after the certificates have been updated to update other services that used those certificates.

deploy/routeros.sh
```diff
@@ -0,0 +1,70 @@
...
+  _info "Trying to push key '$_ckey' to router"
+  scp "$_ckey" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.key"
+  _info "Trying to push cert '$_ccert' to router"
+  scp "$_ccert" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.cer"
+  _info "Trying to push ca cert '$_cca' to router"
+  scp "$_cca" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.ca"
+  # shellcheck disable=SC2029
+  ssh "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST" bash -c "'
+
+/certificate remove $_cdomain.cer_0
```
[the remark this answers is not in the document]
palhaland: I will try to incorporate this in the README, do you have any idea on how the indexing works on RouterOS? Will the latest added certificate always have 0, or is it always at the end?

deploy/README.md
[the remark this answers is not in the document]
kg4zow: The point about the user needing that access is good, my point is that people may read that comment and assume that they also have to do `/ip service enable ftp`. My suggestion is to add another line explaining that while ftp *access* is needed, the ftp *service* is not.

deploy/routeros.sh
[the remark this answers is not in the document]
kg4zow: I just noticed ... what I'm doing on my Mikrotiks is uploading `$_cfullchain.cer` as the server key. This file includes the CA chain, and because it's in the same file, it makes the web server include the CA chain when presenting its certificate to the client. The `www-ssl` server doesn't have a way to configure more than one cert, so I *think* if you upload the files separately, the server will only send the certificate from `server.cer`, without the rest of the CA chain, which could cause problems for some clients.

deploy/README.md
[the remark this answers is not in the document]
palhaland: ah, understood, adding this sentence:
There are no need to enable the ftp service on the router for the deploy script to work as the files are transmitted over scp, however ftp access is needed to store the files on the router.

deploy/routeros.sh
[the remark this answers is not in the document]
kg4zow: If the router doesn't already have any certs, the new one becomes item zero. Otherwise, it becomes the "next number in line". And if you remove an item, they don't get re-numbered until the next `print` command (so if the table has five items and you delete #2, the remaining items are "0,1,3,4" until you `print` them, and then they become "0,1,2,3". It makes sense when you're working interactively, but for scripting, not so much.

This is why I wish that the `/certificate import` command allowed you to set the name during import, so that the numbers didn't matter. :)

deploy/routeros.sh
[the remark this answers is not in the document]
palhaland: will modify to use the _cfullchain instead

aklowther:
@Neilpang any chance this could get merged soon?

aklowther:
@Neilpang ping

Neilpang:
sorry, I was too busy, and missed this PR.
please rebase, I will review again.

Thanks.

palhaland:
@Neilpang pull request have been updated with latest dev branch

deploy/routeros.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
Neilpang: do not use bash

palhaland: changed to /usr/bin/env sh (same as in deploy/gitlab)

aklowther:
@Neilpang ping

deploy/routeros.sh
```diff
@@ -0,0 +1,62 @@
...
+    ROUTER_OS_HOST="$_cdomain"
+  fi
+
+  if [ -z "$ROUTER_OS_USERNAME" ]; then
+    _err "Need to set the env variable ROUTER_OS_USERNAME"
+    return 1
+  fi
+
+  _info "Trying to push key '$_ckey' to router"
+  scp "$_ckey" "$ROUTER_OS_USERNAME@$ROUTER_OS_HOST:$_cdomain.key"
```
Neilpang: why use `scp` ?
why not just run acme.sh on the router, and just use `cp` there ?

aklowther: I didn't write the script but I imagine this is for the same reason you don't run `acme.sh` on every machine you want to have an SSL cert.  The same reason you have an SSH deploy script.  I wouldn't run acme on my router because I don't want my router doing things that it doesn't _have_ to do

kg4zow: > why use `scp` ?
> why not just run acme.sh on the router, and just use `cp` there ?

Because Mikrotik routers don’t have `bash`, or any other “real” shell.

palhaland: @Neilpang, it's like @kg4zow said, Mikrotik does not have a proper shell, and a lot of their routers have no more than 16MB flash and a bit more ram. (https://mikrotik.com/product/RB960PGS). They do have SSH interface, but no regular commands.

Neilpang: if it's just a deploy via ssh.  I think it can be accomplished by the ssh_deploy hook.
Is that true ?

aklowther: No, it can't be accomplished with the ssh_deploy hook, if it could, this PR (thats over a year old) wouldn't continue getting attention. 

@palhaland there are new conflicts that keep popping up while @Neilpang delays merging

palhaland:
@Neilpang anything else missing?

Neilpang:
@palhaland 
please update your usage here:

https://github.com/Neilpang/acme.sh/wiki/deployhooks

palhaland:
@Neilpang Done

PR #781, opened 2017-04-11
Default timeout for _get function

Reonaydo:
This allow avoid hangs if any connection problems


Neilpang:
Why do we need a default timeout here ?

Reonaydo:
I`m not sure about only this place. But I think some rational timout must be in any request to avoid problems

Neilpang:
We don't fix for an non-existing problem.

PR #845, opened 2017-05-15
Ensure that server key is readable only by owner

Firm:


Neilpang:
Sorry, we don't change the permissions:
https://github.com/Neilpang/acme.sh/wiki/how-about-the-private-key-access-modes,--chmod,-or-chown-or-umask

PR #846, opened 2017-05-15
Merge pull request #843 from Neilpang/dev

Firm:
When created key-file should not be group/others readable at least.


Neilpang:
Sorry, we don't change the permissions:
https://github.com/Neilpang/acme.sh/wiki/how-about-the-private-key-access-modes,--chmod,-or-chown-or-umask

PR #858, opened 2017-05-22
Support OVH credentials scoped to a specific zone

yadutaf:
When creating OVH API credentials, one can scope them to a specific subset of routes. Specifically, this allows to limit acme.sh to a specific zone as the zone is part of the URL. This is an important security/safety net feature.

If that's good for you, I plan to update the wiki page wo explain how to create scoped credentials in one shot (with no url to visit for validation)

### DIFF
--- a/dnsapi/dns_ovh.sh
+++ b/dnsapi/dns_ovh.sh
@@ -238,7 +238,7 @@ _get_root() {
       return 1
     fi
 
-    if ! _contains "$response" "This service does not exist" >/dev/null; then
+    if ! _contains "$response" "This service does not exist" >/dev/null && ! _contains "$response" "NOT_GRANTED_CALL" >/dev/null; then
       _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-$p)
       _domain="$h"
       return 0


Neilpang:
Good,  please update the wiki first.
Thanks

yadutaf:
Done: https://github.com/Neilpang/acme.sh/wiki/How-to-use-OVH-domain-api

Neilpang:
Sorry,  the wiki is too hard to understand and use, especially for a starter. 
Is there any official doc ?

Remember, people(especially starters) just want a link to click, and click next and next.
Nobody wants to  copy the example url and paste it to somewhere, then carefully replace the example domain to his domain, then bla bla bla......

It's too detail, technical and geek.  Nobody likes that.

Neilpang:
If the doc can not guide a starter to begin in 10 seconds,  it's not good enough.

yadutaf:
I'm afraid I do not understand your point. I actually simplified it a *lot*. The previous version was barely understandable at all, even requiring manual link copy/paste validation. Please have a look at the diff !

The new version id like:
- generate your credentials
- configure acme.sh
- profite !

It can hardly be made any easier.

Neilpang:
The original guide is started from here: https://eu.api.ovh.com/createApp/,  not: https://api.ovh.com/createToken

Did you followed the original steps ?

Neilpang:
Sorry, I can merge the code,  but the wiki will be reverted.

The original design is to manage all the domain zones within one app key

Because we have only ovh app key for acme.sh, and the user may want to issue cert for all his domain names.

yadutaf:
Thanks for the merge !

Please consider NOT reverting the wiki. I am the original author of the official OVH API wrapper https://github.com/ovh/python-ovh. Although I no longer work for OVH, I have a pretty good understanding of the the inner working of the API engine ;)

I have further simplified it in https://github.com/Neilpang/acme.sh/wiki/How-to-use-OVH-domain-api and removed erroneous mentions of unsupported kimsufi and soyoustart API which do not sell domain names and hence can not support acme.sh.

If you look carefully https://eu.api.ovh.com/createApp/ will require the user to

1. Fill the form
2. copy the credentials + run acme.sh a first time
3. copy/paste a link in the console to the browser
4. validate the link
5. go back to the console and re-run acme.sh

That is a LOT of manual / error-prone steps I strongly advise NOT to document this advanced scenario.

Using https://api.ovh.com/createToken on the other hand, allows to generate all credentials at once, with a single authentication and acme.sh run.

Neilpang:
Good,  I understand.
Let me think again about the wiki.

I wrote the ovh wrapper,   the original steps are the most straightforward and headless steps.
I just tried again with the original steps,  it only requests copy/paste the app key and auth url.  The user don't need to care about the content of the app key and the url.  The app key and url looks very simple.

But if starting from the `creatToken` url,   people would need to replace the example domain in the url.  That's a pain,  you know.

But your idea about requesting access to single zone, instead of all zones, is a good idea.  I think we can implement it in the code,  instead of left the heavy load to the user.

yadutaf:
> But if starting from the creatToken url, people would need to replace the example domain in the url. That's a pain, you know.

OK, I get your point. I've simplified it to have "Quick Start" section which scoped the credentials to the full OVH domain API (but no other APIs) and added a "Security" section with the manual steps, to make it clear that this is optional, while still still strongly recommended.

Gets the best of both worlds

Neilpang:
@yadutaf 
BTW,  is there any difference between `https://eu.api.ovh.com`  and `https://api.ovh.com`

Neilpang:
my account/password is working on `https://eu.api.ovh.com/createApp/`
But not working on `https://eu.api.ovh.com/createToken/?GET=/domain/zone/*&POST=/domain/zone/*&PUT=/domain/zone/`

yadutaf:
Hmm, what do you mean ? There should be no difference regarding the username password. This said, there is a slight bug in the way Github parses URLs and the last "*" was missing. It will cause acme.sh to fail with the generated credentials. I've worked around it in the wiki by using explicit link markup.

I've double tested, my own username / password do work to generate the credentials on both eu.api.ovh.com and api.ovh.com (these are 2 aliases for the same thing. Historical reasons.)

Neilpang:
I got a invalid username password error at this url:

```
https://eu.api.ovh.com/createToken/?GET=/domain/zone/*&POST=/domain/zone/*&PUT=/domain/zone/*
```

Neilpang:
I have an idea.
We should create the auth url in the code, instead of leaving the choice the the user.
See:  https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_ovh.sh#L194
```sh
_ovhdata='{"accessRules": [{"method": "GET","path": "/auth/time"},{"method": "GET","path": "/domain"},{"method": "GET","path": "/domain/zone/*"},{"method": "GET","path": "/domain/zone/*/record"},{"method": "POST","path": "/domain/zone/*/record"},{"method": "POST","path": "/domain/zone/*/refresh"},{"method": "PUT","path": "/domain/zone/*/record/*"}],"redirection":"'$ovh_success'"}'

```
It requires the access to all the zones `*`,  maybe we can add a new env value to control this.
for example:

```
export OVH_ZONE=example.com
acme.sh  --issue -d example.com  --dns dns_ovh  
```
If the `OVH_ZONE` value is set, we will request access to this zone.

What do you think ?

yadutaf:
Looks good! The user will still need to copy/paste 3 values, but we can hardly help with that. And this is much less error-prone.

Side-note: all /auth calls are granted by default and /domain itself does not appear to be needed in the code.

Neilpang:
I just created a new issue:

https://github.com/Neilpang/acme.sh/issues/859

Would you mind if I revert the wiki changes ?  caz it can not work for me now,  I doubt it works for other users.

I will update the wiki when the issue is fixed.

yadutaf:
Maybe create a "staging" wiki page so that the changes will still be around before reverting ? Especially the new structure.

Neilpang:
I just keep your instructions as the `Advanced Usage` in the end.
https://github.com/Neilpang/acme.sh/wiki/How-to-use-OVH-domain-api

And I will also update the wiki when the issue is fixed.

You can keep an eye on issue:  https://github.com/Neilpang/acme.sh/issues/859

Thanks.

ealphonse:
@yadutaf Hello, thank you for the detailed walk-through. Should not one add `DELETE` to the allowed endpoint as it is used by `dns_ovh_rm`?

Neilpang:
@ealphonse   you can edit the wiki page.

ealphonse:
Thank you. Tested and done.

@yadutaf I did not test the syntax for the broad usage. Can you confirm the regexp is ok for DELETE?
```
DELETE=/domain/zone/*/record/*
```

Thank you in advance

yadutaf:
Looks good ! Though, potentially dangerous, should anything go wrong in the script.

Neilpang:
@yadutaf 
If you mind it dangerous,  you can try  dns alias mode:

https://github.com/Neilpang/acme.sh/wiki/DNS-alias-mode

PR #871, opened 2017-06-08
Added views to infoblox dnsapi script

hajhatten:
Views are used to setup "split horizon" dns.
The script only used the default view, and wasn't compatible with our setup.

This change lets you setup a Infoblox_View environment variable. But defaults to the one used before: "default", if not set.


hemskgren:
Great, it was needed.

Neilpang:
@hajhatten 
The `Infoblox_View` variable should be saved if it's not default value, just like `Infoblox_Creds`

hajhatten:
Right, I wasn't sure that was needed. Fixed now.

dnsapi/dns_infoblox.sh
```diff
@@ -94,4 +99,4 @@ dns_infoblox_rm() {
   fi
 }
 
-####################  Private functions below ##################################
+####################  Private functions below ##################################
```
Neilpang: Please add new line at the end of the file.

PR #872, opened 2017-06-09
Add support for deploying to cpanel using perl

a1291762:
Found that deploying to cpanel didn't work. Found a tutorial to deploy ssl to cpanel via API calls. Filled in the blanks and verified on my web host.

No error checks, but it prints out the result as as JSON dump, making errors hopefully obvious if run manually.

I had to use a fqdn rather than localhost to access my cpanel so I made it another variable.

The real trick will be seeing if I can get this to automate in 60 days when my cert updates itself :)


a1291762:
Err... Not sure what I did wrong to get that build failure.

I just want to say, acme.sh was so nice and easy to use, even though I had to fix the deployment script.

PR #895, opened 2017-06-19
Add DuckDNS DNS API

RockyTV:
https://www.duckdns.org/spec.jsp

### DIFF
--- a/dnsapi/dns_duckdns.sh
+++ b/dnsapi/dns_duckdns.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/env sh
+
+# DNS API for DuckDNS
+#
+# Report bugs at https://github.com/RockyTV/acme.sh/issues
+#
+# Set this environment variable to match your DuckDNS account token:
+# DUCKDNS_TOKEN=aaaaaaaa-bbbb-cccc-dddddddddddd
+
+########  Public functions #####################
+# Usage: dns_duckdns_add   _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_duckdns_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  _info "Adding TXT record to DuckDNS domain"
+  _debug "fulldomain: $fulldomain"
+  _debug "txtvalue: $txtvalue"
+
+  if ! _get_root "$fulldomain"; then
+    _err "Domain does not exist."
+    return 1
+  fi
+  
+  if [ -z "$DUCKDNS_TOKEN" ]; then
+    DUCKDNS_TOKEN=""
+    _err "DuckDNS token is not defined."
+    _err "Please export your account token to DUCKDNS_TOKEN and try again."
+    return 1
+  fi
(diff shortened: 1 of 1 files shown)


dnsapi/dns_duckdns.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/bash
```
Neilpang: please use `sh` shebang instead

dnsapi/dns_duckdns.sh
```diff
@@ -0,0 +1,62 @@
...
+  _debug "txtvalue: $txtvalue"
+
+  if [ -z $DUCKDNS_TOKEN ]; then
+    _err "DuckDNS token is not defined."
+    _err "Please export your account token to DUCKDNS_TOKEN and try again."
+    return 1
+  fi
+
+  duckdns_url="https://www.duckdns.org/update?domains=$_domain&token=$DUCKDNS_TOKEN&clear=true"
+  if [[ $(_get $duckdns_url) -ne "OK" ]]; then
```
Neilpang: do not use double bracket

dnsapi/dns_duckdns.sh
```diff
@@ -0,0 +1,62 @@
...
+dns_duckdns_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  _info "Adding TXT record to DuckDNS domain"
+  _debug "fulldomain: $fulldomain"
+  _debug "txtvalue: $txtvalue"
+  
+  if [ -z "$DUCKDNS_TOKEN" ]; then
+    _err "DuckDNS token is not defined."
```
Neilpang: add `DUCKDNS_TOKEN=""`  here.

dnsapi/dns_duckdns.sh
```diff
@@ -0,0 +1,62 @@
...
+    _err "Please export your account token to DUCKDNS_TOKEN and try again."
+    return 1
+  fi
+  _saveaccountconf DUCKDNS_TOKEN "$DUCKDNS_TOKEN"
+  
+  duckdns_url="https://www.duckdns.org/update?domains=$_domain&token=$DUCKDNS_TOKEN&txt=$txtvalue"
```
Neilpang: where  does `$_domain` come from ?

RockyTV: It comes from acme.sh itself. A lot of DNS APIs use it, take a look: https://github.com/Neilpang/acme.sh/search?utf8=%E2%9C%93&q=_domain&type=

Neilpang: you should only use `$fulldomain`,  and detect which part is the root zone, and which part is the  sub domain zone.   do not rely on other variables. in most of the other dns api hooks, there is a function `getroot()`.  please take a look.

RockyTV:
Should be good to go now.

raidenii:
The API URL used in removing the TXT record is wrong, you'll need to include txt in the URI as well, otherwise it will remove the A and AAAA record.
I have forked my version: e7dff47

Neilpang:
duckdns   is already supported,  so close this.

PR #938, opened 2017-07-09
Cleanup Cloudflare dns_cf.sh for general consistency, keeping changes…

abelbeck:
- Cleanup for general consistency, keeping changes to a minimum.
- Consistently use `_egrep_o` ... `_head_n` ... `sed` to extract JSON values
- Use `_contains` instead of `grep`
- Rename key private variables to begin with `_cf_`
- Rename `_get_root` to `_cf_get_zone`, remove unused `_sub_domain` variable
- In `_cf_rest` rename arg variables to be more descriptive, 

Note: The API response `"success":true` seems to occur more often then you would think it should.

Note: To be more robust
```
_contains "$response" '"success":true'
```
could be written as
```
_contains "$response" '"success" *: *true'
```
but I was not sure if you wanted `_contains` to use regex's as a practice.  It currently works fine as is.

I have tested it, issued a couple certificates, monitored the debug output, and it looks good to my eye.

### DIFF
--- a/dnsapi/dns_cf.sh
+++ b/dnsapi/dns_cf.sh
@@ -1,18 +1,22 @@
 #!/usr/bin/env sh
-
 #
-#CF_Key="sdfsdfsdfljlbjkljlkjsdfoiwje"
+# Cloudflare.com API
+# https://api.cloudflare.com/
 #
-#CF_Email="xxxx@sss.com"
+# Pass credentials before "acme.sh --issue --dns dns_cf ..."
+# --
+# export CF_Key="sdfsdfsdfljlbjkljlkjsdfoiwje"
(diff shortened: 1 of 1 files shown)


Neilpang:
Is there any problem that you must modify it ?

abelbeck:
@Neilpang No problem today, but if the JSON output started including a space either side of the colon these changes go a long way to adapt to that.  Allowing `_contains "$response" '"success" *: *true` together with these changes I already made should make this "space proof".

Neilpang:
@abelbeck 
I would like to keep it as is.   Do not touch it if it works.

abelbeck:
I understand, fair enough.

PR #945, opened 2017-07-16
nsupdate -l

MarcelWaldvogel:
Configurationless operation when `nsupdate -l` works

PR #946, opened 2017-07-19
`nsupdate -l`-like mode

MarcelWaldvogel:
Configurationless `nsupdate -l`-like mode
    
When `nsupdate -l` works, `-dnsapi nsupdate` requires no further
configuration. It does not use `nsupdate -l` (would imply more code
and more code changes), but points `nsupdate -k` at the key which
would be used by `nsupdate -l`.

This is IMHO a better path than #945


tgice (commented):
Thanks for sharing.  This worked for me on my one machine running BIND and acme.sh together.

PR #949, opened 2017-07-24
Candlerb/unifi

candlerb:
This adds a deployment script for unifi controller.

Questions:

* Is there a better place to write temporary files than to `/tmp`? Would it be better to use the current working directory that the script is running in?
* Does acme.sh run the deploy script with `set -e` (stop-on-error)? If not, then I would chain together the commands in the script with `&&`

### DIFF
--- a/deploy/unifi.sh
+++ b/deploy/unifi.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env sh
+
+#Here is a script to deploy cert to unifi server.
+
+#returns 0 means success, otherwise error.
+
+#DEPLOY_UNIFI_KEYSTORE="/usr/lib/unifi/data/keystore"
+#DEPLOY_UNIFI_KEYPASS="aircontrolenterprise"
+#DEPLOY_UNIFI_RELOAD="service unifi restart"
+
+########  Public functions #####################
+
+#domain keyfile certfile cafile fullchain
+unifi_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  if ! _exists keytool; then
+    _err "keytool not found"
+    return 1
+  fi
+
+  DEFAULT_UNIFI_KEYSTORE="/usr/lib/unifi/data/keystore"
+  _unifi_keystore="${DEPLOY_UNIFI_KEYSTORE:-$DEFAULT_UNIFI_KEYSTORE}"
+  DEFAULT_UNIFI_KEYPASS="aircontrolenterprise"
+  _unifi_keypass="${DEPLOY_UNIFI_KEYPASS:-$DEFAULT_UNIFI_KEYPASS}"
+  DEFAULT_UNIFI_RELOAD="service unifi restart"
+  _reload="${DEPLOY_UNIFI_RELOAD:-$DEFAULT_UNIFI_RELOAD}"
+
+  _debug _unifi_keystore "$_unifi_keystore"
+  if [ ! -f "$_unifi_keystore" ]; then
+    if [ -z "$DEPLOY_UNIFI_KEYSTORE" ]; then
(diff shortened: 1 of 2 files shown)


Neilpang:
1.  You can use `_mktemp()` function to get a  temp file.

2. No, `set -e` is only for Base,   acme.sh is sh.

deploy/unifi.sh
```diff
@@ -0,0 +1,35 @@
...
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  openssl pkcs12 -export -in "$_ccert" -inkey "$_ckey" \
```
Neilpang: do not use `openssl` command explicitly.  we have a function `toPkcs()`,  you can refactor that function and reuse it.

deploy/unifi.sh
```diff
@@ -0,0 +1,35 @@
...
+   -out /tmp/unifi.p12 -name unifi \
+   -CAfile "$_cca" -caname root -password pass:aircontrolenterprise
+
+  keytool -importkeystore \
```
Neilpang: check if `keytool` command exists first,  using `_exists()` function.

candlerb:
Regarding use of the `toPkcs()` function:

This currently writes its output to `$CERT_PFX_PATH`, but that is not given as an argument to the deploy function.  There are various ways this could be refactored:

* Just pick CERT_PFX_PATH from the environment (that's inconsistent with the other args)
* Pass CERT_PFX_PATH to all deploy functions (ugh, changes them all)
* Make CERT_PFX_PATH an argument toPkcs, and pass in a `_mktemp` value (although I would be fine leaving the pkcs12 file lying around)
* Some other way you'd prefer?

Neilpang:
>> This currently writes its output to $CERT_PFX_PATH, but that is not given as an argument to the deploy function. 

Yes,  I know.  I said : you can refactor the function and then reuse it.

you can refactor a new function, and pass the file path in the function as a parameter.

I want to restrict the usage of `openssl` command.  In future,  it could be easier to replace openssl with keytool or guntls etc.

candlerb:
Refactored.  It's a bit messy with the variable args to openssl, but I don't know another way to do this in shell which is both secure and portable.

I have not exposed the `name`/`caname` functionality to the command line - it's just for internal use.  The unifi keytool import definitely needs it, because without:

```
Warning: Overwriting existing alias unifi in destination keystore
keytool error: java.lang.Exception: Alias <unifi> does not exist
[Mon Jul 24 15:37:30 UTC 2017] keytool failed
[Mon Jul 24 15:37:30 UTC 2017] Error deploy for domain:unifi.example.net
[Mon Jul 24 15:37:30 UTC 2017] Deploy error.
```

deploy/unifi.sh
```diff
@@ -0,0 +1,50 @@
...
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  if ! _exists keytool; then
+    _err "keytool not found"
+    return 1
+  fi
+
+  _pkcs12="$(_mktemp)"
+  _keypass="${DEPLOY_UNIFI_KEYPASS:-aircontrolenterprise}"
```
Neilpang: please use `_savedomainconf()` function to save the "DEPLOY_UNIFI_KEYPASS" and  "DEPLOY_UNIFI_KEYSTORE".

See example:
https://github.com/Neilpang/acme.sh/blob/master/deploy/exim4.sh#L92

candlerb:
I rewrote it based on exim4 deploy script.

However there is a problem with the backup directory not being created:

~~~
[Sun Jul 30 08:43:23 UTC 2017] Backup /usr/lib/unifi/data/keystore to /root/.acme.sh/unifi.example.net/backup/unifi.keystore.bak
cp: cannot create regular file '/root/.acme.sh/unifi.example.net/backup/unifi.keystore.bak': No such file or directory
~~~

And since this error is not trapped, it could result in the file being overwritten with empty data later on when attempting to restore:

~~~
    if cat "$_backup_keystore" >"$_unifi_keystore"; then
~~~

candlerb:
Scrub the keystore backup. We can rely on keytool doing an atomic update.

PR #954, opened 2017-07-26
Fix DNS API scripts on *BSD

jfro:
\n isn't available in all regex/sed

Ran into acme.sh + linode DNS not working on my FreeNAS (FreeBSD) box and found out the sed /{/\n&/ was working. Wouldn't work on OS X either which is BSD based also. This should hopefully fix that and still work on linux.

https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed

### DIFF
--- a/dnsapi/dns_linode.sh
+++ b/dnsapi/dns_linode.sh
@@ -68,7 +68,7 @@ dns_linode_rm() {
   _parameters="&DomainID=$_domain_id"
 
   if _rest GET "domain.resource.list" "$_parameters" && [ -n "$response" ]; then
-    response="$(echo "$response" | tr -d "\n" | awk '{ gsub("{","\n{",$1); print $1 }')"
+    response="$(echo "$response" | tr -d "\n" | sed $'s/{/\\\n&/g')"
 
     resource="$(echo "$response" | _egrep_o "{.*\"NAME\":\s*\"$_sub_domain\".*}")"
     if [ "$resource" ]; then
@@ -128,7 +128,7 @@ _get_root() {
   p=1
 
   if _rest GET "domain.list"; then
-    response="$(echo "$response" | tr -d "\n" | awk '{ gsub("{","\n{",$1); print $1 }')"
+    response="$(echo "$response" | tr -d "\n" | sed $'s/{/\\\n&/g')"
     while true; do
       h=$(printf "%s" "$domain" | cut -d . -f $i-100)
       _debug h "$h"
--- a/dnsapi/dns_linode.sh
+++ b/dnsapi/dns_linode.sh
@@ -68,8 +68,7 @@ dns_linode_rm() {
   _parameters="&DomainID=$_domain_id"
 
   if _rest GET "domain.resource.list" "$_parameters" && [ -n "$response" ]; then
-    response="$(echo "$response" | tr -d "\n" | sed 's/{/\
-&/g')"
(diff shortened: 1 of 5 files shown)


dnsapi/dns_linode.sh
```diff
@@ -68,7 +68,8 @@ dns_linode_rm() {
   _parameters="&DomainID=$_domain_id"
 
   if _rest GET "domain.resource.list" "$_parameters" && [ -n "$response" ]; then
-    response="$(echo "$response" | tr -d "\n" | sed 's/{/\n&/g')"
+    response="$(echo "$response" | tr -d "\n" | sed 's/{/\
```
Neilpang: Please replace the '{' char to some other unused char, and then do a sed, and replace the '{' char back to Lf at last.
for example:

```sh
echo "$response" | tr -d "\n"  | tr '{' "_" | sed 's/_/&{/g' | tr "_" "\n"
```

Neilpang:
please do not modify the other dns apis, unless you have tested.

Thanks.

dnsapi/dns_linode.sh
jfro: hmm, will have to figure out best unused char, I think _ could be a problem since a domain could have a _ at least?

Neilpang: Yes,  it's just an example.
do what you find the best.

jfro:
Couldn't figure out a decent way to make sed do the job, but awk seems to work ok on linux & bsd. Afaik it's also just as widely installed as sed.

Neilpang:
Sorry,  please do not use awk.  It's not crossing platforms.  No one used it before.

jfro:
what's awk not available on? I'm not sure I can safely use a temporary character like suggested earlier, and the only suggestion for \n I've found for BSD sed is the first iteration I did. I'll try to investigate further

jfro:
think this is doable?

`sed $'s/{/\\\n&/g')"`

kballard:
Awk is available on every platform I've seen. I'd also like to know why it's not suitable.

I'm also curious what's wrong with embedding a newline into the sed script. Seems fine to me. Using `tr` twice to swap in some other character just seems janky and easy to screw up.

Neilpang:
>> Using tr twice to swap in some other character just seems janky and easy to screw up.

No,  it's the safest way.

You never know how broken on which our script is running.

Some embeded linux system doesn't contains awk.  And awk's implementation is different on some platforms.

So, if things can  be done with much more simple/common tools like sed/tr,   we never use a more heavy tool like awk.

kballard:
AFAIK there's only 2 seds, so I don't see what the problem with the embedded newline is.

In any case, if you're that worried, how about just using shell built-ins? `IFS=}` plus `read` should let you split up the input and then put it back together however you want.

Neilpang:
Please fix the errors: 
https://travis-ci.org/Neilpang/acme.sh/jobs/258581926

jfro:
Think I got it now, think | is a safe character to use your solution. The literal newline in the original solution would probably be fine too, since the docs of shellcheck says you can ignore that warning for that usage, but travis treated it as an error.

But I think this should finally be alright, don't believe | will likely occur in the response.

Neilpang:
please fix the format errors:

https://travis-ci.org/Neilpang/acme.sh/jobs/259350051

Neilpang:
It seems that here is some problems.
please close this PR and send a new one.

Thanks.

PR #965, opened 2017-08-06
Add deploy hook for ZNC

Xinayder:
I've added a deploy hook to automatically deploy the Let's Encrypt certificate for ZNC.
You need to run `acme.sh` as the same user as ZNC, because otherwise the permissions get borked and ZNC breaks.

The `znc.pem` file is generated with the certificate key, domain certificate and CA certificate. More info: https://wiki.znc.in/Signed_SSL_certificate

The script, by default, uses `/var/lib/znc/.znc` as the ZNC home directory and `znc` as the ZNC user (Arch Linux defaults). The user can override these by setting `ZNC_USER` and `ZNC_DIR` variables, which will be saved to `account.conf`.

### DIFF
--- a/deploy/znc.sh
+++ b/deploy/znc.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env sh
+
+# Deploy Let's Encrypt certs to ZNC
+#
+# Any non-zero returns means something wrong has occurred
+#
+# If you want to use a custom directory and/or user and group owners, you may set the following variables:
+#
+#   $ZNC_DIR_OWNERSHIP - user and group owners for a directory (e.g. export ZNC_DIR_OWNERSHIP="user:group")
+#   $ZNC_DIR - ZNC config directory (e.g. export ZNC_DIR="/home/znc/.znc"), more info (check Misc):
+#                       https://wiki.znc.in/Configuration#File_locations
+
+_ZNC_DIR="/var/lib/znc/.znc"
+_ZNC_DIR_OWNERSHIP="znc:znc"
+
+########  Public functions #####################
+
+#domain keyfile certfile cafile fullchain
+znc_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  # shellcheck disable=SC2034
+  _cfullchain="$5"
+
+  # Workaround for SC2153. Check if ZNC_DIR_OWNERSHIP is set to zero,
+  # then fallback to _ZNC_DIR_OWNERSHIP.
+  if [ -z "$ZNC_DIR_OWNERSHIP" ]; then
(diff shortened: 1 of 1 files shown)


deploy/znc.sh
```diff
@@ -0,0 +1,63 @@
...
+    return 1
+  fi
+
+  # Allow users to override the default ZNC config dir
+  if [ -n "$ZNC_DIR" ]; then
+    _ZNC_DIR="$ZNC_DIR"
+    _info "ZNC config path set to: $_ZNC_DIR"
+
+    # Check if the specified ZNC config dir is owned by the specified ZNC user
+    if [ -z "$(find . -user "$_ZNC_USER" -print -prune -o -prune)" ]; then
```
Neilpang: We have a function "_stat()", can you use that function instead of `find` ?

deploy/znc.sh
```diff
@@ -0,0 +1,63 @@
...
+      _err "The specified ZNC config directory isn't owned by $_ZNC_USER"
+      _err "Please specify the correct directory or correct user"
+      return 2
+    fi
+  fi
+
+  # Save ZNC user and config directory to account.conf
+   _saveaccountconf ZNC_USER "$_ZNC_USER"
```
Neilpang: please use `_savedomainconf()` instead.

deploy/znc.sh
```diff
@@ -0,0 +1,63 @@
...
+   _saveaccountconf ZNC_DIR "$_ZNC_DIR"
+
+  # ZNC certificate file location
+  _znc_cert="$_ZNC_DIR/znc.pem"
+
+  # Please read https://wiki.znc.in/Signed_SSL_certificate
+  _info "Generating ZNC certificate file for $_cdomain"
+  cat "$_ckey" > "$_znc_cert"
```
Neilpang: we need to check the ret code of each `cat`

```sh

cat .....  && cat ...... && cat ....

```

RockyTV:
I've updated the script because `_stat` checks for user and group ownership. Now, the script will check if the specified ownership string `user:group` matches the current user, and if the current user is part of the owner group.

Neilpang:
please fix the format errors first
https://travis-ci.org/Neilpang/acme.sh/jobs/262050938

RockyTV:
Should be good to go now, just waiting for Travis to finish the OS X build that got stuck.

RockyTV:
I think this is ready to be merged. Anything else you want me to change?

Neilpang:
https://travis-ci.org/Neilpang/acme.sh/jobs/263286268

RockyTV:
Formatting errors fixed.

RockyTV:
Anything you want me to change?

RockyTV:
I'll be rebasing this in the next days to the latest commit in dev so it can be merged.

RockyTV:
Rebased.

ldir-EDB0:
Is this likely to be merged?

PR #969, opened 2017-08-11
Update DuckDNS support, fix failure on first call, no longer save the…

abelbeck:
… domain/username as a global, and other tweaks

Remove `API_Params` since on first call `DuckDNS_Domain` was not initialized and it failed.

No longer save `DuckDNS_Domain` but rather calculate it (now `_duckdns_domain`) from `fulldomain` every time.  Multiple domains within the same account should now work.

Fix _egrep_o regex to match a dot, not any character

Tweak comments to not imply ``--insecure`` is always needed, works for me with `curl` and the latest Mozilla ca-bundle.crt.


abelbeck:
@raidenii Any comments on these changes since you created the original ?

I have tested these changes and works well in my testing.

raidenii:
Looks good to me.

PR #973, opened 2017-08-17
Support for busybox in time2str()

mvanini:
'cause busybox's date doesn't work as expected, running acme.sh on it will result in missing date string for renewal. Using awk function "strftime" it works.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1367,6 +1367,10 @@ _time2str() {
     echo "$_t_s_a"
   fi
 
+  #Busybox
+  if echo "$1" | awk '{ print strftime("%c", $0); }' 2>/dev/null; then
+    return
+  fi
 }
 
 _normalizeJson() {


Neilpang:
Is there any other way to fix it in busybox ?
I don't like `awk`.   and ` awk '{ print strftime("%c", $0); }'`  is only working busybox.   `strftime` is not defined in Linux.

Neilpang:
`time2str` is just for printable text for the user.   it will not affect the real renewal time.

mvanini:
@Neilpang, it also work in a CentOS/RHEL (i've just tried).
Busybox is very limited shell, I've tried many other solutions with no success.
I know that this function is only for printable, but it is usefull to read (and understand) renewal-date

Neilpang:
I tried on my ubuntu:
```
root@ed:~# echo 1502977392 | awk '{ print strftime("%c", $0); }'
awk: line 2: function strftime never defined

```

mvanini:
but on your Ubuntu "date -u -d@$0" works which mean that you never use awk... this is only a fallback

PR #976, opened 2017-08-17
Create a dns api for the german domain registrar Schlund Technologies

hmb:
The german domain registrar Schlund Technologies provides an XML interface to manipulate their DNS servers. The new dns api file dns_schlundtech.sh provides support for this registrar.
DE: https://www.schlundtech.de/
EN: https://www.schlundtech.com/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: shell
-sudo: required
 dist: trusty
 
 os:
--- a/deploy/haproxy.sh
+++ b/deploy/haproxy.sh
@@ -233,7 +233,6 @@ haproxy_deploy() {
           -header Host${_header_sep}\"${_ocsp_host}\" \
           -respout \"${_ocsp}\" \
           -verify_other \"${_issuer}\" \
-          -no_nonce \
           ${_cafile_argument} \
           | grep -q \"${_pem}: good\""
         _debug _openssl_ocsp_cmd "${_openssl_ocsp_cmd}"
--- a/dnsapi/dns_df.sh
+++ b/dnsapi/dns_df.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env sh
+
+########################################################################
+# https://dyndnsfree.de hook script for acme.sh
+#
+# Environment variables:
+#
+#  - $DF_user      (your dyndnsfree.de username)
+#  - $DF_password  (your dyndnsfree.de password)
+#
+# Author: Thilo Gass <thilo.gass@gmail.com>
+# Git repo: https://github.com/ThiloGa/acme.sh
+
+#-- dns_df_add() - Add TXT record --------------------------------------
+# Usage: dns_df_add _acme-challenge.subdomain.domain.com "XyZ123..."
+
+dyndnsfree_api="https://dynup.de/acme.php"
+
+dns_df_add() {
+  fulldomain=$1
+  txt_value=$2
+  _info "Using DNS-01 dyndnsfree.de hook"
+
+  DF_user="${DF_user:-$(_readaccountconf_mutable DF_user)}"
+  DF_password="${DF_password:-$(_readaccountconf_mutable DF_password)}"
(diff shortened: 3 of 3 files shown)


Neilpang:
Please fix the format errors first https://travis-ci.org/Neilpang/acme.sh/jobs/265691704

jan-z:
Whats the status on this PR? I would like to help. Are the changes ready for testing?

hmb:
I use it in my production environment since August. Don't know what's missing for it to get merged.

dnsapi/dns_schlundtech.sh
```diff
@@ -0,0 +1,200 @@
...
+    _saveaccountconf SLTEC_context "${SLTEC_context}"
+    _saveaccountconf SLTEC_server "${SLTEC_server}"
+    return 0
+  fi
+}
+
+_SLTEC_split_domain() {
+  _ST_split_fulldomain="$1"
+
+  _SLTEC_domain="$(echo "$_ST_split_fulldomain" | sed 's/.*\.\([^.]*\.[^.]*\)/\1/')"
```
Neilpang: do not just split the domain.
please use the api of Schlundtech to get a domain list, then check which part is the root zone.

otherwise you will get confused by `example.uk`  and `example.co.uk`

dnsapi/dns_schlundtech.sh
```diff
@@ -0,0 +1,200 @@
...
+      </zone>
+    </task>
+  </request>"
+}
+
+_SLTEC_send_request() {
+  _ST_send_request="$1"
+  _ST_send_url="$2"
+
+  _SLTEC_response="$(curl -s -H "Content-type: text/xml" --data-binary "${_ST_send_request}" "${_ST_send_url}")"
```
Neilpang: do not use `curl` directly here.
please use `_get()` or `_post()` function instead.

dnsapi/dns_schlundtech.sh
hmb: Yep, you're right. I'll check into this,

Neilpang: here is a guide for you:  https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide

dnsapi/dns_schlundtech.sh
hmb: Thanx, I checked that out and dug into the curl function. The `_post` function, when using `curl`, sends data with the `--data` switch. According to the manpage that's the same as `--data-ascii`. This is **not** the same as using the `--data-binary` switch. Unfortunately the curl manpage is not very precise about it. Actually it's written in the desciption of `--data-binary`: `Data is posted in a similar manner as --data-ascii does, except that newlines are preserved and conversions are never done.` This implies that `--data` removes newlines which in fact it does. I just checked it out with a small php program. IMHO the `_post` function should use the `--data-binary` switch to ensure that the data given is passed to the server as is, without changes. Only the caller knows exactly which format is needed. The `_post` function should pass that data on to the server transparently.
I'll create a ticket, just for the records.

nieselregencom:
The script also works for the AutoDNS XML API of InterNetX. (InterNetX is one of the largest domain registrars in Germany.)

tohn:
What's the status with this PR?

hmb:
I fixed both the issues, looking up the subdomain recursively and using the _post function some time ago.

hmb:
@nieselregencom Yes, InterNetX is the technical Provider of Schlundtech. That's why it works.

Neilpang:
please check this guide first:  https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide

tohn:
@meinname I created a [PR](https://github.com/Neilpang/acme.sh/pull/1529) to do this :-)
@Neilpang Your [mentioned guide](https://github.com/Neilpang/acme.sh/pull/976#issuecomment-375652022) is not available anymore?

tohn:
Hello, what's the status on this PR? I'd love to see this merged into master :)

PR #981, opened 2017-08-21
Add a very simple deployment script using scp.

hmb:
Add a very simple deployment script to move certificates to a remote server. The deployment uses scp and simply drops certs into a target directory. A second server side script has to take care of proper installation and service restart. Meant for usecases where the ssh user has minimal permissions only.

### DIFF
--- a/deploy/scp.sh
+++ b/deploy/scp.sh
@@ -1,0 +1,145 @@
+#!/usr/bin/env sh
+
+##########################################################################
+# This is a very simple deployment script to move certificates to a remote
+# server. The deployment uses scp (the remote cp method of ssh) and simply
+# drops certs into a target directory. Targets have the original scp format
+# e.g. like:
+#
+#   server.com:/var/spool/acme.sh/certs/
+#   user@server.com:/var/spool/acme.sh/certs/
+#   configuredserver:/var/spool/acme.sh/certs/
+#
+# You may use something like "configuredserver" which is the name of a host
+# configuration in the ~/.ssh/config file. If you have a more complex setup
+# like different ports, identity files, users or hostnames you are strongly
+# encouraged to use an entry in your ~/.ssh/config file. This saves this
+# little script from reimplementing every possible scp switch.
+#
+# You might wanto to configure ssh on the target server to use a special
+# account with key based authentication and allow scp only. Have a further
+# look at the rssh shell to allow scp only. You might as well put the user
+# into a chroot.
+#
+# The main reason for this form of deployment is, that the acme.sh script
+# can run in a safe and controlled environment. The acme.sh script needs
+# detailed and sensitive information e.g. like your acme private keys or
+# your dns providers credentials. Information like this you certainly don't
+# want to have lying around on your public webserver.
+#
+# Further deployment of the certificates should be handled by a cron job on
+# the remote server. That remote script could then move the new cert's to
+# their proper position, set file owner and permissions and restart the
+# belonging service.
+#
+# An example script for apache (on debian systems) might be:
+#
+#    #!/usr/bin/env sh
+#    chown root:root /var/spool/acme.sh/certs/*
+#    mv /var/spool/acme.sh/certs/* /etc/apache2/ssl.crt/
+#    systemctl restart apache2
+#
+# To avoid misunderstandings, this script is NOT like other deployment
+# scripts that target a specific type of server (apache/cyrus/exim/...)
+# and do all ssl configuration for you. With this script YOU do all your
+# ssl configuration on your target server yourself. Then, and only after
+# the target server is properly configured, you use this script to deploy
+# the forthcoming LE certificates.
+
+# When called for the first time use the following env vars to setup the
+# configuration. The vars will be stored on a per domain basis.
+#DEPLOY_SCP_CA_TARGET="user@server.com:/etc/apache2/ssl.crt"
+#DEPLOY_SCP_KEY_TARGET="user@server.com:/etc/apache2/ssl.key"
+#DEPLOY_SCP_CERT_TARGET="user@server.com:/etc/apache2/ssl.crt"
+#DEPLOY_SCP_FULLCHAIN_TARGET="user@server.com:/etc/apache2/ssl.crt"
+
+########  public functions #####################
+
+#domain keyfile certfile cafile fullchain
+scp_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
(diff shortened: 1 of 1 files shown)


Neilpang:
WE already have a ssh deploy PR, which I will merge soon. please just wait.
https://github.com/Neilpang/acme.sh/pull/586

I prefer ssh deploy.

hmb:
You're right, and I certainly noticed the ssh deploy script that has been in there for a while. What I need is a quick and very simple solution to copy an issued cert onto the server. Very much like the inbuilt "copy to a certain path" switch, just to a remote server. My requirements are very different and more those of a "paranoid server admin". I want to give as little access as possible to s.o. outside of the server. I certainly do not want to:

a) run a foreign script on the server (no offense please)
b) have an ssh user to require more right's than the bare necessity (certainly no service restart and no access to actively used certificate directories like /etc/apache2/ssl.crt)
c) no server side console

Of course one could tweak the ssh deploy into all that, but restricting ssh to scp nicely solves all the above points in a single step. It greatly simplifies the whole process. Other than placing the burden of having to fit many admins requirements on the ssh deployment script, the scp solution opens the door to quickly copy a cert to the server and give the admin the freedom to do with it what he likes, by writing his own personal server side part of the deployment script.

Apart from that - again no offense please - the last commit (other than merges of the upstream) has happened in March and the last comment happened in May. So the ssh deployment PR looked pretty much stalled to me.

In my opinion there's room for both and after I wrote the script I thought I might as well make it publicly available to everyone. This way the user has his own choice of one of the two approaches.

Neilpang:
"To deploy a cert" means 2 things usually:

1.  Copy the cert/key to the remote server.  Yes,  you can do this steps by either  `scp` or `ssh`.  it's no problem.

2. And, usually, you will need to reload the target service to make the copied cert/key take effect.  For example,  you must reload nginx/apache service after the cert/key is copied.    However,  this step can only be down by `ssh`. 


If you use only `scp`, you will face a lot of problems, since you can not reload the services.

but if you also use `ssh` to reload the service,   why not use `ssh deploy` hook, instead of yours ?  
The ssh deploy hook only depends on `ssh`.  no `scp` is needed there.

And in the ssh deploy hook,  it also needs the minimum access, as you expected.  It only requires the `write` access of the target `cert/key` file, and/or the access to reload the service.


The only advantage of a `scp` deploy hook I could see is that:  some system may forbidden `ssh(shell)` access,  but only allow `scp` access. but that's a rare case.

I have been working on `ssh protocol` for years,  it's more common that a system can forbidden `scp` access, but only allow `ssh` access.

So,  `ssh` hook is a more common and more reliable solution than `scp`.

hmb:
Neil,
to explain a little more, I'm using this in a mass hosting environment. But let's take it step by step:

> "To deploy a cert" means 2 things usually:
> 
> Copy the cert/key to the remote server. Yes, you can do this step by either scp or ssh. it's no problem.

agreed

> And, usually, you will need to reload the target service to make the copied cert/key take effect. For example, you must reload nginx/apache service after the cert/key is copied.

agreed

>However, this step can only be done by ssh.

No, this can be done by any script running in a cron job on the server (like logrotate). No need for external access.

> If you use only scp, you will face a lot of problems, since you can not reload the services.

If you deploy hundreds of certs in a hosting environment you'll face problems with the ssh method. With ssh (like it's implemented right now) the service get's restarted after every single deployment. With the scp method the certs get copied to the server one by one let's say once a day at a dedicated time. The server then is reloaded only once way later after the copy process. This suffices easily as you probably renew your certs days if not weeks before the old ones expire.

> but if you also use ssh to reload the service, why not use ssh deploy hook, instead of yours ?
> The ssh deploy hook only depends on ssh. no scp is needed there.
> 
> And in the ssh deploy hook, it also needs the minimum access, as you expected. It only requires the write access of the target cert/key file, and/or the access to reload the service.

The right to to reload the server is not "only minimum access", this is a severe access right, needing root or at least some kind of sudo. Opens a big door for anyone who controls the host running the acme script. E.G denial of service by reloading the service any 10 seconds. No way to give that right to a customer in a mass hosting environment.

> The only advantage of a scp deploy hook I could see is that: some system may forbidden ssh(shell) access, but only allow scp access. but that's a rare case.

This is exactly the case in  my hosting environment. I host sites for my customers. I don't give them a shell, just scp/sftp in a changeroot. They usually don't even use that on the commandline, but use tools like filezilla to copy their data.

> I have been working on ssh protocol for years, it's more common that a system can forbidden scp access, but only allow ssh access.

Agreed. Same here. But that is like telling those who use only scp: "Sorry guys your setup is not common. We don't support your setup then."

> So, ssh hook is a more common and more reliable solution than scp.

More common agreed, but not interchangeable.

IMHO both really have different usage scenarios. The ssh method is the case for a small host setup running sth. like a couple of websites and maybe a mail server. There you have a controlled environment, on the server as well as on the machine running the acme.sh.

In a hosting environment however with many cert exchanges a day, restarting the server after every deployment is just not feasible. You may have many different users deploying their own certs, you don't want to give them any rights they don't really need. Like shell access, access to certain directories, or even restarting the server. Those users might even be restricted to a changeroot. So they can use the scp method to copy their certs to a dedicated directory, cannot do any harm otherwise and the server is reloaded once a day by a server-side script.

Just to mention: I'm very grateful to you and everybody else who contributed to the acme.sh script. It's the Swiss army knife to implement a simple and automated LE process, without having to run sth. like python on your webserver. The hooks and apis you provided let me extend and tweak it to my needs in less than a day. I can run it on an external host or allow others (my clients) to run it and generate the certs they want and need with minimal impact on my production machine. And that alone is great!

If you think - which is your absolute right as creator and maintainer - that the scp solution for some reason or another is not contributing to the spirit of the acme.sh script I do accept that. I'm still happy and grateful that I can use it ;-)

And anyway my little script is no rocket science either. Anyone running a setup like mine can write it in less than an hour. So it's probably no harm, if it's not included.

Cheers Holger

Neilpang:
sorry first.
I also appreciate your contributing very much.

yes,  scp may be a solution in your case,  but it can not provide a way to reload the service.

In most of the cases, it's required to reload the service.  Otherwise, it could make a user(especially for a starter ) confused: why scp deploy success, but the cert is still not valid.

Yes,  to reload service may require extra permissions(maybe root access), but the `ssh` deploy hook just provides a `full` solution.

hmb:
I'm currently cleaning up old tasks. might as well close this now.

PR #1067, opened 2017-10-09
Fix install issue

snielson:
Result 0 of cp and chmod would sometimes fail, even though the results were truly 0.
Storing results into _status solved the problem for me


snielson:
Still having mixed results for some reason. Probably OS issue. Closing pull request

PR #1069, opened 2017-10-11
Skip private hosted zones correctly

ksperling:
This fixes the cases where a private zone is a sub-domain of a
public zone, e.g. foo.com (public) and vpc.foo.com (private).

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -105,13 +105,9 @@ _get_root() {
         return 1
       fi
 
-      if _contains "$response" "<Name>$h.</Name>"; then
-        hostedzone="$(echo "$response" | sed 's/<HostedZone>/#&/g' | tr '#' '\n' | _egrep_o "<HostedZone><Id>[^<]*<.Id><Name>$h.<.Name>.*<PrivateZone>false<.PrivateZone>.*<.HostedZone>")"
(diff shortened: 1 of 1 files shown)


Neilpang:
Can you please give me more details about this case ?

ksperling:
Let's say I have hosted zones `foo.com` (public) and `vpc.foo.com` (private), and want to get a cert for `xyz.vpc.foo.com`. The current code will start by looking for a hosted zone for `vpc.foo.com`, and `_contains "$response" "<Name>$h.</Name>"` returns true. Extracting the actual hosted zone ID from the response then fails, because only at that point does the code check that the zone it's about to return is not private. It then logs that as "Error, can not get hostedzone." and gives up.

With my change, `vpc.foo.com` will be checked first as before, but no hosted zone id will be extracted (the matching zone is private). The loop then continues on to check `foo.com`, and correctly finds and uses that zone.

Neilpang:
@ksperling 
got it.  thanks.

I just made a fix https://github.com/Neilpang/acme.sh/commit/5f6e3da766ef6ffcefca8d5ced3df4b2fcdd7a62

Can you please try branch `fix1069` ?

```sh
export BRANCH=fix1069

acme.sh --upgrade
```

please if it works for you in your case and in a normal case.

Thanks.

ksperling:
Yep, works for me in both cases. Thanks!

Neilpang:
fixed,  please try the latest code.

PR #1077, opened 2017-10-20
small Docker output enhancement

max2711:
In Docker daemon mode the container outputs crond messages every 15 min. This PR removes the unused run-parts crontab jobs and unneeded crond output messages. This also helps NAS systems to enter HD standby mode.


Neilpang:
we can not `grep acme.sh` here.
because `acme.sh --cron` command can output many messages that don't contain `acme.sh` word.

max2711:
The commit does not grep the crond messages or ` acme.sh --cron` output. It removes the cronjobs in the crontab that generate the unneeded crond messages.

Neilpang:
good

PR #1111, opened 2017-11-12
Adding tinydns format when printing the TXT record needed for DNS verification

kg4zow:
This PR causes `acme.sh` to print the TXT record needed for DNS verification, in the format needed by [tinydns](https://cr.yp.to/djbdns/tinydns.html).

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3524,6 +3524,7 @@ issue() {
           _info "$(__red "Add the following TXT record:")"
           _info "$(__red "Domain: '$(__green "$txtdomain")'")"
           _info "$(__red "TXT value: '$(__green "$txt")'")"
+          _info "$(__red "tinydns: $(__green "'${txtdomain}:${txt}:300")")"
           _info "$(__red "Please be aware that you prepend _acme-challenge. before your domain")"
           _info "$(__red "so the resulting subdomain will be: $txtdomain")"
           continue


neilpang:
Acme.sh should be neutral/agnostic, shouldn't it? If I use TinyDNS, will I still need to add other specific DNS services like XXdns?

kg4zow:
I wish you had asked about this several years ago.

[djbdns](https://cr.yp.to/djbdns.html) is a DNS server package, similar to BIND but better in many ways. `tinydns` is the "authoritative nameserver" component of djbdns, it's not a *service* with an API to add or remove records. tinydns uses a [single `data` file](https://cr.yp.to/djbdns/tinydns-data.html) to hold all DNS records, and "compiles" that into a `data.cdb` file that the running `tinydns` process actually reads from.

Think about servers running tinydns the same way you'd think about servers running BIND, except you don't have to parse a bunch of config files just to figure out which data file to add the verification record to, then parse the contents of that file to ensure you're adding the `IN TXT` line to the right *part* of the file (think about how `$ORIGIN` lines work) ... you just add a single line to a single file, and run a single command.

Another significant difference between djbdns and BIND is that djbdns (tinydns) doesn't support "zone transfers" the way that BIND does. Instead, if you need multiple tinydns servers to serve the same data, you copy the `data` file to each server whenever the data is updated. This can be done using any mechanism you like - I use `scp` with SSH key authentication, as documented [here](https://qmail.jms1.net/djbdns/replication.shtml).

Anyway.

There are two ways to use `acme.sh` with `tinydns` ...

* Use "manual DNS" mode with the "`--yes-i-know..." option, find the TXT record *value* from the output, manually build the data line that `tinydns` needs, then rebuild the `data.cdb` file like you have to do with every change in DNS records.
* Write a DNS plugin to do this automatically.

When I first started using `acme.sh`, I added the change from this PR to save myself from having to manually build the TXT record in the `tinydns` data file. At the time I was managing about 40 certs, so this one-line change saved me about half an hour.

After you rejected this PR, I looked into writing a DNS plugin, and ended up writing one that not only handles adding and removing records, but also provides a "commit" function (see #1173) so that if I'm renewing certificates with several names, instead of "add/push, add/push, add/push", the high-level process became "add, add, add, push". This was a HUGE improvement, since "add" takes a fraction of a second, but "push" can take anywhere from 10 seconds to two minutes, depending on how many remote nameservers are involved and how long it takes to copy the file to each one.

Again, this PR just adds a single `_info` call that prints the actual TXT record, in the format that `tinydns` needs. It allows the user to copy/paste the entire record, rather than having to manually build every record by hand (which is the most labour-intensive part of the process).

neilpang:
Fair question, and the answer is better than I made it sound: acme.sh already ships plugins of exactly this shape. `dns_nsd.sh` appends the record to a zone file and runs `Nsd_Command` (e.g. `sudo nsd-control reload`); `dns_maradns.sh` appends to the zone file and HUPs duende. Neither is a service with an API either -- both are local file-based servers, same as tinydns. So `dns_tinydns.sh` (append to `data`, run your make/push from a config var) is squarely in scope, and you already have one written. A PR would be welcome -- tinydns users would get add and remove automatically, rather than a line to copy/paste.

That's the reasoning behind keeping the print out of core: manual mode shows the canonical name and value, and a format-specific line has no natural stopping point (BIND, NSD, MaraDNS, knot, PowerDNS each have their own). The nsd and maradns users ended up with plugins instead.

On #1173: the commit hook only pays off once the plugin exists -- `dns_nsd.sh` reloads per record today, which is the accepted tradeoff. If `dns_tinydns.sh` lands first, the commit hook becomes a measurable optimization on top of it, which makes it much easier to justify the rebase.

PR #1117, opened 2017-11-16
fix: updated DEFAULT_AGREEMENT to latest URL so it works

aphexddb:


RockyTV:
This can be fixed by reinstalling acme.sh from git. I tried using an existing install and it would fail every time until I installed acme.sh from git.

Neilpang:
DEFAULT_AGREEMENT is removed now.

PR #1136, opened 2017-12-07
 default to domain as friendlyName for --toPkcs

freddy36:
This makes the --toPkcs mode more useful by using the domain as default for the PKCS frindyName.
Before this it wasn't set at all, so this shouldn't break anything.


freddy36:
@Neilpang you've any comment on this?

PR #1140, opened 2017-12-09
added dns api support for servercow

jhartlep:
added [dns api](https://cp.servercow.de/plugin/support_manager/knowledgebase/view/34/dns-api-v1/) support for [servercow](https://servercow.de)


jhartlep:
hmm... can anybody tell me, why the build fails?

jhartlep:
found it - and hopefully now fixed...

Issue #1122, opened 2017-11-27
Digital Ocean DNS API not tidying up


xyleth:
Hello,

When I'm using the Digital Ocean DNS API to issue certificates the process mostly works and the cert gets issues, but it fails at the end with the following error:

[Mon 27 Nov 10:09:14 UTC 2017] Using digitalocean dns validation - remove record
[Mon 27 Nov 10:09:16 UTC 2017] no record and no nextpage in digital ocean DNS removal
[Mon 27 Nov 10:09:16 UTC 2017] Error removing txt for domain:_acme-<DOMAIN>

and leaves behind the validation TXT records in my DNS Zone file which need to be cleared out by hand.  I've run a debug on a test cert request and include the output below.  I'm afraid my shell scripting isn't up to debugging the code directly but if there's anything else I can do to help please let me know - this is an great tool and I'm happy to provide feedback.

I am using the latest version (updated last night) and I've searched the issues page for similar key words and can't find anything hence raising the new ticket.  The only change I have made to the debug output is to obscure my API key with DEADBEEF for reasons that I hope are obvious.

Steps to reproduce
------------------
acme.sh --issue --dns dns_dgon -d test.local.xyleth.co.uk --debug 2

Debug log
-----------------
[Mon 27 Nov 10:26:00 UTC 2017] Lets find script dir.
[Mon 27 Nov 10:26:00 UTC 2017] _SCRIPT_='/home/pi/.acme.sh/acme.sh'
[Mon 27 Nov 10:26:00 UTC 2017] _script='/home/pi/.acme.sh/acme.sh'
[Mon 27 Nov 10:26:00 UTC 2017] _script_home='/home/pi/.acme.sh'
[Mon 27 Nov 10:26:00 UTC 2017] Using config home:/home/pi/.acme.sh
[Mon 27 Nov 10:26:00 UTC 2017] LE_WORKING_DIR='/home/pi/.acme.sh'
[Mon 27 Nov 10:26:00 UTC 2017] Using config home:/home/pi/.acme.sh
[Mon 27 Nov 10:26:00 UTC 2017] ACME_DIRECTORY='https://acme-v01.api.letsencrypt.org/directory'
[Mon 27 Nov 10:26:00 UTC 2017] _ACME_SERVER_HOST='acme-v01.api.letsencrypt.org'
[Mon 27 Nov 10:26:00 UTC 2017] DOMAIN_PATH='/home/pi/.acme.sh/test.local.xyleth.co.uk'
[...]

PR #1142, opened 2017-12-11
Fix problem that digitalocean api failed to remove record

maomihz:
issue #1122

Changed regex from `\d` to `[[:digit:]]`, hopefully it's all good.

Also use `_saveaccountconf_mutable` instead of `_saveaccountconf`.

### DIFF
--- a/dnsapi/dns_dgon.sh
+++ b/dnsapi/dns_dgon.sh
@@ -20,12 +20,22 @@
 dns_dgon_add() {
   fulldomain="$(echo "$1" | _lower_case)"
   txtvalue=$2
+
+  DO_API_KEY="${DO_API_KEY:-$(_readaccountconf_mutable DO_API_KEY)}"
+  # Check if API Key Exist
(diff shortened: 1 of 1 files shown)


GawainLynch:
This resolved the issue for me, and _seems_ to be working correctly. 

Thanks, @maomihz, I wasn't looking forward to having to manually go through and clean up :+1:

Neilpang:
please do not use `[[:digit:]]`,  use `[0-9]` instead.

maomihz:
I changed that.

itssimple:
I would do the API key control as a separate function, so that it is easier to change that code. But other than that the changes look good (And I look forward to not clearing those txt-records manually)

EnJens:
When using challenge-domain for multiple domains, this bug breaks issuance eventually.
I ended up with 21 TXT records and apparently that breaks letsencrypts ability to verify domains ;)
Any ETA on getting this in? @Neilpang

PR #1187, opened 2018-01-15
Add InternetX autoDNS dns api support

auerswald:


Neilpang:
please add it into the dns api list of the main readme file too.
Thanks.

Neilpang:
Is the autoDNS only used in InternetX ?  or   is there any other companies use autoDNS ?

auerswald:
Internetx manages millions of domains worldwide with its domain management software called "autoDNS" via a reller program. All functions in this software can be used via an xml api called "AutoDNS XML API".

the hoster hosteurope has a dns administration in its domain administration software and also calls it "autoDNS". i don't know if it is the same system.

Issue #1217, opened 2018-01-23
--standalone with --local-address arg bind to 0.0.0.0


hcouplet:
Steps to reproduce
------------------
On a machine with multiple public ip and 0.0.0.0:80 already used.

/root/.acme.sh/acme.sh --issue -d my.domain.com \
--standalone --local-address X.X.X.X --listen-v4

In the code ,there is a comment just before calling socat saying :   #todo  listen address

stevebovy:
Strange; I just did an upgrade,  and I did not get any updated changes

stevebovy:
The fix for this must also be applied the the socket-check;  because a socket open on one interface:80
should not effect a socket:xx open on a different interface;  if this is fixed correctly we should not
get an error warning

stevebovy:
>> UPGRADE does not indicate that this is FIXED <<<

./acme.sh --home . --upgrade
[Fri Feb 16 20:27:57 PST 2018] Installing from online archive.
[Fri Feb 16 20:27:57 PST 2018] Downloading https://github.com/Neilpang/acme.sh/archive/master.tar.gz
[Fri Feb 16 20:27:59 PST 2018] Extracting master.tar.gz
tar: warning: skipping header 'g'
[Fri Feb 16 20:28:00 PST 2018] Installing to .
cp: 'acme.sh' and './acme.sh' are the same file
[Fri Feb 16 20:28:00 PST 2018] Install failed, can not copy acme.sh
[Fri Feb 16 20:28:00 PST 2018] Upgrade success!
admin-280@RT-AC66U_B1:/tmp/mnt/Data/acmesh#

stevebovy:
Warning !!!

#The patched Merged above is incorrect

stevebovy:
The following additional change should be made >>

>> Add $_local_address

_ss() {
  _port="$1"

  if _exists "ss"; then
    _debug "Using: ss"
    ss -ntpl 2>/dev/null | grep "$_local_address:$_port "
    return 0
  fi
  
  if _exists "netstat"; then
    _debug "Using: netstat"
    if netstat -h 2>&1 | grep "\-p proto" >/dev/null; then
      #for windows version netstat tool
      netstat -an -p tcp | grep "LISTENING" | grep "$_local_address:$_port "
    else
      if netstat -help 2>&1 | grep "\-p protocol" >/dev/null; then
        netstat -an -p tcp | grep LISTEN | grep "$_local_address:$_port "
[...]

PR #1218, opened 2018-01-23
Standalone listen address fix. #1217

hcouplet:
This fix issue #1217 .

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1984,9 +1984,15 @@ _startserver() {
     _NC="$_NC -6"
   fi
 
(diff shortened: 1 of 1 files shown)


acme.sh
```diff
@@ -1984,9 +1984,15 @@ _startserver() {
     _NC="$_NC -6"
   fi
 
+  SOCAT_OPTIONS=TCP-LISTEN:$Le_HTTPPort,crlf,reuseaddr,fork
+
+  #Adding bind to local-address
+  if [ "$_local_address" ]; then
```
Neilpang: do not use '$_local_address'.   please use the local variable "$ncaddr" instead.

PR #1226, opened 2018-01-26
Create Edgemax deploy script.

orangevan:


deploy/edgemax.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
Neilpang: please change the shebang to the same as the others'

deploy/edgemax.sh
```diff
@@ -0,0 +1,80 @@
...
+###
+################################################################
+
+#This file name is "edgemax.sh"
+#So, here must be a method   edgemax_deploy()
+#Which will be called by acme.sh to deploy the cert
+#returns 0 means success, otherwise error.
+
+########  Public functions #####################
+function atexit() {
```
Neilpang: remove the `function` keyword

deploy/edgemax.sh
```diff
@@ -0,0 +1,80 @@
...
+    SAVE=${vyatta_sbindir}/vyatta-save-config.pl
+    _info "Setting Certificate parameter."
+    $SET service gui cert-file /config/auth/le-cert.pem
+    $COMMIT
+    $SAVE
+  else
+    _info "EdgeMax cert-file already set to $lighttpd_pem"
+  fi
+  _info Restarting lighttpd
+  sudo kill -SIGTERM "$(cat /var/run/lighttpd.pid)"
```
Neilpang: do the `sudo` need to input password ?
do not use sudo in the script.
if you really to do something as root,   install and run acme.sh as root.  do not use sudo in your script.

orangevan: On these particular devices (edgemax), in their build of linux sudo does not require a password.  It is however required to restart the web service that the certificate is being installed to though i'd rather all the other tasks the script does not run as root.  I will double verify that I can't reboot the webserver as the standard user but I am fairly confident that sudo is required.

deploy/edgemax.sh
orangevan: I'm confused on this request.  When I look at all of the other deploy scripts, they are just 
#!/bin/bash
Same as this one.
What do you want me to change?

deploy/edgemax.sh
orangevan: removed. Tested, still works.  Thanks.

deploy/edgemax.sh
orangevan: I have checked and I cannot reboot the web service (needed for deployment) without root privileges but for installation and execution of tasks/cronjobs etc ideally on these devices i'd like them to be user level.

deploy/edgemax.sh
Neilpang: https://github.com/Neilpang/acme.sh/blob/master/deploy/vsftpd.sh#L1

deploy/edgemax.sh
Neilpang: If the sudo requires inputing password,  the cronjob will be broken.
Never never never use sudo.

orangevan: the sudo does not require inputting password

Issue #1214, opened 2018-01-22
Instructions to generate tsig key for knot api not working.


skid9000:
Hi,

In the README.md of the dnsapi, the knot part said to generate a tsig key with :

`keymgr tsig generate acme_key algorithm hmac-sha512 > /etc/knot/acme.key`

However, the command return (in the file)

`Invalid parameter: acme_key
Error (invalid parameter)`

<img src="https://puush.tuto-craft.com/5w972w.png">

Tested on 2 servers running Debian 9.3 with Knot 2.6.4

skid9000:
OK, i fond THE solution : 

`keymgr tsig generate -t acme_key hmac-sha512`

This command is working !

skid9000:
Anyone can open a PR to fix this ?

PR #1248, opened 2018-02-07
Fix the command to generate tsig key for knot api

skid9000:
The command in the readme didn't work as i said in #1214 so i fixed it.


skid9000:
Can anyone merge this ?
Thx.

PR #1289, opened 2018-02-18
added option to set alias for PKCS12

eddyhub:
I need to set the alias, so I added this option. Before this the alias was 1 and is still the default with --name parameter.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1211,16 +1211,17 @@ _toPkcs() {
 toPkcs() {
   domain="$1"
   pfxPassword="$2"
+  pfxName="$3"
   if [ -z "$domain" ]; then
-    _usage "Usage: $PROJECT_ENTRY --toPkcs -d domain [--password pfx-password]"
+    _usage "Usage: $PROJECT_ENTRY --toPkcs -d domain [--password pfx-password] [--name pfx-name/pfx-alias]"
     return 1
   fi
 
-  _isEcc="$3"
+  _isEcc="$4"
 
   _initpath "$domain" "$_isEcc"
 
-  _toPkcs "$CERT_PFX_PATH" "$CERT_KEY_PATH" "$CERT_PATH" "$CA_CERT_PATH" "$pfxPassword"
+  _toPkcs "$CERT_PFX_PATH" "$CERT_KEY_PATH" "$CERT_PATH" "$CA_CERT_PATH" "$pfxPassword" "$pfxName"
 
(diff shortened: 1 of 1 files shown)


acme.sh
```diff
@@ -5746,6 +5748,10 @@ _process() {
         _password="$2"
         shift
         ;;
+      --name)
```
Neilpang: do not use `--name`.
use something like: `--pfx-alias` instead.

acme.sh
```diff
@@ -6016,7 +6022,7 @@ _process() {
     uninstallcronjob) uninstallcronjob ;;
     cron) cron ;;
     toPkcs)
-      toPkcs "$_domain" "$_password" "$_ecc"
+      toPkcs "$_domain" "$_password" "$_name" "$_ecc"
```
Neilpang: place the new parameter at the last position.

acme.sh
```diff
@@ -1192,13 +1192,13 @@ _toPkcs() {
...
-  pfxName="$6"
+  pfxAlias="$6"
   pfxCaname="$7"
 
   if [ "$pfxCaname" ]; then
-    ${ACME_OPENSSL_BIN:-openssl} pkcs12 -export -out "$_cpfx" -inkey "$_ckey" -in "$_ccert" -certfile "$_cca" -password "pass:$pfxPassword" -name "$pfxName" -caname "$pfxCaname"
-  elif [ "$pfxName" ]; then
-    ${ACME_OPENSSL_BIN:-openssl} pkcs12 -export -out "$_cpfx" -inkey "$_ckey" -in "$_ccert" -certfile "$_cca" -password "pass:$pfxPassword" -name "$pfxName"
+    ${ACME_OPENSSL_BIN:-openssl} pkcs12 -export -out "$_cpfx" -inkey "$_ckey" -in "$_ccert" -certfile "$_cca" -password "pass:$pfxPassword" -name "$pfxAlias" -caname "$pfxCaname"
+  elif [ "$pfxAlias" ]; then
```
Neilpang: I don't understand this change.
Is the `pfxName` same as `pfxAlias` ?

eddyhub: As far as i know the alias field is a unique string to identify the key entry.  I am used to oracles keystore tool and oracle named this field _alias_. I reverted the change.

eddyhub:
@Neilpang: Is there any way I can help you to get this changes merged?

PR #1292, opened 2018-02-19
Let AWS DNS API driver pull creds from instance metadata

mal:
Add code to allow the AWS DNS API driver pull the necessary IAM role credentials from the AWS EC2 instance metadata endpoint when required.

This is designed as a non-breaking change as it only takes effect when the driver was otherwise going to fail with missing credentials. It also fails fast (in the event that there's no attached role, and/or it's not running on an EC2 instance) inline with the behavior established in `botocore` .

[Final fallback reference](https://github.com/boto/botocore/blob/9b46cfbecbc86f700dfc59e546178d57b390fb28/botocore/credentials.py#L98) - It's the last provider, only activated if all others fail
[Metadata request timeout reference](https://github.com/boto/botocore/blob/30c96a24796e423b25108e9c72d2c541d4bdf617/botocore/session.py#L95-L99) - 1 second by default

Additionally if activated it prevents the key and secret from being saved. This is because the credentials pulled from the metadata are only valid for a certain period of time before they are rotated. Not saving them ensures the code will fallback to fetching the instance metadata again fresh for renewals and removals.

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -9,7 +9,6 @@
 
 AWS_HOST="route53.amazonaws.com"
 AWS_URL="https://$AWS_HOST"
-AWS_METADATA_URL="http://169.254.169.254/latest/meta-data"
 
 AWS_WIKI="https://github.com/Neilpang/acme.sh/wiki/How-to-use-Amazon-Route53-API"
 
@@ -36,7 +35,7 @@ dns_aws_add() {
   fi
 
   #save for future use, unless using a role which will be fetched as needed
-  if [ -z "$_using_instance_role" ]; then
+  if [ -z "$_using_role" ]; then
     _saveaccountconf_mutable AWS_ACCESS_KEY_ID "$AWS_ACCESS_KEY_ID"
     _saveaccountconf_mutable AWS_SECRET_ACCESS_KEY "$AWS_SECRET_ACCESS_KEY"
   fi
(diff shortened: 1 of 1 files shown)


myoung34:
This only works on EC2 instances, will not work as expected under ECS containers which use a different method. I wont ask you to scope creep that by adding it, but maybe change the naming convention for others so it seems more obvious?

maybe by specifically naming the things with `ec2` in it?

mal:
I'll move the `AWS_METADATA_URL` var because I agree that right now that's ambiguous, but hopefully you'll agree that the `instance` specifier elsewhere is sufficient as it is in `botocore`. For reference they have [`InstanceMetadataFetcher` and `ContainerMetadataFetcher`](https://github.com/boto/botocore/blob/9b46cfbecbc86f700dfc59e546178d57b390fb28/botocore/credentials.py#L42-L43) classes to handle the two cases.

Future work could add a `_use_container_role` function to extend this work in a clear understandable way.

myoung34:
@mal I actually removed my comment because after re reading it `instance` was kind of as good as I was thinking. But that's a good change to the var name as well

mal:
FWIW I gave the docs on ECS metadata a look over and it should be pretty simple to add, but I'm not in a position to be able to test it, so rather than add it to this PR, if this one gets accepted I'll make a separate PR for the ECS support and request that someone test it in my place.

PR #1305, opened 2018-02-24
DNSMadeEasy ACMEv2 support

justmwa:
fixes for new API, along with non-english locales (timestamp generated for auth would fail)

### DIFF
--- a/dnsapi/dns_me.sh
+++ b/dnsapi/dns_me.sh
@@ -43,17 +43,17 @@ dns_me_add() {
     return 1
   fi
 
-#    _info "Adding record"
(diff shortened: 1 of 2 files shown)


dnsapi/dns_me.sh
```diff
@@ -45,8 +45,7 @@ dns_me_add() {
 
   count=$(printf "%s\n" "$response" | _egrep_o "\"totalRecords\":[^,]*" | cut -d : -f 2)
```
Neilpang: is the "count" still needed ?

Neilpang:

unclespeedo:
Is there anything specific that dnsmadeeasy environments should test? This is what we use and I'm happy to test/provide feedback if requested.

PR #1325, opened 2018-03-03
Added Conditions to call apache and created module to work with pdns Mysql backend.

danc403:
acme.sh:
Added apache2 and httpd command conditions to apachePath()

dns_pdnsMysql.sh:
Calls pdns Mysql backend to verify Let's Encrypt.

dnsapi/README.md:
Added section 5a describing usage of dns_pdnsMysql.sh

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2452,8 +2452,12 @@ _apachePath() {
   if ! _exists apachectl; then
     if _exists apache2ctl; then
       _APACHECTL="apache2ctl"
+    elif _exists apache2; then #added
+      _APACHECTL="apache2" #added
+    elif _exists httpd; then #added
+      _APACHECTL="httpd" #added
     else
-      _err "'apachectl not found. It seems that apache is not installed, or you are not root user.'"
+      _err "'apachectl (or apache2 or httpd) not found. It seems that apache is not installed, or you are not root user.'"
       _err "Please use webroot mode to try again."
       return 1
     fi
--- a/dnsapi/dns_pdnsMysql.sh
+++ b/dnsapi/dns_pdnsMysql.sh
@@ -21,14 +21,14 @@ dns_pdnsMysql_add() {
   txtvalue=$2
 
   if [ -z "$PDNS_Host" ]; then
-    PDNS_Url=""
+    PDNS_Host=""
     _err "You didn't specify PowerDNS Mysql address."
     _err "Please set PDNS_Host and try again."
     return 1
   fi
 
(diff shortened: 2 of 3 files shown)


danc403:
Sorry but I don't understand this error:

The command "if [ "$TRAVIS_OS_NAME" = "linux" ]; then git diff --exit-code && echo "shfmt OK" ; fi" exited with 1.

Can you explain what it means?
All other tests seem to pass except this one.
Thank You,
Dan

dnsapi/dns_pdnsMysql.sh
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env sh
+
+#PowerDNS Mysql backend
+#
+#
+PDNS_Host="example.com"
```
Neilpang: do not hardcode these values.

dnsapi/dns_pdnsMysql.sh
```diff
@@ -0,0 +1,190 @@
...
+PDNS_Port=3306
+PDNS_User="username"
+PDNS_Pass="password"
+PDNS_Database="powerdns"
+PDNS_Ttl=60
+
+DEFAULT_PDNS_TTL=60
+
+if ! _exists mysql; then
+  _err "'mysql not found. It seems that mysql client is not installed.'"
```
Neilpang: move this check into the  add() function.

Neilpang:
please check this guide:

acme.sh
```diff
@@ -2452,8 +2452,12 @@ _apachePath() {
   if ! _exists apachectl; then
     if _exists apache2ctl; then
       _APACHECTL="apache2ctl"
+    elif _exists apache2; then #added
```
Neilpang: why is this change needed ?

Neilpang:
what is the difference with pdns api:  https://github.com/Neilpang/acme.sh/tree/master/dnsapi#5-use-powerdns-embedded-api-to-automatically-issue-cert

danc403:
The changes in acme.sh are for backwards compatibility. httpd is the
Apachectl equivalent on all my CentOS 7 machines.

I will look at the DNS script today. Made those changes on my phone but it
may have not come out like I planned.

danc403:
The difference in pdns is also backwards compatible. Older versions had no
api. It is only useful for rhel machines that still ship with older
versions of pdns. Rhel 7 still only goes to pdns v3 without adding
repository which many businesses won't allow.

PR #1348, opened 2018-03-12
Add Rancher server cert deploy hook

CMoH:
Here's a deploy hook script integrating acme.sh cert deployment with [rancher](https://rancher.com).

 I'm about to make a renew hook as well, but there seem to be no appropriate place for it in the source tree. Are there any plan to add one?

### DIFF
--- a/deploy/rancher.sh
+++ b/deploy/rancher.sh
@@ -1,0 +1,109 @@
+#!/usr/bin/env sh
+
+# Deploy certificates to rancher environmentsx
+
+# here are the defaults, overridable via env vars
+#
+#export RANCHER_CONFIG=${HOME}/.rancher/cli.json
+#export RANCHER_ENV=
+
+# usage:
+# - download rancher-cli from your rancher server and use it to create cli.json
+#   the format of the file is quite simple, so you can just create your own
+# ! also run chmod 600 ~/.rancher/cli.json, since rancher-cli doesn't
+# - for multiple servers override RANCHER_CONFIG
+# - for multiple environments on a server set RANCHER_ENV appropriately
+#   otherwise the one selected within cli.json is used
+
+# example
+# acme.sh --deploy -d my.website.com --deploy-hook rancher --debug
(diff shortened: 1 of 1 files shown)


Neilpang:
@CMoH   what is a ` renew hook` ?  if you want to do something when the cert is renewed,  you can do it in the deploy hook.  coz the deploy hook will be called everytime when the cert is renewed.

CMoH:
@Neilpang I didn't understand this aspect. In my mind I have two different scenarios: one where I have to "deploy" a new certificate onto one or multiple targets (sometimes a rancher server/environment, sometimes a baremetal service) and another scenario where that certificate was already deployed and is only "renewed". The difference between the two is that the renew hook would examine the target environment and redeploy the certificate only if it belongs here, while the deploy hook deploys it no matter what the existing state is. In other words, the deploy hook also assigns a certificate to an environment, and the renew hook relies on that information to know it has to redeploy it there or not.

My intention was to have an imperative command `acme.sh ... --deploy ...`  which deploys a new certificate, and an automated cron job with `acme.sh ... --renew-hook ...` where the renew hook verifies where that cert is already deployed and updates it, restarts services etc. The fact

Would you see any value in this perspective? If acme.sh sticks to having a single script used both in initial deployment and in renewals then I have to modify this script and issue a new PR.

Neilpang:
The deploy hook or renew hook or another parameters are saved in the domain conf.  When the cert is renewed in future, the hooks will be re-called again.

Actually,  there is no `renew` in ACME protocol,  we just re-issue a new cert in 60 days, and run all the hooks with all the parameters you gave when you issued the cert for the first time.

So,
1. Do not modify the default cron job,    It's silly to add `--renew-hook` parameter to the cronjob, it's useless at all.  Any hooks or parameters will be gracefully processed by the default cronjob.  It's smart enough.
2. Everytime you deploy the cert, don't forget to reload/restart the service. otherwise the cert may not be used.

CMoH:
Thanks. Your insight puts quite a dent in my plan though. It is doable, but I have to rework the entire idea. I'll get back with a PR later.

PR #1358, opened 2018-03-13
Properly escape domains

zauguin:
Fixes expansion of wildcards in domain names if some files in the `pwd` match the wildcard.


Neilpang:
thanks for your PR,  but I just fixed it in another way.
please try again with the latest master code.
Thanks.

PR #1379, opened 2018-03-14
If ACCOUNT_URL is empty when we need it, re-read it. Fixes #1360

### LINKED ISSUE #1360: fix acme v2 issue
https://community.letsencrypt.org/t/acme-v2-missing-jws-header-with-acme-sh/55665/3

jim-p:
Fix #1360 by re-reading the account URL if it is empty when needed for ACME v2

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1826,6 +1826,17 @@ _send_signed_request() {
       if [ "$url" = "$ACME_NEW_ACCOUNT" ] || [ "$url" = "$ACME_REVOKE_CERT" ]; then
         protected="$JWK_HEADERPLACE_PART1$nonce\", \"url\": \"${url}$JWK_HEADERPLACE_PART2, \"jwk\": $jwk"'}'
       else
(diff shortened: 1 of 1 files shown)


Neilpang:
but why is ACCOUNT_URL  empty  here ?

jim-p:
It may be a quirk of how it is executed on pfSense, in which case I can just keep that fix local there. It almost always fails the first time it is run but works the second time. The difference being that the first time, the account data, csr, keys, etc, have just been written to the filesystem for the first time. When run the second time, those files already exist.

The logs show the ACCOUNT_URL is there earlier in the process, but somehow by the point it is needed, it is no longer set in the environment. Following through the code I didn't see anything obvious that pointed to why, however, but re-reading it worked and seemed like a better solution since it did not rely on something else reading it sooner and happening to place it in the environment.

Neilpang:
@jim-p   please try again with latest code. it's fixed.

PR #1392, opened 2018-03-16
DNS-Manual: better documentation in script

cfstras:
Change the hint that tells you how to use DNS manual (second run needs to be --renew)


cfstras:
:+1:

FernandoMiguel:
i think this should actually be avoided at all cost.
it doesnt auto renew, will cause issues to users in 90 days.

it should be made VERY clear and VERY hard to run --dns when no api provider is given

cfstras:
There is a clear notification in the output that says it doesn't auto-renew.

If your DNS setup makes it hard to use the DNS-API, for whatever reason, and you're just trying to get LE to work for a prototype before you start working on using the API (which might need some development), this is a really useful feature.

Please don't make stuff hard to use on purpose.

PR #1394, opened 2018-03-16
dns_pdns.sh: Allow "." as root zone

rafaelgieschke:
This pull request allows to use the (root) zone "." as zone in dns_pdns.sh. Differently from all other zones, the zone "." is specially encoded as "=2E" (Quoted-Printable) in PowerDNS REST URLs.

### DIFF
--- a/dnsapi/dns_pdns.sh
+++ b/dnsapi/dns_pdns.sh
@@ -90,7 +90,7 @@ set_record() {
   full=$2
   txtvalue=$3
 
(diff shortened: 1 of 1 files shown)


Neilpang:
will this bring in any compatibility issue ?

rafaelgieschke:
This should (and did so in my tests) behave the same as before when no zone "." exists in PowerDNS.

E.g., issuing a certificate for "test.example.com":

Previous behavior: check for zone "test.example.com.", check for zone "example.com.", check for zone "com.", error.

Behavior with this PR: check for zone "test.example.com.", check for zone "example.com.", check for zone "com.", check for zone "." (encoded as "=2E"), error.

rafaelgieschke:
Thanks! 🎉

PR #1419, opened 2018-03-19
Fix removing record bug on alidns

sempr:
If request  `example.com` and `*.example.com` at the same time, it will create two TXT record `_acme-challenge.example.com` on the dns.

When remove the record, it will get two at the first request and failed to delete the first, as well as the second one.

So just delete one record and then the second will work well.

More beautiful fix: 
1. If I can get the record value, i can delete each value at the time it should be delete.
1. Just loop the record ids and remove all the records with the same names.

### DIFF
--- a/dnsapi/dns_ali.sh
+++ b/dnsapi/dns_ali.sh
@@ -185,7 +185,7 @@ _clean() {
     return 1
   fi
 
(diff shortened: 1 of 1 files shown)


Neilpang:
删除记录的时候,  最好根据指定的 txt 值来删除, 而不要写死只删除第一条, 或者最后一条.
txt 记录值是通过第二个参数传进来的.

PR #1429, opened 2018-03-21
Fixed DNSAPI for PowerDNS to support wildcard certificates

softcat:
Modified DNSAPI for PowerDNS to support wildcard certificates

### DIFF
--- a/dnsapi/dns_pdns.sh
+++ b/dnsapi/dns_pdns.sh
@@ -88,9 +88,20 @@ set_record() {
   _info "Adding record"
   root=$1
   full=$2
-  txtvalue=$3
+  new_challenge=$3
 
-  if ! _pdns_rest "PATCH" "/api/v1/servers/$PDNS_ServerId/zones/$root" "{\"rrsets\": [{\"changetype\": \"REPLACE\", \"name\": \"$full.\", \"type\": \"TXT\", \"ttl\": $PDNS_Ttl, \"records\": [{\"name\": \"$full.\", \"type\": \"TXT\", \"content\": \"\\\"$txtvalue\\\"\", \"disabled\": false, \"ttl\": $PDNS_Ttl}]}]}"; then
+  _pdns_rest "GET" "/api/v1/servers/$PDNS_ServerId/zones/$root"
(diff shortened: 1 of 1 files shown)


softcat:
Ok, I'll have a look into the travis errors

softcat:
They should be fixed now.

dnsapi/dns_pdns.sh
```diff
@@ -88,17 +88,22 @@ set_record() {
   _info "Adding record"
   root=$1
   full=$2
-  txtvalue=$3
+  new_challenge=$3
+
+  _pdns_rest "GET" "/api/v1/servers/$PDNS_ServerId/zones/$root"
+  _record_string=""
+  _build_record_string "$new_challenge"
+  _existing_challenges=$(echo "$response" | _normalizeJson | grep -Po "\"name\":\"$fulldomain\\K.*?}]" | grep -Po 'content\":\"\\"\K[^\\]*')
```
Neilpang: do not use `grep -Po`,  please use our function `_egrep_o()` instead.

Neilpang:
how about the `dns_pdns_rm()` function ?
please remove the record by the txt value.
https://github.com/Neilpang/acme.sh/issues/1261

softcat:
> how about the dns_pdns_rm() function ?
please remove the record by the txt value.

I didn't touch the dns_pdns_rm() function, it always removed all "_acme_challenge" from the $root zone disregarding their content. The PowerDNS API does not support deleting entries by their content, just by name.

But I will use the adapted code from adding entries, so I will parse through all entries, delete all from PowerDNS and add the not to be removed entris again. Unfortunately not really elegant.

dnsapi/dns_pdns.sh
softcat: Ok, I will have to adapt the regex because _egrep_o() seems not to support non greedy modifiers which is enabled by the -P flag. Working on it...

softcat:
Grep is now not used anymore and I reworked the dns_pdns_rm() function to remove the records by txt value.

ikheetleon:
Great, thanks a lot! I've just tested the new pdns api script and it works for me.

PR #1437, opened 2018-03-23
Disable line wrapping on base64 conversion to prevent auth failure

james-gibson:
Base64 by default wraps lines at 76 chars, the line break was resolving as the incorrect auth string.
~See http://man7.org/linux/man-pages/man1/base64.1.html for details on cmd flag~

Edit:  Amended commit to reflect pr feedback.

### DIFF
--- a/dnsapi/dns_namecom.sh
+++ b/dnsapi/dns_namecom.sh
@@ -123,7 +123,7 @@ _namecom_login() {
   # Auth string
   # Name.com API v4 uses http basic auth to authenticate
   # need to convert the token for http auth
-  _namecom_auth=$(printf "%s:%s" "$Namecom_Username" "$Namecom_Token" | base64)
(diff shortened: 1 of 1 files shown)


dnsapi/dns_namecom.sh
```diff
@@ -123,7 +123,7 @@ _namecom_login() {
   # Auth string
   # Name.com API v4 uses http basic auth to authenticate
   # need to convert the token for http auth
-  _namecom_auth=$(printf "%s:%s" "$Namecom_Username" "$Namecom_Token" | base64)
+  _namecom_auth=$(printf "%s:%s" "$Namecom_Username" "$Namecom_Token" | base64 -w 0)
```
Neilpang: we have a function `_base64()`,  please use that instead.

james-gibson:
@Neilpang Great catch!  I did not realize that was available, I've updated the code as requested.  Thanks for the quick feedback.

PR #1440, opened 2018-03-23
Improve readability of list command

ATLief:
Display the default key length of domain certificates when listing certificates if no key length is explicitly set. Wrap value in square brackets to indicate default value, or double quotes to indicate an explicit value.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4534,7 +4534,12 @@ list() {
         _initpath "$d" "$_isEcc"
         if [ -f "$DOMAIN_CONF" ]; then
           . "$DOMAIN_CONF"
-          printf "%s\n" "$Le_Domain${_sep}\"$Le_Keylength\"${_sep}$Le_Alt${_sep}$Le_CertCreateTimeStr${_sep}$Le_NextRenewTimeStr"
+          if [ -z "$Le_Keylength" ]; then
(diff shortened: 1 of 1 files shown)


ATLief:
@Neilpang I'd appreciate if you could give me some feedback.

acme.sh
```diff
@@ -4538,7 +4538,12 @@ list() {
         _initpath "$d" "$_isEcc"
         if [ -f "$DOMAIN_CONF" ]; then
           . "$DOMAIN_CONF"
-          printf "%s\n" "$Le_Domain${_sep}\"$Le_Keylength\"${_sep}$Le_Alt${_sep}$Le_CertCreateTimeStr${_sep}$Le_NextRenewTimeStr"
+          if [ -z "$Le_Keylength" ]; then
```
Neilpang: don't need so complicated.
just one line:

```
printf "%s\n" "$Le_Domain${_sep}\"${Le_Keylength:-$DEFAULT_DOMAIN_KEY_LENGTH}\"${_sep}$Le_Alt${_sep}$Le_CertCreateTimeStr${_sep}$Le_NextRenewTimeStr"
```

ATLief: Thanks! I tested it out, and it doesn't have quite the same functionality (no square brackets for default values). Do you ever intent to merge this pull request, using my original changes or the ones you suggested?

Neilpang: your original code  is too overkill.

PR #1452, opened 2018-03-27
Fix several Shellcheck SC1117 warnings.

neheb:
\ is supposed to escape several characters but not in these cases.

https://github.com/koalaman/shellcheck/wiki/SC1117

As this is a stylistic issue, I avoided several cases.


Neilpang:
I don't like it.

Issue #1086, opened 2017-10-26
split problem dns_freedns.sh


rbijos:
Hello! 
The split part of the script caused me trouble.
In this part

# split our full domain name into two parts...
  i="$(echo "$fulldomain" | tr '.' ' ' | wc -w)"
  i="$(_math "$i" - 1)"
  top_domain="$(echo "$fulldomain" | cut -d. -f "$i"-100)"
  i="$(_math "$i" - 1)"
  sub_domain="$(echo "$fulldomain" | cut -d. -f -"$i")"

it was not trimming my domain right.
I have a subdomain.domain.tld.cc (cc=country code)
It was trimming as subdomain.domain       tld.cc 
A workaround in my case was to put 2 for the first 1.
it worked as should subdomain        domain.tld.cc 

It should be another way to trim or to put a var to specify the domain or a documentation.
Thanks

mcgr0g:
guy, just read this one http://freedns.afraid.org/dynamic/ and ready copy curl script with your AUTH token to your router.

pisaman2:
Yep I'm having a similar problem with a .co.uk domain 

I'm gonna try changing the 1 to 2 now and see how that works

felixgonsug:
I manage to change the 1 for the 2 and int worked, thanks.

dkerr64:
@rbijos can you please test the freedns api script here... https://raw.githubusercontent.com/dkerr64/acme.sh/FreeDNS/dnsapi/dns_freedns.sh

dkerr64:
PR #2285 opened.

rbijos:
> @rbijos can you please test the freedns api script here... https://raw.githubusercontent.com/dkerr64/acme.sh/FreeDNS/dnsapi/dns_freedns.sh

It worked after the first time I tried, on a total of 6 tries on 3 different domains  (I think the error was on the server not the script).  I tried both domain.tld.cc and domain.tld.

First try error log. (only the error part, domain name changed to test). TXT records have been created correctly at freedns.
[qui mai 23 07:20:04 -03 2019] All success, let's return
[qui mai 23 07:20:04 -03 2019] ok, let's start to verify
[qui mai 23 07:20:04 -03 2019] Verifying: test.adv.br
[qui mai 23 07:20:04 -03 2019] d='test.adv.br'
[...]

PR #1454, opened 2018-03-27
freedns: fix domain splitting for multi-part domains

frebib:
It seems that the way domains were split before was using some unnecessarily complex maths which was broken for domains >3 parts (e.g `_acme-challenge.test.domain.com`)

I assume what the correct behaviour should be is the following:
```
sub_domain='_acme-challenge'
top_domain='everything.after.the.first.dot.com'
```

This PR fixes for domains >3 parts and still works for simple `sub.domain.com` domains

### DIFF
--- a/dnsapi/dns_freedns.sh
+++ b/dnsapi/dns_freedns.sh
@@ -51,68 +51,38 @@ dns_freedns_add() {
   # sub_domain is the element before the first dot
   top_domain="$(echo "$fulldomain" | cut -d. -f 2-)"
   sub_domain="$(echo "$fulldomain" | cut -d. -f 1)"
+  DNSdomainid=
 
-  _debug "top_domain: $top_domain"
-  _debug "sub_domain: $sub_domain"
+  while [ -z "$DNSdomainid" ]; do
+    _debug "top_domain: $top_domain"
+    _debug "sub_domain: $sub_domain"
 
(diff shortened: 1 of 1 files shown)


martgras:
Does that also work if I have separate zone for a subdomain.
Let's say I have  example.com.    I need hostnames like www.devorg.example.com mail.devorg.example.com 
www.sales.example.com etc.    
Of course I can create the records in the example.com (like www.devorg) but  the devorg team wants to manage entries in their subdomain .   For security reasons I don't want them to have control over anything else than devorg.example.com.    The solution is creating a new DNS zone for devorg.example.com   
This becomes more important with the [DNS alias mode](https://github.com/Neilpang/acme.sh/wiki/DNS-alias-mode) where I can use a subdomain for the validation records without the need to grant access to the complete example.com zone

frebib:
I'm not entirely sure, are you able to test it?
Did it work before this patch?

martgras:
I never used freedns.
Most implenations start splitting at the first dot. Then check if the domain is in the list of domains from your provider. 
If not split the domain again until you have a match.

frebib:
This addresses the issue in #1086.

As for checking each part in turn, that will need some testing.
I'd suggest a separate issue should be opened to track that? It looks like it would be reasonably significant rewrite of the code in this function

PR #1485, opened 2018-04-03
fix bad dns_ali api SignatureNonce

sadoneli:


Neilpang:
we have 2 functions: `_utc_date()` and `_time()`,  can you use one of them instead ?

Neilpang:
@sadoneli 
How is going on ?

PR #1500, opened 2018-04-08
Fix dns_selectel work with unicode domains

popstas:
For example, I want to issue certificate for domain `тест.рф` (punycode `xn--e1aybc.xn--p1ai`)
Domain not contains in response from selectel because it escaped in response: `\u0442\u0435\u0441\u0442.\u0440\u0444` and `_contains` function cannot found it.

This PR does:

- replace domain to utf before `_get_root` with `idn -u` for `_contains` works
- unescape response with `echo -e` for `_contains` works
- convert domain to punycode before selectel api request with `idn` for api works

I think that would be better to escape domain instead of unescape response, but I don't know how to convert `тест.рф` to `\u0442\u0435\u0441\u0442.\u0440\u0444`.

Tested with all of: `тест.рф`, `xn--e1aybc.xn--p1ai`, `example.com`

### DIFF
--- a/dnsapi/dns_selectel.sh
+++ b/dnsapi/dns_selectel.sh
@@ -104,7 +104,7 @@ dns_selectel_rm() {
 # _domain=domain.com
 # _domain_id=sdjkglgdfewsdfg
 _get_root() {
-  domain=$1
+  domain="$(echo "$1" | idn -u)"
 
   if ! _sl_rest GET "/"; then
     return 1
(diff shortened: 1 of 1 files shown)


popstas:
Replaced `echo -e` to `printf '%b'` due shellcheck.

dnsapi/dns_selectel.sh
```diff
@@ -104,7 +104,7 @@ dns_selectel_rm() {
 # _domain=domain.com
 # _domain_id=sdjkglgdfewsdfg
 _get_root() {
-  domain=$1
+  domain="$(echo "$1" | idn -u)"
```
Neilpang: do not use command `idn`,  please use `_idn()` function instead.

popstas: Replaced `echo "$domain" | idn` to `_idn "$domain"`, tested.
But `idn -u` converts domain from punycode to utf, `_idn` make utf -> punycode convert, I cannot replace it.

popstas:
Should I do anything else for merge PR?

PR #1501, opened 2018-04-08
Add Support for inwx.de mobile tan

jkroepke:
Add support for inwx.de 2FA/Mobile TAN authentication.

### DIFF
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -641,6 +641,14 @@ acme.sh --issue --dns dns_inwx -d example.com -d www.example.com
 
 The `INWX_User` and `INWX_Password` settings will be saved in `~/.acme.sh/account.conf` and will be reused when needed.
 
+If your account is secured by mobile tan you have also defined the shared secret.
+
+```
+export INWX_Shared_Secret="shared secret"
+```
+
+You may need to re-enable the mobile tan to gain the shared secret.
+
 ## 34. User Servercow API v1
 
 Create a new user from the servercow control center. Don't forget to activate **DNS API** for this user.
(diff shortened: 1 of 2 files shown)


Neilpang:

jkroepke:
Done.

apollo13:
@jkroepke This is now breaking for me for accounts without TFA enabled; the response contains:
```
<member><name>tfa</name><value><string>0</string></value></member>
```
for TFA disabled and if TFA is enabled:
```
<member><name>tfa</name><value><string>GOOGLE-AUTH</string></value></member>
```

This means the check here: https://github.com/Neilpang/acme.sh/blob/084de9d8e0178f7d49c01572925bcb524d0e67e3/dnsapi/dns_inwx.sh#L161 is not enough and should actually check that it is not 0

jkroepke:
Eugh, thanks for the clarification.

So, you want to make a PR or should I do it?

apollo13:
My bash knowledge is limited at best; if you could do one that would be great.

apollo13:
@jkroepke I've contacted INWX support to clarify their api docs (the description of the actual values for tfa misses); in the meantime their support told me:
```
The variable tfa returns either 'GOOGLE-AUTH' if 2FA is activated, otherwise, it returns 0 / false.
```

apollo13:
@jkroepke Did you find some time to fix the code?

jkroepke:
Sorry, I forgot it.

Hopefully it's good for now. It would be helpfully, if you can test the patch.

apollo13:
@jkroepke Thanks, this looks good to me (I do not have 2fa to test). I wonder what will happen though if inwx introduces whitespace between the xml tags?

jkroepke:
Yes. I would be break again.

I'm look for a solution but a lot of checkes in acme.sh would be break too if other services implement whitespaces.

I can't find a helper function in acme.sh which removes/trim safe whitespaces from xml.

PR #1529, opened 2018-04-17
Include schlundtech in dns_autodns.sh

tohn:
As mentioned in [this comment](https://github.com/Neilpang/acme.sh/pull/976#issuecomment-379898107), this PR outsources the AUTODNS_API, so one can use the API for InternetX and Schlund Technologies (and also updates the documentation).

### DIFF
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -680,14 +680,15 @@ And now you can issue certs with:
 acme.sh --issue --dns dns_namesilo --dnssleep 900 -d example.com -d www.example.com
 ```
 
-## 36. Use autoDNS (InternetX)
+## 36. Use autoDNS (InternetX or Schlund Technologies)
 
-[InternetX](https://www.internetx.com/) offers an [xml api](https://help.internetx.com/display/API/AutoDNS+XML-API)  with your standard login credentials, set them like so:
+[InternetX](https://www.internetx.com/) offers an [xml api](https://help.internetx.com/display/API/AutoDNS+XML-API)  with your standard login credentials. You can also use this API for [Schlund Technologies](https://www.schlundtech.de/). Set them like so:
 
 ```
 export AUTODNS_USER="yourusername"
 export AUTODNS_PASSWORD="password"
 export AUTODNS_CONTEXT="context"
+export AUTODNS_API="https://gateway.autodns.com" # or export AUTODNS_API="https://gateway.schlundtech.de"
 ```
 
 Then you can issue your certificates with:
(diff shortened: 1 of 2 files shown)


hmb:
While the idea of using the same code for different providers that happen to use the same interface is certainly a good one, it still comes with some implications we need to think about. Of course the code has to be maintained only once which as such is a great benefit already.

On the other hand let's have a look at the architecture of the acme.sh script. It's a single script, that uses plugins to 'speak' to different providers. Those plugins are files in a subdir. The point is, these plugins currently are **provider specific** not **api specific**. Now when someone connects to provider **foo** she uses a script called **dns_foo.sh**. If provider **bar** happens to use the same interface 'by accident' it would be a break of the current paradigm to simply use the **dns_foo.sh** script.

Well, one could probably live with the fact that it would not be evident to the new user, why she has to use **dns_foo.sh** to connect to **bar**. Not nice but practical. But this will break compatibility and maintainability. If provider **bar** get's sold to another company - exactly that happened to SchlundTech - which will use a different interface, all domains using **dns_foo.sh** will break with no simple solution to fix them. All domains will then need manual interaction in their config files to switch to the new interface.

The whole point of the acme.sh script is to automate all the tricky bits of LetsEncrypt certificate generation. If provider **bar** changes it's interface one would suspect that only the **dns_bar.sh** script has to be adjusted to the new interface, going to great lengths to migrate to the new interface without user  interaction.

IMO it makes sense to introduce an intermediate layer that separates **provider** from **api**. An **api** script just takes care of a specific **api**, independently of the **provider** using it. The **provider** script simply uses the **api** script. If the provider switches the api, the **provider** script only has to be adjusted to use a differen **api**, taking care of possibly needed migration.

While this at first sounds like a simple task: "Just rename current provier files to api files and symlink provider files to the renamed api files", it does have more implications. I won't elaborate on them now, we could discuss them in upcoming comments.

tohn:
@hmb Thanks a lot for your point of view. I consider the switch from provider to api also a good solution, thanks.
Imho the best option would be to have another repo for these api scripts, written as modular as possible, so that even other ACME clients like [getssl](https://github.com/srvrco/getssl/) can use them. To be honest this annoyed me the most with the release of ACMEv2, that every client now has to write their own api scripts ...
@Neilpang since this is your client, what are your thoughts on this?

Neilpang:
please make sure if  `AUTODNS_API` is not set by user, then use the default value: "https://gateway.autodns.com"

Neilpang:
@hmb   @tohn 
I don't think the api scripts can cross other projects,  since our goal and compatibility is not same.

But we do have plans to make a new structure to share code between the dns scripts easily.
such : https://github.com/Neilpang/acme.sh/pull/1313
I will merge it soon.

tohn:
Since the comment of @hmb convinced me, I'll close this PR.

PR #1531, opened 2018-04-17
Rancher deploy hook both for deploy and renew

CMoH:
This is a second attempt (see https://github.com/Neilpang/acme.sh/pull/1348) at managing rancher certificates via acme.sh deploy hook, this time both for initial deployment and updates when certificates are renewed. Rancher details are stored in acme's domain configuration files.

The script comments contains documentation for users on how to use this deploy hook to achieve both goals.


CMoH:
The check seems to have failed due to a libc6 dependency error.

PR #1536, opened 2018-04-18
Fix Renew Skipped Return code

lpeet:
Set $_ret to $rc ("2") when "$rc" = "$RENEW_SKIP":
$_ret is set to 0 above and is not set to 2 on "Skipped". Script exits with return code 0 when it should be 2.


lpeet:
Bad pull request.

PR #1543, opened 2018-04-20
Fix DNSimple when zone has > 100 records

Rid:
The _get_records function currently returns the first 100 records. As our TXT is added most recently, if you have > 100 records it will not be returned.

I've changed the function to sort by ID DESC, so it will always return the latest 100 records.

### DIFF
--- a/dnsapi/dns_dnsimple.sh
+++ b/dnsapi/dns_dnsimple.sh
@@ -152,7 +152,7 @@ _get_records() {
   sub_domain=$3
 
   _debug "fetching txt records"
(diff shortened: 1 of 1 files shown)


Neilpang:
I would suggest make it  `2000` or `5000` etc.

Rid:
Thanks @Neilpang, however the DNSimple API has a hard limit of 30 records per request anyway, so in reality this will only output the 30 latest records.

However it's unlikely that between the adding of the TXT record and subsequent removal that > 30 records would be created.

Rid:
@Neilpang would you be able to merge this? At the moment we have >10k TXT records because they're not being removed.

PR #1584, opened 2018-05-03
Add HAProxy deploy implementation and documentation

dwatrous:

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -263,6 +263,12 @@ You must specify the path where you want the concatenated key and certificate ch
 export DEPLOY_HAPROXY_PEM_PATH=/etc/haproxy
 ```
 
+You may optionally define the command to reload HAProxy. The value shown below will be used as the default if you don't set this environment variable.
+
+```sh
+export DEPLOY_HAPROXY_RELOAD="/usr/sbin/service haproxy restart"
+```
+
 You can then deploy the certificate as follows
 ```sh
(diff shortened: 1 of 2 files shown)


dwatrous:
I'm not sure how to read the failure in Travis CI. Any pointers on how I can update my script to pass this test?

mnordhoff:
I think it wants you to adjust the whitespace in two places to pass a linter?

https://travis-ci.org/Neilpang/acme.sh/jobs/374463228#L395

Neilpang:
how about restarting/reloading the haproxy ?
Is it necessary to restart haproxy when the cert is changed ?

dwatrous:
@Neilpang I used `--reloadcmd` for that. I guess I haven't verified that when the cron job runs and renews the certificate, if `--reloadcmd` runs after the deploy, but I know that the deploy step runs with renew. It seemed more sensible to use the reload argument than to introduce reload functionality directly into the script, especially since different Linux distros may use different service management scripts...

Neilpang:
no,   deploy hook should be able to live without `--reloadcmd`.  they are not expected to work together.

dwatrous:
I added reload logic to the deploy script.

dwatrous:
The way acme.sh is implemented, you must first `--issue` the certificate and then `--deploy` it. However, upon renewal it performs both the renew and deploy steps.

dwatrous:
@Neilpang do you have any other requirements before merging this?

PR #1591, opened 2018-05-10
Update haproxy deploy hook

andrewheberle:
Updated version of the haproxy deploy hook.

This version differs from the initial version as follows:

1. Requires specifying the PEM filename rather than a path
2. Supports creating a ".ocsp" file to support OCSP stapling in HAProxy
3. Supports creating a ".issuer" file, which is the issuing CA's certificate (this is needed for the OCSP support)
4. Corrects the order of key, certificate, issuer in the PEM file in the initial version (although HAProxy does not seem to care)
5. Supports mulit-cert bundles to allow HAProxy to serve RSA or ECC certificates on the same listener/frontend.  This creates the above ".pem", ".issuer" and ".ocsp" files with a ".rsa" or ".ecc" suffix depending on the type of the certificate.  This does require HAProxy was compiled against a version of OpenSSL that supports this though.


andrewheberle:
I've made some changes so the initial haproxy deploy hook variable of "DEPLOY_HAPROXY_PEM_PATH" can be used.

This defaults to "/etc/haproxy" in this version, and the name of the PEM file is set via the "DEPLOY_HAPROXY_PEM_NAME" env var (defaults to "<domain>.pem" as per the initial hook by dwatrous).

The above makes all values have some sort of "sane-ish" default.  The only part I'm not sure of is my default for the reload command (DEPLOY_HAPROXY_RELOAD) being "systemctl reload haproxy".

Ideally some heuristics here to divine the init system in use may be useful, or alternatively have the default as "/bin/true" so it is a no-op by default which requires the user to set the relevant env var to define the reload action.

smeijer:
Or document that the reload cmd should be specified with `--reloadcmd` when auto reload is required / wished for

```
acme.sh --issue --reloadcmd "systemctl reload-or-restart haproxy"
```

There already is a `reload flag`, so why introduce another one?

andrewheberle:
Deploy hooks are expected to work independently of the reloadcmd option based on a comment by @Neilpang in pull request #1584

lapo-luchini:
Since [HAProxy 1.8](http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.1-crt) (and [1.9](http://cbonte.github.io/haproxy-dconv/1.9/configuration.html#5.1-crt)):

> With BoringSSL and Openssl >= 1.1.1 multi-cert is natively supported, no need to bundle certificates. ECDSA certificate will be preferred if client support it.

Which makes part of the "bundle" support unnecessary (with very recent installations), but still the current version of the deploy hook writes RSA and ECC on the same target file, overwriting each other so some part of this PR would still be very nice to have.

Neilpang:
sorry, for the delay.
but please fix the conflicts first.

Thanks.

andrewheberle:
I have resolved the conflicts in this PR (from the README file only).

I am happy to submit changes to the deploy hook Wiki page to reflect the changes in this PR if it's merged.

@lapo-luchini As far as the usefulness of the "bundle" support...we use it internally here with CentOS 7 which includes a version of openssl is too old for this...but for newer systems? Agreed, it's not particularly useful apart from ensuring RSA and ECC certs don't clobber each other

PR #1616, opened 2018-05-21
DNS plugin for Danish service gratisdns.dk

loial:
Currently only supports primary domains. My use case does not involve
secondary domains so I'm not sure how it behaves, and cannot test it.
Might be as simple as turning all "primary"-references into a variable
that's either "primary" or "secondary", and make an extra check for this
in _get_domain...

Cookie handling heavily inspired by freedns plugin, including caching
the cookie in the config file, so we can rm without re-authenticating


loial:
Oops, should have made a pull for dev instead... sorry, will try again...

PR #1636, opened 2018-05-30
dns_yandex multiple domains fix

lam015:
Issue cert for "example.com" and "*.example.com" creates equal "_acme-challenge.example.com" records. Script dns_yandex.sh gets several "record_id" in one answer (separated with newline) and so it fails to remove DNS TXT-record which is not needed anymore.
This litte fix limits several lines of "record_id" only to the first one.

Thanks to Neilpang for syntax fixing.


lam015:
I have a website in production ("example.com" and "www.example.com" at the same time) and this fix allows to use acme.sh with dns01 challenge through dns_yandex. Please merge if it he does not interfere for ideological reasons.

Neilpang:
please merge the latest code.

Issue #1328, opened 2018-03-05
php -r error in cpanel_uapi.sh - Error in argument 1, char 2: option not found r


williatf:
Steps to reproduce
------------------
Running cron job with GoDaddy cPanel:
`~/.acme.sh/acme.sh --deploy --domain <domain> --deploy-hook cpanel_uapi`

I get an error stating that the system could not parse the certificate.

Traced the error to cpanel_uapi.sh lines 38 & 39:
```
  _cert=$(php -r "echo urlencode(\"$_certstr\");")
  _key=$(php -r "echo urlencode(\"$_keystr\");")
```

The php version that cron is using is CGI-FCGI, which does not have the -r option:

$  which -a php
```
/usr/local/bin/php
/usr/bin/php
```

$  /usr/bin/php -v
`PHP 5.6.32 (cgi-fcgi) (built: Dec  4 2017 12:21:23)`

$  /usr/bin/php -h
```
Usage: php [-q] [-h] [-s] [-v] [-i] [-f <file>]
       php <file> [args...]
  -a               Run interactively
  -b <address:port>|<port> Bind Path for external FASTCGI Server mode
  -C               Do not chdir to the script's directory
  -c <path>|<file> Look for php.ini file in this directory
  -n               No php.ini file will be used
  -d foo[=bar]     Define INI entry foo with value 'bar'
  -e               Generate extended information for debugger/profiler
  -f <file>        Parse <file>.  Implies `-q'
  -h               This help
  -i               PHP information
  -l               Syntax check only (lint)
  -m               Show compiled in modules
  -q               Quiet-mode.  Suppress HTTP Header output.
  -s               Display colour syntax highlighted source.
  -v               Version number
  -w               Display source with stripped comments and whitespace.
  -z <file>        Load Zend extension <file>.
  -T <count>       Measure execution time of script repeated <count> times.
```

I fixed it by hard-coding "/usr/local/bin/php" as the php version to use.
**Extract of cpanel_uapi.sh**
```
  if ! _exists /usr/local/bin/php; then
    _err "The command php is not found."
    return 1
  fi
  # read cert and key files and urlencode both
  _certstr=$(cat "$_ccert")
  _keystr=$(cat "$_ckey")
  _cert=$(/usr/local/bin/php -r "echo urlencode(\"$_certstr\");")
  _key=$(/usr/local/bin/php -r "echo urlencode(\"$_keystr\");")
```

[...]

PR #1650, opened 2018-06-02
Issue #1328 bug fix

santerikannisto:
Eliminated php dependency with a private function for urlencode using sed. Php had failed on godaddy due to multiple php instances and naturally cron using the one without the necessary -r option.


santerikannisto:
Has a posix incompatilbility with the "N" option in sed.  Works on some environments and not from others.

PR #1664, opened 2018-06-08
Issue #1328 bug fix v2

santerikannisto:
Eliminated php dependency with a private function for urlencode using sed. Php had failed on godaddy due to multiple php instances and naturally cron using the one without the necessary -r option. Compared to previous PR the sed code is now POSIX and should work on all environments.

### DIFF
--- a/deploy/cpanel_uapi.sh
+++ b/deploy/cpanel_uapi.sh
@@ -2,8 +2,14 @@
 # Here is the script to deploy the cert to your cpanel using the cpanel API.
 # Uses command line uapi.  --user option is needed only if run as root.
 # Returns 0 when success.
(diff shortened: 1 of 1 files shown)


santerikannisto:
@Neilpang This has been waiting here now 3 weeks. It's been successfully used by quite a many people and I have not received bug reports. How about merging it so that I can close my GitHub account? Thanks.

Neilpang:
we have a `_url_encode()` function: https://github.com/Neilpang/acme.sh/blob/master/acme.sh#L565
Can you use that one instead ?

santerikannisto:
Sure. Thanks for the tip. Looks like it works at least on my hosting accounts. I will make a new pull request right away.

PR #1667, opened 2018-06-11
Make domain available to reloadcmd

alexbowers:
Le_Domain is empty, but `domain` is not. Exporting the `domain` will make it available.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4721,6 +4721,7 @@ _installcert() {
       export CA_CERT_PATH
       export CERT_FULLCHAIN_PATH
       export Le_Domain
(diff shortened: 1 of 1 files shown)


alexbowers:
@Neilpang Any feedback on this?

alexbowers:
@Neilpang Any feedback on this change?

We are having to use a forked version for now, and would like to go back to the primary version to make maintenance easier on our end.

Neilpang:
do not use `export domain`.
If in your case Le_Domain  is empty,  let's make it not empty.
show me your case and log.

PR #1673, opened 2018-06-13
Create dns_leaseweb.sh

m-boone:
Add support for Leaseweb (http://developer.leaseweb.com). This is based on dns_cloudns and I cannot really take credit for it. 

I have tested it and it works. I have the following notes:
-Sometimes Leaseweb gives back no response while I would expect an errorMessages (so i now check for empty response or errorMessage). 
-the acme.sh _Post() function does not return base64decoded response when base64 is used. Not a real big problem. 
-I'm not sure if any other dnsapi provides this functionality. But with this implementation two simultaneous running processes for the same domain will not work (2 TXT files for _acme-challenge at the same time).

### DIFF
--- a/dnsapi/dns_leaseweb.sh
+++ b/dnsapi/dns_leaseweb.sh
@@ -1,0 +1,163 @@
+#!/usr/bin/env sh
+
+# Author: Boyan Peychev <boyan at cloudns dot net>
+# Modified for Leaseweb: M-Boone
+
+LEASEWEB_API="https://api.leaseweb.com/hosting/v2/domains"
+
+########  Public functions #####################
+
+#Usage: dns_leaseweb_add   _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_leaseweb_add() {
+  
+  if ! _dns_leaseweb_init_check; then
+    return 1
+  fi
+  
+  zone="$(_dns_leaseweb_get_zone_name "$1")"
+  if [ -z "$zone" ]; then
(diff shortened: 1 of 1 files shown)


Neilpang:
check this guide first:  https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide

m-boone:
Ok, thank for the doc, that helped a lot. 

Dev Guide:
1. The cloudflare dns api is a recommended reference: Check, rebuilt with cloudfare as base
2. The script file name must be myapi.sh: Check
3. The file can be placed in acme.sh/ folder, or in acme.sh/dnsapi/ subfolder: Check, in /dnsapi/
4.  The file shebang must be sh not bash: Check
5. There must be 2 functions in your script: Check
6.1 Get the full domain and the txt record: Check
6.2 You must save your username and password in the add function: Check
6.3 Detect which part is your root zone: Check, Tested with domain.com, *.domain.com, test.domain.com, do not have .co.uk domain 
6.4 Call your dns api to add txt record: Check, wildcard works with 2 entries
6.5 Process the api response: Check
7. Guide for the rm function: Check
8. Please also check this bug to support V2 wildcard cert: Check
9. Update the docs to include your dns api and usage: Check
Style Guidelines: indentation, trailing spaces, Checked with shellcheck, travis build. 

Leaseweb API description: Leaseweb works with one TXT record per domain which they call a records set. In this records set you add the individual entries. So for this to work we need to check if the TXT records set exists, if not create it. If it does exist then we need to send all previous entries + the new entry as an update to the TXT recordset. 

Last remarks: Shell is a new language for me so there might be some places that a code review could help rough out some beauty errors but it seems to works as intended.

Neilpang:
please merge the latest code.

Neilpang:
sorry  please merge the latest code.

PR #1679, opened 2018-06-16
save fritzbox deploy credentials mutable

moritz-h:
The credentials for fritzbox deploy could not be overwritten by a newer env variable.
This PR changes this to use `_saveaccountconf_mutable` / `_readaccountconf_mutable` instead, like in the wiki: https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide#2-you-must-save-your-username-and-password-in-the-add-function

### DIFF
--- a/deploy/fritzbox.sh
+++ b/deploy/fritzbox.sh
@@ -32,9 +32,9 @@ fritzbox_deploy() {
     return 1
   fi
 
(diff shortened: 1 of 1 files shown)


moritz-h:
@Neilpang I think this fix is still valid. Is there any more work required to make this PR acceptable?

PR #1681, opened 2018-06-17
Fixes socat version output when --debug flag is set.

navossoc:
This fix make `socat` outputs his version, not his help information.

[]'s


Neilpang:
we need all the help message , not only the version.

PR #1700, opened 2018-06-29
Deploy key for exim4 with chmod 600

CoolDuke:
Key file was being deployed with world-readable access.
This patch sets the key's file modes to 600.


CoolDuke:
Obviously exim4 reads the key as the user it started up with.
Changing the group is not an option as it's not permitted as unprivileged user, so this module would need sudo to work properly. Closing this as this would depend on the system's environment.

PR #1745, opened 2018-07-24
Added dns_lexicon_rm command

oldium:
Remove created TXT record when finished. Works with `lexicon` version 2.3.0 and later.


oldium:
Hm, the lexicon API still does not work in all cases. When you do `--issue` with additional domain, it does not take the `PROVIDER` from the old one. I will update the pull request to read config values when the are empty.

oldium:
Now I think it works as intended - the LEXICON_{_PROVIDER_}_{_SOMETHING_} is read from account config if not specified in the environment data, also PROVIDER is read from domain config if not specified.

PR #1750, opened 2018-07-26
Fix inwx account without Mobile TAN

jkroepke:
Its a follow up from https://github.com/Neilpang/acme.sh/pull/1501#issuecomment-393827632

I'm using now the same detection like https://github.com/inwx/php-client/blob/master/INWX/Domrobot.php#L71 .


apollo13:
This got reverted in https://github.com/Neilpang/acme.sh/commit/68a290c34752c2aa0b913332467b7a5f2c001111 again. What was the reason for reverting it?

PR #1786, opened 2018-08-13
Create a draft version of pleskxml API for DNS01

stilez:
This DNS01 method allows acme.sh to set DNS TXT records using the Plesk XML API described at:
    https://docs.plesk.com/en-US/12.5/api-rpc/about-xml-api.28709
and more specifically: 
    https://docs.plesk.com/en-US/12.5/api-rpc/reference.28784

This may be needed if the DNS provider doesn't make the standard Plesk API available to a user - in which case, Plesk can't be configured using usual means or RFC1236. But the XML API is often still left accessible, and is well documented, so it can be used instead, for acme.sh purposes.

Note that I have tested the API calls work as expected, and is sanely structured, but I'm not an experienced shell coder. So I expect any number of stylistic and other amendments/corrections. It also includes a ton of debug code which won't hurt to leave in while checking.

Please help me to knock this module into shape for merging, and be gentle with me on it - I'm sure it needs a lot of attention! If edits are obvious, feel free to edit without asking if quicker.

Thanks :)

### DIFF
--- a/dnsapi/dns_pleskxml.sh
+++ b/dnsapi/dns_pleskxml.sh
@@ -72,27 +72,25 @@ _DBG_EARLY_CHECK_MODE() {
   fi
 }
 
-# arg1 = LINENO
-# arg2 = severity level (1=least serious, 3=most serious)
+# arg1 = severity level (1=least serious, 3=most serious)
 #   By design if DBG level is 9 for a MESSAGE, the message is ALWAYS shown, this is used for _info and _err
-# arg3 = message
+# arg2 = message
 _DBG() {
-  if [ "$2" -eq 9 ] || ([ "$_pleskxml_DBG_LEVEL" -gt 0 ] && [ "$2" -ge "$_pleskxml_DBG_LEVEL" ]); then
-    case $2 in
+  if [ "$1" -eq 9 ] || ([ "$_pleskxml_DBG_LEVEL" -gt 0 ] && [ "$1" -ge "$_pleskxml_DBG_LEVEL" ]); then
+    case $1 in
       1) _pleskxml_severity='MAX_DETAIL' ;;
       2) _pleskxml_severity='DETAIL' ;;
       3) _pleskxml_severity='INFO' ;;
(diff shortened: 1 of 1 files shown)


stilez:
Not sure what the eerror above refers to. The log states **"The command "if [ "$TRAVIS_OS_NAME" = "linux" ]; then git diff --exit-code && echo "shfmt OK" ; fi" exited with 1"** but I can't find a declaration of this variable in other dns_*.sh files and thew code above both runs, and also passes the online ShellCheck, so I'm not sure what to make of it.....

Neilpang:
it's formatting errors.  space key,  and indent etc.

Neilpang:
check this guide https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide

stilez:
I need some help here, as I'm completely unfamiliar with the core code, and the dev info/wiki still leaves gaps that I have to address to get this properly integrated and complete, ready to pull. 

For simplicity I'll tackle these a few at a time:

**USER INPUTS, VARIABLES AND cURL ARGS:**

I'm currently using the following variables: (1) the API user name+password, (2) the API URI, (3) a flag to allow user+pass to be sent "clear" and not require https (eg, allow http for development/testing within ones own LAN), (4) any additional cURL args that their use-case may require, and (5) a debug output flag.

1. The username+pw are saved as described in the dev info and dns_CF. But what should I do with the rest? Are variable names for these already assigned that I should reuse? Should I save all of these manually in the dns module's code, for future renewal, like the user+pw, and if not, which should I save?
2. What if any input sanitising can I assume the core code has already done, and what sanitising should I do myself on the inputs?
3. In this DNS API, various API options _including the account login credentials (user+pw)_ must be passed within HTTP headers. How do I do that with the inbuilt _post function, or else what workaround do you prefer - should I temporarily copy the code for _post and modify it to add the extra args?

stilez:
Rewritten from scratch, to be PRed shortly, hence closing this PR

PR #1807, opened 2018-08-21
freedns: fix domain splitting for multi-part domains

frebib:
This should address issue #1086.
Supercedes previous PR #1454.

This should fix the situation mentioned in https://github.com/Neilpang/acme.sh/pull/1454#issuecomment-376578643

### DIFF
--- a/dnsapi/dns_freedns.sh
+++ b/dnsapi/dns_freedns.sh
@@ -51,68 +51,38 @@ dns_freedns_add() {
   # sub_domain is the element before the first dot
   top_domain="$(echo "$fulldomain" | cut -d. -f 2-)"
   sub_domain="$(echo "$fulldomain" | cut -d. -f 1)"
+  DNSdomainid=
 
-  _debug "top_domain: $top_domain"
-  _debug "sub_domain: $sub_domain"
+  while [ -z "$DNSdomainid" ]; do
+    _debug "top_domain: $top_domain"
+    _debug "sub_domain: $sub_domain"
 
-  # Sometimes FreeDNS does not return the subdomain page but rather
-  # returns a page regarding becoming a premium member.  This usually
-  # happens after a period of inactivity.  Immediately trying again
-  # returns the correct subdomain page.  So, we will try twice to
-  # load the page and obtain our domain ID
-  attempts=2
-  while [ "$attempts" -gt "0" ]; do
-    attempts="$(_math "$attempts" - 1)"
-
-    htmlpage="$(_freedns_retrieve_subdomain_page "$FREEDNS_COOKIE")"
+    found_domainid="$(_freedns_find_domain "$FREEDNS_COOKIE" "$top_domain")"
     if [ "$?" != "0" ]; then
-      if [ "$using_cached_cookies" = "true" ]; then
-        _err "Has your FreeDNS username and password changed?  If so..."
-        _err "Please export as FREEDNS_User / FREEDNS_Password and try again."
-      fi
-      return 1
-    fi
-
-    subdomain_csv="$(echo "$htmlpage" | tr -d "\n\r" | _egrep_o '<form .*</form>' | sed 's/<tr>/@<tr>/g' | tr '@' '\n' | grep edit.php | grep "$top_domain")"
-    _debug3 "subdomain_csv: $subdomain_csv"
-
-    # The above beauty ends with striping out rows that do not have an
-    # href to edit.php and do not have the top domain we are looking for.
-    # So all we should be left with is CSV of table of subdomains we are
-    # interested in.
-
-    # Now we have to read through this table and extract the data we need
-    lines="$(echo "$subdomain_csv" | wc -l)"
-    i=0
-    found=0
-    DNSdomainid=""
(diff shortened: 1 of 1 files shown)


JonnyTech:
This works perfectly for my .co.uk domains - @Neilpang can this be merged please?

frebib:
I understand that there might be some code-style issues with this PR, if so please leave a review and I will get them fixed and squashed

chrishas35:
My FreeDNS domain is a subdomain of a .net.  Had the same issue and this PR solves it.  Thanks @frebib.

frebib:
Bump @Neilpang, any reason this (and the other 100+ PRs) can't be merged?

dnsapi/dns_freedns.sh
```diff
@@ -47,73 +47,42 @@ dns_freedns_add() {
...
-
-    htmlpage="$(_freedns_retrieve_subdomain_page "$FREEDNS_COOKIE")"
+    found_domainid="$(_freedns_find_domain "$FREEDNS_COOKIE" "$top_domain")"
     if [ "$?" != "0" ]; then
-      if [ "$using_cached_cookies" = "true" ]; then
-        _err "Has your FreeDNS username and password changed?  If so..."
-        _err "Please export as FREEDNS_User / FREEDNS_Password and try again."
+      # Check if there is a dot in the top_domain
+      # If not, we can't split any more and should assume the domain doesn't exist
+      if ! echo "$top_domain" | grep -qF '.'; then
```
Neilpang: could you please not use `grep -qF` options here ?
it's dangerous

dnsapi/dns_freedns.sh
```diff
@@ -245,6 +214,73 @@ _freedns_login() {
...
+    htmlpage="$(_freedns_retrieve_subdomain_page "$FREEDNS_COOKIE")"
+    if [ "$?" != "0" ]; then
+      if [ "$using_cached_cookies" = "true" ]; then
+        _err "Has your FreeDNS username and password changed?  If so..."
+        _err "Please export as FREEDNS_User / FREEDNS_Password and try again."
+      fi
+      return 1
+    fi
+
+    subdomain_csv="$(echo "$htmlpage" | tr -d "\n\r" | _egrep_o '<form .*</form>' | sed 's/<tr>/@<tr>/g' | tr '@' '\n' | grep edit.php | grep "$top_domain")"
```
Neilpang: if you trim `\r\n` before  piping to `_egrep_o()` function,  it will not work on solaris.

On solaris,  `_egrep_o` is implemented by `sed`, but `sed` on solaris can not work without `\n`.

please consider replacement

Neilpang:
sorry,  I was just too busy to review code.

dkerr64:
This PR just came to my attention.  As the original author of the freedns script a user recently reached out to me with this bug.  And I just implemented a fix.  When considering creating a PR I find that a fix is already proposed.  We solve it in a similar manner... looping through all the possible TLD's until one is found to work.  Only I did it in-line whereas this PR creates a new function -- which is probably more elegant.  Moral is that I should have looked for an existing issue before trying to fix it myself.

All that said.  I have found additional problems that need addressed.  Specifically when we grep for "www.example.com" we (I) forgot that dot is a wildcard.  The correct way to grep would be to first escape the dots.
`search_domain="$(echo "$top_domain" | sed 's/\./\\./g')"`
and then use $search_domain where we grep or sed.

Also during testing with a new domain I found that the domain name is not tightly bound by
`<td> and </td>`
tags.  So...
`_contains "$line" "<td>$full_domain</td>"`
fails.  And the fix is...
`_contains "$line" "<td>$search_domain.*</td>"`
which allows for text immediately following the domain to be present (which it is when first created and before FreeDNS has validated that the DNS is set correctly).

Okay, now what.  I could create a PR with my own version of a fix, and @Neilpang  can chose between them, or @frebib can address the questions raised by @Neilpang and my comments above.

Thoughts?

dkerr64:
For reference, my version of the fix is here... https://github.com/dkerr64/acme.sh/commit/465189c0b47fb9c9939f4d87ff2196bb3a5aa37c

David

frebib:
It's been a long time since I looked at this, and I don't use freedns any more. Please feel free to adapt/ignore this PR and open another one. I don't have the time or bother to take this any further. I'm glad someone else is picking this up again :)

dkerr64:
Thank you @frebib. I will do some cleanup on my version of the fix and then submit a new pull request.  This is a pretty serious bug so I want to get that done soon -- I just wish it had come to my attention sooner.
Thanks, David

dkerr64:
I have made the cleanups that I wanted to do.  If anyone is monitoring this can you please test the version here... https://raw.githubusercontent.com/dkerr64/acme.sh/FreeDNS/dnsapi/dns_freedns.sh
Thank you.

felixgonsug:
I tested and it works after adding the flag "--dnssleep 120". FreeDNS has a propgation time of 60 seconds, so the default of 20 is to short. In previous versions the default was 120, but that problem is not related to this PR. The PR is working correctly, for multi TLD and for single TLD. Sorry for the dealy of the response and thanks for fixing this.

PR #1814, opened 2018-08-25
Add reloadcmd support for cron flag.

hexfusion:
It is very convenient to allow `reloadcmd` to be passed with `cron` flag. As the `reloadcmd` flag already exists this PR just adds support for `cron` usage.

ref: #1002

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4713,19 +4713,7 @@ _installcert() {
   fi
 
   if [ "$_reload_cmd" ]; then
-    _info "Run reload cmd: $_reload_cmd"
-    if (
-      export CERT_PATH
-      export CERT_KEY_PATH
-      export CA_CERT_PATH
-      export CERT_FULLCHAIN_PATH
-      export Le_Domain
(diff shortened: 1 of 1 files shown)


hexfusion:
Kind ping @Neilpang any input on this?

uqs:
This is quite unfortunate, I was also under the impression I could alter the reloadcmd from the crontab whenever I need it to change. Especially as a first time user one has to play around a bit with getting it right, so it's surprising that it gets hardcoded during the initial run.

Can you please make it error out when --cron and --reloadcmd are given at the same time?

mcbsys:
I understand the intention to leave cron for scheduling, but I'm also wishing there was a way to do a "global" command during the certificate renewal process (namely restart the web server so the certificate will become active). I have 10 domains on a server. One call to acme.sh from cron checks and updates all domains, which is great. But if I want to run the same Le_ReloadCmd no matter which cert is reissued, I have to update 10 .conf files? Or can I put Le_ReloadCmd in /root/.acme.sh/account.conf?

PR #1834, opened 2018-09-10
Create zimbra.sh hook file

JimDunphy:
support for zimbra deploy-hook


JimDunphy:
Verified with zimbra 8.7 and above. Would not work with earlier versions as zmcertmgr command needs to run as root. This is documented in provided link to wiki.zimbra.com in comments inside the hook file.

PR #1844, opened 2018-09-14
Fixing incompatibilties with Pebble

kmadac:
Pebble is official lightweight test server from Letsencrypt.

One difference between boulder and pebble is that http header names start with lower-case instead of upper-case. As this behavior is in compliance with RFC acme.sh should be able to work in both cases.

Another fix is that `certificate` field is not part of finalize-order response, but we need to get certificate url from request to order URL taken from Location header.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1829,7 +1829,7 @@ _send_signed_request() {
 
       _debug2 _headers "$_headers"
 
-      _CACHED_NONCE="$(echo "$_headers" | grep "Replay-Nonce:" | _head_n 1 | tr -d "\r\n " | cut -d ':' -f 2)"
+      _CACHED_NONCE="$(echo "$_headers" | grep -i "Replay-Nonce:" | _head_n 1 | tr -d "\r\n " | cut -d ':' -f 2)"
       _debug2 _CACHED_NONCE "$_CACHED_NONCE"
     else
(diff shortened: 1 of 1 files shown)


kmadac:
Travis CI job has been stucked on `brew update && brew install socat` for 10 minutes and then it was killed by travis. Can anybody run the build again?

kmadac:
Travis CI job has been stucked on `brew update && brew install socat` for 10 minutes and then it was killed by travis. Can anybody run the build again?

acme.sh
```diff
@@ -3267,7 +3267,7 @@ _regAccount() {
   fi
 
   _debug2 responseHeaders "$responseHeaders"
-  _accUri="$(echo "$responseHeaders" | grep "^Location:" | _head_n 1 | cut -d ' ' -f 2 | tr -d "\r\n")"
+  _accUri="$(echo "$responseHeaders" | grep -i "^Location:" | _head_n 1 | cut -d ' ' -f 2 | tr -d "\r\n")"
```
Neilpang: as for the `grep -i`,  I remember not clearly that it's not supported on solaris,  but I will try later.  my solaris VM was broken.

Neilpang:
`grep -i` is ok.

can you update this PR ?

PR #1870, opened 2018-10-04
Add deployment script for Mikrotik systems

spack971:
Add a script to deploy certificates on Mikrotik/RouterOS
systems. Certificate will be deployed for the `www-ssl` service and
optionally the `api-ssl` service.

### DIFF
--- a/deploy/mikrotik.sh
+++ b/deploy/mikrotik.sh
@@ -1,0 +1,130 @@
+#!/usr/bin/env sh
+#
+# mikrotik.sh
+# ===========
+#
+# Deploy certificate to a Mikrotik RouterOS system using the SSH protocol.
+#
+# The script must be able to connect to the remote host without a password.
+# Therefore only public key authentication method is available and SSH keys must
+# have been exchanged prior to running this script.
+#
+#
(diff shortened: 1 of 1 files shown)


aklowther:
hopefully this gets more attention than #706

Neilpang:

Thanks.

spack971:
@Neilpang I should have fixed all the CI issues however it is now failing on `shfmt` and a file `test.sh`. I don't know where this file is coming from. Any idea?

aklowther:
How does this differ from #706 ?

spack971:
@aklowther in essence both will accomplish the same which is setting up a certificate on a routeros based device.

My implementation allows finer SSH client configuration using environment variables. It is possible to provide additional options or specify which port to use.

In general some useful options available via the environment variables.

aklowther:
IMO: I think it would be less confusing if your changes modified that deployment method rather than duplication with additions.

spack971:
@aklowther By the time I made this pull request no routeros deploy method was available yet. I just realized that #706 got merged so indeed I will change this deploy method instead.

PR #1875, opened 2018-10-10
Update alpine

taogre:
change alpine version from 3.6 to 3.8


ovalseven8:
@Neilpang: Note that Alpine 3.6 only gets security updates until May 2019.

Neilpang:
@ovalseven8  It's already updated to 3.9 by : https://github.com/Neilpang/acme.sh/pull/2086

PR #1882, opened 2018-10-17
Updated dns_nsupdate to use multiple dns update keys

SicorDev:
Added the option to use multiple dns update keys via naming convention.
The solution is backward compatible and completely optional.

You are now able to specify a folder, where your keys are located. The dns_api will try to read the keyfile based on the domain name and use it instead of the default `NSUPDATE_KEY`.

If there is no folder/key, nothing changes and the regular `NSUPDATE_KEY` will be used.


panachoi:
I think this is a really good idea; the only other reasonable solution would be to possibly move all of the NSUPDATE variables into the domain/domain.conf file instead.

PR #1921, opened 2018-11-13
Add missing bind-tools package

nakermann1973:
The bind_tools package is required for dns_nsupdate to work


nakermann1973:
`docker image inspect` shows that the size = 26702245 - about 26 MB

PR #1925, opened 2018-11-16
Fix Vscale hostedzone

kuk:
In case of >1 domains `"{.*\"name\":\s*\"$h\".*}")"` matches `{id: 1, name: a}, {id: 2, name: $h}` that leads to wrong `id` (we look for id=2, but match id=1)

`{[^{]*\"name\":\s*\"$h\"[^}]*}` matches just `{id: 2, name: $h}`

### DIFF
--- a/dnsapi/dns_vscale.sh
+++ b/dnsapi/dns_vscale.sh
@@ -102,7 +102,7 @@ _get_root() {
         return 1
       fi
 
-      hostedzone="$(echo "$response" | _egrep_o "{.*\"name\":\s*\"$h\".*}")"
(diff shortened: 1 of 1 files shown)


kuk:
@Neilpang could you please take a look at this PR. Currently Vscale DNS integration is broken, this PR fixes it

Neilpang:
there can be more graceful fixes, for example:

```
echo  '{id: 1, name: a}, {id: 2, name: $h}' | tr "{" "\n" | grep "\"name\":\s*\"$h\".*}"
```
Split the text into multiple lines before greping them.

Because `grep  "[^{]" ` is not always safe on all the platforms.

kuk:
Thank you, changed the code

PR #1937, opened 2018-11-28
Add native Vultr API support

oefd:
For those that don't want to have to use lexicon.

I'm not too experience with sh scripting but I had this working correctly with both of bash and dash, and `shellcheck -s sh dns_vultr.api` has no complaints.


Neilpang:
please merge the latest code and check this guide: https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide

Neilpang:
please merge latest code.

PR #1944, opened 2018-12-02
Update Linode API to v4

titanofold:
Linode API has made breaking changes that are resolved by this
update. No user action is required.

Additionally, related README.md entry updated to include new cloud
manager interface.

### DIFF
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -264,9 +264,18 @@ when needed.
 ## 14. Use Linode domain API
 
 First you need to login to your Linode account to get your API Key.
-[https://manager.linode.com/profile/api](https://manager.linode.com/profile/api)
 
-Then add an API key with label *ACME* and copy the new key.
+  * [Classic Manager](https://manager.linode.com/profile/api)
+
+   Under "Add an API key", Give the new key a "Label" (we recommend *ACME*),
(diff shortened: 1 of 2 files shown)


Neilpang:
Hi,
Can this new api be compatible with the old one ?
I just want to know if it's possible.

Thanks.

steven-ryan:
It is not compatible.
The API key is replaced by a personal access token, and the requests/responses are different

[Here are some notes on migrating](https://www.linode.com/docs/platform/api/upgrade-to-linode-api-v4/)

In Summary:
- Previous versions of the API allowed users to submit request and method parameters via URL parameters; this is no longer supported.
- Authentication is now done through the Authorization header rather than through the query string
- Modernized language is used throughout the API: “data centers” are now “regions”, “distributions” are now “images”, and “plans” are now “types”.

titanofold:
Sorry! I didn't mean to ghost on this. I had finals that kept me a bit busy.

I didn't realize I had made an assumption. The Cloud manager is just pitched as an alternative interface, and Linode didn't make any statements (in the cloud manager) that the API key generators in the classic and Cloud interface were not the same.

Looks like the best way forward is to revert the change to dns_linode.sh and make a new dns_linode_v4.sh (or however it should be named).

PR #1989, opened 2018-12-27
Added fixes so BuyPass Go ACME server can be used

mysteq:
Had to add a few tweaks to be able to use the ACME server for BuyPass GO, as it seems to not quite conform to the ACME v1 specs implemented currently in acme.sh.

Resources:
https://api.buypass.com/acme/directory
https://www.buypass.com/ssl/resources/go-ssl-technical-specification

The tweaks should not affect existing functionality, and I've done some basic testing on CentOS.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1822,12 +1822,9 @@ _send_signed_request() {
     _debug3 _request_retry_times "$_request_retry_times"
     if [ -z "$_CACHED_NONCE" ]; then
       _headers=""
-      if [ "$ACME_NEW_NONCE" ] || [ "$BUYPASS" ]; then
+      if [ "$ACME_NEW_NONCE" ]; then
+        _debug2 "Get nonce with HEAD. ACME_NEW_NONCE" "$ACME_NEW_NONCE"
         nonceurl="$ACME_NEW_NONCE"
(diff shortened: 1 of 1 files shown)


Neilpang:
I think we should be able to remove any "buypass" or "BUYPASS" string from the code.
please go ahead.

acme.sh
```diff
@@ -13,6 +13,8 @@ _SCRIPT_="$0"
 
 _SUB_FOLDERS="dnsapi deploy"
 
+BUYPASS_CA="https://api.buypass.no/acme/directory"
```
Neilpang: remove it

m-ueberall:
Apologies if this has been mentioned elsewhere, but I did not find it: Where are certificates being kept if you use the exact same parameters/(sub)domains but reference different server URIs/CAs?
I'm not sure the resulting files can coexist in ~/.acme.sh/sub.domain.tld[_ecc]--will users get to see a warning when existing files from a different CA would be overwritten or should this at least be mentioned somewhere else?

emmyforeignsss:
https://emmyforeignsss.github.io.diff

emmyforeignsss:
mysteq:buypass-support

Issue #2000, opened 2019-01-02
dns_hostingde doesn't work with challenge-alias and more than one domain


afaenger:
Steps to reproduce
------------------

acme.sh --issue --test --force -d example1.com --challenge-alias other-domain.com --dns dns_hostingde -d '*.example1.com'

This will throw UNKNOWN API ERROR
It works only when one domain is used or when the first domain has been validated already.

Workaround: this seems to be a timing issue when more than one TXT record is added/removed via the hosting.de API. When inserting a small delay in between, then it seems to work.
E.g. adding a "sleep 5" right before "return 0" in _hostingde_addRecord() and _hostingde_removeRecord() 

Debug log
-----------------

```
[Wed Jan  2 14:56:44 CET 2019] _main_domain='example1.com'
[Wed Jan  2 14:56:44 CET 2019] _alt_domains='*.example1.com'
[Wed Jan  2 14:56:44 CET 2019] Using config home:/root/.acme.sh
[Wed Jan  2 14:56:44 CET 2019] Using stage ACME_DIRECTORY: https://acme-staging-v02.api.letsencrypt.org/directory
[Wed Jan  2 14:56:44 CET 2019] ACME_DIRECTORY='https://acme-staging-v02.api.letsencrypt.org/directory'
[Wed Jan  2 14:56:44 CET 2019] DOMAIN_PATH='/root/.acme.sh/example1.com'
[Wed Jan  2 14:56:44 CET 2019] Using ACME_DIRECTORY: https://acme-staging-v02.api.letsencrypt.org/directory
[Wed Jan  2 14:56:44 CET 2019] _init api for server: https://acme-staging-v02.api.letsencrypt.org/directory
[Wed Jan  2 14:56:44 CET 2019] GET
[Wed Jan  2 14:56:44 CET 2019] url='https://acme-staging-v02.api.letsencrypt.org/directory'
[Wed Jan  2 14:56:44 CET 2019] timeout=
[Wed Jan  2 14:56:44 CET 2019] _CURL='curl -L --silent --dump-header /root/.acme.sh/http.header  -g '
[Wed Jan  2 14:56:45 CET 2019] ret='0'
[Wed Jan  2 14:56:45 CET 2019] ACME_KEY_CHANGE='https://acme-staging-v02.api.letsencrypt.org/acme/key-change'
[Wed Jan  2 14:56:45 CET 2019] ACME_NEW_AUTHZ
[...]

PR #2001, opened 2019-01-02
hosting.de API: waiting for API zoneStatus active

membero:
references issue #2000

new _hostingde_getZoneStatus gets zone status from API. _hostingde_addRecord and _hostingde_removeRecord wait until the zone can be updated.

### DIFF
--- a/dnsapi/dns_hostingde.sh
+++ b/dnsapi/dns_hostingde.sh
@@ -90,7 +90,7 @@ _hostingde_addRecord() {
   _hostingde_getZoneStatus
   _debug "Result of zoneStatus: '${zoneStatus}'"
   while [ "${zoneStatus}" != "active" ]; do
-    sleep 5
+    _sleep 5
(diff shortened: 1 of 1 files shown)


dnsapi/dns_hostingde.sh
```diff
@@ -74,8 +74,26 @@ _hostingde_getZoneConfig() {
...
+  _debug "zoneStatus '${zoneStatus}'"
+  return 0
+}
+
 _hostingde_addRecord() {
   _info "Adding record to zone"
+  _hostingde_getZoneStatus
+  _debug "Result of zoneStatus: '${zoneStatus}'"
+  while [ "${zoneStatus}" != "active" ]; do
+    sleep 5
```
Neilpang: call `_sleep()` function instead.

membero: now using _sleep() instead of sleep

PR #2023, opened 2019-01-10
Use cross-platform grep pattern for Namecheap API. Fixes #2022

### LINKED ISSUE #2022: dns_namecheap.sh egrep pattern not compatible with BSD grep
Attempting to use dns_namecheap.sh to validate DNS-01 fails on FreeBSD due to an incorrect regex pattern when parsing the API output. The log output shows that it failed to parse anything from the host record entries.

```
[Thu Jan 10 12:34:16 EST 2019] _host='echo "<host HostId="153420153" Name="@" Type="A" Address="xx.xx.xx.xx" MXPref="60" TTL="1800" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" /'
[Thu Jan 10 12:34:16 EST 2019] hostid
[Thu Jan 10 12:34:16 EST 2019] hostname
[Thu Jan 10 12:34:16 EST 2019] hosttype
[Thu Jan 10 12:34:16 EST 2019] hostaddress
[Thu Jan 10 12:34:16 EST 2019] hostmxpref
[Thu Jan 10 12:34:16 EST 2019] hostttl
```

BSD egrep does not like patterns that use `\s` for space, so the patterns such as `'\sHostId="[^"]*'` fail to match anything. Changing it to a character class for whitespace works with both BSD and GNU grep: `'[[:space:]]HostId="[^"]*'` and allows the validation to proceed.

```
[Thu Jan 10 15:42:45 EST 2019] _host='echo "<host HostId="153420153" Name="@" Type="A" Address="xx.xx.xx.xx" MXPref="60" TTL="1800" AssociatedAppTitle="" FriendlyName="" IsActive="true" IsDDNSEnabled="false" /'
[Thu Jan 10 15:42:45 EST 2019] hostid='153420153'
[Thu Jan 10 15:42:45 EST 2019] hostname='@'
[Thu Jan 10 15:42:45 EST 2019] hosttype='A'
[Thu Jan 10 15:42:45 EST 2019] hostaddress='xx.xx.xx.xx'
[Thu Jan 10 15:42:45 EST 2019] hostmxpref='60'
[Thu Jan 10 15:42:45 EST 2019] hostttl='1800'
```

I'll work up a PR shortly.

jim-p:
Use cross-platform grep pattern for Namecheap API to fix #2022

The original pattern used `\s` which does not work the same with BSD grep. Using a character class for whitespace, `[[:space:]]`, functions with BSD and GNU grep.

### DIFF
--- a/dnsapi/dns_namecheap.sh
+++ b/dnsapi/dns_namecheap.sh
@@ -161,12 +161,12 @@ _namecheap_parse_host() {
   _host=$1
   _debug _host "$_host"
 
-  _hostid=$(echo "$_host" | _egrep_o '\sHostId="[^"]*' | cut -d '"' -f 2)
-  _hostname=$(echo "$_host" | _egrep_o '\sName="[^"]*' | cut -d '"' -f 2)
-  _hosttype=$(echo "$_host" | _egrep_o '\sType="[^"]*' | cut -d '"' -f 2)
-  _hostaddress=$(echo "$_host" | _egrep_o '\sAddress="[^"]*' | cut -d '"' -f 2)
-  _hostmxpref=$(echo "$_host" | _egrep_o '\sMXPref="[^"]*' | cut -d '"' -f 2)
-  _hostttl=$(echo "$_host" | _egrep_o '\sTTL="[^"]*' | cut -d '"' -f 2)
+  _hostid=$(echo "$_host" | _egrep_o ' HostId="[^"]*' | cut -d '"' -f 2)
+  _hostname=$(echo "$_host" | _egrep_o ' Name="[^"]*' | cut -d '"' -f 2)
+  _hosttype=$(echo "$_host" | _egrep_o ' Type="[^"]*' | cut -d '"' -f 2)
+  _hostaddress=$(echo "$_host" | _egrep_o ' Address="[^"]*' | cut -d '"' -f 2)
+  _hostmxpref=$(echo "$_host" | _egrep_o ' MXPref="[^"]*' | cut -d '"' -f 2)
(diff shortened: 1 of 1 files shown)


Neilpang:
sorry, but why not use a plain ` ` space char here ?

jim-p:
`' '` only matches one literal space, where `\s` or `[[:space:]]` match any whitespace character, no matter what it may be. It's also more clear to explicitly state what's being matched, it's easy to overlook a ` `, especially depending on the font and spacing.

Neilpang:
Is it parsing  html or xml content ?
if it's xml, I don't think there would be any other type of space.
what do you think ?

jim-p:
It's supposed to be XML. It may work with ` ` but I'd prefer to err on the side of caution and use the character class. I'm willing to try it the other way, though. As long as it produces the expected output on all platforms it should be OK.

Neilpang:
The function may use `sed`  on solaris 
```sh
_egrep_o() {
  if ! egrep -o "$1" 2>/dev/null; then
    sed -n 's/.*\('"$1"'\).*/\1/p'
  fi
}
```

Is your fix working there ?

jim-p:
I don't have access to anything that runs Solaris to check currently

Neilpang:
what I asked is  whether `'[[:space:]]'` is working for `sed` command.

I just tried, it's not working.

So, please use plain `' '` space char instead.

jim-p:
The character class works for me in `sed` on both FreeBSD (12.0) and Linux (Mint 19.1).

I can redo the PR with a space, but it seems like a step backward.

PR #2032, opened 2019-01-14
Dev: add support for All-Inkl.com ACME V2 request

4lexk:
please add acme V2 validation method for all-inkl.com

hook example: https://github.com/o1oo11oo/dehydrated-all-inkl-hook
API: http://kasapi.kasserver.com


Neilpang:
It's not a PR, it should be an issue.

PR #2043, opened 2019-01-19
Check for DNS records to apply, if not then sleep

raunsbaekdk:
Hi @Neilpang 

This commit adds another check before failing hard:

[Sun Jan 20 00:30:38 CET 2019] Sleep 30 seconds for the txt records to take effect
[Sun Jan 20 00:31:09 CET 2019] Verifying:DOMAIN.TLD
[Sun Jan 20 00:31:12 CET 2019] DOMAIN.TLD:Verify error:DNS problem: NXDOMAIN looking up TXT for _acme-challenge.DOMAIN.TLD
[Sun Jan 20 00:31:12 CET 2019] Removing DNS records.
[Sun Jan 20 00:31:13 CET 2019] Don't need to remove.
[Sun Jan 20 00:31:13 CET 2019] Please add '--debug' or '--log' to check more details.
[Sun Jan 20 00:31:13 CET 2019] See: https://github.com/Neilpang/acme.sh/wiki/How-to-debug-acme.sh

What it does is run a dig lookup on the domain, if the DNS record still has not been applied, it will sleep for 3 more seconds. It runs two dig commands, the first one gets the nameserver for the domain and the second one gets the TXT record. Using the nameserver prevents caching the result.

This helps on two problems:

1. You can lower the DNS sleep time
2. Getting SSL certificate does not fail hard

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3882,6 +3882,17 @@ $_authorizations_map"
 
     _info "Sleep $(__green $Le_DNSSleep) seconds for the txt records to take effect"
     _sleep "$Le_DNSSleep"
+
+    while true; do
+      doesTxtExist=$(dig @$(dig @8.8.8.8 $_main_domain ns +short | head -n1) TXT $txtdomain | grep "$txt")
+      if [ -z "$doesTxtExist" ]; then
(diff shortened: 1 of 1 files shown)


Neilpang:
https://github.com/Neilpang/acme.sh/issues/2015

PR #2053, opened 2019-01-24
Added dns_doapi.sh

Manawyrm:
Hi,

this new adapter adds support for the official LE API for the domain hosting company do.de. 
There is already support in acme.sh in the dns_do.sh adapter, but this API has some downsides: 
- it's only available to business/enterprise customers
- it requires you to save the full admin credentials (username/pw) in the server config of each machine
- old API uses SOAP/XML and is marked for deprecation

The new API is explicitly meant to set _acme-challenge records and provides authentication via a token mechanism. 
If you want to test this API, I will gladly send you a (test) token. Just tell me a mail address to send it to. 

The plugin was written according to the dev guidelines and also supports wildcard certifiates.

PS: I'm not 100% sure about the naming, but I didn't want to break compatibilty with the existing API. Suggestions for a different filename are welcome. 

Thanks,
Tobias

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,7 @@ You don't have to do anything manually!
 1. Exoscale.com API (https://www.exoscale.com/)
 1. PointDNS API (https://pointhq.com/)
 1. Active24.cz API (https://www.active24.cz/)
+1. do.de API (https://www.do.de/)
 
 And:
 
--- a/dnsapi/README.md
+++ b/dnsapi/README.md
@@ -1139,6 +1139,22 @@ You can then issue certs by using:
 ```acme.sh --issue --dns dns_pointhq -d example.com -d www.example.com
 ```
 
+## 59. Use do.de API
+
+Create an API token in your do.de account.
+
+Set your API token:
+```
+export DO_LETOKEN='FmD408PdqT1E269gUK57'
+```
+
+To issue a certificate run:
+```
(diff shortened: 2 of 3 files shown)


Manawyrm:
Apart from the fact that my name is dns_doapi.sh instead of doapi.sh (like it suggets in #2 of the guide), I think it follows all the guidelines. 

It can handle v2 wildcard and wildcard domains. I've successfully issued such a certificate. 
dig screenshot

Manawyrm:
Hi,

it can handle all these domains. The hoster API automatically handles that, and only needs to be provided with the FQDN.
It's a specialized LetsEncrypt API, build for this purpose. 

The API is already on the supported list as "Domain-Offensive/Resellerinterface/Domainrobot API". Don't need to mention it twice.

Neilpang:
Ok,

please add it to the list still. So that the list count keep same as the count in the dnsapi readme.

```sh
60.  do.de API
```

In the dns api readme, please move to `60`, not 59.

Thanks.

Neilpang:
OK.
please create a new issue:  https://github.com/Neilpang/acme.sh/issues

```sh
Report bug to do.de dns api
```

And watch to that issue,  if in future any bug is found,  it will be reported there.

Thanks.

Manawyrm:
OK, created #2057 and followed the issue 😄

PR #2076, opened 2019-02-03
Support for MyDevil.net

ahwayakchih:
Hi,

I needed to get a wildcard cert for my domain and found your project. Thanks to the nice documentation and example files i added DNS and deploy scripts and now i can access any random subdomain through HTTPS just fine :). Thanks!

I tried to follow instructions found at https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide.
Let me know if something else is needed.

One thing i was not sure about is where `acme.sh` is run usually. All the APIs seem to be external, so it looks like `acme.sh` can be run locally (on dev's comp, for example) and then deployed.
Scripts i wrote use `devil` command, which is available on target servers. So, `acme.sh` has to be run there. I mentioned it in DNS README, but i'm not sure that's enough.

~*Edit:* I did not create an issue for the API yet, as i'm not sure it will be merged.~
*Edit 2:* Issue for bug reports created at #2079 

Thanks again for a very nice script,
Regards

### DIFF
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -54,6 +54,6 @@ mydevil_deploy() {
 # Usage: ip=$(mydevil_get_ip domain.com)
 #        echo $ip
 mydevil_get_ip() {
-  devil dns list "$1" | awk '{print $3"\t"$7}' | grep "^A$(printf '\t')" | awk '{print $2}' || return 1
+  devil dns list "$1" | cut -w -s -f 3,7 | grep "^A$(printf '\t')" | cut -w -s -f 2 || return 1
   return 0
 }
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -27,6 +27,11 @@ mydevil_deploy() {
   _debug _cca "$_cca"
   _debug _cfullchain "$_cfullchain"
 
+  if ! _exists "devil" ; then
+    _err "Could not find 'devil' command."
+    return 1
+  fi
+
   ip=$(mydevil_get_ip "$_cdomain")
   if [ -z "$ip" ]; then
     _err "Could not find IP for domain $_cdomain."
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -5,7 +5,7 @@
 # MyDevil.net already supports automatic Let's Encrypt certificates,
 # except for wildcard domains.
 #
-# This script depends on `devil dns` that MyDevil.net provides,
+# This script depends on `devil` command that MyDevil.net provides,
 # which means that it works only on server side.
 #
 # Author: Marcin Konicki <https://ahwayakchih.neoni.net>
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -14,12 +14,12 @@
 
 # Usage: mydevil_deploy domain keyfile certfile cafile fullchain
 mydevil_deploy() {
-  local _cdomain="$1"
-  local _ckey="$2"
-  local _ccert="$3"
-  local _cca="$4"
-  local _cfullchain="$5"
-  local ip=""
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+  ip=""
 
   _debug _cdomain "$_cdomain"
   _debug _ckey "$_ckey"
@@ -49,8 +49,6 @@ mydevil_deploy() {
 # Usage: ip=$(mydevil_get_ip domain.com)
 #        echo $ip
 mydevil_get_ip() {
-  local domain=$1
-
-  devil dns list "$domain" | awk '{print $3"\t"$7}' | grep "^A$(printf '\t')" | awk '{print $2}' || return 1
+  devil dns list "$1" | awk '{print $3"\t"$7}' | grep "^A$(printf '\t')" | awk '{print $2}' || return 1
   return 0
 }
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -27,7 +27,7 @@ mydevil_deploy() {
   _debug _cca "$_cca"
   _debug _cfullchain "$_cfullchain"
 
-  if ! _exists "devil" ; then
+  if ! _exists "devil"; then
     _err "Could not find 'devil' command."
     return 1
   fi
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -28,7 +28,7 @@ mydevil_deploy() {
   _debug _cfullchain "$_cfullchain"
 
   ip=$(mydevil_get_ip "$_cdomain")
-  if [ -z "$ip" ] ; then
+  if [ -z "$ip" ]; then
     _err "Could not find IP for domain $_cdomain."
     return 1
   fi
--- a/deploy/mydevil.sh
+++ b/deploy/mydevil.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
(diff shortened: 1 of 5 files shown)


dnsapi/README.md
```diff
@@ -1172,6 +1172,32 @@ acme.sh --issue --dns dns_doapi -d example.com -d *.example.com
...
+```sh
+devil binexec on
+```
+
+Install acme.sh somewhere on your mydevil host account, probably in your home directory.
+Once it is installed, add it to your `~/bin` directory (and make sure it exists first):
+
+```sh
+mkdir ~/bin
+ln -s /path/to/installed/directory/of/.acme.sh/acme.sh ~/bin/acme.sh
```
Neilpang: It's net necessary to link in the bin folder.

dnsapi/dns_mydevil.sh
```diff
@@ -0,0 +1,100 @@
...
+# This script depends on `devil dns` that MyDevil.net provides,
+# which means that it works only on server side.
+#
+# Author: Marcin Konicki <https://ahwayakchih.neoni.net>
+#
+########  Public functions #####################
+
+#Usage: dns_mydevil_add   _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_mydevil_add() {
+  local fulldomain=$1
```
Neilpang: do not use `local` keyword.  it's bash only.

dnsapi/dns_mydevil.sh
```diff
@@ -0,0 +1,100 @@
...
+  local txtvalue=$2
+  local domain=""
+
+  _info "Using mydevil"
```
Neilpang: please use `_exists()` function to check if command `devil` is already existing.

dnsapi/dns_mydevil.sh
```diff
@@ -0,0 +1,100 @@
...
+    _info "Removing record $id from domain $domain"
+    devil dns del "$domain" "$id" || _err "Could not remove DNS record."
+  done
+}
+
+####################  Private functions below ##################################
+
+# Usage: mydevil_check_record "_acme-challenge.www.domain.com" || _err "Invalid record name"
+mydevil_check_record() {
+  local record=$1
```
Neilpang: you can not use `_acme-challenge.` prefix for subdomain and domain.
we have  dns alias mode: https://github.com/Neilpang/acme.sh/wiki/DNS-alias-mode

The full domain you get here is not necessarily starting with `_acme-challenge.`

It's recommended to `list` all he zones by `devil` and compare with each one to check which part is root domain and which is subomain.

see:  https://github.com/Neilpang/acme.sh/wiki/DNS-alias-mode

dnsapi/dns_mydevil.sh
```diff
@@ -0,0 +1,100 @@
...
+  esac
+}
+
+# Usage: domain=$(mydevil_get_domain "_acme-challenge.www.domain.com" || _err "Invalid domain name")
+#        echo $domain
+mydevil_get_domain() {
+  local fulldomain=$1
+  local domain=""
+
+  for domain in $(devil dns list | grep . | awk '{if(NR>1)print $1}'); do
```
Neilpang: do not use `awk`

dnsapi/README.md
ahwayakchih: It's not necessary, but it makes all the commands in examples a lot easier to read :).
Also `~/bin` is supported by default by MyDevil.net, e.g., `PATH` includes it by default.

dnsapi/dns_mydevil.sh
ahwayakchih: OK. I removed them.
This script is run by bash anyway, so i'm guessing you want to keep the code "universal"? In which case, should i change also `#!/bin/bash` to `#!/bin/sh`?

dnsapi/dns_mydevil.sh
ahwayakchih: OK.

dnsapi/dns_mydevil.sh
ahwayakchih: This function is only to make sure that the record is valid, and not some real, non-acme related record that should not be touched.

To get the root domain name, another function is used.

Is there a variable that keeps the prefix part of the name, that i could check?

dnsapi/dns_mydevil.sh
ahwayakchih: I'm using it in a couple more places. Can i ask why it should not be used? It makes things a lot easier.
If i stop using `awk`, can i use `jq` instead? Both are installed on MyDevil.net. I could change from using `devil` command to using unix socket to read/write settings, but it outputs JSON, so `jq` would be necessary (i do not want to write JSON parser in bash, sorry :).

ahwayakchih: @Neilpang I've replaced calls to `awk` with calls to `cut` and `tail`.

dnsapi/dns_mydevil.sh
ahwayakchih: @Neilpang i removed whole function and changed code so it does not depend on `_acme-challenge` prefix at all. Instead it makes sure to only remove `TXT` records that match both `txtdomain` and `txtvalue`.

dnsapi/README.md
ahwayakchih: @Neilpang i rewrote that part of README.

deploy/mydevil.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
Neilpang: do not use `bash`,  see the dev guide

dnsapi/dns_mydevil.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
Neilpang: bash

deploy/mydevil.sh
ahwayakchih: @Neilpang fixed

dnsapi/dns_mydevil.sh
ahwayakchih: @Neilpang fixed

dnsapi/dns_mydevil.sh
```diff
@@ -0,0 +1,97 @@
...
+#
+# This script depends on `devil` command that MyDevil.net provides,
+# which means that it works only on server side.
+#
+# Author: Marcin Konicki <https://ahwayakchih.neoni.net>
+#
+########  Public functions #####################
+
+#Usage: dns_mydevil_add   _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_mydevil_add() {
```
Neilpang: the name should be `dns_deveil`,  not `dns_mydeveil`,  please update all the names.

ahwayakchih: @Neilpang host name is mydevil.net and they are the ones who work on `devil` tool (AFAIK it's a set of custom scripts and it's not published anywhere). You can read more about it at: https://translate.google.com/translate?sl=pl&tl=en&u=https%3A%2F%2Fwiki.mydevil.net%2FDevil

Unless there's some other host, that uses the same exact tool (i do not know about any) it's probably better to keep the name as `mydevil` - just in case there is another `devil` API somewhere out there, that's completely different.

I cannot guarantee that it will work on any other host - i tested it only on mydevil.net.

Neilpang: alright,  I didn't noticed that.

ahwayakchih:
@Neilpang Thanks :)!

PR #2077, opened 2019-02-04
hosting.de: better parsing of json responses

membero:
fixes an error if customer does not have access to dns-groups
return 1 if one subfunction fails

### DIFF
--- a/dnsapi/dns_hostingde.sh
+++ b/dnsapi/dns_hostingde.sh
@@ -13,6 +13,7 @@ dns_hostingde_add() {
   txtvalue="${2}"
   _debug "Calling: _hostingde_addRecord() '${fulldomain}' '${txtvalue}'"
   _hostingde_apiKey && _hostingde_getZoneConfig && _hostingde_addRecord
+  return $?
 }
 
 dns_hostingde_rm() {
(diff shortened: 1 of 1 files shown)


dnsapi/dns_hostingde.sh
```diff
@@ -38,6 +40,18 @@ _hostingde_apiKey() {
...
 
+_hostingde_parse() {
+  find="${1}"
+  if [ "${2}" ]; then
+    notfind="${2}"
+  fi
+  if [ "${notfind}" ]; then
+    _egrep_o \""${find}\":.*" | grep -v "${notfind}" | cut -d ':' -f 2 | cut -d ',' -f 1 | tr -d '[:space:]'
+  else
+    _egrep_o \""${find}\":.*" | cut -d ':' -f 2 | cut -d ',' -f 1 | tr -d '[:space:]'
```
Neilpang: please do not use `'[:space:]'`, it's not supported crossing all the platforms.
please just use plain space char `' '` instead.

Thanks.

membero: Hi Neil,
I updated the tr command.

PR #2106, opened 2019-02-23
Fix verification for namecheap domains not *owned* by the calling user

EverPresent:
The NameCheap DNS plugin is using NameCheap's `domains.getList` api call to try and verify that the user has access to the requested domain.  That works great as long as the user is the primary owner of the domain.  If the user is just a user who has been granted administrative DNS rights to the domain, then the `domains.getList` api will not include that domain in the response.  See http://disq.us/p/1q6v9x9 for NameCheap's official comment (short story, getList only returns domains the user owns, not ones they've been granted admin rights for).

I changed the code so that it will first attempt to verify that the user has permissions to the domain by calling the original `domains.getList` api endpoint.  If that verification fails, then it falls back to calling the `domains.dns.getHosts` api endpoint directly.

### DIFF
--- a/dnsapi/dns_namecheap.sh
+++ b/dnsapi/dns_namecheap.sh
@@ -76,6 +76,22 @@ dns_namecheap_rm() {
 # _sub_domain=_acme-challenge.www
 # _domain=domain.com
 _get_root() {
+  fulldomain=$1
+
+  if ! _get_root_by_getList "$fulldomain"; then
+    _debug "Failed domain lookup via domains.getList api call. Trying domain lookup via domains.dns.getHosts api."
+    # The above "getList" api will only return hosts *owned* by the calling user. However, if the calling
+    # user is not the owner, but still has administrative rights, we must query the getHosts api directly.
+    # See this comment and the official namecheap response: http://disq.us/p/1q6v9x9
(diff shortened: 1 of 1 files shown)


dnsapi/dns_namecheap.sh
```diff
@@ -94,6 +110,10 @@ _get_root() {
       #not valid
       return 1
     fi
+    if ! _contains "$h" "\\."; then
```
EverPresent: I added this small guard branch because $h was being set to the TLD, which obviously can not be the only part of the domain name.  However, since my domain was not being returned in the response from the `getList` api call, the `if` statement a few lines down was finding that $response *does* contain the string "com" (unsurprisingly).  It was then setting the _sub_domain and _domain variables both to "com" (obviously wrong) and returning success.

With my update, this method will correctly return false for domains which are not returned in $response.  The _get_root method will then fall back to trying another method of verifying that this user has access to the requested domain.

Neilpang:
Good.

Would you like to create an issue for namecheap dns api?  so that users can report bugs to it.

Thanks.

EverPresent:
Sure.  I just created the issue at https://github.com/Neilpang/acme.sh/issues/2107

Issue #798, opened 2017-04-18
please add postfix support


amaxcz:
please add postfix deploy/reload script and examples how to configure it
Thanks.

TonyHoyle:
Here is a simple one I knocked together in a few minutes.  It copies the certs into /etc/postfix/certs then calls postfix reload to reload them.

`#Here is a script to deploy cert to postfix server.

#returns 0 means success, otherwise error.

########  Public functions #####################

#domain keyfile certfile cafile fullchain
postfix_deploy() {
  _cdomain="$1"
  _ckey="$2"
  _ccert="$3"
  _cca="$4"
  _cfullchain="$5"

  _debug _cdomain "$_cdomain"
  _debug _ckey "$_ckey"
  _debug _ccert "$_ccert"
  _debug _cca "$_cca"
  _debug _cfullchain "$_cfullchain"

  _ssl_path="/etc/postfix/certs"
  if ! mkdir -p "$_ssl_path"; then
    _err "Can not create folder:$_ssl_path"
    return 1
  fi

  cp $_ckey $_ssl_path
  cp $_ccert $_ssl_path
  cp $_cca $_ssl_path
  cp $_cfullchain $_ssl_path

  postfix reload
 
  return 0
}`

neilpang:
Closing as not planned for lack of interest -- PR welcome if someone wants to pick it up.

Issue #285, opened 2016-09-09
Add deploy hook support to deploy certs to nginx/apache/exim4/dovecot/cpanel/WebFaction  etc.


Neilpang:
It turns out that so many users don't know how to deloy the certs at all.

Just similar like what we did for dns api hooks,  I'd like to add hooks to deploy certs to a number of servers,  such as: nginx/apache/exim4/dovecot/cpanel/WebFaction  etc.

It may be used like:

```
#issue the cert
acme.sh  --issue  -d mydomain.com  -w ...........

#deply the certs
acme.sh  --deploy  -d mydomain.com  --nginx

# or
acme.sh  --deploy  -d mydomain.com  --apache

# or
acme.sh  --deploy  -d mydomain.com  --webfaction

# etc.
```

It's an early thought, but let's see.

Theliel:
Hi @Neilpang 

Some good news for cpanel. I'd successful deploy my test cert in one domain. cpanel API info is more or less clear.

cpanel API use 3 auth options, but only web tokens or plain user/pass dont required root or WHM access (so in theory, should work with most of all cpanel account). I believe that Web token is useless here (cpanel give you a session token after loggin), so best options should be user/pass. Port 2083 use SSL connection (security measures)

The problem... i believe perl or php dependence is needed. In my case, a simple perl script.
Note: All without root access, only SSH, on Godaddy Shared Hosting Linux Cpanel. Encode and UTF is needed only for parsing the output, no needed for deploy. LWP is needed because the header need to include the auth.  

My skills with shell scripts are limited, so i dont know if perl/php can be bypasses and use a pure shell, or in the worse case, maybe acme.sh could add perl/php scripts too.

```
#!/usr/bin/perl
use strict;
use LWP::UserAgent;
use LWP::Protocol::https;
use Encode;
use utf8;
use JSON;
use MIME::Base64;

my $username = "cpanel_username";
my $password = "cpanel_pass";
 
my $request = "https://127.0.0.1:2083/execute/SSL/install_ssl";

$ENV{PERL_LWP_SSL_VERIFY_HOSTNAME} = 0;

my $ua = LWP::UserAgent->new();

$ua->default_header(
    'Authorization' => 'Basic ' . MIME::Base64::encode("$username:$password"),
);

my ( $cert, $key );
{
    local $/;
    open ( my $fh, '<', '/path/to/example.com.cer' );
    $cert = <$fh>;
    close $fh;
  
[...]

PR #2108, opened 2019-02-25
Feature/deploy postfix, dovecot, and both together

jordanwiseman:
I wanted to setup a simple dovecot/postfix based mail server with LetsEncypt certs and liked the idea of using acme.sh.  As the deploy script for dovecot was a stub, and there was nothing for postfix, I made some using the vsftpd deploy hook as a template.

> this also seems to help #285 and #798


jordanwiseman:
one docker build still shows as queued, but cirrus ci says it finished in 8 minuets?  https://cirrus-ci.com/task/4846212919328768

Neilpang:
why both together ? I don't like it.

jordanwiseman:
basically, for when both services use the same hostname (i.e., on a single-server setup using something like mail.example.com instead of imap. and smtp.).  discussions on best practice aside, one deploy script with one cert is convenient in this case.

Neilpang:
you can  set multiple  deploy hooks for a single cert:

```
acme.sh --deploy    -d example.com   --deploy-hook  postfix  --deploy-hook  dovecot 
```

you don't need to make them together.

jordanwiseman:
awesome, thanks for the info!  somehow i missed that.

gassan:
I think it would be better to use doveconf, doveadm, postconf etc

For exampe:
```
# returns dovecot.conf path
doveconf -n | head -n 1 | awk '{print $NF}'
```

```sh
# if dovecot running reload after certs are updated
if [ "$(doveadm instance list | awk 'FNR == 2 {print $1F}')" = "yes" ]; then
  doveadm reload
fi
`
``

etc

PR #2123, opened 2019-02-26
Populate $Le_Domain in _installcert() if empty

### LINKED ISSUE #2122: $Le_Domain not populated when using --issue for use by reload command
Around about here:
https://github.com/Neilpang/acme.sh/blob/f84103918a0baeb3d1809a9eff07de958fdfd197/acme.sh#L4829

When using `--issue`, `$Le_Domain` is not populated yet for usage by the command passed to `--reloadcmd`. The domain is passed as `$1` to `_installcert` where the reload command is called, but it never does anything with it.

mjbnz:
This allows the reload command to use $Le_Domain when called during initial cert --issue

(This should resolve #2122)

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4758,6 +4758,10 @@ _installcert() {
   _reload_cmd="$6"
   _backup_prefix="$7"
 
+  if [ "$Le_Domain" = "" ]; then
+    Le_Domain="$_main_domain"
+  fi
+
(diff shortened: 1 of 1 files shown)


Neilpang:
fixed by https://github.com/Neilpang/acme.sh/commit/b3f6129718bf0e7b7f352344b7149c725cf1576b

PR #2175, opened 2019-03-19
dns_gdnsdk: Fixed stupid regex error, want literal "-", not a range

loial:


Neilpang:
is this enough ?

loial:
Yeah, this fixes the sed error "Invalid range end", by placing '-' at the end, instead of between . and _

Issue #2132, opened 2019-03-02
Rate limited with Route53


qcybb:
I just got rated limited using acme.sh with Amazon's Route 53. It appears the script is trying to renew my 9 domains too quickly.

Is there a way to slow things down between each domain renewal attempt?

`
[Sat Mar  2 06:00:03 PST 2019] Response error:<?xml version="1.0"?>< ErrorResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/"><Error><Type>Sender</Type><Code>Throttling</Code><Message>Rate exceeded</Message></Error><RequestId>******-****-****</RequestId></ErrorResponse>
`

Moser-ss:
I think I have the same issue, in my case that happens when I try to add DNS records.
I create a PR for that  https://github.com/Neilpang/acme.sh/pull/2197

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

neilpang:
Fixed in ea6a3c09: dns_aws sleeps between Route53 calls and supports AWS_DNS_SLOWRATE=<seconds> to further slow down and avoid throttling.

PR #2197, opened 2019-03-29
Add aws api slowrate 

Moser-ss:
PR to address issue #2132 , it adds a new arg `--dnsslowrate` that will insert the sleep command inside of dns_aws.sh in the method dns_aws_add.
The unit of `--dnsslowrate` is second

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3974,9 +3974,16 @@ $_authorizations_map"
             return 1
           fi
 
-          if ! $addcommand "$txtdomain" "$txt"; then
-            _err "Error add txt for domain:$txtdomain"
-            return 1
+          if [ "$addcommand" = "dns_aws_add" -a -n "$_dnsslowrate" ] ; then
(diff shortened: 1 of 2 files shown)


Neilpang:
you don't need a new option.

you can use a new env like:  `AWS_DNS_SLOWRATE`

Moser-ss:
So are you suggesting to create the env `AWS_DNS_SLOWRATE` instead of an argument?
Any reason why?

Neilpang:
I'm not sure we need such a `slow rate` mode at all.

It's just a common way to talk to your dns api script.  acme.sh won't add any dedicated commandline option for a specify dns api script.

If you want to pass some options to dns api script, use env variable.

Moser-ss:
Thanks for the feedback I will change the script to use env variable.
I notice the need for `slow rate` because the AWS has a rate limit of 5 requests per minute and when I try to create a certificate with 100 domains it fails because of the rate limit

Moser-ss:
Can I open a new PR with the changes?

PR #2205, opened 2019-04-02
CloudFlare dns invalid domain

kirpichiki:
Currently dns_cf generates headers like this: 'X-Auth-Email: "sample@mail.com"'. Cloudflare API responses 400 BadRequest for quoted headers with message "Invalid format for X-Auth-Email header".

#1980

### DIFF
--- a/dnsapi/dns_cf.sh
+++ b/dnsapi/dns_cf.sh
@@ -182,8 +182,11 @@ _cf_rest() {
   data="$3"
   _debug "$ep"
 
-  export _H1="X-Auth-Email: $CF_Email"
-  export _H2="X-Auth-Key: $CF_Key"
+  email_trimmed=$(echo $CF_Email | tr -d '"')
+  key_trimmed=$(echo $CF_Key | tr -d '"')
+
+  export _H1="X-Auth-Email: $email_trimmed"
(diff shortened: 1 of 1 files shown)


Neilpang:
why it generates with quoted email ?

kirpichiki:
I didn't realize that until your comment. The "bug" appears in docker environment.
As mentioned in the docs, I tried to append environment variables CF_Key and CF_Email.
Now watch the trick:
```
(in docs)
export CF_Key="somekey"
```
```
(in my docker-compose)
environment:
  - CF_Key="somekey"
```

BUT the first command results variable content **`somekey`** and the second - **`"somekey"`**.
So, this is not even acme.sh bug, but the carelessness mistake of mine.
However, this mistake is really hard to determine and I think I am not the only one who catches it. And my fool-proof fix still in force.
As alternative option we can mention this problem in docs, but this solution will be less effective.

kirpichiki:
And in light of recent events, I thought the fix should be at a lower level, more global.

anglinb:
Was trying to figure this out, ended up with `CF_Token` instead of `CF_Key` and it works for me, hopefully this finds someone coming from google.

PR #2207, opened 2019-04-02
dns_hostingde.sh zoneConfig missing templateValues object

Kimmax:
These changes are not yet reflected in the official API documentation, but I could verify that this field is required by testing and via mail exchange with an employee.

### DIFF
--- a/dnsapi/dns_hostingde.sh
+++ b/dnsapi/dns_hostingde.sh
@@ -97,7 +97,7 @@ _hostingde_getZoneConfig() {
       zoneConfigDnsServerGroupId=$(echo "${curResult}" | _hostingde_parse "dnsServerGroupId")
       zoneConfigEmailAddress=$(echo "${curResult}" | _hostingde_parse "emailAddress")
       zoneConfigDnsSecMode=$(echo "${curResult}" | _hostingde_parse "dnsSecMode")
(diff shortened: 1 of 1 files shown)


Neilpang:
Does this change break the current official api ?

Kimmax:
It is already broken, this change should fix it again. I tested it on both a domain which is tied to a template and with one that is not, both worked fine. Ran it against 32 domains after that, did also work without problems

membero:
Wonderful. I had not in mind that zones might be tied to templates. I can confirm that this patch fixes the problem also for the hosting.de endpoint as both endpoints point to the same system.

membero:
Hello @Kimmax,

there is a bug report with your PR:  https://github.com/Neilpang/acme.sh/issues/2058

I tested a little bit and if I replace _hostingde_parse_no_strip_whitespace with _hostingde_parse, everything runs fine (with and without a template). Could you explain why you implemented _hostingde_parse_no_strip_whitespace?

Could you please test the attached test?

[hostingde-template-patch.txt](https://github.com/Neilpang/acme.sh/files/3156770/hostingde-template-patch.txt)

Kimmax:
@membero Hey, sorry for the late response, I have been busy all month. I'll look into this as soon as I can!

PR #2210, opened 2019-04-07
Fix gcloud most-specific zone match

chasefox:
Most specific zone selected by deepest sub-domain (how many '.' in the domain)
rather than seemingly irrelevant count of the number of characters within the zone.

### DIFF
--- a/dnsapi/dns_gcloud.sh
+++ b/dnsapi/dns_gcloud.sh
@@ -139,7 +139,7 @@ _dns_gcloud_find_zone() {
     --format="value(name, dnsName)" \
     --filter="$filter" \
     | while read -r dnsName name; do
-      printf "%s\t%s\t%s\n" "$(awk -F"." '{print NF-1}' <<< "$name")" "$dnsName" "$name"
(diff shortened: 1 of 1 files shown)


Neilpang:
sorry,  can you please explain more ?

chasefox:
In google cloud DNS, you have zones which have a _zone name_ and a _dns name_. The _zone name_ is a unique arbitrary value that describes the zone. The _dns name_ is  the FQDN suffix of the zone.

So let's consider two zones:

- Zone `bbb` with dns name `xxx.example.com.`
- Zone `aaa` with dns name `example.com.`

pre-commit of this PR would match both zones because it compared the length of characters of the zone name. In this example, both zone names have a length of characters of 3, and the less-specific zone `aaa` would be chosen due to the alphabetical precedence of the zone name.

This patch changes the zone selection to look at the _dns name_ and count not just the length of characters, but the depth of the domain based on the number of `.` characters in the dns name.

Neilpang:
good,  would you like to create an issue to handle future bugs for gcloud ?

Thanks.

Neilpang:
see this:

PR #2236, opened 2019-04-28
Correct parsing error in sed

mod242:
I corrected an stupid parsing error in the sed which lead to problems when first issuing an certificate (happened after I removed -e in the last PR)


Neilpang:
what is dns_schlundtech.sh  ?

mod242:
Oh man, I am so stupid, I wanted to seperate this into two pull requests: One that corrects an parsing error in my dns_ddnss script and a second one for a new API for the DNS-Provider "schlundtech.de". Should I close this PR and generate two new ones? So sorry for generating this overhead for you, Neil

mod242:
Removed dns_schlundtech from this PR. So this is the change I really wanted to commit.

PR #2255, opened 2019-05-09
dns_acmeproxy: Username/password no longer required

mdbraber:
Due to a new version of `acmeproxy` that offers the option to use IP-based authentication on the server side, username and password are no longer required (but still possible) on the client side.

### DIFF
--- a/dnsapi/dns_acmeproxy.sh
+++ b/dnsapi/dns_acmeproxy.sh
@@ -46,15 +46,6 @@ _acmeproxy_request() {
     return 1
   fi
 
-  ## Check for the credentials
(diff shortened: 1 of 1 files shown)


dnsapi/dns_acmeproxy.sh
```diff
@@ -46,15 +46,6 @@ _acmeproxy_request() {
     return 1
   fi
 
-  ## Check for the credentials
```
Neilpang: Can you add a message here ?  something like: `_info "ACMEPROXY_USERNAME" is not set , try to use ip based authentication.`

mdbraber:
Update the logic to only use authentication when ACMEPROXY_USERNAME and ACMEPROXY_PASSWORD are set and give appropriate information / help when not set

PR #2259, opened 2019-05-11
Add mail notify

honzahommer:
Notify hook that uses sendmail.


Neilpang:
I don't like a dedicated hook for sendmail only.
please implement it in  the `mail.sh` hook: https://github.com/Neilpang/acme.sh/blob/master/notify/mail.sh

In this hook,  please try to use both `mail` command and `sendmail` command.

if `mail` exists, we use `mail`,   if `sendmail` exists  we use `sendmail`.

what do you think ?

Neilpang:
it can also try with `mutt` if `mail` or `sendmail` neither exists.
see: 
https://tecadmin.net/ways-to-send-email-from-linux-command-line/

Neilpang:
please make sure you tested EVERY  commands, such as:  `mail`, `sendmail` or `mutt`

honzahommer:
> I don't like a dedicated hook for sendmail only.
> please implement it in the `mail.sh` hook: https://github.com/Neilpang/acme.sh/blob/master/notify/mail.sh
> 
> In this hook, please try to use both `mail` command and `sendmail` command.
> 
> if `mail` exists, we use `mail`, if `sendmail` exists we use `sendmail`.
> 
> what do you think ?

Yeah, great idea. I move sendmail to mail and add mail notify support. But I need to check if mail command works correctly - it's not done yet.

honzahommer:
> please make sure you tested EVERY commands, such as: `mail`, `sendmail` or `mutt`

I tested all methods, works fine.

notify/mail.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env sh
 
-# support local mail app
+#Support local mail app
+
+#MAIL_FROM="yyyy@gmail.com"
+#MAIL_TO="yyyy@gmail.com"
```
Neilpang: we need a public `MAIL_BIN` variable.   The user can choose which app to use, if both `mail` and `sendmail` are installed.

The MAIL_BIN variable is optional.

notify/mail.sh
```diff
@@ -10,6 +13,69 @@ mail_send() {
...
+  MAIL_FROM="${MAIL_FROM:-$(_readaccountconf_mutable MAIL_FROM)}"
+  if [ -z "$MAIL_FROM" ]; then
+    MAIL_FROM="$USER@$(hostname -f)"
+    _info "The MAIL_FROM is not set, so use the default value: $MAIL_FROM"
+  fi
+  _saveaccountconf_mutable MAIL_FROM "$MAIL_FROM"
+
+  MAIL_TO="${MAIL_TO:-$(_readaccountconf_mutable MAIL_TO)}"
+  if [ -z "$MAIL_TO" ]; then
+    MAIL_TO="$(_readaccountconf ACCOUNT_EMAIL)"
```
Neilpang: `ACCOUNT_EMAIL`  can be empty too.
please check and throw error here.

Neilpang:
please create a new issue for reporting bugs, like this:

notify/mail.sh
honzahommer: - Both e-mails are now validated,
- MAIL_FROM is not required,
- MAIL_TO is not saved if ACCOUNT_EMAIL is used.

notify/mail.sh
honzahommer: Done.

honzahommer:
> please create a new issue for reporting bugs, like this:
> https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide#10-please-create-a-new-issue-for-future-bugs

https://github.com/Neilpang/acme.sh/issues/2263

Neilpang:
please  update your usage here: 

notify/mail.sh
```diff
@@ -10,6 +14,107 @@ mail_send() {
   _debug "_content" "$_content"
   _debug "_statusCode" "$_statusCode"
 
-  _err "Not implemented yet."
-  return 1
+  unset -f _MAIL_BIN _MAIL_BODY _MAIL_CMD
```
Neilpang: you don't need `unset` anything,   the `mail_send()` function should be called in an isolated shell, so, the variable should be empty already.  please double check.

notify/mail.sh
```diff
@@ -10,6 +14,107 @@ mail_send() {
...
+
+  MAIL_BIN="${MAIL_BIN:-$(_readaccountconf_mutable MAIL_BIN)}"
+  if [ -n "$MAIL_BIN" ] && ! _exists "$MAIL_BIN"; then
+    _err "It seems that the command $MAIL_BIN is not in path."
+    return 1
+  fi
+  _MAIL_CMD=$(_mail_cmnd)
+  if [ -n "$MAIL_BIN" ]; then
+    _saveaccountconf_mutable MAIL_BIN "$MAIL_BIN"
```
Neilpang: add  `else`  here.  

```
else 
  _clearaccountconf "MAIL_BIN"
fi
```

honzahommer:
> please update your usage here:
> https://github.com/Neilpang/acme.sh/wiki/notify

https://github.com/Neilpang/acme.sh/wiki/notify#3-set-notification-for-mail

PR #2269, opened 2019-05-17
Challenge route is not a POST-as-GET, should have empty json object

martijnjanssen:
To verify the validity of a challenge, there should be an empty body in the post request. The route for a challenge is not listed as a POST-as-GET route in the acme v2 specification.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4289,7 +4289,7 @@ $_authorizations_map"
       sleep 2
       _debug "checking"
       if [ "$ACME_VERSION" = "2" ]; then
-        _send_signed_request "$uri"
+        _send_signed_request "$uri" "{}"
       else
         response="$(_get "$uri")"
       fi


Neilpang:
is there any problem you saw?

please give me the reference to the spec.

Thanks.

cpu:
> To verify the validity of a challenge, there should be an empty body in the post request.

That's to initiate a challenge, like in `__trigger_validation`: https://github.com/Neilpang/acme.sh/blob/0c6bf51c7e3d3e00e2c57a05c34074688194acb6/acme.sh#L3577

To poll a challenge to see what its status is (like is happening on the line you're changing) **is** a POST-as-GET. It shouldn't have a `{}` body in the JWS.

I don't think this PR matches the RFC.

martijnjanssen:
Wait, you're right. Otherwise it's an attempt to initiate another challenge. I'll close the pr!

PR #2274, opened 2019-05-19
Pass arguments to installOnline function

honzahommer:
Allow specify options in install online function as referenced in issue https://github.com/Neilpang/get.acme.sh/issues/2. 

Backward compatibility with actual allowed options, related PR https://github.com/Neilpang/get.acme.sh/pull/5.


msva:
@Neilpang ping?

I'd very like to have that possibility too (for example, to be able to don't install cron and profile on `curl|sh` installation).

Although, I personally not really like the logic in the PR code, but assuming it is work is intended - let it be, imo.

honzahommer:
@msva I know, little messy, but necessary for backward compatibility :-(

PR #2276, opened 2019-05-20
updated to work with one.com

der-berni:
rev command not found on OpenWrt
CURL does not work, using wget
JSESSIONID replaced with OneSIDCrmAdmin
CSRF_G_TOKEN not needed

### DIFF
--- a/dnsapi/dns_one.sh
+++ b/dnsapi/dns_one.sh
@@ -110,19 +110,19 @@ dns_one_rm() {
 # _sub_domain=_acme-challenge.www
 # _domain=domain.com
 _get_root() {
-  domain=$1
+  domain="$1"
   i=2
   p=1
   while true; do
     h=$(printf "%s" "$domain" | cut -d . -f $i-100)
-	
+
     if [ -z "$h" ]; then
       #not valid
       return 1
     fi
-    
+
     response="$(_get "https://www.one.com/admin/api/domains/$h/dns/custom_records")"
-    
+
     if ! _contains "$response" "CRMRST_000302"; then
       _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-$p)
       _domain="$h"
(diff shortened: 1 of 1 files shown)


dnsapi/dns_one.sh
```diff
@@ -4,48 +4,68 @@
 # one.com ui wrapper for acme.sh
 # Author: github: @diseq
 # Created: 2019-02-17
+# Fixed by: @der-berni
+# Modified: 2019-05-20
 #
-#     export ONECOM_User="username"
-#     export ONECOM_Password="password"
+#     export ONECOM_USER="username"
```
Neilpang: do not change the name from  `ONECOM_User`  to  `ONECOM_USER`.
The existing users will kill you.

der-berni: That was my fault. I misread.
https://github.com/Neilpang/acme.sh/pull/2098#discussion_r258946332

See my second commit.

dnsapi/dns_one.sh
```diff
@@ -41,11 +57,15 @@ dns_one_add() {
   postdata="$postdata&password1=$ONECOM_Password"
   postdata="$postdata&loginTarget="
   #_debug postdata "$postdata"
-
+  
+  #CURL does not work
+  local tmp_USE_WGET=$ACME_USE_WGET
+  ACME_USE_WGET=1
```
Neilpang: No,  never select to use wget or curl in your script.
It's control by acme.sh

if curl does not work,  you can display an `_err` message to ask the user to use wget wih `--use-wget`

Neilpang: I still don't understand why curl is not working ?

der-berni: the response from curl is every time the login page.
the same results on other devices.
i will try to get more info

der-berni: Without the follow redirects (Option) "-L" it works with curl

der-berni: i fixed to work with curl

Neilpang:
fix the CI errors first.

Neilpang:
would you like to create an issue for bug report: 

PR #2278, opened 2019-05-21
SES support

raven-kg:
Add Amazon SES support as backend to send notifications

### DIFF
--- a/notify/ses.sh
+++ b/notify/ses.sh
@@ -58,7 +58,7 @@ ses_send() {
   _saveaccountconf_mutable AWS_SES_FROM "$AWS_SES_FROM"
 
   _date="$(date -R)"
-  _signature="$(echo -n "$_date" | ${ACME_OPENSSL_BIN:-openssl} dgst -sha256 -hmac "$AWS_SECRET_KEY" -binary | base64 -w 0)"
+  _signature="$(printf "%s" "$_date" | _hmac sha256 $AWS_SECRET_KEY | _base64)"
   _endpoint="https://email.$AWS_REGION.amazonaws.com"
 
   export _H1="X-Amzn-Authorization: AWS3-HTTPS AWSAccessKeyId=$AWS_ACCESS_KEY, Algorithm=HmacSHA256, Signature=$_signature"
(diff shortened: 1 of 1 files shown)


notify/ses.sh
```diff
@@ -0,0 +1,80 @@
...
+  AWS_SES_FROM="${AWS_SES_FROM:-$(_readaccountconf_mutable AWS_SES_FROM)}"
+  if [ -z "$AWS_SES_FROM" ]; then
+    AWS_SES_FROM=""
+    _err "You didn't specify an email to AWS_SES_FROM receive messages."
+    return 1
+  fi
+  _saveaccountconf_mutable AWS_SES_FROM "$AWS_SES_FROM"
+
+  _date="$(date -R)"
+  _signature="$(echo -n "$_date" | ${ACME_OPENSSL_BIN:-openssl} dgst -sha256 -hmac "$AWS_SECRET_KEY" -binary | base64 -w 0)"
```
Neilpang: don't use  `openssl`,  we have `_hmac()` function for use.

Here are good examples:

https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_aws.sh#L298

raven-kg: Changed it to use _hmac and _base64 functions. Works fine on my side, but there is one question regarding to variable naming. 
Here I used AWS_ACCESS_KEY and AWS_SECRET_KEY as variable names, the dns_aws.sh script uses AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY as the same variable names. Should I use the same variable names in both scripts or will it be better to make them different (SES account may differ from route53 account)?

pingce: Please use the same name: AWS_ACCESS_KEY_ID 

The name was borrowed from AWS official SDK,   that means, the user doesn't need to set it explicitly.  We can get the value automatically.

Neilpang: yes, please use the same name.

raven-kg: Done)

szhu25:
Hey @raven-kg, do you still plan to continue your work on this notification hook? I would like to make something similar...

Issue #1835, opened 2018-09-10
Cant generate certificate in freeDNS that has multiple top level domains.


felixgonsug:
Im trying to issu a cert to *.mydomain.org.uy with the freedns api, but when generating its gives an error.

Found domain api file: /var/db/acme/.acme.sh/dnsapi/dns_freedns.sh
Add TXT record using FreeDNS
Domain org.uy not found at FreeDNS
Retry loading subdomain page (1 attempts remaining)
Domain org.uy not found at FreeDNS
Error add txt for domain:_acme-challenge.mydomain.org.uy
Please add '--debug' or '--log' to check more details.
See: https://github.com/Neilpang/acme.sh/wiki/How-to-debug-acme.sh

the domain i own is mydomain.org.uy not .org.uy, is this contemplated in acme?

Thanks in advance.

felixgonsug:
#1086 Here is  solution

frebib:
I opened a PR with a fix awaiting a merge: https://github.com/Neilpang/acme.sh/issues/1807

dkerr64:
@felixgonsug can you please test the version of freedns api script here... https://raw.githubusercontent.com/dkerr64/acme.sh/FreeDNS/dnsapi/dns_freedns.sh
Thank you

dkerr64:
PR #2285 opened

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

neilpang:
Fixed in 2.8.2 (c74686a1, PR #2285): dns_freedns was rewritten and handles multi-level public suffixes like .org.uy.

PR #2285, opened 2019-05-23
Fixes to dns_freedns.sh for multi-part top level domain names

dkerr64:
Fix for multi-part top level domains not working.  Fixes issue #1086 and issue #1835 
Also cleaned up the code to simplify and hopefully make it more robust.

### DIFF
--- a/dnsapi/dns_freedns.sh
+++ b/dnsapi/dns_freedns.sh
@@ -58,7 +58,7 @@ dns_freedns_add() {
     _debug "sub_domain: $sub_domain"
     _debug "top_domain: $top_domain"
 
-    DNSdomainid="$(_freedns_domain_id $top_domain)"
+    DNSdomainid="$(_freedns_domain_id "$top_domain")"
     if [ "$?" = "0" ]; then
       _info "Domain $top_domain found at FreeDNS, domain_id $DNSdomainid"
       break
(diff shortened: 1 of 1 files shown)


dnsapi/dns_freedns.sh
```diff
@@ -138,76 +96,44 @@ dns_freedns_rm() {
   FREEDNS_COOKIE="$(_read_conf "$ACCOUNT_CONF_PATH" "FREEDNS_COOKIE")"
```
Neilpang: why you need to save the cookie ?
And never use `_read_conf()` in your code.

dkerr64: Once you login to FreeDNS for the first time you are issued a cookie to use for all future requests.  The cookie becomes invalid if user changes their password.  You don't want to save the userid/password and login every time, instead you use the cookie on every request.  This diff/compare also clips of the comments inline in the code that explained this. namely...
```
# Need to read cookie from conf file again in case new value set
# during login to FreeDNS when TXT record was created.
# acme.sh does not have a _readaccountconf() function
FREEDNS_COOKIE="$(_read_conf "$ACCOUNT_CONF_PATH" "FREEDNS_COOKIE")"
```
Of course as I look at acme.sh now, I see that you do have a readaccountconf() function that did not exist when I originally wrote this.  So, I can change to use that.  I also see a "mutable" version... can you explain those please?
Thanks

dkerr64: Made the change to use _readaccountconf()

Neilpang:
please create an issue for bug report:  https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide#10-please-create-a-new-issue-for-future-bugs

And then add your issue link to the doc: https://github.com/Neilpang/acme.sh/wiki/dnsapi
example:  https://github.com/Neilpang/acme.sh/wiki/dnsapi

dkerr64:
> please create an issue for bug report: https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide#10-please-create-a-new-issue-for-future-bugs
> 
> And then add your issue link to the doc: https://github.com/Neilpang/acme.sh/wiki/dnsapi
> example: https://github.com/Neilpang/acme.sh/wiki/dnsapi

Done.

PR #2293, opened 2019-05-26
Notify mail update

honzahommer:
Fix:
- https://github.com/Neilpang/acme.sh/issues/2263#issuecomment-494920987
- https://github.com/Neilpang/acme.sh/issues/2263#issuecomment-495089133

Feat:
- allow to disable e-mail validation `MAIL_NOVALIDATE` (allow to use e-mail address without 'at'; local delivery, etc.)

### DIFF
--- a/notify/mail.sh
+++ b/notify/mail.sh
@@ -62,7 +62,7 @@ mail_send() {
 
   contenttype="text/plain; charset=utf-8"
   subject="=?UTF-8?B?$(echo "$_subject" | _base64)?="
-  result=$({ _mail_body | _mail_cmnd; } 2>&1)
+  result=$({ echo "$(_mail_body)" | eval $(_mail_cmnd); } 2>&1)
 
   # shellcheck disable=SC2181
   if [ $? -ne 0 ]; then
--- a/notify/mail.sh
+++ b/notify/mail.sh
@@ -28,13 +28,11 @@ mail_send() {
     return 1
   fi
   _MAIL_BIN=$(_mail_bin)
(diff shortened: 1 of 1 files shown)


notify/mail.sh
```diff
@@ -55,8 +62,9 @@ mail_send() {
 
   contenttype="text/plain; charset=utf-8"
   subject="=?UTF-8?B?$(echo "$_subject" | _base64)?="
-  result=$({ echo "$_MAIL_BODY" | eval "$_MAIL_CMD"; } 2>&1)
+  result=$({ _mail_body | _mail_cmnd; } 2>&1)
```
zuexo: This does not work for me. Had to change it to suggested to make it work.
```suggestion
  result=$({ echo "$(_mail_body)" | eval $(_mail_cmnd); } 2>&1)
```

zuexo:
I've tested it with `sendmail` and `mail` bin and it seems to work after changing line 65. But I did notice one thing. Using `mail` command to send email doesn't use MAIL_FROM if set.

honzahommer:
> I've tested it with `sendmail` and `mail` bin and it seems to work after changing line 65. But I did notice one thing. Using `mail` command to send email doesn't use MAIL_FROM if set.
> 
> Edit: I don't have mutt installed, but I expect it also doesn't use MAIL_FROM.

There are many implementations of mail command, so there is no chance to implement MAIL_FROM :-(

honzahommer:
@Neilpang https://github.com/Neilpang/acme.sh/issues/2263#issuecomment-496131539

PR #2299, opened 2019-05-28
allowing for greater than 100 zones to be returned in a single call.

mianosm:
Worked with @syndex101 and UltraDNS customer support to allow for all zones to be returned instead of a paginated method.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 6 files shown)


Neilpang:
does `1000` means all the zones ?

mianosm:
This does not match the documentation now that I'm reading the user guide available via Neustar. The offset is the position in the list for the first returned element (default is "0", which makes sense.

The limit parameter will simply return 100 by default, and could likely be figured out with a bit of logic, how may times to iterate over the list of zones. I'll do more testing with a lower limit to suss this out a bit more thoroughly.

mianosm:
> does `1000` means all the zones ?

No, it doesn't. I've found a bit more information in the documentation. Will look into iteratively calling for the zones so that, in the case that a large number of zones are present this will still work.

PR #2301, opened 2019-05-29
Deploy hook for Sophos XG

andrewheberle:
This deploy hook allows a certificate to be pushed to a Sophos XG appliance.

The certificate is not assigned to any use/services. This must be done by the admin.

The initial deployment will be an "add" but subsequent renewals will "update" the existing certificate.

It is expected the first deployment will fail on the "update" step ("update" is tried first as renewing a certificate will happen every two months, but issuing a new certificate will only happen rarely, maybe only once), but then it will attempt the "add" step to add the new certificate.

### Basic Usage

```
env DEPLOY_SOPHOSXG_HOST="IPADDRESS:4444" \
  DEPLOY_SOPHOSXG_USER="admin" \
  DEPLOY_SOPHOSXG_PASSWORD="adminpassword"
  acme.sh -d <domain> --deploy --deploy-hook sophosxg
```

### Full options/env vars

- **DEPLOY_SOPHOSXG_HOST**: REQUIRED - Appliance admin UI "host:port"
- **DEPLOY_SOPHOSXG_USER**: REQUIRED - Admin user
- **DEPLOY_SOPHOSXG_PASSWORD**: REQUIRED - Admin password
- **DEPLOY_SOPHOSXG_NAME**: OPTIONAL - Defaults to "<domain>" - Friendly name for certificate
- **DEPLOY_SOPHOSXG_PFX_PASSWORD**: OPTIONAL - Defaults to "s0ph0sXG" - password for temporary Pkcs file
- **DEPLOY_SOPHOSXG_HTTPS_INSECURE**: OPTIONAL - Defaults to "1" - Don't verify HTTPS on appliance (often Admin UI is self-signed)

### Other Requirements

1. API access needs to be enabled on the Sophos XG appliance
2. Requires "curl" to work...purely because I am not sure how to do the same process with "wget".


deploy/sophosxg.sh
```diff
@@ -0,0 +1,197 @@
...
+#domain keyfile certfile cafile fullchain
+sophosxg_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  # check for curl first
+  if _exists "curl"; then
```
Neilpang: why curl is required ?
Can you please use `_get()` or `_post()` functions instead ?

andrewheberle: I've updated my branch to remove this requirement.

I am now building the required "multipart/form-data" request body so it can be done via "_post"

andrewheberle:
Apologies for the bunch of commits...but this should now be good to go.

deploy/sophosxg.sh
```diff
@@ -0,0 +1,188 @@
...
+  _do_req_name="$5"
+  _do_req_pfxpass="$6"
+  _do_req_host="$7"
+
+  # static values - as variables in case these need to change
+  _do_req_boundary="SOPHOSXGPOST"
+  _do_req_certfile="certificate.p12"
+
+  # dont verify certs if config set
+  _do_req_old_HTTPS_INSECURE="${HTTPS_INSECURE}"
```
Neilpang: don't need to save _do_req_old_HTTPS_INSECURE

the shell is isolated,  you won't affect others.

deploy/sophosxg.sh
```diff
@@ -0,0 +1,188 @@
...
+  _cfullchain="$5"
+
+  # Some defaults
+  DEFAULT_SOPHOSXG_PFX_PASSWORD="s0ph0sXG"
+  DEFAULT_SOPHOSXG_NAME="$_cdomain"
+  DEFAULT_SOPHOSXG_HTTPS_INSECURE="1"
+
+  if [ -f "$DOMAIN_CONF" ]; then
+    # shellcheck disable=SC1090
+    . "$DOMAIN_CONF"
```
Neilpang: dont load DOMAIN_CONF

deploy/sophosxg.sh
andrewheberle: ack

PR #2369, opened 2019-06-30
Initial support for Synology DSM deployhook

tresni:
This allows you to update a key on a Synology DSM using the existing API.
Handles restarting the necessary services the certificate is attached to and all other internal stuff (copying the certificate around, etc.)

This is way less error prone than most articles I've found on how to update a Synology DSM certificate.

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -39,6 +39,7 @@ synology_dsm_deploy() {
   # Get Username and Password, but don't save until we successfully authenticate
   SYNO_Username="${SYNO_Username:-$(_getdeployconf SYNO_Username)}"
   SYNO_Password="${SYNO_Password:-$(_getdeployconf SYNO_Password)}"
+  SYNO_Create="${SYNO_Create:-$(_getdeployconf SYNO_Create)}"
   if [ -z "$SYNO_Username" ] || [ -z "$SYNO_Password" ]; then
     SYNO_Username=""
     SYNO_Password=""
@@ -108,8 +109,8 @@ synology_dsm_deploy() {
   id=$(echo "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\"id\":\"\([^\"]*\).*/\1/p")
   _debug2 id "$id"
 
-  if [ -z "$id" ]; then
-    _err "Unable to find certificate: $SYNO_Certificate"
+  if [ -z "$id" ] && [ -z "$SYNO_Create" ]; then
+    _err "Unable to find certificate: $SYNO_Certificate and \$SYNO_Create is not set"
     return 1
   fi
 
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -22,7 +22,7 @@
 
 _syno_get_cookie_data() {
   _debug2 Cookie "$1"
-  _debug3 grep "$(grep "\W$1=" "$HTTP_HEADER" | grep "^Set-Cookie:" | _tail_n 1 | _egrep_o "$1=[^;]*;" | tr -d ';' )"
+  _debug3 grep "$(grep "\W$1=" "$HTTP_HEADER" | grep "^Set-Cookie:" | _tail_n 1 | _egrep_o "$1=[^;]*;" | tr -d ';')"
   grep "\W$1=" "$HTTP_HEADER" | grep "^Set-Cookie:" | _tail_n 1 | _egrep_o "$1=[^;]*;" | tr -d ';'
 }
 
@@ -105,7 +105,7 @@ synology_dsm_deploy() {
   _info "Getting certificates in Synology DSM"
   response=$(_post "api=SYNO.Core.Certificate.CRT&method=list&version=1" "$_base_url/webapi/entry.cgi")
   _debug3 response "$response"
-  id=$(printf "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\"id\":\"\([^\"]*\).*/\1/p")
+  id=$(echo "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\"id\":\"\([^\"]*\).*/\1/p")
   _debug2 id "$id"
 
   if [ -z "$id" ]; then
@@ -117,8 +117,8 @@ synology_dsm_deploy() {
   _savedeployconf SYNO_Certificate "$SYNO_Certificate"
 
   default=false
-  if printf "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\([^{]*\).*/\1/p" | grep -q -- 'is_default":true'; then
-      default=true
+  if echo "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\([^{]*\).*/\1/p" | grep -q -- 'is_default":true'; then
+    default=true
   fi
   _debug2 default "$default"
 
@@ -132,14 +132,15 @@ synology_dsm_deploy() {
   content="$content${nl}--$delim${nl}Content-Disposition: form-data; name=\"desc\"${nl}${nl}${SYNO_Certificate}"
   content="$content${nl}--$delim${nl}Content-Disposition: form-data; name=\"as_default\"${nl}${nl}${default}"
   content="$content${nl}--$delim--${nl}"
-  content="$(printf "%b_" "$content")";content="${content%_}" # protect trailing \n
+  content="$(printf "%b_" "$content")"
+  content="${content%_}" # protect trailing \n
 
   _info "Upload certificate to the Synology DSM"
   response=$(_post "$content" "$_base_url/webapi/entry.cgi?api=SYNO.Core.Certificate&method=import&version=1&SynoToken=$token" "" "POST" "multipart/form-data; boundary=${delim}")
   _debug3 response "$response"
 
-  if ! printf "$response" | grep -q '"error":'; then
-    if printf "$response" | grep -q '"restart_httpd":true'; then
+  if ! echo "$response" | grep -q '"error":'; then
+    if echo "$response" | grep -q '"restart_httpd":true'; then
       _info "http services were restarted"
     else
       _info "http services were NOT restarted"
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -74,11 +74,11 @@ synology_dsm_deploy() {
   # use jq because I'm too lazy to figure out what is required to parse json
   # by hand.  Also it seems to be in place for Synology DSM (6.2.1 at least)
   for x in curl jq; do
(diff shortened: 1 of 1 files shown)


herbetom:
I think, such a deploy hook for DSM is great. But on a quick look into the code, I saw, that you use the function `_saveaccountconf_mutable` to store the username and password. But these stores those, central?!
If I have multiple DSM installations and want to use different usernames and passwords, then I can't do that. Or am I mistaken? Wouldn't it be better to store those informations on a per domain base with `_savedomainconf`?

tresni:
I wasn't sure which would be more common, multiple DSMs with differing logins or multiple DSMs all sharing the same login information.  My belief was it would be the later (ie they would use some unified login information.)

It wouldn't be too hard to change this to be login-per-DSM, I only have 1 so doesn't matter to me either way ;-)

herbetom:
Well, you're probably right. Most people would use the same username and password. But for those who don't, it's problematic. But if you store them on a domain basis, it's maybe a bit more complicated to process them. But it is possible. 

I would certainly use different users and passwords for each device (if I had multiple). For example, if one DS is a backup for another. There should be absolutely no way to log into the backup DS with information stored on the productive DS. But if you need to use the same user and password, especially for an admin user, it may be possible (although the passwords are hopefully only stored hashed on the DS) that some kind of virus logs the username and password while the certificates are renewed. Not very likely. But an unnecessary risk.

And since the pull request currently isn't merged, it is really easy to change this behavior now, before it becomes a problem for someone.

Neilpang:
please use `_savedomainconf` instead.

deploy-hook is domain-specific.

tresni:
@Neilpang using `_savedomainconf` instead of `_saveaccountconf_mutable` now.

King3R:
@Neilpang When will it be merged to dev or master branch?

deploy/synology_dsm.sh
```diff
@@ -0,0 +1,145 @@
...
+
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+
+  _debug _cdomain "$_cdomain"
+
+  # Get Username and Password, but don't save until we successfully authenticate
+  SYNO_Username="${SYNO_Username:-$(_readdomainconf SYNO_Username)}"
```
Neilpang: please just use `_getdeployconf` and `_savedeployconf`.
Don't use `xxxdomainconf()` or `xxxaccountconf()`.

see examples:

https://github.com/acmesh-official/acme.sh/blob/master/deploy/docker.sh

deploy/synology_dsm.sh
```diff
@@ -0,0 +1,145 @@
...
+  if [ -z "${SYNO_Certificate}" ]; then
+    _err "SYNO_Certificate needs to be defined (with the Certificate description name)"
+    return 1
+  fi
+  _debug SYNO_Certificate "$SYNO_Certificate"
+
+  # We can't use _get or _post because they lack support for cookies
+  # use jq because I'm too lazy to figure out what is required to parse json
+  # by hand.  Also it seems to be in place for Synology DSM (6.2.1 at least)
+  for x in curl jq; do
```
Neilpang: cookies are just http headers,
Here is an example:  https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_cyon.sh#L133

json is just a string. we have a lot of examples processing json strings.

Neilpang:
please add usage here:

please also read this guide, which will help you:

deploy/synology_dsm.sh
tresni: Happy to change this, I was using _readdomainconf and _savedomainconf based on [your previous request](https://github.com/acmesh-official/acme.sh/pull/2369#issuecomment-520194731).

tresni:
@Neilpang I have addressed your comments in 52a168b.

I wrote the [documentation](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#20-deploy-the-cert-into-synology-dsm) and opened #2727 for a Bug tracker

tresni:
https://github.com/acmesh-official/acme.sh/wiki/Synology-NAS-Guide could likely be updated too to utilize the deployhook, but not sure if that's required (or wanted.)

deploy/synology_dsm.sh
```diff
@@ -0,0 +1,152 @@
...
+
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+
+  _debug _cdomain "$_cdomain"
+
+  # Get Username and Password, but don't save until we successfully authenticate
+  SYNO_Username="${SYNO_Username:-$(_getdeployconf SYNO_Username)}"
```
Neilpang: please see how `_getdeployconf()` is used in the example script:
https://github.com/acmesh-official/acme.sh/blob/master/deploy/docker.sh

Use the same way. 
Thanks.

deploy/synology_dsm.sh
```diff
@@ -0,0 +1,152 @@
...
+  if [ -z "$id" ] && [ -z "$SYNO_Create" ]; then
+    _err "Unable to find certificate: $SYNO_Certificate and \$SYNO_Create is not set"
+    return 1
+  fi
+
+  # we've verified this certificate description is a thing, so save it
+  _savedeployconf SYNO_Certificate "$SYNO_Certificate"
+
+  default=false
+  if echo "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\([^{]*\).*/\1/p" | grep -q -- 'is_default":true'; then
```
Neilpang: don't use `-q` option for `grep`

deploy/synology_dsm.sh
```diff
@@ -0,0 +1,154 @@
...
+  _savedeployconf SYNO_Password "$SYNO_Password"
+  _debug token "$token"
+
+  _info "Getting certificates in Synology DSM"
+  response=$(_post "api=SYNO.Core.Certificate.CRT&method=list&version=1" "$_base_url/webapi/entry.cgi")
+  _debug3 response "$response"
+  id=$(echo "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\"id\":\"\([^\"]*\).*/\1/p")
+  _debug2 id "$id"
+
+  # shellcheck disable=SC2154
```
Neilpang: what is this `SC2154` for ?

deploy/synology_dsm.sh
```diff
@@ -0,0 +1,154 @@
...
+
+  default=false
+  if echo "$response" | sed -n "s/.*\"desc\":\"$SYNO_Certificate\",\([^{]*\).*/\1/p" | grep -q -- 'is_default":true'; then
+    default=true
+  fi
+  _debug2 default "$default"
+
+  _info "Generate form POST request"
+  nl="\015\012"
+  delim="--------------------------$(date +%Y%m%d%H%M%S)"
```
Neilpang: if you just need a random string as a delimiter,  try to use `_time()` or `_utc_date()` function with some `tr` or `sed` replacement.

deploy/synology_dsm.sh
Neilpang: https://github.com/koalaman/shellcheck/wiki/SC2154
try to fix it as possible as you can,  instead of just disable it.

tresni: "Fixing" this would be doing an empty assignment to `$SYNO_Create`.  The problem is that this variable is dynamically allocated by `_getdeployconf` above, but shellcheck doesn't know that.  So there's no actual issue here, hence disabling that specific linter for that specific line.

deploy/synology_dsm.sh
tresni: `-q` is a POSIX option that should be supported on all implementations.  It's used because the JSON returned can be rather large.  While POSIX `-q` only guarentees no output, most implementations also stop parsing at the first match.  Suppressing the output is easy enough with a pipe redirect, but the potential benefit of not having to parse a large JSON object can't be otherwise replicated to my knowledge.  If you really want it removed, let me know, but I think there's good reason for keeping it.

deploy/synology_dsm.sh
tresni: This was already fixed.

deploy/synology_dsm.sh
tresni: Changed in 7963709

deploy/synology_dsm.sh
tresni: Used variable expansion in d07172a to address this.

Neilpang: see example:
https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_cf.sh#L29

deploy/synology_dsm.sh
Neilpang: use  redirect instead `>>/dev/null`.
The less option is used, the more portable it will be.
Never use unnecessary options.  you will not like troubles.

tresni: Addressed in 1b475cf

deploy/synology_dsm.sh
tresni: That example doesn't work for `_getdeployconf`, but like I said, this has been addressed in d07172a

King3R:
@Neilpang @tresni Many thanks for your work.

meiser79:
Great work, thanks! Any chance to support 2-factor authentification?

lippertmarkus:
@meiser79 I added OTP/2-factor auth support in #2782 :)

DJGauthier:
Hi Community,

I am trying to use the Synology DSM hook (synology_dsm.sh) on my DS1010+ DSM 5.2.
I know it's a 8 years old version but my hardware does not support new versions of DSM.
I use the DDNS name, (no DNS)

The script is working well until logon to the DSM.
**here my commands:**
export SYNO_Username='myadmin'
export SYNO_Password='mypassword'
export SYNO_Hostname='mysno.synology.me'
export SYNO_Scheme="https"
export SYNO_Port="5001"
export SYNO_Certificate="Lets Encrypt certificate"
acme.sh -d "mysyno.synology.me" --deploy --deploy-hook synology_dsm --webroot /var/lib/letsencrypt --staging --debug 2 --log acme_hook.log

**bottom of the logs messages:**
[Thr Jun 22 10:44:37 CEST 2023] Unable to authenticate to mysyno.synology.me:5001 using https.
[Thr Jun 22 10:44:37 CEST 2023] Check your username and password.
[Thr Jun 22 10:44:37 CEST 2023] Error deploy for domain:mysyno.synology.me
[Thr Jun 22 10:44:37 CEST 2023] Deploy error.

I tried the url in my browser but only the logon mask appear and will not filled up.
for local with IP:
https://myip:5001/webman/index.cgi?account=myadmin&passwd=myadminpassword&enable_syno_token=yes
I fill manually the mask with the same user user and password, and I can login in.


Can someone help me with this error

Best regards

PR #2383, opened 2019-07-11
Fix multi ip issue

FlyveHest:
Fixes issues when you have multiple IPs on the box, and have specified a specific local IP for the ALPN validation server.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3182,7 +3182,10 @@ _on_before_issue() {
       _netprc="$(_ss "$_checkport" | grep "$_checkport")"
       netprc="$(echo "$_netprc" | grep "$_checkaddr")"
       if [ -z "$netprc" ]; then
-        netprc="$(echo "$_netprc" | grep "$LOCAL_ANY_ADDRESS")"
+        # Only check for local bindings when you aren't looking for a specific IP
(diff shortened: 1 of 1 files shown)


acme.sh
```diff
@@ -3182,7 +3182,10 @@ _on_before_issue() {
       _netprc="$(_ss "$_checkport" | grep "$_checkport")"
       netprc="$(echo "$_netprc" | grep "$_checkaddr")"
```
Neilpang: imagine you have 2 ipaddresses:   1.1.1.1  and 2.2.2.2
you have specified  _checkaddr=1.1.1.1

We first check if someone else is listening at  `11.1.1:443`
```
netprc="$(echo "$_netprc" | grep "$_checkaddr")"
```

If not found, `netprc` is empty,  we then check if someone else is listening at `0.0.0.0:443`
```
if [ -z "$netprc" ]; then
  netprc="$(echo "$_netprc" | grep "$LOCAL_ANY_ADDRESS")"
fi
```

I'm not sure why you need this fix here.

If someone is listening at `0.0.0.0:443`,   you are not allowed to listen at neithe `1.1.1.1:443`  nor `2.2.2.2:443`.

what do you think ?

PR #2384, opened 2019-07-11
Fix gratisdns api

FlyveHest:
Fixed an issue when deleting the TXT record on GratisDNS, SED did not like matching on the multiline HTML response, so strip all newlines before letting _egrep_o work on it

### DIFF
--- a/dnsapi/dns_gdnsdk.sh
+++ b/dnsapi/dns_gdnsdk.sh
@@ -159,7 +159,7 @@ _successful_update() {
 _findentry() {
   #returns id of dns entry, if it exists
   _myget "action=dns_primary_changeDNSsetup&user_domain=$_domain"
-  _id=$(echo "$_result" | _egrep_o "<td>$1</td>\s*<td>$2</td>[^?]*[^&]*&id=[^&]*" | sed 's/^.*=//')
+  _id=$(echo "$_result" | tr '\n' ' ' | _egrep_o "<td>$1</td>\s*<td>$2</td>[^?]*[^&]*&id=[^&]*" | sed 's/^.*=//')
   if [ -n "$_id" ]; then
     _debug "Entry found with _id=$_id"
     return 0


Neilpang:
No,  we can not trim `\n` here.

because `_egrep_o` function depends on `sed`, when running on `solaris`.  but the `sed` tool on solaris depends on `\n` to work.

please consier other solutions.

FlyveHest:
Do you have any mechanics for detecting OS built in? In its current state this does not work on Ubuntu, the RM part of this DNS API can't find the ID its supposed to delete, and thus, does not delete anything.

I'll try and take a look and see if the RegEx can be modified

Neilpang:
It's not recommended to detect OS in the script.
We should avoid platform specific code.

There are a lot of tricks in this project to help you.

Neilpang:
you can show me the html content, and the target string you want to grep.
I can also help you.

FlyveHest:
The problem is in the _findentry function in the dns_gdnsdk.sh API script.

The function is called with the domain name ($1) and verification TXT string ($2)

This is what comes back from the curl call, i've removed most of the excess data, and changed the domain and ID to be generic (domain.tld, ACMETEXTENTRYSTRING and the ID the function should return as 12345)

> <tr>
>                 <td>_acme-challenge.domain.tld</td>
>                 <td>ACMETEXTENTRYSTRING</td>
>                 <td>1</td>
>                 <td>
>                   <div class="btn-group btn-group-sm float-right" role="group" aria-label="Andre/slet record">
>                                           <a href="?action=dns_primary_record_edit_txt&id=12345&user_domain=domain.tld" class="btn btn-secondary">Ændre</a>
> 
>                     <a href="?action=dns_primary_delete_txt&id=12345&user_domain=domain.tld" class="btn btn-danger">Slet</a>

If you strip the newlines before piping it to _egrep_o, it works, if the newlines are in, it doesn't work on Ubuntu 18, and the function does not return an ID, leaving the TXT record on the domain.

Neilpang:
Here is an example for how we process  html tables:
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_he.sh#L129

Neilpang:
any update  ?

PR #2407, opened 2019-07-26
Notify xmpp

fgma:
Send notifications via xmpp (using sendxmpp)

sendxmpp needs to be installed and configured for the current user running acme.sh according to https://manpages.ubuntu.com/manpages/disco/en/man1/sendxmpp.1p.html#configuration%20file

### DIFF
--- a/notify/xmpp.sh
+++ b/notify/xmpp.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env sh
+
+#Support xmpp via sendxmpp
+
+#XMPP_BIN="/usr/bin/sendxmpp"
+#XMPP_BIN_ARGS="-n -t --tls-ca-path=/etc/ssl/certs"
+#XMPP_TO="zzzz@example.com"
+#XMPP_NOVALIDATE=""
+
+xmpp_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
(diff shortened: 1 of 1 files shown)


fgma:
@Neilpang Do you have any feedback concerning this pull request?

Neilpang:
can you please update the usage here: 

And create a new issue for bug reporting ?

fgma:
@Neilpang I've documented it on https://github.com/Neilpang/acme.sh/wiki/notify and opened an issue for bug reporting: https://github.com/Neilpang/acme.sh/issues/2425

Let me know if there is anything left for me to do.

Neilpang:
Is it possible to pass these values on commandline?
```
username: example
jserver: example.com
password: xxxxxxxx
```

fgma:
@Neilpang Yes, this is possible but from a security point of view I would not consider doing so as other users on the system are able to access passwords that way. See also https://unix.stackexchange.com/questions/8223/can-other-users-view-the-arguments-passed-to-a-command

Neilpang:
@fgma  I see but please support command line, just for optional choices.

It's possible that the user wants to use a different username/password/server from the system global default settings.

Of cause,  they are optional, just for advanced users to use.

fgma:
@Neilpang For this use-case you can use a different configuration file for sendxmpp using

`
export XMPP_BIN_ARGS="-t /path/to/sendxmpprc_for_acme_sh'"
`

which does what you want a keeps the credentials secret.

PR #2416, opened 2019-07-31
IFTTT Webhooks Notification

lcdtyph:
Thanks for this amazing project!

I've simply implemented the ifttt webhooks notify hook.

This hook needs `IFTTT_API_KEY` and `IFTTT_EVENT_NAME`. The api key can be obtained at [https://ifttt.com/maker_webhooks](https://ifttt.com/maker_webhooks), and the event name can be set while creating an ifttt applet with webhooks.

This notification method could be a bridge to make acme.sh work with tons of IFTTT services.

### DIFF
--- a/notify/ifttt.sh
+++ b/notify/ifttt.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+
+#Support ifttt.com webhooks api
+
+#IFTTT_API_KEY="xxxx"
+#IFTTT_EVENT_NAME="yyyy"
+
+#IFTTT_SUBJECT_KEY="value1|value2|value3"      #optional, use "value1" as default
+#IFTTT_CONTENT_KEY="value1|value2|value3"      #optional, use "value2" as default
+
+_IFTTT_AVAIL_MSG_KEYS="value1,value2,value3"
+
+# subject  content statusCode
+ifttt_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
+  _debug "_subject" "$_subject"
+  _debug "_content" "$_content"
+  _debug "_statusCode" "$_statusCode"
+
+  IFTTT_API_KEY="${IFTTT_API_KEY:-$(_readaccountconf_mutable IFTTT_API_KEY)}"
+  if [ -z "$IFTTT_API_KEY" ]; then
+    IFTTT_API_KEY=""
+    _err "You didn't specify a ifttt webhooks api key IFTTT_API_KEY yet."
+    _err "You can get yours from https://ifttt.com"
+    return 1
+  fi
+  _saveaccountconf_mutable IFTTT_API_KEY "$IFTTT_API_KEY"
+
+  IFTTT_EVENT_NAME="${IFTTT_EVENT_NAME:-$(_readaccountconf_mutable IFTTT_EVENT_NAME)}"
+  if [ -z "$IFTTT_EVENT_NAME" ]; then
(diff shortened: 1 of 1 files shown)


Neilpang:
please update the usage here:

notify/ifttt.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
```
Neilpang: use `sh`  instead of `bash`

lcdtyph: Done, thanks!

lcdtyph:
Should I open an issue for bug reporting?

Neilpang:
yes

notify/ifttt.sh
```diff
@@ -0,0 +1,84 @@
...
+    return 1
+  fi
+
+  if [ "$IFTTT_SUBJECT_KEY" = "$IFTTT_CONTENT_KEY" ]; then
+    IFTTT_SUBJECT_KEY=""
+    IFTTT_CONTENT_KEY=""
+    _err "The IFTTT_SUBJECT_KEY must not be same as IFTTT_CONTENT_KEY."
+    return 1
+  fi
+  _saveaccountconf_mutable IFTTT_SUBJECT_KEY "$IFTTT_SUBJECT_KEY"
```
Neilpang: don't save the values here.

we should only save the values when the users set a value.

we should not save it if it was empty but used the default value.

lcdtyph:
I've created the issue, https://github.com/Neilpang/acme.sh/issues/2421, and appended to the wiki already.

notify/ifttt.sh
```diff
@@ -0,0 +1,92 @@
...
+
+  _IFTTT_CONTENT_KEY="${IFTTT_CONTENT_KEY:-$(_readaccountconf_mutable IFTTT_CONTENT_KEY)}"
+  IFTTT_CONTENT_KEY="$_IFTTT_CONTENT_KEY"
+  if [ -z "$IFTTT_CONTENT_KEY" ]; then
+    IFTTT_CONTENT_KEY="value2"
+    _info "The IFTTT_CONTENT_KEY is not set, so use the default value2 as key."
+  elif ! _hasfield "$_IFTTT_AVAIL_MSG_KEYS" "$IFTTT_CONTENT_KEY"; then
+    _err "The IFTTT_CONTENT_KEY \"$IFTTT_CONTENT_KEY\" is not available, should be one of $_IFTTT_AVAIL_MSG_KEYS"
+    IFTTT_CONTENT_KEY=""
+    return 1
```
Neilpang: add  here
```
else
    _saveaccountconf_mutable xxxxxxxx
```

notify/ifttt.sh
```diff
@@ -0,0 +1,92 @@
...
+  fi
+
+  if [ "$IFTTT_SUBJECT_KEY" = "$IFTTT_CONTENT_KEY" ]; then
+    IFTTT_SUBJECT_KEY=""
+    IFTTT_CONTENT_KEY=""
+    _err "The IFTTT_SUBJECT_KEY must not be same as IFTTT_CONTENT_KEY."
+    return 1
+  fi
+
+  if [ -n "$_IFTTT_SUBJECT_KEY" ]; then
```
Neilpang: no

notify/ifttt.sh
lcdtyph: I was worrying about saving incorrect values, so I thought it's better to save them after all checks.

Neilpang: if the value is set by the user,  just save it.

if it's from a default value,   don't save it.

PR #2433, opened 2019-08-09
[DNSAPI/BUG] Fixes naive top domain detection for KnotDNS

felipewd:
The current implementation of top/root domain is wrong for some TLDs.

In order to properly detect the top domain, lets query the NS through
the SOA record and get the top domain being validated.

This fixes incorrect top domain detection under TLDs such as 'com.br',
'co.uk', etc.

This solution could be used for other DNS APIs as well, since it's pretty straightforward.

### DIFF
--- a/dnsapi/dns_knot.sh
+++ b/dnsapi/dns_knot.sh
@@ -72,19 +72,17 @@ EOF
 # _domain=domain.com
 _get_root() {
   domain=$1
-  i="$(echo "$fulldomain" | tr '.' ' ' | wc -w)"
-  i=$(_math "$i" - 1)
 
-  while true; do
-    h=$(printf "%s" "$domain" | cut -d . -f "$i"-100)
-    if [ -z "$h" ]; then
-      return 1
-    fi
-    _domain="$h"
-    return 0
-  done
-  _debug "$domain not found"
-  return 1
+  d=$(dig soa "$domain" | grep -v ^\; | grep SOA | awk '{print $1}')
+  if [ -z "${d}" ]; then
+    _debug "$domain not found"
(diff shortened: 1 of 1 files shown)


Neilpang:
Sorry for the issue.

I'm not familiar with nsupdate at all.  but is there any local config file or from the nsupdate commandline parameters to get a full list of all the root domains ?
So that you can check which part is the correct root domain from the full domain list.
This is the best recommended way.

However, at the worst,  if you can not get the full domain list and have to check the NS record of the domain,   do not use `dig`, it's not portable.  we have a built in function `_ns_lookup_cf`, which you can use to lookup any kinds of records.  For example:

```
_ns_lookup_cf   youdomain.com.br   NS
```

felipewd:
hey @Neilpang,

Thanks for the feedback. First of all, this is a great project, but I just digged into the code because of this issue. So, sorry for not knowing about this function.

As for the 1st alternatives (know all valid TLDs and checking against those), I quite frankly do not a correct way of checking this, that's why I went for the `dig` approach. 

I've tested the function you talked about, and for `test.bbc.co.uk` it returned:

```
"Answer":[
"name": "test.bbc.co.uk.", "type": 6, "TTL": 3600, "data": "ns.bbc.co.uk. hostmaster.bbc.co.uk. 2019072901 1800 600 864000 14400"
]
```

Didn't provide the root domain in that case, then.

I'm thinking now this module could take the same (simpler)  approach as the `nsupdate` module and accept an optional `KNOT_ZONE` external variable and fallback to this provided zone instead of relying on the automatic one. Thoughts?

Thanks!

Issue #2414, opened 2019-07-29
_egrep_o broken on docker images


codestation:
This bug was reported on #1364 and a pull request was made that solved the problem on #1366 but it was never merged and instead a workaround was pushed.

Right now on on the latest version (2.8.2) dns_linode.sh and dns_linode_v4.sh (and maybe others) are broken because this bug (the domain cannot be found) Can #1366 be properly merged? I made a test case to further demonstrate that the current implementation is broken because the pipe cannot be read twice.

Steps to reproduce
------------------

`test.sh`

```sh
#!/bin/sh

_egrep_o1() {
  egrep -o "$1" 2> /dev/null
}

_egrep_o2() {
  sed -n 's/.*\('"$1"'\).*/\1/p'
}

_egrep_o() {
  if ! egrep -o "$1" 2>/dev/null; then
    sed -n 's/.*\('"$1"'\).*/\1/p'
  fi
}

_egrep_o_fix() {
  stdin=$(cat)
  if ! echo "$stdin" | egrep -o "$1" 2>/dev/null; then
    echo "$stdin" | sed -n 's/.*\('"$1"'\).*/\1/p'
  fi
}

echo "egrep -o test:"
echo '{"pages": 1, "results": 2, "data": [
{"id": 1, "domain": "example.org"},
{"id": 2, "domain": "example.com"},
{"id": 3, "domain": "example.net"}], "page": 1}' | _egrep_o1  '{.*"domain":\s*"example.com".*}'

echo "sed -n test:"
echo '{"pages": 1, "results": 2, "data": [
{"id": 1, "domain": "example.org"},
{"id": 2, "domain": "example.com"},
{"id": 3, "domain": "example.net"}], "page": 1}' | _egrep_o2  '{.*"domain":\s*"example.com".*}'

echo "_egrep_o() test:"
echo '{"pages": 1, "results": 2, "data": [
{"id": 1, "domain": "example.org"},
{"id": 2, "domain": "example.com"},
{"id": 3, "domain": "example.net"}], "page": 1}' | _egrep_o  '{.*"domain":\s*"example.com".*}'

echo "_egrep_o_fix() test:"
echo '{"pages": 1, "results": 2, "data": [
{"id": 1, "domain": "example.org"},
{"id": 2, "domain": "example.com"},
{"id": 3, "domain": "example.net"}], "page": 1}' | _egrep_o_fix  '{.*"domain":\s*"example.com".*}'
```

```
[...]

Issue #2385, opened 2019-07-11
Bug in _egrep_o() when extended regex passed in and sed emulation used


dkerr64:
_egrep_o() function accepts extended regex and on systems that do not have egrep uses sed to emulate egrep.  This is failing on the specific regex I was using in dns_freedns.sh.  See https://github.com/Neilpang/acme.sh/issues/2305 for the problem I was working on.

The problem is that egrep requires extended regex, and if you pass in non-extended it may fail if there are special characters.  But on systems that do not have egrep, the _egrep_o function uses sed to emulate egrep.  This fails with the extended regex, but works with the regular regex.

Below script illustrates the problem... I split your _egrep_o function into two to illustrate the problem on systems that have egrep (ie, removed the test for failure with egrep)

```
#!/usr/bin/env sh

_egrep_o1() {
  egrep -o "$1" 2>/dev/null
}

_egrep_o2() {
  sed -n 's/.*\('"$1"'\).*/\1/p'
}

txt="<tr><td>example.com</td><tdalign=right><ahref=/subdomain/edit.php?edit_domain_id=1234567>[add]</a></td></tr></table></td></tr>"

echo "Test 1 (grep -o)"
echo "$txt" | grep -o "edit\.php?edit_domain_id=[0-9a-zA-Z]*"
echo "Test 2 (egrep -o with non-extended regex, should fail)"
echo "$txt" | _egrep_o1 "edit\.php?edit_domain_id=[0-9a-zA-Z]*"
echo "Test 3 (sed emulation of egrep -o with non-extended regex, should fail like test 2"
echo "$txt" | _egrep_o2 "edit\.php?edit_domain_id=[0-9a-zA-Z]*"
echo "Test 4 (egrep -o with extended regex, should work"
echo "$txt" | _egrep_o1 "edit\.php\?edit_domain_id=[0-9a-zA-Z]+"
echo "Test 5 (sed emulation of egrep -o with extended regex, should work like test 4"
echo "$txt" | _egrep_o2 "edit\.php\?edit_domain_id=[0-9a-zA-Z]+"
echo "End"
exit
```

codestation:
This bug also breaks dns_linode.sh and dns_linode_v4.sh when used on the docker image. Seems like `_egrep_o` is broken on alpine.

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

PR #2434, opened 2019-08-10
Work around bug in _egrep_o() function

dkerr64:
_egrep_o() function accepts extended regex and on systems that do not have egrep uses sed to emulate egrep. This is failing on the specific regex I was using before my last commit... The problem is that I fixed it by passing in non-extended regex which then fails on systems that do have egrep. So I am no longer using _egrep_o.
See bugs #2385 and #2414 
Note... even if those bugs are fixed, I would like to merge this PR and stop using _egrep_o()
Thanks

### DIFF
--- a/dnsapi/dns_freedns.sh
+++ b/dnsapi/dns_freedns.sh
@@ -305,7 +305,7 @@ _freedns_domain_id() {
 
     domain_id="$(echo "$htmlpage" | tr -d "[:space:]" | sed 's/<tr>/@<tr>/g' | tr '@' '\n' \
       | grep "<td>$search_domain</td>\|<td>$search_domain(.*)</td>" \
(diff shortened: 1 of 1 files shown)


Neilpang:
the `grep -o`  option is not always available on all the POSIX platforms.

Try to avoid using `grep -o`. 
here is an example about how we process html:
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_he.sh

dkerr64:
> the `grep -o` option is not always available on all the POSIX platforms.
> 
> Try to avoid using `grep -o`.
> here is an example about how we process html:
> https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_he.sh

Okay, then I replace grep -o with sed.

dkerr64:
@Neilpang can you please merge this.
Thank you.
David

Issue #2480, opened 2019-09-12
Report bugs to OPNsense dns API


blablup:
This is the place to report bugs in the OPNsense  DNS API.

If you experience a bug, please report it in this issue.

Thanks!

wwebers:
Hi,

I experience an "invalid domain" error with the latest version of the "os-bind" plugin. However, I found they added the following part to the domain query response:

```
"masterip": "",
  "allownotifyslave": {
    "": {
      "value": "",
      "selected": 1
    }
  },
"domainname": "....",
```

as such, I added the following to the pattern matching part inside the ```dns_opnsense.sh``` script:

```
\"masterip\":\"[^\"]*\",\"allownotifyslave\":{\"\":{[^}]*}},\"domainname\":\"${h}\""
```

That way it works again.

fraenki:
There seems to be a bug in the most recent version:
https://github.com/acmesh-official/acme.sh/issues/3735

benyamin-codez:
> I found a few more for v25.7 per PR [#6492](https://github.com/acmesh-official/acme.sh/pull/6492)...

PR #6503 replaced #6492 and has been merged into dev...

PR #2481, opened 2019-09-12
Add OPNsense Bind API Support

blablup:
Hi,

I've added dns api support for the OPNsense Bind API. (#2480 )

### DIFF
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -114,11 +114,11 @@ rm_record() {
         _opns_rest "POST" "/service/reconfigure" "{}"
         _debug "Record deleted"
       else
-        _err "Error deleting record $host from domain $fulldomain"
+        _err "Error deleting record $_host from domain $fulldomain"
         return 1
       fi
     else
-      _err "Error deleting record $host from domain $fulldomain"
+      _err "Error deleting record $_host from domain $fulldomain"
       return 1
     fi
   else
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -234,7 +234,7 @@ _opns_check_auth() {
   if [ -z "$OPNs_Api_Insecure" ]; then
     OPNs_Api_Insecure="0"
   else
-  #save the api addr and key to the account conf file.
+    #save the api addr and key to the account conf file.
     _saveaccountconf_mutable OPNs_Api_Insecure "$OPNs_Api_Insecure"
   fi
   export HTTPS_INSECURE="${OPNs_Api_Insecure}"
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -236,7 +236,7 @@ _opns_check_auth() {
   else
     _saveaccountconf_mutable OPNs_Port "$OPNs_Port"
   fi
-  
+
   if ! printf '%s' "$OPNs_Api_Insecure" | grep -q '^[01]$'; then
     _err 'OPNs_Api_Insecure specified but not 0/1 value'
     return 1
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -4,15 +4,19 @@
 #https://docs.opnsense.org/development/api.html
 #
 #OPNs_Host="opnsense.example.com"
-#OPNs_Port="443"
+#OPNs_Port="443" (optional, defaults to 443 if unset)
 #OPNs_Key="qocfU9RSbt8vTIBcnW8bPqCrpfAHMDvj5OzadE7Str+rbjyCyk7u6yMrSCHtBXabgDDXx/dY0POUp7ZA"
 #OPNs_Token="pZEQ+3ce8dDlfBBdg3N8EpqpF5I1MhFqdxX06le6Gl8YzyQvYCfCzNaFX9O9+IOSyAs7X71fwdRiZ+Lv"
-#OPNs_Api_Insecure=0     # Set 1 for insecure and 0 for secure -> difference is whether ssl cert is checked for validity (0) or whether it is just accepted (1)
+#OPNs_Api_Insecure=0 (optional, defaults to 0 if unset) # Set 1 for insecure and 0 for secure -> difference is whether ssl cert is checked for validity (0) or whether it is just accepted (1)
 
 ########  Public functions #####################
 #Usage: add _acme-challenge.www.domain.com "123456789ABCDEF0000000000000000000000000000000000000"
 #fulldomain
 #txtvalue
+OPNs_DefaultPort=443
+OPNs_DefaultApi_Insecure=0
+
+
 dns_opnsense_add() {
   fulldomain=$1
   txtvalue=$2
@@ -168,7 +172,7 @@ _opns_rest() {
   key=$(echo "$OPNs_Key" | tr -d "\n\r" | _url_encode)
   token=$(echo "$OPNs_Token" | tr -d "\n\r" | _url_encode)
 
-  opnsense_url="https://${key}:${token}@${OPNs_Host}:${OPNs_Port}/api/bind${ep}"
+  opnsense_url="https://${key}:${token}@${OPNs_Host}:${OPNs_Port:-$OPNs_DefaultPort}/api/bind${ep}"
   export _H1="Content-Type: application/json"
   if [ ! "$method" = "GET" ]; then
     _debug data "$data"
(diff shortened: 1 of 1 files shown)


dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,261 @@
...
+_opns_rest() {
+  method=$1
+  ep=$2
+  data=$3
+  #Percent encode user and token
+  key=$(echo "$OPNs_Key" | tr -d "\n\r" | _url_encode)
+  token=$(echo "$OPNs_Token" | tr -d "\n\r" | _url_encode)
+
+  opnsense_url="https://${key}:${token}@${OPNs_Host}:${OPNs_Port}/api/bind${ep}"
+  export _H1="Content-Type: application/json"
```
Neilpang: I think this line is not needed.

blablup: I tried it already, but it gave errorors without this line.

blablup:
Do you plan to merge this at some point?

Neilpang:
don't save the value to account.conf if it has a default value.
for example,   if OPNs_Api_Insecure is not set or `0`, don't save it.
OPNs_Port also.

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,265 @@
...
+  if [ -z "$OPNs_Host" ]; then
+    OPNs_Host="localhost"
+    _err "You don't specify OPNsense address."
+    return 1
+  else
+    _saveaccountconf_mutable OPNs_Host "$OPNs_Host"
+  fi
+
+  if [ -z "$OPNs_Port" ]; then
+    OPNs_Port="443"
```
Neilpang: when `_opns_check_auth()`  is called from the `dns_opnsense_add()` function,  OPNs_Port is empty, and not saved.

but when it's called from `dns_opnsense_rm()` function, the `OPNs_Port` is 443, not empty.  it's then saved.

stilez: The `if...else...fi` statement at lines 228 - 231 should probably read as follows:

```
  if [ -z "$OPNs_Port" ]; then
    _info 'OPNSense port not specified. Defaulting to using port 443'
    OPNs_Port="443"
  fi
   _saveaccountconf_mutable OPNs_Port "$OPNs_Port"
```

But I don't have access to make or propose this edit.

blablup: look at the comment 

> don't save the value to account.conf if it has a default value.
> for example, if OPNs_Api_Insecure is not set or `0`, don't save it.
> OPNs_Port also.

That is why i changed it to if...else...fi.
I have to change a little bit. Maybe I'll use different variables for default values, I'll do it soon.  Had a lot to do past weeks.

stilez: In that case, the easiest fix is to

1) Add this line: `OPNs_DefaultPort=443` somewhere sensible near the top.

2) Change the comment at line 7 to.read `...(optional, defaults to 443 if unset)`, for clarity.

3) Replace `$OPNs_Port` by `${OPNs_Port:-$OPNs_DefaultPort}` in line 171

4) Replace the code at lines 228-231 by:

```
  if [ -z "$OPNs_Port" ]; then
    _info 'OPNSense port not specified. Defaulting to using port 443'
 else
    _saveaccountconf_mutable OPNs_Port "$OPNs_Port"
  fi
 ```

The effect is - if a port is specified it's saved. If not, OPNs_Port remains unset/zero,so it isn't saved as Neilpang points out. And if it's null, then the value of `$OPNs_DefaultPort` is substituted at call-time.

**Comment by stilez** [comment]
Lines 221, 243, and 252 all detect a variable is unset/empty/zero, and if so, they assign a value (local host or empty string) and then return an error. 

If it's going to crash out anyway, there's little point setting the variable values, and especially, doesnt seem a good point in setting them to "" when they're already -z. Especially if local host is empty, it's set to.local host and then immediately bails out with an error, that seems wrong.  Surely just error and return if values aren't valid - I.e., just delete those 3 assignment lines? 

(If they do need to be reset for any reason, reset them all, not just the one that's wrong, maybe)

File: dnsapi/dns_opnsense.sh (hunk 2 above)
Comment (blablup) [reply dnsapi/dns_opnsense.sh]: Thanks for the tips. 

That's much cleaner then my thought about it.

Comment (stilez) [reply dnsapi/dns_opnsense.sh]: Actually from Neilpang's comment on my own recent PR I should also suggest you replace `grep -q` with `grep ... >/dev/null`. Sorry!!

File: dnsapi/dns_opnsense.sh [hunk 3]
```diff
@@ -0,0 +1,268 @@
+#!/usr/bin/env sh
+
+#OPNsense Bind API
+#https://docs.opnsense.org/development/api.html
+#
+#OPNs_Host="opnsense.example.com"
+#OPNs_Port="443" (optional, defaults to 443 if unset)
+#OPNs_Key="qocfU9RSbt8vTIBcnW8bPqCrpfAHMDvj5OzadE7Str+rbjyCyk7u6yMrSCHtBXabgDDXx/dY0POUp7ZA"
+#OPNs_Token="pZEQ+3ce8dDlfBBdg3N8EpqpF5I1MhFqdxX06le6Gl8YzyQvYCfCzNaFX9O9+IOSyAs7X71fwdRiZ+Lv"
+#OPNs_Api_Insecure=0 (optional, defaults to 0 if unset) # Set 1 for insecure and 0 for secure -> difference is whether ssl cert is checked for validity (0) or whether it is just accepted (1)
```
stilez: Split this line so explanatory comment is easily readable?

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+
+  if ! rm_record "$fulldomain" "$txtvalue"; then
+    return 1
+  fi
+
+  return 0
+}
+
+set_record() {
+  _info "Adding record"
```
stilez: Maybe include `$fulldomain` and `$new_challenge` in the _info message? Could help if you get issue reports.

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+    else
+      return 1
+    fi
+  fi
+
+  if echo "$_return_str" | _egrep_o "\"result\":\"saved\"" >/dev/null; then
+    _opns_rest "POST" "/service/reconfigure" "{}"
+    _debug "Record created"
+  else
+    _err "Error createing record $_record_string"
```
stilez: Typo: "creating"

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+    return 1
+  fi
+
+  return 0
+}
+
+rm_record() {
+  _info "Remove record"
```
stilez: As above, maybe move _info below the next 2 lines, and include $fulldomain and $new_challenge in the _info message?

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+  _debug _domainid "$_domainid"
+  _uuid=""
+  if _existingchallenge "$_domain" "$_host" "$new_challenge"; then
+    # Delete
+    if _opns_rest "POST" "/record/delRecord/${_uuid}" "\{\}"; then
+      if echo "$_return_str" | _egrep_o "\"result\":\"deleted\"" >/dev/null; then
+        _opns_rest "POST" "/service/reconfigure" "{}"
+        _debug "Record deleted"
+      else
+        _err "Error delteting record $fulldomain"
```
stilez: Typo: "deleting"

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+        return 1
+      fi
+    else
+      _err "Error delteting record $fulldomain"
```
stilez: Typo: "deleting"

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+  method=$1
+  ep=$2
+  data=$3
+  #Percent encode user and token
+  key=$(echo "$OPNs_Key" | tr -d "\n\r" | _url_encode)
+  token=$(echo "$OPNs_Token" | tr -d "\n\r" | _url_encode)
+
+  opnsense_url="https://${key}:${token}@${OPNs_Host}:${OPNs_Port:-$OPNs_DefaultPort}/api/bind${ep}"
+  export _H1="Content-Type: application/json"
+  if [ ! "$method" = "GET" ]; then
```
stilez: Maybe add a _debug line to report that we are about to try calling the API. Because otherwise the log shows a "response" but no indication of the call it relates to, which is strange.

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+  OPNs_Api_Insecure="${OPNs_Api_Insecure:-$(_readaccountconf_mutable OPNs_Api_Insecure)}"
+
+  if [ -z "$OPNs_Host" ]; then
+    _err "You don't specify OPNsense address."
+    return 1
+  else
+    _saveaccountconf_mutable OPNs_Host "$OPNs_Host"
+  fi
+
+  if ! printf '%s' "$OPNs_Port" | grep -q '^[0-9]*$'; then
```
stilez: Replace all `grep -q` in the code, by `grep ... >/dev/hnull`

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+    _err 'OPNs_Api_Insecure specified but not 0/1 value'
+    return 1
+  elif [ -n "$OPNs_Api_Insecure" ]; then
+    _saveaccountconf_mutable OPNs_Api_Insecure "$OPNs_Api_Insecure"
+  fi
+  export HTTPS_INSECURE="${OPNs_Api_Insecure:-$OPNs_DefaultApi_Insecure}"
+
+  if [ -z "$OPNs_Key" ]; then
+    _err "You don't specify OPNsense api key id."
+    _err "Please set you OPNs_Key and try again."
```
stilez: Typo: you -> your

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+    return 1
+  else
+    _saveaccountconf_mutable OPNs_Key "$OPNs_Key"
+  fi
+
+  if [ -z "$OPNs_Token" ]; then
+    _err "You don't specify OPNsense token."
+    _err "Please create you OPNs_Token and try again."
```
stilez: Same as lines 248-249

dnsapi/dns_opnsense.sh
```diff
@@ -0,0 +1,268 @@
...
+    return 1
+  else
+    _saveaccountconf_mutable OPNs_Token "$OPNs_Token"
+  fi
+
+  if ! _opns_rest "GET" "/general/get"; then
+    _err "Can't Access OPNsense"
```
stilez: Not helpful to user. Cannot access what aspect of OPNsense?
Maybe reword: "Call to OPNsense API interface failed. Unable to access OPNsense API."

blablup:
What about this MR??

PR #2482, opened 2019-09-12
README.md update - acme.sh needs root/sudoer

adamczi:
Changes reflect the fact that in order to have a fully functional acme.sh script that renews certificates automatically, you must run the `--install-cert` command as root or at least add your `--reloadcmd` command to your /etc/sudoers (or /etc/doas or whatever). I think this was a needed clarification in Readme.


Neilpang:
no,   acme.sh  doesn't require root.
If the current user is a non-root, and you need to do something as root, you must configure the command to not require root.

PR #2488, opened 2019-09-17
new option: --allow-sudo

bjmgeek:
This is for https://github.com/Neilpang/acme.sh/issues/2462

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -6650,6 +6650,9 @@ _process() {
       --force-color)
         export ACME_FORCE_COLOR=1
         ;;
+      --allow-sudo)
+        export ALLOW_SUDO=1
+        ;;
       --ecc)
         _ecc="isEcc"
         ;;
@@ -6790,7 +6793,7 @@ _process() {
 
   if [ "${_CMD}" != "install" ]; then
     if [ "$__INTERACTIVE" ] && ! _checkSudo; then
-      if [ -z "$FORCE" ]; then
+      if [ -z "$FORCE" -a -z "$ALLOW_SUDO" ]; then
         #Use "echo" here, instead of _info. it's too early
         echo "It seems that you are using sudo, please read this link first:"
         echo "$_SUDO_WIKI"


Neilpang:
can you please let me know your use case: why you need sudo ?

hxx-fetch:
In my case I have legacy scripts for provisioning/management webservices etc. where later acme.sh was added to it. When running this scripts from console instead of cron this parameter would be useful.

MirLach:
I created similar patch locally with the same parameter --allow-sudo. Using --force just to be able to restore previous functionality is too dangerous because it allows overwriting of keys etc. Separate option is much better.
We are using `acme.sh` as unprivileged user but it should be run as this user by all devops. It means "devopA" runs `acme.sh` as user "certbot", "devopB" also runs `acme.sh` as user "certbot" thru `sudo`. Cronjob is under "certbot" too. All configs and certificates are stored in one place under one UID. 

I know you don't like `sudo` but why people able to configure and use `sudo` should have life more complicated is beyond me.

joeshaw:
For my use case, I am using https://github.com/j-c-m/ubnt-letsencrypt which uses acme.sh to install certs on Ubiquiti EdgeRouters.  The script does things like stopping services, making iptables changes, etc. which all require root.

It writes the files to /config, which is completely owned by root on these devices AFAICT.

acme.sh
```diff
@@ -6790,7 +6793,7 @@ _process() {
 
   if [ "${_CMD}" != "install" ]; then
     if [ "$__INTERACTIVE" ] && ! _checkSudo; then
-      if [ -z "$FORCE" ]; then
+      if [ -z "$FORCE" -a -z "$ALLOW_SUDO" ]; then
```
tbehling: Hello @bjmgeek, I want to thank you for opening this PR, as I have the same need.

I noticed one style nit, that `-a` in Bash is a deprecated style, due to some syntax ambiguity; what is commonly preferred is `&&`, like the following:

```suggestion
      if [ -z "$FORCE" ] && [ -z "$ALLOW_SUDO" ]; then
```

bjmgeek:
If you know you have bash, you could use [[ test1 && test2 ]] as well.  The
[ test1 ] && [ test2 ] syntax has the advantage of being more portable.  I
didn't know -a had been deprecated.

ThreeFx:
Is there anything preventing this from getting merged @Neilpang? I've been wanting this for a while now.

Neilpang:
@ThreeFx 

Don't use sudo.

if you need root,  please su to root first.

```
acme.sh --uninstall
sudo su
curl https://get.acme.sh | sh

bash
acme.sh --issue ....
```

MirLach:
Yeah... advising users to run everything as root and piping unverified curl output to sh.. that is security habits of 2020.

ThreeFx:
That's not gonna happen, I'm sorry @Neilpang. I'm not running code as root if I don't have to, and using `sudo` allows me to use a dedicated `acme` user which can _only_ restart my webserver when it needs to and nothing else.

I'm a big fan of the project though! Maybe I'll maintain a fork of it with the sudo option added myself :)

Neilpang:
>> allows me to use a dedicated acme user which can only restart my webserver when it needs to and nothing else.

You can configure to use command like: `sudo nginx restart` not requiring password.

Then use `sudo` with `reloadcmd`:

```
acme.sh  --install-cert -d example.com   .......   --reloadcmd "sudo nginx restart"


```


The reason why we don't recommend to use `sudo acme.sh --issue ....` is that the cronjob is like: 
```
10 0 * * * "/root/.acme.sh"/acme.sh --cron --home "/root/.acme.sh" > /dev/null
```

There is no `sudo` here,  so, the cronjob will not succeed when the cert is renewed.

I have explained many thousands of times, but  there a many users mis-used `sudo`.
I don't want them finally find that the cert is not renewed automatically in 60 days and then come and complain here.

ThreeFx:
I understand that, but in order for this to work I have to pass --force to the script, which _always_ renews certificates, regardless of their expiration time, which is not what I want.

beam:
And it impossible to use --cron with sudo, with --force cert is renewed on every run :(

PR #2508, opened 2019-09-26
Fix Acme.sh script gets hung

tomasji:
when curl is run with '-X HEAD' and '--data ""' using http1.1
the server responds: "no chunk, no close, no size. Assume close
to signal end" and hungs.

This fixes the curl calls with when -X HEAD is used.


Neilpang:
I think it's fixed already in the latest code.

PR #2509, opened 2019-09-26
Notify mail envelope and body from tweak

honzahommer:
See related issue Neilpang/acme.sh#2496.

https://github.com/Neilpang/acme.sh/blob/909441bc5770df6da1e986ec9cf9ceb11078f108/notify/mail.sh#L101

If `$MAIL_FROM` is matching pattern `^.*<\(.*\)>$` (Name <<email@address>>), then string between angle brackets is used as envelope.

### DIFF
--- a/notify/mail.sh
+++ b/notify/mail.sh
@@ -98,7 +98,9 @@ _mail_cmnd() {
   case $(basename "$_MAIL_BIN") in
     sendmail)
       if [ -n "$MAIL_FROM" ]; then
-        echo "'$_MAIL_BIN' -f '$MAIL_FROM' '$MAIL_TO'"
+        _MAIL_ENV="$(echo "$MAIL_FROM" | sed 's/^.*<\(.*\)>$/\1/')"
+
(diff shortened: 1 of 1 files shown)


Neilpang:
Is this only valid for `sendmail` ?

myelsukov:
> Is this only valid for `sendmail` ?

No. It's a standard: https://tools.ietf.org/html/rfc5322#section-3.4

Neilpang:
please make sure you have tested all the supported MAIL_BIN,  and then please update the usage here:
https://github.com/Neilpang/acme.sh/issues/2518#issuecomment-538474232

honzahommer:
> Is this only valid for `sendmail` ?

Actually it is used only with sendmail command.

Neilpang:
@honzahommer 

Can we support it for all the MAIL_BIN ?

We can have another new env variable,  like: `MAIN_FROM_NAME`
it's up to you.

Thanks.

PR #2531, opened 2019-10-06
Fix for dns_vultr.sh

moose-kazan:
Script "dns_vultr.sh" have small bug in function "_get_root". For example:
I expect what it return _domain=test.com for name  "_acme-challenge.test.com", but i got error: "Invalid domain".

This pull request fix this issue.


Neilpang:
thanks.

PR #2532, opened 2019-10-06
Adding dns_rimu.sh based on dns_zonomi.sh with changed endpoint

rroseselavy42:
Zonomi and Rimu have the same  API with different endpoints. Instead of editing the ZM_Api variable every time I needed it I just created dns_rimu.sh based on the dns_zonomi.sh
I would be more than happy to contribute an edit to the wiki to accommodate RimuHosting instead of adding another dns_ file but I thought this might be cleaner. 

Thanks! Apologies if anything wrong here, first pull request :-)

### DIFF
--- a/dnsapi/dns_rimu.sh
+++ b/dnsapi/dns_rimu.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env sh
+
+#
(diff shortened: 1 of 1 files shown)


Neilpang:
check this guide first:

PR #2537, opened 2019-10-09
Use more widely supported options for the "tr" command line utility in dns_freedns.sh

### LINKED ISSUE #2536: Report bugs to freedns dns api
Add support for Synology routers while using [dnsapi/dns_freedns.sh](https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_freedns.sh). Synology router scripting stacks do not have a version of the "tr" command line utility that supports character class representation option i.e. the "[:space:]" :

`tr -d "[:space:]"` should be replaced with `tr -d "\r\n"` inside of [dnsapi/dns_freedns.sh](https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_freedns.sh) for better support across embedded linux systems and routers.

Existing good usage:
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_freedns.sh#L161

Usages to replace:
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_freedns.sh#L306
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_freedns.sh#L352

Steps to reproduce
------------------
Use a Synology router to run acme.sh

Debug log
-----------------
The script silently fails and simply doesn't parse the freedns web page correctly.

master-nevi:
…by removing the use of the character class representation option. Fixes #2536

See #2536 for more info.

### DIFF
--- a/dnsapi/dns_freedns.sh
+++ b/dnsapi/dns_freedns.sh
@@ -303,7 +303,7 @@ _freedns_domain_id() {
       return 1
     fi
 
-    domain_id="$(echo "$htmlpage" | tr -d "[:space:]" | sed 's/<tr>/@<tr>/g' | tr '@' '\n' \
+    domain_id="$(echo "$htmlpage" | tr -d "\r\n" | sed 's/<tr>/@<tr>/g' | tr '@' '\n' \
       | grep "<td>$search_domain</td>\|<td>$search_domain(.*)</td>" \
       | _egrep_o "edit\.php\?edit_domain_id=[0-9a-zA-Z]+" \
       | cut -d = -f 2)"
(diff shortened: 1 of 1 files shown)


dkerr64:
This should be updated to also remove tabs and spaces as well, and maybe even vertical tabs and form feeds.  While things may be working just fine now, the subsequent greps and seds assume that spaces have been removed.  FreeDNS does not presently have spaces in places that the code is sensitive to, but if that changed in the future this would fail.  Hence if "[:space:]" won't work then I would replace with " \r\n\t\v\f" (note blank space within the string"
David

dkerr64:
Correction to my previous comment.  You DO NEED to explicitly have a space character in the string as there ARE spaces that need to be removed for later greps to work.

Neilpang:
@dkerr64 
Sorry,  please create a new PR to fix it.

Thanks.

PR #2551, opened 2019-10-22
fix error

fjh1997:

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,8 +200,9 @@ acme.sh --install-cert -d example.com \
 **Nginx** example:
 ```bash
 acme.sh --install-cert -d example.com \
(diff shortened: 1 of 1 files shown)


Neilpang:
it's not an error.
nginx usually uses the fullchain.pem instead of the cert.pem.

fjh1997:
> it's not an error.
> nginx usually uses the fullchain.pem instead of the cert.pem.

However，I have try the 

> it's not an error.
> nginx usually uses the fullchain.pem instead of the cert.pem.

Thank you! I's my mistake,I have changed the file extension from cert.pem to cert.crt, and It causes fullchain file not work,after I change it back, it works,thanks again!

PR #2560, opened 2019-10-25
Avoid API throttling errors in AWS DNS plugin

scottkof:
The AWS DNS (Route53) API currently has a rate limit of 5 requests per second. Requests in excess of this rate return "Throttling/Rate exceeded" errors. This issue has been reported a number of times, see:
https://github.com/Neilpang/acme.sh/issues/1506
https://github.com/Neilpang/acme.sh/issues/2132
https://github.com/Neilpang/acme.sh/issues/2382

This pull request addresses the issue by adding one-second pauses in the form of `sleep 1`. 

Adding a DNS entry to Route53 without error requires four API requests. Removing a DNS entry without error also requires four API requests. When this code is merged, there will be a one-second pause after the fourth API request. Whether or not there are errors, this PR adds two seconds per domain to issue/renewal.

Reference:
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -6,6 +6,8 @@
 #AWS_SECRET_ACCESS_KEY="xxxxxxx"
 
 #This is the Amazon Route53 api wrapper for acme.sh
+#All `sleep` commands are included to avoid Route53 throttling, see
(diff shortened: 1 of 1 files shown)


Neilpang:
please use `_sleep`  function instead.

scottkof:
Done.

PR #2569, opened 2019-11-01
Add openssh package for ssh.sh deploy support

johannrichard:
* `acme.sh`'s `ssh.sh` is probably one of the deploy hooks that's most versatile
* By default, it's not installed on `alpine` docker images and therefore is lacking in the `acme.sh` docker image
* This change adds the `openssh-client` package and therefore the `ssh` and associated commands

BTW: `acme.sh` is awesome. 🥇

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:3.10
 RUN apk update -f \
   && apk --no-cache add -f \
   openssl \
-  openssh \
+  openssh-client \
   coreutils \
   bind-tools \
   curl \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM alpine:3.10
 RUN apk update -f \
   && apk --no-cache add -f \
   openssl \
+  openssh \
```
Neilpang: I think we only need  `openssh-client` here.

Thanks.

Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM alpine:3.10
 RUN apk update -f \
   && apk --no-cache add -f \
   openssl \
+  openssh-client \
```
johannrichard: As per your comment.

PR #2589, opened 2019-11-14
Add docker console logging

christianbur:
i have the problem that the acme.sh container does not show logs (docker-compose logs acme.sh), to fix this i changed the crond parameters. 

```
/ # crond --help
BusyBox v1.30.1 (2019-06-12 17:51:55 UTC) multi-call binary.

Usage: crond -fbS -l N -d N -L LOGFILE -c DIR

	-f	Foreground
	-b	Background (default)
	-S	Log to syslog (default)
	-l N	Set log level. Most verbose 0, default 8
	-d N	Set log level, log to stderr
	-L FILE	Log to FILE
	-c DIR	Cron dir. Default:/var/spool/cron/crontabs
```


Neilpang:
don't use `crond -f` option. it can not handle kill signals,  suche as `docker kill/stop`

PR #2598, opened 2019-11-20
fixing nginx wildcard includes with relative path

runapp:
When the file containing `server_name` tag lies in `/etc/nginx/vhosts.d`, acme.sh won't find it properly unless `cwd()=="/etc/nginx"`.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3067,7 +3067,6 @@ _setNginx() {
     _err "Can not find conf file for domain $d"
     return 1
   fi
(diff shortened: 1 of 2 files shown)


runapp:
Nice, someone has chosen to write more generic codes to fix it in 0a4ef171351db5eb7f02341e231801e80026d33e, but it never succeeded and has severe problem. Now the whole nginx module is unusable when `include` is used.

PR #2604, opened 2019-11-26
Added trailing slash to end of each line of DEPLOY_SCRIPT_CMD

cngarrison:
After upgrading RouterOS to v6.45, the script commands passed to the router could not be parsed correctly (syntax error). 

A semi-colon `;` was added to the end of each command, and backslash `\` added to end of each line.


herbetom:
I can confirm that this pull request solves the mentioned problem. 
Since the deploy-hook isn't working https://github.com/Neilpang/acme.sh/issues/2344#issuecomment-514880505 with current versions of routeros it would be really nice if this pull request could soon be accepted, so that the other problems of the hook https://github.com/Neilpang/acme.sh/pull/2413 could be solved as well.

PR #2606, opened 2019-11-27
Don't exit with issue failure if in DNS manual mode

ianw:
In our environment we use DNS manual mode and take the TXT record
output of acme.sh and process it with Ansible to install the records
(then we call renew later when the records have been pushed to the DNS
servers by a whole bunch of other bits).

One problem is that after getting/showing the TXT records, acme.sh
always returns 1.  This makes it difficult to tell if there is
actually an error condition.

Since we have set the manual-mode flag, not installing the DNS records
is an expected correct result.  This proposes checking for the
FORCE_DNS_MANUAL flag and, if present, returning 0 in this case.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4262,7 +4262,13 @@ $_authorizations_map"
       _err "Please add the TXT records to the domains, and re-run with --renew."
       _on_issue_err "$_post_hook"
       _clearup
-      return 1
+      if [ -n "$FORCE_DNS_MANUAL" ]; then
+        # If asked to be in manual DNS mode, don't flag lack of DNS
+        # entries as a failure.
+        return 0
+      else
+        return 1
+      fi
     fi
 
   fi


Neilpang:
if you just want to add txt record by yourself,  why not use a custom api hook ?

https://github.com/Neilpang/acme.sh/wiki/dnsapi#use-custom-api

ianw:
@Neilpang well the DNS manual mode does exactly what we want, which is give us a TXT record we can install, without any other dependencies (no virtuatlenv's etc to manage was the appeal of acme.sh :)

Not that I expect you to dive into our configuration system but it is documented at [1] with links ot the roles, etc. for reference.  A central Ansible server calls acme.sh on each host to get a certificate, takes notes of the TXT records required, then deploys this to the bind configuration on the  nameservers (and reloads) and finally runs the renewal process on the remote hosts to get the certs.

So it's all completely automated, but not in a "call a DNS API from the client" way.  I imagine it's probably a common-ish pattern for anyone in an ansible/puppet type environment where you have a central authorised server.

In short, not knowing if we're seeing a real error or not is basically the only problem with DNS manual mode for us :)

[1] https://docs.openstack.org/infra/system-config/letsencrypt.html

ianw:
@neilpang I had a further play to see if I could use the API in some way to simulate manual mode as you suggested, but it doesn't work.  If the ``myapi_add`` function returns with 0, then acme.sh will pause for a bit and then try to issue the certificates:

https://github.com/Neilpang/acme.sh/blob/ef15e55947f0df966822cdbede6a1bd572970eb6/acme.sh#L4270

I can't pause or wait in the plugin because the server orchestration will hang forever.  acme.sh needs to show us the required TXT records, stop, and the server orchestration will come back later after it has installed the dns entries and re-run with renewal.

If I return an error, then it's the same situation of it *not* actually being an error and messing up checking if there is actually something wrong getting the TXT records.

We've set the override flags and DNS manual mode is working and doing what it says and what we want.  That's why I don't think it should be raising an error exit when it's working as expected.

Neilpang:
@ianw 
Will it work for you if we return a new error code for dns manual mode?
Maybe we can `return 3` for dns manual mode.

what do you think?

ianw:
@neilpang yes any different exit code would work, thanks.  

That would mean 0 is ok, 1 is error, 2 is no renewal and (theoretically) 3 is dns-manual-mode-but-ok.

Personally, I think that 0 is most correct, because it is working as expected.  But I can certainly match another code and consider that OK too.

ianw:
@neilpang; any chance we could get a resolution on this?  I'd most certainly prefer to not have to run with any sort of fork if not necessary

PR #2609, opened 2019-11-28
Enable docker console logging

christianbur:
i have the problem that the acme.sh container does not show logs (docker-compose logs acme.sh), with this PR I fix this.


seoester:
#2659 may enable #2589 to work without problems.

PR #2614, opened 2019-12-02
Adding abillity to deploy cert to Palo Alto Networks Firewall via API.

paulmnguyen:

### DIFF
--- a/deploy/panos.sh
+++ b/deploy/panos.sh
@@ -94,7 +94,6 @@ panos_deploy() {
   _cdomain="$1"
   _ckey="$2"
   _cfullchain="$5"
-  
   # PANOS ENV VAR check
   if [ -z "$PANOS_USER" ] || [ -z "$PANOS_PASS" ] || [ -z "$PANOS_HOST" ]; then
     _debug "No ENV variables found lets check for saved variables"
--- a/deploy/panos.sh
+++ b/deploy/panos.sh
@@ -13,61 +13,75 @@
 
 # This function is to parse the XML
 parse_response() {
-  status=$(echo "$1" | sed 's/^.*"\([a-z]*\)".*/\1/g')
-  message=$(echo "$1" | sed 's/^.*<result>\(.*\)<\/result.*/\1/g')
+  type=$2
+  if [ $type = 'keygen' ]; then
+    status=$(echo "$1" | sed 's/^.*\(['\'']\)\([a-z]*\)'\''.*/\2/g')
+    if [ "$status" = "success" ]; then
+      panos_key=$(echo "$1" | sed 's/^.*\(<key>\)\(.*\)<\/key>.*/\2/g')
+      _panos_key=$panos_key
+      message='PAN-OS key is set.' 
+    else
+      message="PAN-OS Key could not be set."
+    fi
+  else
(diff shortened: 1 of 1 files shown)


paulmnguyen:
There is an error in Travis regarding "git diff --exit-code"

How do I fix this?

Neilpang:
please add your usage to this page:
https://github.com/Neilpang/acme.sh/wiki/deployhooks

Neilpang:
please read this guide and create a bug:

paulmnguyen:
I have added the usage and created a bug for future bug reports.

deploy/panos.sh
```diff
@@ -0,0 +1,166 @@
...
+
+# This function is to deploy the cert to the Firewall
+deploy_cert() {
+  #Checking for IP and Key
+  if [ -z "$_panos_host" ] && [ -z "$_panos_key" ]; then
+    _err "No IP or Key found. Can not deploy certs."
+    return 1
+  else # IP and key found lets import cert, key and commit.
+    _debug "Deploying Cert"
+    cert_response=$(curl -k --form file=@"$_cfullchain" "https://$_panos_host/api/?type=import&category=certificate&certificate-name=$_cdomain&format=pem&key=$_panos_key")
```
Neilpang: don't use `curl`,  please use `_get()` or `_post()` function instead.
we need to support wget.

paulmnguyen: @Neilpang How do I make a form-data file post using your  _post() function?

Neilpang: it's submitting a form to upload cert and key.  we can use `_post()` to simulate this action.
It's just about the content-type and the postbody.

content-type is just a header, which is easy to set.
postbody is urlencoded file,  which is  percentage sign`%` splited hex encoding bytes.

We have a `_url_encode()` function to do so.

paulmnguyen:
Reopening

Neilpang:
please fix the ci errors.

deploy/panos.sh
paulmnguyen: Thanks Neil refactored the code to use _post

deploy/panos.sh
```diff
@@ -0,0 +1,144 @@
...
+      _err "PANOS_HOST not defined."
+      return 1
+    fi
+  else
+    _debug "PANOS HOST is set. Save to domain conf."
+    _panos_host="$PANOS_HOST"
+    _savedomainconf _panos_host "$_panos_host"
+  fi
+  # Retrieve stored variables
+  _panos_user="$(_readaccountconf_mutable PANOS_USER)"
```
Neilpang: do not use `_readaccountconf_mutable/_saveaccountconf_mutable`,
please use `_getdeployconf/_savedeployconf` instead.

example:
https://github.com/acmesh-official/acme.sh/blob/master/deploy/docker.sh#L67

paulmnguyen: Should I also replace _savedomainconf as well?

paulmnguyen: decided to just use savedeployconf for all 3.

paulmnguyen:
A user stated that there is an error when deploying the certificates to Panorama. I will need to resolve that issue before we can merge this code.

paulmnguyen:
Fixed the deployment issue to panorama.

PR #2628, opened 2019-12-11
Add DNS API support for Constellix

woutd:
The following plugin adds support for the Constellix DNS API using bulk modify calls.

Currently, `dns_constellix_add` will fail if the TXT record already exists. Is this a problem? Should it be removed first or it is preferred that a manual intervention is needed in this case?

Please let me know about any issues or required changes.

Kind regards
Wout


Neilpang:
please check this guide:

PR #2661, opened 2020-01-02
Add --verify-server option

blry:
Add --verify-server option which is used to force to renew a cert when ACME Directory Resource URI changes.


Neilpang:
why need this ?

blry:
I'll describe my case. We have an instance with CI/CD setup, and we don't have access using ssh. 
There's a shell script in Docker which runs every 6hours acme.sh with environment option for acme directory (e.g. $TESTING_ACME equals to --testiglng) to check if the certificate needs renew.
In some time we decided to use real certificate, however when we disabled TESTING_ACME variable to generate certificate with real CA, it responded that actual certificate renew will be in a month. 

Theoretically there was an option to make a commit with acme.sh --force and then reverse it, but what if we had more servers? It's a bad practice to do so.

PR #2670, opened 2020-01-04
Fix pfsense linode

sreyemnayr:
Here are the changes I made to fix the Linode (v4) API on my Netgate pfSense device. It required both replacing \s with hard spaces AND escaping the open brackets. I put them in separate commits since I'm not sure if the open-bracket escaping will cause problems on other systems. The space replacement should be fine everywhere, however. 
I can also do a search-replace for the other APIs if you'd like, but I can't test them since I don't use those services for nameservers... I assume that the same problem will persist no matter what.

### DIFF
--- a/dnsapi/dns_linode_v4.sh
+++ b/dnsapi/dns_linode_v4.sh
@@ -74,7 +74,7 @@ dns_linode_v4_rm() {
   if _rest GET "/$_domain_id/records" && [ -n "$response" ]; then
     response="$(echo "$response" | tr -d "\n" | tr '{' "|" | sed 's/|/&{/g' | tr "|" "\n")"
 
-    resource="$(echo "$response" | _egrep_o "{.*\"name\": *\"$_sub_domain\".*}")"
+    resource="$(echo "$response" | _egrep_o "\{.*\"name\": *\"$_sub_domain\".*}")"
     if [ "$resource" ]; then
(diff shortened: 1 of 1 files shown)


dnsapi/dns_linode_v4.sh
```diff
@@ -74,9 +74,9 @@ dns_linode_v4_rm() {
   if _rest GET "/$_domain_id/records" && [ -n "$response" ]; then
     response="$(echo "$response" | tr -d "\n" | tr '{' "|" | sed 's/|/&{/g' | tr "|" "\n")"
 
-    resource="$(echo "$response" | _egrep_o "{.*\"name\":\s*\"$_sub_domain\".*}")"
+    resource="$(echo "$response" | _egrep_o "\{.*\"name\": *\"$_sub_domain\".*}")"
```
Neilpang: why bracket `\{` is escaped?

sreyemnayr: It's the only way it worked on the FreeBSD pfSense appliance I was using. I think it has something to do with the POSIX vs PCRE regular expressions implemented in grep for the OS, but that's a little outside of my depth of knowledge. I can only say that it worked only when both the { were escaped AND the \s were converted to hard spaces. Doing one or the other did not result in success.

sreyemnayr: Based on this answer: https://unix.stackexchange.com/a/445641 
It may be best to do [{] rather than escaped for cross-platform compatibility

PR #2671, opened 2020-01-05
dns_lexicon.sh: Add extra variable _API_KEY

Rayzilt:
Lexicon TransIP module uses LEXICON_TRANSIP_API_KEY for authentication.
(https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/transip.py)

A simple search reveals more modules that uses LEXICON_${PROVIDER}_API_KEY variable.
(https://github.com/AnalogJ/lexicon/search?q=--auth-api-key&unscoped_q=--auth-api-key)


Neilpang:
can you update the usage here:
https://github.com/Neilpang/acme.sh/wiki/How-to-use-lexicon-DNS-API

Rayzilt:
I've added an TransIP usage section to the wiki.

PR #2679, opened 2020-01-07
 Implement forced V_TYPE DNS for wildcard domains.

notjames:
Might need to be more involved than this, but this works for my use case.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4071,7 +4071,6 @@ $_authorizations_map"
       _debug "_currentRoot" "$_currentRoot"
 
       vtype="$VTYPE_HTTP"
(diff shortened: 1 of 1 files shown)


Neilpang:
no, it doesn't make any sense to silently use V_TYPE DNS in the code when the user doesn't specify a dns method for a wildcard domain.

If we detect it's a wildcard, but the user doesn't specify a dns method,  we should just report an error and exit.
The user should correct his command line parameters.

notjames:
The problem with the code as it is, though, is that it defaults to type HTTP when DNS is specified as was in my case. The PR submitted fixed that issue.

Issue #1773, opened 2018-08-08
dns_yandex does not remove txt-record


krivochenko:
Steps to reproduce
------------------
1. set environment variable PDD_Token
2. run `/root/.acme.sh/acme.sh --issue -d krivochenko.ru --dns dns_yandex --accountemail "all@krivochenko.ru" --test --debug 2`
3. after issue cert I steel seeing TXT-record for _acme-challenge

Debug log
-----------------

```
[Wed Aug  8 16:53:47 UTC 2018] Removing DNS records.
[Wed Aug  8 16:53:47 UTC 2018] txt='DyBlcf-IrUZwfFzkoU4AtfkP5xqu9z9cLFA9nPQ30iU'
[Wed Aug  8 16:53:47 UTC 2018] d_api='/root/.acme.sh/dnsapi/dns_yandex.sh'
[Wed Aug  8 16:53:47 UTC 2018] _d_alias
[Wed Aug  8 16:53:47 UTC 2018] Calling: dns_yandex_rm() '_acme-challenge.krivochenko.ru'
[Wed Aug  8 16:53:47 UTC 2018] GET
[Wed Aug  8 16:53:47 UTC 2018] url='https://pddimp.yandex.ru/api2/admin/domain/domains?page=1&on_page=20'
[Wed Aug  8 16:53:47 UTC 2018] timeout=
[Wed Aug  8 16:53:47 UTC 2018] _CURL='curl -L --silent --dump-header /root/.acme.sh/http.header  --trace-ascii /tmp/tmp.IePBmf  -g '
[Wed Aug  8 16:53:48 UTC 2018] ret='0'
[Wed Aug  8 16:53:48 UTC 2018] res1='{"direction":"asc", "on_page": 20, "success":"ok", "domains":[{"status":"added", "from_registrar":"no", "name":"afigol.ru", "ws_technical":"no", "logo_enabled": true, "m
aster_admin": true, "nsdelegated": true, "emails-max-count": 1000, "emails-count": 2, "dkim-ready": true, "logo_url":"", "stage":"added", "aliases":[]}, {"status":"added", "from_registrar":"no", "name":"coo
kery-book.com", "ws_technical":"no", "logo_enabled": true, "master_admin": true, "nsdelegated": true, "emails-max-count": 1000, "emails-count": 4, "workspace":"yes", "dkim-ready": true, "logo_url":"", "stag
[...]

PR #2690, opened 2020-01-12
Fix dns records removing after usage.

nstepa:
Issue #1773


Neilpang:
Can you please explain more details ?
Thanks.

nstepa:
Existing code doesn't remove added DNS records, so ppl ending up having several _acme-challenge subdomains for their domain.

PR #2692, opened 2020-01-14
fix for ddnss.de updates

helbgd:


Neilpang:
why is this necessary?

It seems the old url is still working

helbgd:
the upd.php script that is triggered is not reachable from ddnss.de
if you call www.ddnss.de there seems to be a bug in the upd.php script once you update the txt record it also deletes the ipv4 address from the dns record at the ddnss.de DNS servers.
if you do the same with ip4 instead of www it also updates the ipv4 record correctly.

This seems to be happening once the dyndns txt update is triggered via ipv6 thenn it deletes the ipv4 record.

PR #2717, opened 2020-02-05
Add support for octoDNS (multi-provider)

josvo:
This PR adds an API for octoDNS (see https://github.com/github/octodns) so that you are able to push your TXT record to multiple providers. It is basically just a wrapper for other APIs so make sure the APIs of your provider works and you have exported (or saved) their credentials before calling this API.

### DIFF
--- a/dnsapi/dns_octodns.sh
+++ b/dnsapi/dns_octodns.sh
@@ -25,7 +25,7 @@ dns_octodns_add() {
   fi
 
   #save the providers list to the account conf file.
-  _saveaccountconf OCTODNS_PROVIDERS "$OCTODNS_PROVIDERS"
+  _saveaccountconf_mutable OCTODNS_PROVIDERS "$OCTODNS_PROVIDERS"
 
   for element in $(echo "$OCTODNS_PROVIDERS" | tr "_" ' '); do
     _debug element "$element"
--- a/dnsapi/dns_octodns.sh
+++ b/dnsapi/dns_octodns.sh
@@ -39,7 +39,11 @@ dns_octodns_add() {
 
     addcommand="dns_${element}_add"
     _debug addcommand "$addcommand"
-    $addcommand "$fulldomain" "$txtvalue"
+    if ! $addcommand "$fulldomain" "$txtvalue"; then
+      _err "Adding record via $element API was not successful!"
+      _err "Please check if this API works."
+      return 1
+    fi
   done
 
   _info "Finished adding records via octoDNS API"
@@ -74,7 +78,12 @@ dns_octodns_rm() {
 
     rmcommand="dns_${element}_rm"
     _debug rmcommand "$rmcommand"
-    $rmcommand "$fulldomain" "$txtvalue"
+    if ! $rmcommand "$fulldomain" "$txtvalue"; then
+      _err "Removing record via $element API was not successful!"
+      _err "You might have to remove the key manually."
+      _err "Please check if this API works."
+      return 1
+    fi
   done
 
   _info "Finished deleting records via octoDNS API"
--- a/dnsapi/dns_octodns.sh
+++ b/dnsapi/dns_octodns.sh
@@ -26,14 +26,12 @@ dns_octodns_add() {
 
   #save the providers list to the account conf file.
   _saveaccountconf OCTODNS_PROVIDERS "$OCTODNS_PROVIDERS"
-  return 1
 
-  IFS='_' read -r -a used_providers <<< "$OCTODNS_PROVIDERS"
-  for element in "${used_providers[@]}"
-  do
+  for element in $(echo "$OCTODNS_PROVIDERS" | tr "_" ' '); do
     _debug element "$element"
     sourcecommand="$_SUB_FOLDER_DNSAPI/dns_${element}.sh"
 
+    # shellcheck disable=SC1090
     if ! . "$sourcecommand"; then
       _err "Load file $sourcecommand error. Please check your api file and try again."
       return 1
@@ -44,7 +42,7 @@ dns_octodns_add() {
     $addcommand "$fulldomain" "$txtvalue"
   done
 
-  _info "Finished adding records via octoDNS"
+  _info "Finished adding records via octoDNS API"
   
   return 0
 }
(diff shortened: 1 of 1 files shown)


Neilpang:
please read this guide first:

josvo:
> please read this guide first:
> https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide

@Neilpang  Thanks for you commen:
2. The script file is named:  dns_octodns.sh
3. file is placed in acme.sh/dnsapi
4. The file shebang is sh
5. the add and rm functions are in the script
9. I'll update the wiki as soon as this get's merged - doesn't make much sense as long as it is not merged. 
Style guides: script is checked with shellcheck.net

If you have a hint were I should change something, please provide some more information and I'll be happy to work on it until it fullfills your standards.

dnsapi/dns_octodns.sh
```diff
@@ -0,0 +1,92 @@
...
+  _debug txtvalue "$txtvalue"
+  if [ -z "$OCTODNS_PROVIDERS" ]; then
+    OCTODNS_PROVIDERS=""
+    _err "You didn't specify OCTODNS_PROVIDERS yet."
+    _err "Please specifiy your providers and try again."
+    return 1
+  fi
+
+  #save the providers list to the account conf file.
+  _saveaccountconf OCTODNS_PROVIDERS "$OCTODNS_PROVIDERS"
```
Neilpang: don't use `_saveaccountconf`

see the example about what and how we use:
https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_cf.sh

dnsapi/dns_octodns.sh
```diff
@@ -0,0 +1,92 @@
...
+
+  for element in $(echo "$OCTODNS_PROVIDERS" | tr "_" ' '); do
+    _debug element "$element"
+    sourcecommand="$_SUB_FOLDER_DNSAPI/dns_${element}.sh"
```
Neilpang: don't need to load the source file.
When you come here,  you are already loaded.

dnsapi/dns_octodns.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env sh
+
+# This API is a wrapper for other API so that you
+# are able to put your TXT record to multiple providers.
+# The functionality is in the responsibility of the respective API
+# therefore please check if your used APIs work.
+# e.g. for using NS1 and AWS Route53, use:
+# OCTODNS_PROVIDERS=nsone_aws
```
Neilpang: why?
why we need your api to load other providers ?

josvo: That is because this is just a wrapper for other APIs. The idea of octodns is have dns as code and provide the possibility to use multiple dns providers for your domains. And when using multiple providers, in my example NS1 and AWS Route53, we need to put the acme TXT Record on both providers at once (or we get a 50% chance of failing to create the let's encrypt certificate). 
Let's take the domain swisstxt.org for example. When you query the NS records, you get
```;; ANSWER SECTION:
swisstxt.org.		3600	IN	NS	ns-544.awsdns-04.net.
swisstxt.org.		3600	IN	NS	dns4.p07.nsone.net.
swisstxt.org.		3600	IN	NS	ns-1660.awsdns-15.co.uk.
swisstxt.org.		3600	IN	NS	ns-181.awsdns-22.com.
swisstxt.org.		3600	IN	NS	ns-1490.awsdns-58.org.
swisstxt.org.		3600	IN	NS	dns2.p07.nsone.net.
swisstxt.org.		3600	IN	NS	dns3.p07.nsone.net.
swisstxt.org.		3600	IN	NS	dns1.p07.nsone.net.
```
That is why I created this wrapper API where you can define which providers are in use and the acme script will automatically put the TXT records to all listed providers (and delete them after use).

dnsapi/dns_octodns.sh
josvo: We call the other providers from dns_octodns.sh directly and not from acme.sh. And in dns_octodns.sh, the other provider scripts are not sourced (I remember having errors when not sourcing them here). 
So the idea is to source only the the API files for the providers that are specified. 
See also my next comment.

dnsapi/dns_octodns.sh
josvo: Right; I took dns_nsone.sh as an example. This API uses `saveaccountconf` unfortunately.
I change it to `_saveaccountconf_mutable`.

josvo:
Hello @Neilpang 
Is there anything I can do for this PR? Please let me know.
Thanks for your efforts!

Neilpang:
I need more time to think about it.

josvo:
@Neilpang: I this functionality not needed, not wanted or implemented otherwise?

Neilpang:
sorry, it was closed by accident.

josvo:
@Neilpang No problem :smiley: 
Please let me know if I can do something here - we are already using this in production for quite some time.

dnsapi/dns_octodns.sh
Neilpang: why not use the function `_findhook()` to search for the dns api:
https://github.com/acmesh-official/acme.sh/blob/2a2d556551e9266a3924da205c1ede55d89a689d/acme.sh#L4578

dnsapi/dns_octodns.sh
```diff
@@ -0,0 +1,92 @@
...
+  _debug txtvalue "$txtvalue"
+  if [ -z "$OCTODNS_PROVIDERS" ]; then
+    OCTODNS_PROVIDERS=""
+    _err "You didn't specify OCTODNS_PROVIDERS yet."
+    _err "Please specifiy your providers and try again."
+    return 1
+  fi
+
+  #save the providers list to the account conf file.
+  _saveaccountconf_mutable OCTODNS_PROVIDERS "$OCTODNS_PROVIDERS"
```
Neilpang, line 28: dont' save to account conf.   save it to the domain conf.

dnsapi/dns_octodns.sh
```diff
@@ -0,0 +1,92 @@
...
+    _debug element "$element"
+    sourcecommand="$_SUB_FOLDER_DNSAPI/dns_${element}.sh"
+
+    # shellcheck disable=SC1090
+    if ! . "$sourcecommand"; then
+      _err "Load file $sourcecommand error. Please check your api file and try again."
+      return 1
+    fi
+
+    addcommand="dns_${element}_add"
```
Neilpang, line 40: the  $element  should contain the `dns_` prefix.

Issue #2398, opened 2019-07-20
[DNS mode] Cloudflare New API Tokens


mxvin:
Hello,
Cloudflare just releasing new API Tokens that can specify each API key for it's usage (Access Permission), that more secure than using Global API key.

As stated on https://api.cloudflare.com/#getting-started-requests (section API Tokens) implementing new API is just dirt easy by sending header `"Authorization: Bearer <token>"` instead of Global API that using header `X-AUTH-EMAIL: ` & `X-AUTH-KEY: `

It's quite possible for adding new variable on account.conf like `CF_API_Tokens=<tokens>` and make some logic on [dns_cf.sh](https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_cf.sh) that can deal with both new API Tokens & Global API header payload.

Regards.

mxvin:
I did 'some' changes to the `dns_cf.sh` file and do integrate both Global API and new API Tokens in one script. Adding some variables: `CF_UseGlobalAPI` defining what API method to use, and `CF_Account_ID` because new auth method need your Account ID for zones query.

If you feel curious or just wanna try and look around what I do, [here you go.](https://github.com/mxvin/acme.sh/commit/00e059b7f75103dd507eb170a03befb9be2746ee) Hope you like it. :)

Neilpang:
fixed.  please try with the latest `dev` branch.

```sh
acme.sh --upgrade -b dev

```

Berzerker:
I believe this needs to be based on Zone ID, rather than account ID. Running a general account ID query with a specific zone being allowed (with no other zones being allowed) gives you an error that the account cannot list zones since the default API call tries to list all the entire account's information when it only has access to a specific zone.

`{"success":false,"errors":[{"code":0,"message":"Requires permission \"com.cloudflare.api.account.zone.list\" to list zones of the account"}],"messages":[],"result":null}%`

Running this with a specific Zone ID allows permission to list that specific zone, but even permissions across all zones will allow you to list that specific zone as well.

Running the API call with `.../client/v4/zones/<zoneid>` for that specific zone rather than `.../client/v4/zones?account.id=<accountid>` allows for more granular permissioning without reduced functionality.

AllenBalaj:
[...]

PR #2728, opened 2020-02-10
add support for using a Zone ID

artooro:
Per comments in issue #2398 this pull request is to support using Cloudflare Zone IDs which allows an API token to be restricted to a specific domain which is not possible with the current implementation.

### DIFF
--- a/dnsapi/dns_cf.sh
+++ b/dnsapi/dns_cf.sh
@@ -153,8 +153,8 @@ _get_root() {
       if _contains "$response" '"success":true'; then
         _domain=$(printf "%s\n" "$response" | _egrep_o "\"name\":\"[^\"]*\"" | cut -d : -f 2 | tr -d \" | head -n 1)
         if [ "$_domain" ]; then
-          _cutlength=$(expr ${#domain} - ${#_domain} - 1)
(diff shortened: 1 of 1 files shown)


artooro:
I'll fix the shellcheck issue hopefully tonight.

msg7086:
We were using this version but we saw occasional failures on removing entries.

```
[Wed 19 Feb 2020 05:56:33 AM UTC] Removing DNS records.
[Wed 19 Feb 2020 05:56:33 AM UTC] Removing txt: xxyyzz for domain: _acme-challenge.xxyyzz.example.com
[Wed 19 Feb 2020 05:56:34 AM UTC] Error removing txt for domain:_acme-challenge.xxyyzz.example.com
```

I cannot reproduce this issue reliably. When I enable debug mode and test for a few times it worked. On production it sometimes fails.

Did you hit the same issue?

artooro:
@msg7086 I've had some issues which appeared to be transient cloudflare issues, and that happened before this patch.
In order to get a full understanding we'd have to see debug logs of course.

msg7086:
Thanks. If I manage to get a debug log I'll let you know.

PR #2736, opened 2020-02-18
Make socat debug output show version + features instead of help text

xpac1985:
When using `--debug`, the output shows the version of `openssl`, `nginx` etc.
For `socat`, it uses `-h` which shows the help text, instead of `-V`, which shows version + feature infos.


xpac1985:
Should've built that change against the dev branch... I opened a new PR, pls ignore this.

PR #2737, opened 2020-02-18
Make socat debug output show version + features instead of help text

xpac1985:
When using --debug, the output shows the version of openssl, nginx etc.
For socat, it uses -h which shows the help text, instead of -V, which shows version + feature infos.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -207,7 +207,7 @@ _dlg_versions() {
 
   echo "socat:"
   if _exists "socat"; then
-    socat -h 2>&1
+    socat -V 2>&1
   else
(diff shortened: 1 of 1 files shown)


Neilpang:
why is this necessary ?

xpac1985:
The function is supposed to show diagnosis data, e.g. versions of the tools being used. openssl, nginx and apache all correctly show their version numbers etc.
But socat is used with the `-h` switch - which makes it show the help output, but not the version number or any other useful details.
My commit changes this to use the `-V` switch, so socat actually prints the version number.

PR #2740, opened 2020-02-20
haproxy deploy script now compatible with OpenSSL v1.1+

xpac1985:
The OpenSSL OCSP request for haproxy deployment breaks from OpenSSL v1.1.0 on.
The format of the `-header` option has been changed and does now contain a `=` instead of a whitespace.
Other projects have hit the same issue:
https://github.com/nghttp2/nghttp2/issues/742

This commit determines the OpenSSL/LibreSSL version and then adjusts the request accordingly.
Also removed the duplicate command line and added some more debug output.


xpac1985:
Fixed formatting and created new pull request.

PR #2741, opened 2020-02-20
Add anchor to make link more specific

adrian5:
Nicer than to discover it by scrolling down; the section in the wiki looks almost identical to the section in the README.md where the user just came from.


Neilpang:
sorry,  it's not identical.
It was intended to link to the whole page, instead of an anchor to the "advanced" seciton.

PR #2749, opened 2020-02-23
Updates to ssh_deploy hook

dkerr64:
This update fixes a bug and adds ability to split all the commands sent to a remote host from a single SSH call into multiple SSH calls.

With some remote hosts, notably VMware ESXi (which uses busybox sh), the default shell threw an error when the command string sent was too long.  The solution is to split the remote commands into multiple SSH calls.  This is implemented with a new variable DEPLOY_SSH_BATCH_MODE which defaults to "yes" (emulating the existing behavior).  But when set to "no" the commands on the remote server are split into multiple calls.

Example for deploying to VMware ESXi 6.7...
```
export DEPLOY_SSH_USER="root"
export DEPLOY_SSH_SERVER="vmwareesxi.example.com"
export DEPLOY_SSH_BATCH_MODE="no"
export DEPLOY_SSH_FULLCHAIN="/etc/vmware/ssl/rui.crt"
export DEPLOY_SSH_KEYFILE="/etc/vmware/ssl/rui.key"
export DEPLOY_SSH_REMOTE_CMD="/etc/init.d/hostd restart"
```
The bug fix is to allow DEPLOY_SSH_BACKUP and DEPLOY_SSH_BATCH_MODE to correctly update when changed.  Previously once set to "no" they could never be set back to "yes"


Neilpang:
It's rare case.  Most of the users should not take care of it at all.

How about naming it as something like: `Le_Deploy_ssh_single_mode`.
Default is empty(or if not set).

Only in this case, the user set `Le_Deploy_ssh_single_mode=1`

what do you think?

dkerr64:
> It's rare case. Most of the users should not take care of it at all.
> 
> How about naming it as something like: `Le_Deploy_ssh_single_mode`.
> Default is empty(or if not set).
> 
> Only in this case, the user set `Le_Deploy_ssh_single_mode=1`
> 
> what do you think?

So you are suggesting replacing Le_Deploy_ssh_batch_mode with Le_Deploy_ssh_single_mode?  I'm not sure that I find that a good idea as it is ambiguous... does single mean send all commands in a single remote call (same as batch) or does single mean send each command separately (as a single command)?  Batch on the other hand is a recognized term in computer science and unambiguous.

Neilpang:
`Single` means ` send each command separately`.

So that its default value can be empty, all the commands are sent together.
Once set  `Single=1`, you can sent the commands one by one.

Thanks.

dkerr64:
> `Single` means ` send each command separately`.
> 
> So that its default value can be empty, all the commands are sent together.
> Once set `Single=1`, you can sent the commands one by one.
> 
> Thanks.

Sorry, I am still not getting it.  Why change to `Single=1` rather than `Batch="no"`.  There would never be a case where Single is empty/null as it would be set to some value (0/1 or yes/no) on testing whether `DEPLOY_SSH_BATCH_MODE` variable had been set. 

Right now I have
```
if [ "$DEPLOY_SSH_BATCH_MODE" = "no" ]; then
    Le_Deploy_ssh_batch_mode="no"
  elif [ -z "$Le_Deploy_ssh_batch_mode" ] || [ "$DEPLOY_SSH_BATCH_MODE" = "yes" ]; then
    Le_Deploy_ssh_batch_mode="yes"
  fi
  _savedomainconf Le_Deploy_ssh_batch_mode "$Le_Deploy_ssh_batch_mode"
```
And you are suggesting...
```
  if [ "$DEPLOY_SSH_BATCH_MODE" = "no" ]; then
    Le_Deploy_ssh_single_mode=1
  elif [ -z "$Le_Deploy_ssh_single_mode" ] || [ "$DEPLOY_SSH_BATCH_MODE" = "yes" ]; then
    Le_Deploy_ssh_single_mode=0
  fi
  _savedomainconf Le_Deploy_ssh_single_mode "$Le_Deploy_ssh_single_mode"
```
which seems to be nothing more than just changing variable names and from string to integer.  What am I missing?

Thanks.

Neilpang:
I just don't like the value `yes` or `no`.
And, as a default value, I'd prefer "not do something"(empty, or '0'),  instead of `do something`('yes').
So, the variable name should be something like `***Single|non-batch` node,  instead of `no-single|batch` mode.

dkerr64:
> I just don't like the value `yes` or `no`.
> And, as a default value, I'd prefer "not do something"(empty, or '0'), instead of `do something`('yes').
> So, the variable name should be something like `***Single|non-batch` node, instead of `no-single|batch` mode.

I will change the default, but I prefer a variable name like `Le_Deploy_ssh_multi_call="yes"` which is less ambiguous than `single`.   I think this is more clear, indicating that we are splitting the commands sent to the remote host into multiple calls.

The default value for the variable must be "no" it cannot be empty or undefined as that means use whatever pervious value was stored in the conf file.  So `export DEPLOY_SSH_MULTI_CALL=""` means use previously set value, and if never set then default to "no"

I am not switching to 1/0 from yes/no.  This is an end user setting that is saved in the conf files along with many other strings.  I use yes/no for `Le_Deploy_ssh_backup` and to change to 0/1 for this setting would be inconsistent and less clear to end user.

Neilpang:
Well, I think `Le_Deploy_ssh_multi_call` is better and acceptable.

>> The default value for the variable must be "no" it cannot be empty or undefined

Sorry, I just don't get your point about why it can not be empty.
Just think about the existing ssh deploy hook users, who upgrade to this new version.  There is no such variable defined in the conf file, so, we must deal with Empty(undefined) case.

let me know what you think.

Thanks.

dkerr64:
> Well, I think `Le_Deploy_ssh_multi_call` is better and acceptable.
> 
> > > The default value for the variable must be "no" it cannot be empty or undefined
> 
> Sorry, I just don't get your point about why it can not be empty.
> Just think about the existing ssh deploy hook users, who upgrade to this new version. There is no such variable defined in the conf file, so, we must deal with Empty(undefined) case.
> 
> let me know what you think.
> 
> Thanks.

I have made the updates.  The `DEPLOY_SSH_MULTI_MODE` exported variable can be null and in that case means use whatever the previously stored value was, and if not previously set then default to "no".  Thus this variable will have three possible states, yes, no and null.  So I cannot use null to be equivalent to no.

This is stored in `Le_Deploy_ssh_multi_call` which is saved in the domain .conf file.  The first time the code is executed there will be no value saved in the conf file, but every subsequent call will have a value of either yes or no and it will not change unless the deploy hook is called with `DEPLOY_SSH_MULTI_MODE` set to something other than null. `Le_Deploy_ssh_multi_call` can have two possible states, yes or no.  I do not want to use null to be equivalent to no as then it would not match what can be passed in from `DEPLOY_SSH_MULTI_MODE`

David.

Neilpang:
Sorry, but I mean if the value is default value, we don't need to save it in the conf.

If `DEPLOY_SSH_MULTI_CALL` is not defined,  it's empty,   we don't need to set it to `no` and then save it to the conf.
Just don't save it.   If it's empty, keep it empty.

dkerr64:
> Sorry, but I mean if the value is default value, we don't need to save it in the conf.
> 
> If `DEPLOY_SSH_MULTI_CALL` is not defined, it's empty, we don't need to set it to `no` and then save it to the conf.
> Just don't save it. If it's empty, keep it empty.

So to be clear, you would like me to use `_cleardomainconf()` to remove any setting from the file when using the default?  Correct?  And for consistency I should do the same for `Le_Deploy_ssh_backup` when that is set to default value?

When looking for the function to clear a setting I came across `_savedeployconf()` and `_getdeployconf() `could you explain what they are intended to be used for please?  I found only one deploy hook that uses them.

Thanks.

Neilpang:
>> And for consistency I should do the same for Le_Deploy_ssh_backup when that is set to default value?

It's not necessary for the existing variables.


>> I came across _savedeployconf() and _getdeployconf() could you explain what they are intended to be used for please? 

It's a new recommended way to save and use variables for deploy hooks.  Any new deploy hooks are required to use these methods.
For existing deploy hooks,   you don't have to change,  you may use the current way with domainconf.


>> you would like me to use _cleardomainconf() to remove any setting from

Yes, if it is needed.


Thanks.

dkerr64:
> > > And for consistency I should do the same for Le_Deploy_ssh_backup when that is set to default value?
> 
> It's not necessary for the existing variables.
> 
> > > I came across _savedeployconf() and _getdeployconf() could you explain what they are intended to be used for please?
> 
> It's a new recommended way to save and use variables for deploy hooks. Any new deploy hooks are required to use these methods.
> For existing deploy hooks, you don't have to change, you may use the current way with domainconf.
> 
> > > you would like me to use _cleardomainconf() to remove any setting from
> 
> Yes, if it is needed.
> 
> Thanks.

Updated to clear out value from config file if needed.  Is there documentation anywhere on recommended methods for deploy hooks?

Unless there is anything else you see, I think this is ready to merge.  Can you squash all the commits in this branch together when you merge, or would you like a clean pull request from me?

Neilpang:
I just noticed there is a hardcoded path:

```
 _backupprefix="$_homedir/.acme_ssh_deploy/$_cdomain-backup"
```

don't use this hardcoded path.  because the home dir char `~` is only defined in bash.

please use this dir instead:
```
 _backupprefix="$DOMAIN_BACKUP_PATH"
```

dkerr64:
> I just noticed there is a hardcoded path:
> 
> ```
>  _backupprefix="$_homedir/.acme_ssh_deploy/$_cdomain-backup"
> ```
> 
> don't use this hardcoded path. because the home dir char `~` is only defined in bash.
> 
> please use this dir instead:
> 
> ```
>  _backupprefix="$DOMAIN_BACKUP_PATH"
> ```

I cannot use `$DOMAIN_BACKUP_PATH` as that points to a directory on the local machine (system on which acme.sh is running) when what I need is to save a backup of the certificates on the remote machine (the target of the ssh calls).

I do agree that hard coding is bad practice and so what I think I should do is allow for a `export DEPLOY_SSH_BACKUP_PATH="<path>"` setting which would then be saved in the config file as `Le_Deploy_ssh_backup_path`.  However there still needs to be some default hard coded in the deploy hook code... because default for backup is "yes" so  unless I change the default backup to "no" there needs to be some path specified when the user fails to export a path.  Using a hidden directory under homedir feels reasonable default.

Now if `~` is not always valid I could just remove it... as ssh is going to connect to the remote system and default to the home directory I can probably get away with simply `_backupprefix=".acme_ssh_deploy/$_cdomain-backup"`

Thoughts?

dkerr64:
Nudge @Neilpang thoughts on this?  Anything else you see that should be changed?
Thanks
David

Neilpang:
don't forget to update the usage changes if any in the doc.
https://github.com/acmesh-official/acme.sh/wiki/deployhooks#3-deploy-the-cert-to-remote-server-through-ssh-access

dkerr64:
> don't forget to update the usage changes if any in the doc.
> https://github.com/acmesh-official/acme.sh/wiki/deployhooks#3-deploy-the-cert-to-remote-server-through-ssh-access

Thank you.  Yes I will go update the wiki.
David.

Issue #900, opened 2017-06-26
Docker daemon does not run with PID 1


hlavki:
Docker daemon (crond) doesn't run with PID 1 so when you run docker stop, it waits (10 seconds by default) and then kills it.

```
PID   USER     TIME   COMMAND
    1 root       0:00 sh /entry.sh daemon
    6 root       0:00 crond -f
```

Neilpang:
Thank you for pointing this out.
I will fix it soon.

Thanks.

PR #2757, opened 2020-02-27
Have crond as PID 1 in container

andrewheberle:
There was a previous issue (#900) regarding this however the PR to closed that issue did not actually make `crond` run as PID 1 inside the container when running in "daemon" mode.

This PR changes the `/entry.sh` script slightly to run `crond` via `exec`.


Neilpang:
`crond`  can not process `ctrl+c` signal.

That's why I used a shell loop to process signals.

PR #2782, opened 2020-03-08
Add OTP support to Synology DSM deployhook

lippertmarkus:
Follow-up to #2369. Adding support for OTP / 2-factor-auth for the Synology DSM deployment hook.

The added URL parameter `device_id` is ignored if OTP is not activated in DSM. Therefore the empty value when not setting the environment variable doesn't cause a problem.

I would add the following to the [wiki entry](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#20-deploy-the-cert-into-synology-dsm) if the PR get's merged:

---

**With OTP (2-Factor-Authentication)**

Use your browser to sign in with the admin account you want to use. When entering the OTP code, check the "Save this device" checkbox and continue. Get the device ID from the cookie `did` (Left click on the lock to the left side of the URL -> Cookies and Copy the content of the `did` cookie). Set the environment variable to the cookie value:
```sh
export SYNO_DID=A1Bjk...
```

---

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -39,6 +39,7 @@ synology_dsm_deploy() {
   _getdeployconf SYNO_Username
   _getdeployconf SYNO_Password
   _getdeployconf SYNO_Create
+  _getdeployconf SYNO_DID
   if [ -z "$SYNO_Username" ] || [ -z "$SYNO_Password" ]; then
     SYNO_Username=""
     SYNO_Password=""
@@ -100,6 +101,7 @@ synology_dsm_deploy() {
   # Now that we know the username and password are good, save them
   _savedeployconf SYNO_Username "$SYNO_Username"
   _savedeployconf SYNO_Password "$SYNO_Password"
+  _savedeployconf SYNO_DID "$SYNO_DID"
   _debug token "$token"
 
   _info "Getting certificates in Synology DSM"
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -15,6 +15,7 @@
 # SYNO_Scheme - defaults to http
 # SYNO_Hostname - defaults to localhost
 # SYNO_Port - defaults to 5000
+# SYNO_DID - device ID to skip OTP - defaults to empty
 #
 #returns 0 means success, otherwise error.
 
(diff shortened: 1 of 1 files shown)


deploy/synology_dsm.sh
```diff
@@ -79,7 +81,7 @@ synology_dsm_deploy() {
 
   # Login, get the token from JSON and session id from cookie
   _info "Logging into $SYNO_Hostname:$SYNO_Port"
-  response=$(_get "$_base_url/webman/login.cgi?username=$SYNO_Username&passwd=$SYNO_Password&enable_syno_token=yes")
+  response=$(_get "$_base_url/webman/login.cgi?username=$SYNO_Username&passwd=$SYNO_Password&enable_syno_token=yes&device_id=$SYNO_DID")
```
Neilpang: how about the existing users, who didn't have `SYNO_DID` saved before?
Can they renew their certs automatically ?

lippertmarkus: @Neilpang yes, an empty value for the URL parameter doesn't cause any problem.

lippertmarkus:
@Neilpang Do you have further thoughts on the PR or can we get this merged? :)

Neilpang:
Please add the usage in this page:

lippertmarkus:
@Neilpang @tresni Thanks, I added it to the deployhooks page.

Maybe we should update [the wiki entry about the Synology setup](https://github.com/acmesh-official/acme.sh/wiki/Synology-NAS-Guide) as well to use the new deployment hook instead?

I also just published [a blog post](https://lippertmarkus.com/2020/03/14/synology-le-dns-auto-renew/) describing the setup in detail. We could also link to that instead. 

What do you both think?

vizkoze:
I succesfully use the deployhook on my synology dsm 6.1.7 from an external server. My useraccount used to renew the certificate is using 2 factor auth.
When using this deployhook with 2FA, I always have to export a new did-cookie every two months because the did has a validity of only one month.
Is there a way to make that did-cookie permanent or to automate a new cookie retrieval? How do you guys do it?

tresni:
This may be possible using a similar flow to https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_cyon.sh#L167-L193 but it's definitely not been setup that way today.  That would need to be investigated and a new PR opened.

hubgitmm:
Hi @lippertmarkus,
I noticed that the device id could change and it is mandatory to automate the cert renewal when 2FA is activated ...

Is there a way to retrieve the NAS device ID from the command line ?
My Synology cert renewal sheduled task failed beacause, sometime, the synology device id changed...

Thanks in advance.

lippertmarkus:
@hubgitmm not a way that I know of. Also the Synology API docs don't describe one.

hubgitmm:
Thank you @lippertmarkus for your reply.
So, if I understand, there is no way to automate the synology cert renew when 2FA is activated on the "cert renewal" account...

PR #2798, opened 2020-03-24
add support for upgrade from tag

oliverblaha:
Since 2.8.4 the upgrade mechanism compares the commit hash to prevent unnecessary upgrades.
Before this, you could also specify a tag to explicitly upgrade to - but the comparison fails as it can't find tags within "heads".

Workaround: use --force. However, this will always update, even if not needed.

This change simply falls back to tags if no branch with the given name can be found, thus allows to specify a tag instead of a branch. At the same time, this will prevent updating if the same tag has already been installed.


Neilpang:
thanks

PR #2799, opened 2020-03-25
fix(deploy/panos): data format improvements

btorresgil:
It was discovered in testing that PAN-OS <9.0 has slightly different
requirements for the multipart/form-data format and requires the `type`
parameter to be passed in the URL. These changes should work for all
PAN-OS versions now.


btorresgil:
Thanks for merging this so quickly!

PR #2804, opened 2020-03-25
fix dns_dp.sh invalid domian

ubitsp:
Recently, dnspod returned a Chinese response by default, which caused the judgment in the code to be invalid, so use lang = en to force an English response and replace "Action completed successful "with "Operation successful"


Neilpang:
这个貌似已经修过了, 请试试最新的 code

houseme:
现在又出现了这个问题了。

PR #2815, opened 2020-03-29
skip installation when source files are not newer than destination files

oliverblaha:
With this change, certificate files are only installed if at least one of them is newer than the destination file. If so, all of the files will be installed again - otherwise the installation will be skipped and the script exits with code 2, just as it does when skipping during renew or issue. Installation can be forced with '--force'.

This change helps with idempotent automation tasks: It won't perform changes unless required, and will report if it did change anything by an exit code of 0.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -5198,6 +5198,16 @@ installcert() {
   _installcert "$_main_domain" "$_real_cert" "$_real_key" "$_real_ca" "$_real_fullchain" "$_reload_cmd"
 }
 
+_checkSkip() {
+  _src="$1"
+  _dst="$2"
+
+  if [ ! "$_dst" ]; then return 0; fi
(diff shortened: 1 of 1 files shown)


Neilpang:
we don't need to check if the file is newer.

if a renewal happens,  it must update the file.

oliverblaha:
The change is not for renewal, but for --install-cert!
So if you call this now, it will always install the certs, even if they were not renewed.

That means if you try to do automated deployment, an external script cannot detect if a new certificate has been installed or not - it can't even check the modification time, as it is not preserved during the installation.

For automated deployments (not talking about the automatic renewal that acme.sh offers, but automatically deploying acme.sh and doing the first installation of a cert), it's desirable to be idempotent, means when it runs the first time, it will install everything, and if you run it again, nothing changes. This PR would allow to do so, e.g. with Ansible.

oliverblaha:
@Neilpang - is there any further clarification I can provide?

PR #2834, opened 2020-04-03
Fix Docker crontab Entry

tomalok:
* avoids using the hardcoded default minute set when image is built - everyone gets a different minute every time they `docker run neilpang/acme.sh:latest daemon`
* allows cron job to use LE_CONFIG_HOME environment variable passed at runtime instead of hardcoding the image build-time default.
```
% docker run --rm -e LE_CONFIG_HOME=/tmp acme.sh:tomalok daemon
[Fri Apr  3 01:43:56 UTC 2020] Removing cron job
[Fri Apr  3 01:43:56 UTC 2020] LE_WORKING_DIR='/root/.acme.sh'
[Fri Apr  3 01:43:56 UTC 2020] Installing cron job
```
```
% docker exec heuristic_poitras crontab -l
56 0 * * * "/root/.acme.sh"/acme.sh --cron --home "/root/.acme.sh" --config-home "/tmp" > /dev/null
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,8 @@ RUN for verb in help \
 
 RUN printf "%b" '#!'"/usr/bin/env sh\n \
 if [ \"\$1\" = \"daemon\" ];  then \n \
+ acme.sh --uninstall-cronjob \n \
+ acme.sh --install-cronjob --config-home \"\$LE_CONFIG_HOME\" \n \
  trap \"echo stop && killall crond && exit 0\" SIGTERM SIGINT \n \
  crond && while true; do sleep 1; done;\n \
 else \n \


Neilpang:
why do you need to change the config home at runtime?
you can mount the volume to any folder on the host.  it doesn't make any sense do that.

tomalok:
I have a Docker Swarm running multiple services (haproxy, nginx, apache, dovecot, etc. etc.) each on multiple nodes.  In order to persist dynamic configuration, all nodes mount an NFS volume, but each service container _only_ mounts the portion of the NFS volume that it needs.

For the `acme.sh` service (single container on single node), I am mounting `/mnt/nfs/swarm/certs` from the node into the container at `/certs`.  This directory contains subdirectories specific to each of the other services (`haproxy/`, `nginx/`, etc.) in which certifcates for those services get installed.  There is also a `acme.sh/` subdirectory, which is `acme.sh`'s configuration home.

`/mnt/nfs/swarm/certs` is mounted inside the `acme.sh` container, instead of `/mnt/nfs/swarm/certs/acme.sh`, so that `acme.sh` can re/install certificates for all services, on all the Swarm nodes they're currently running on, as necessary.  This frees me from having to make _N_ additional mounts into the `acme.sh` container to be able to deploy certs to _N_ services.

When I run the `acme.sh` container, I'm setting the `LE_CONFIG_HOME` environment variable to `/certs/acme.sh`.  When I `docker exec` into the container to manually work with certs, this environment variable is properly respected.

But, I recently noticed that my certificates had *NOT* been updated as expected.  `crond` was indeed running inside the `acme.sh` container, so I checked the `crontab` entry, and noticed that `--config-home "/acme.sh"` option was hard-coded, ignoring the `LE_CONFIG_HOME` environment variable.  This is set when the container image is built.

I also noticed that time that the cron job runs at is also hard coded at build time.  So, everyone with the same system time zone who uses the same build of the `neilpang/acme.sh:latest` docker image in `daemon` mode has their cron jobs run at the same time.

I particularly like the fact that `acme.sh` itself is _extremely_ flexible and configurable.  It would be a good thing, in my opinion, if that flexibility was preserved when it's being run by `crond` inside a Docker container.

This PR is limited to just deal with the hardcoded cron job launch time and `LE_CONFIG_HOME`, the "itches" that I need to "scratch".  However, it is not inconceivable that someone else may have a valid use case requiring a different `LE_WORKING_DIR` when they `docker run neilpang/acme.sh:latest daemon`.

It's configurable when it's not running inside Docker, why not allow it to be so when running inside?

tomalok:
An improved fix (between dev and dev branches) is in PR #2838.

PR #2843, opened 2020-04-07
Update to work with new one.com procedure

der-berni:
Since some Months, its no longer possible to add TXT Records with the Name "_acme-challenge" to the base domain.
To override the fallback value, you must use a CNAME and proxy it.
For example.
CNAME _acme-challenge.yourdomain.com => proxy_acme-challenge.yourdomain.com
The TXT Records have to be created on proxy_acme-challenge.yourdomain.com

Since the default CNAME TTL is 3600 seconds, it is recommended to leave the CNAME record.
But if you would like to use the build-in SSL (for your web-site etc.) from one.com, you have to delete the record.

A new variable "ONECOM_KeepCnameProxy" you can set in the account.conf is used to keep the CNAME record.
By default the CNAME record will be removed.

For ex.: SAVED_ONECOM_KeepCnameProxy='1'
to keep the CNAME Record and speedup the process.

### DIFF
--- a/dnsapi/dns_one.sh
+++ b/dnsapi/dns_one.sh
@@ -6,7 +6,10 @@
 # Created: 2019-02-17
 # Fixed by: @der-berni
 # Modified: 2020-04-07
-#
+#     
+#     Use ONECOM_KeepCnameProxy to keep the CNAME DNS record
+#     export ONECOM_KeepCnameProxy="1"
+#     
 #     export ONECOM_User="username"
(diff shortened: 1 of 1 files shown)


Neilpang:
please also update the usage here:

https://github.com/acmesh-official/acme.sh/wiki/dnsapi#77-use-your-onecom-credentials-as-you-would-login-into-the-control-panel

der-berni:
done

PR #2863, opened 2020-04-17
Filter out blank lines

mod242:
Response from the provider has changed so that there are blank lines at the end, which leads to the result can not be parsed correctly.


mod242:
Maybe same issue as PR 2864? I think the CI isn't working correctly here (at least I don't see a reason in the codechange why ci is breaking)

PR #2864, opened 2020-04-17
Fix: allow removal of email address as contact

siwyd:
It seems the current code doesn't allow for removing the email address
from the contact field. This fixes that. This only removes the email
address if an explicit empty email address is specified on the command
line or in the account.conf file. If it is left unspecified on the
command line it still just uses whatever was configured in the
account.conf.


siwyd:
Mmm unsure what's wrong with the CI?

siwyd:
Thx!

Issue #2809, opened 2020-03-26
Add native notification support for Microsoft Teams


ccommisso:
With [Microsoft Teams becoming one of most popular](https://www.theverge.com/2019/7/11/20689143/microsoft-teams-active-daily-users-stats-slack-competition) (if not the most popular) enterprise messaging app it would be nice if there was a way to configure notifications to MS Teams in a native fashion (similar to the way you can already do with Slack).

Im sure there is a way to get Acme.sh to send notification to MS Teams using IFTTT but it would be nice if the functionally was provided out of the box.

Neilpang:
I never used Microsoft Teams.
Can you please show me the dev doc ?  That will save a lot of me time.

ccommisso:
[Main Webhook page](https://docs.microsoft.com/en-us/microsoftteams/platform/webhooks-and-connectors/what-are-webhooks-and-connectors)

I assume you would nee to create an [incoming connector](https://docs.microsoft.com/en-us/microsoftteams/platform/webhooks-and-connectors/how-to/add-incoming-webhook)

More [detailed technical info](https://docs.microsoft.com/en-us/microsoftteams/platform/webhooks-and-connectors/how-to/connectors-using)

If you are not using MS Teams I would be happy to validate this for you or help out any way I can.

Neilpang:
I never registered/used Microsoft Teams.

please do everything you can to help save time.

melkypie:
This should be closed as #2872 has been merged for a while now.

PR #2872, opened 2020-04-18
Add Microsoft Teams notify

honzahommer:
This PR adds support for sending webhooks to a Microsoft Teams channel requested in issue https://github.com/acmesh-official/acme.sh/issues/2809.

* Usage: https://github.com/acmesh-official/acme.sh/wiki/notify#11-set-notification-for-microsoft-teams

### DIFF
--- a/notify/teams.sh
+++ b/notify/teams.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env sh
+
+#Support Microsoft Teams webhooks
+
+#TEAMS_WEBHOOK_URL=""
+#TEAMS_THEME_COLOR=""
+#TEAMS_SUCCESS_COLOR=""
+#TEAMS_ERROR_COLOR=""
+#TEAMS_SKIP_COLOR=""
+
+teams_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
+  _debug "_statusCode" "$_statusCode"
+
+  TEAMS_WEBHOOK_URL="${TEAMS_WEBHOOK_URL:-$(_readaccountconf_mutable TEAMS_WEBHOOK_URL)}"
+  if [ -z "$TEAMS_WEBHOOK_URL" ]; then
+    TEAMS_WEBHOOK_URL=""
+    _err "You didn't specify a Microsoft Teams webhook url TEAMS_WEBHOOK_URL yet."
+    return 1
+  fi
+  _saveaccountconf_mutable TEAMS_WEBHOOK_URL "$TEAMS_WEBHOOK_URL"
+
+  TEAMS_THEME_COLOR="${TEAMS_THEME_COLOR:-$(_readaccountconf_mutable TEAMS_THEME_COLOR)}"
+  if [ -n "$TEAMS_THEME_COLOR" ]; then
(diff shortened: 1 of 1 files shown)


notify/teams.sh
```diff
@@ -0,0 +1,79 @@
...
+      _color="$TEAMS_SUCCESS_COLOR"
+      ;;
+    1)
+      _color="$TEAMS_ERROR_COLOR"
+      ;;
+    2)
+      _color="$TEAMS_SKIP_COLOR"
+      ;;
+  esac
+  _color="$(echo "${_color:-$TEAMS_THEME_COLOR}" | tr -cd '[:xdigit:]')"
```
Neilpang, line 61: can you not use `[:xdigit:]` ?

notify/teams.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env sh
+
+#Support Microsoft Teams webhooks
+
+#TEAMS_WEBHOOK_URL=""
+#TEAMS_THEME_COLOR=""
+#TEAMS_SUCCESS_COLOR=""
+#TEAMS_ERROR_COLOR=""
+#TEAMS_SKIP_COLOR=""
```
Neilpang, line 9: Are all these  COLORs required? or optional?

Maybe we can give some default values to make it more easy-to-use.

For example, default color for Success can be green or yellow.
Error color can be RED
Skip color can be green or yellow.

honzahommer: Colors are optional, I think we shouldn't set defaults.

notify/teams.sh
honzahommer: Replaced with `a-fA-F0-9`.

notify/teams.sh
Neilpang: why not?
The user can only set one TEAMS_WEBHOOK_URL variable. it will work out-of-the-box.

honzahommer: Done.

notify/teams.sh
```diff
@@ -0,0 +1,87 @@
...
+  fi
+
+  _data="{\"title\": \"$_subject\","
+  if [ -n "$_color" ]; then
+    _data="$_data\"themeColor\": \"$_color\", "
+  fi
+  _data="$_data\"text\": \"$_content\"}"
+
+  if _post "$_data" "$TEAMS_WEBHOOK_URL"; then
+    # shellcheck disable=SC2154
```
Neilpang, line 78: why disable SC2154 ?

honzahommer:
```sh
$ shellcheck notify/teams.sh

In notify/teams.sh line 78:
    if ! _contains "$response" error; then
                    ^-------^ SC2154: response is referenced but not assigned.

For more information:
  https://www.shellcheck.net/wiki/SC2154 -- response is referenced but not as...
```

Neilpang: see how we did that:
https://github.com/acmesh-official/acme.sh/blob/9190fdd42c5332f8821ce3f0de91cf0d18fa07d5/dnsapi/dns_cf.sh#L223-L231

honzahommer: You are right, I am idiot :-) Fixed.
```sh
  if response=$(_post "$_data" "$TEAMS_WEBHOOK_URL"); then
```

honzahommer:
Issue created https://github.com/acmesh-official/acme.sh/issues/2918.

PR #2879, opened 2020-04-21
Allow old and new API response for OPNsense Plugin

### LINKED ISSUE #2480: Report bugs to OPNsense dns API

blablup:
CLOSES #2480


blablup:
It is an or. so either it is ",\"allownotifyslave\":{\"\":{[^}]*}}" or it is only the ",". Use a ")?" would also work, I don't use that often, don't know why.

Issue #2878, opened 2020-04-21
[Bug] When using acme.sh in docker - can't deploy to Hashicorp Vault


pashinin:
Steps to reproduce
------------------

```
acme.sh  .....   --deploy --deploy-hook vault_cli
```

Log
-----------------

```
/root/.acme.sh/acme.sh: line 53: write: not found
[Tue Apr 21 06:18:10 UTC 2020] Error deploy for domain:<domain>
[Tue Apr 21 06:18:10 UTC 2020] Deploy error.
```

That happens on [this line](https://github.com/acmesh-official/acme.sh/blob/master/deploy/vault_cli.sh#L53) for me just because there is no vault binary in docker image.

I suggest using curl to make requests to Vault since it (curl) is already in Dockerfile.
So Vault deployment will work in Docker.

PS: I wonder why I didn't see [this message](https://github.com/acmesh-official/acme.sh/blob/master/deploy/vault_cli.sh#L48)

pashinin:
Was fixed in https://github.com/acmesh-official/acme.sh/pull/3111
For KV2 see other pull requests.

PR #2919, opened 2020-05-13
Use curl in vault_cli.sh, not vault binary

pashinin:
vault binary is not in a Docker image but curl is.
So it will be possible to do deployment from a Docker container.

https://github.com/acmesh-official/acme.sh/issues/2878


Neilpang:
this hook is named vault_cli.sh, so it's for vault cli.

If you want to not use  cli,  please create a new hook, name vault.sh

PR #2923, opened 2020-05-13
Add TransIP provider

mdbraber:
Adding support for TransIP since they recently switched to API v6 which is now fully REST-based (and easier to implement than the previous SOAP API). Just upon submitting this pull request I noticed that @biker-ruud has also written a TransIP implementation (https://github.com/biker-ruud/acme.sh) which is great! 

This implementation takes a bit more dynamic approach which I think works well in more circumstances:

- Dynamic domain name - you don't have to fix the domainname as it will look it up as supposed to from the command line using a `_get_root` function
- Uses a private key file to generate short time tokens. TransIP tokens only last for max 1 month and must otherwise be manually generated in the API dashboard which is a chore. Supplying your private key file using this implementation of the API takes care of generating the tokens.
- All parts from the API dev guide are implemented
- A fix is included for the main acme.sh to also allow for private keys starting with "BEGIN PRIVATE KEY" (which is also valid)

Happy to combine the best of both worlds and get this merged! @biker-ruud @Neilpang

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1003,7 +1003,7 @@ _sign() {
 
   _sign_openssl="${ACME_OPENSSL_BIN:-openssl} dgst -sign $keyfile "
 
-  if egrep -o "BEGIN( RSA)? PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
+  if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1; || if grep "BEGIN PRIVATE KEY" "$keyfile" >/dev/null 2>&1;
     $_sign_openssl -$alg | _base64
   elif grep "BEGIN EC PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
     if ! _signedECText="$($_sign_openssl -sha$__ECC_KEY_LEN | ${ACME_OPENSSL_BIN:-openssl} asn1parse -inform DER)"; then
--- a/acme.sh
+++ b/acme.sh
@@ -1003,7 +1003,7 @@ _sign() {
 
   _sign_openssl="${ACME_OPENSSL_BIN:-openssl} dgst -sign $keyfile "
 
-  if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1; || if grep "BEGIN PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
+  if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1 || grep "BEGIN PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
     $_sign_openssl -$alg | _base64
   elif grep "BEGIN EC PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
     if ! _signedECText="$($_sign_openssl -sha$__ECC_KEY_LEN | ${ACME_OPENSSL_BIN:-openssl} asn1parse -inform DER)"; then
--- a/acme.sh
+++ b/acme.sh
@@ -1003,7 +1003,7 @@ _sign() {
 
   _sign_openssl="${ACME_OPENSSL_BIN:-openssl} dgst -sign $keyfile "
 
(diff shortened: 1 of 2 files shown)


mdbraber:
I haven't (yet) updated the documentation in the Wiki awaiting merge for this, so users won't get confused between different versions.

acme.sh
```diff
@@ -1003,7 +1003,7 @@ _sign() {
 
   _sign_openssl="${ACME_OPENSSL_BIN:-openssl} dgst -sign $keyfile "
 
-  if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
+  if egrep -o "BEGIN( RSA)? PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
```
Neilpang, line 1006: why is this change necessary?

mdbraber: Trying to not repeat code I saw that there already was `_sign` function. The private key as being used by TransIP doesn't start with "BEGIN RSA PRIVATE KEY" but rather with "BEGIN PRIVATE KEY" (which is also valid), therefore - instead of incorporating a duplicate of the sign command I propose to also allow the `_sign` command to work for this type of key (and as it's only an extension it doesn't break anything)

mdbraber:
Created a bug report to report bugs for this provider #2949. @neilpang anything else that needs to be done before we can get this merged?

dnsapi/dns_transip.sh
```diff
@@ -0,0 +1,161 @@
...
+  if [ "$retcode" != "0" ]; then
+    _err "error $ep"
+    return 1
+  fi
+  _debug2 response "$response"
+  return 0
+}
+
+_transip_get_token() {
+  nonce=$(openssl rand -hex 12)
```
Neilpang, line 96: don't use `openssl` here.
If you just want a random number,  please use `_time()` or `_utc_date()` with `_digest()` method.

It could look like:
```
echo "TRSIP$(_time)" | _digest sha1 hex
```

acme.sh
Neilpang: don't use `egrep`, it's not portable.

please use 2  'if' conditions. like:


```
if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1; || if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1;
```

mdbraber: Happy to fix this. But there's also a _egrep_o() function in acme.sh on line 878 which uses egrep (which issued at various other places). I'm not tied to egrep in anyway so I'll make the changes if that works best.

dnsapi/dns_transip.sh
mdbraber: Done

acme.sh
Neilpang: `_egrep_o` is just for searching simple regex strings.  It's not reliable or robust enough.

mdbraber: Fixed

acme.sh
```diff
@@ -1003,7 +1003,7 @@ _sign() {
 
   _sign_openssl="${ACME_OPENSSL_BIN:-openssl} dgst -sign $keyfile "
 
-  if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1; then
+  if grep "BEGIN RSA PRIVATE KEY" "$keyfile" >/dev/null 2>&1; || if grep "BEGIN PRIVATE KEY" "$keyfile" >/dev/null 2>&1;
```
Neilpang, line 1006: "then" is missing

Neilpang:
try to issue a new cert after you finish the changes. just to make sure the change doesn't break anything.

mdbraber:
- All fixes implemented
- Passes all the checks / CI builds
- No unnecessary dependencies
- Issuing a new certificate succeeded with all changes applied:

`[Tue May 19 23:07:39 CEST 2020] Cert success.`

LGTM :-) @Neilpang

biker-ruud:
I've changed the wiki documentation according to this implementation.

PR #2961, opened 2020-05-25
Fix broken grep so that One984HOSTING_COOKIE actually gets set, and isn't left empty.

DerVerruckteFuchs:
In `_1984hosting_login()`, the login succeeds just fine. However, `One984HOSTING_COOKIE` fails to get set. This is because the HTTP header has `set-cookie` in all lower case, while the grep is looking for `Set-Cookie` with capitalization. As a result, `One984HOSTING_COOKIE` is left empty, since the output for grep is empty. Correcting the case for the grepped string to lower case results in grep actually finding something, and thus `One984HOSTING_COOKIE` gets set properly.

### DIFF
--- a/dnsapi/dns_1984hosting.sh
+++ b/dnsapi/dns_1984hosting.sh
@@ -168,7 +168,7 @@ _1984hosting_login() {
   _debug2 response "$response"
 
   if [ "$response" = '{"loggedin": true, "ok": true}' ]; then
-    One984HOSTING_COOKIE="$(grep '^Set-Cookie:' "$HTTP_HEADER" | _tail_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
+    One984HOSTING_COOKIE="$(grep '^set-cookie:' "$HTTP_HEADER" | _tail_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
     export One984HOSTING_COOKIE
     _saveaccountconf_mutable One984HOSTING_COOKIE "$One984HOSTING_COOKIE"
     return 0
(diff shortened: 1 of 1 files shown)


dnsapi/dns_1984hosting.sh
```diff
@@ -168,7 +168,7 @@ _1984hosting_login() {
   _debug2 response "$response"
 
   if [ "$response" = '{"loggedin": true, "ok": true}' ]; then
-    One984HOSTING_COOKIE="$(grep '^Set-Cookie:' "$HTTP_HEADER" | _tail_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
+    One984HOSTING_COOKIE="$(grep '^set-cookie:' "$HTTP_HEADER" | _tail_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
```
Neilpang, line 171: why not use `"grep -i "` option?

DerVerruckteFuchs: I was assuming this was a permanent change in how HTTP headers behaved and didn't think much of it. After looking at it more deeply though, using `grep -i` is probably more appropriate due to the differences between HTTP/1.1 and HTTP/2. I've been getting HTTP/2 headers where lower case header fields are the default, but HTTP/1.1 is case insensitive. Any combination of upper/lower case is valid for HTTP/1.1, but could be invalid for HTTP/2. Using `grep '^SET-COOKIE:\|Set-Cookie:\|Set-cookie:\|set-cookie:'` could probably be fine for more sane looking header fields. It would just break again if it were anything else though.

DerVerruckteFuchs: I went ahead and took care of it.

PR #2964, opened 2020-05-27
To work in cronjob

vitaly80:
${_domain} was empty in cronjob. Replace it by fulldomain without _acme-challenge. at begin string.

Now CronJob works perfectly excellent for me. Thanks all.

### DIFF
--- a/dnsapi/dns_regru.sh
+++ b/dnsapi/dns_regru.sh
@@ -14,6 +14,8 @@ REGRU_API_URL="https://api.reg.ru/api/regru2"
 dns_regru_add() {
   fulldomain=$1
   txtvalue=$2
+  mydomain=${fulldomain#*_acme-challenge.}
+  _info "Name of domain ${mydomain}"
 
   REGRU_API_Username="${REGRU_API_Username:-$(_readaccountconf_mutable REGRU_API_Username)}"
   REGRU_API_Password="${REGRU_API_Password:-$(_readaccountconf_mutable REGRU_API_Password)}"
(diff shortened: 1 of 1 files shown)


dnsapi/dns_regru.sh
```diff
@@ -14,6 +14,8 @@ REGRU_API_URL="https://api.reg.ru/api/regru2"
 dns_regru_add() {
   fulldomain=$1
   txtvalue=$2
+  mydomain=${fulldomain#*_acme-challenge.}
```
Neilpang, line 17: you should use the api to detect which part is the root zone.

https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide#3-detect-which-part-is-your-root-zone

vitaly80: Hello, Neil! But in line 33 

`%22subdomain%22:%22_acme-challenge%22`

That is therefore I thinking, that root = fulldomain - subdomain (in this case _acme-challenge).

Neilpang: That was a mistake,   now is the time to fix it.

Neilpang: any update ?

vitaly80: > 
> 
> any update ?

Hello. I'm sorry, but I don't find how to got subdomain and rootzone from fulldomain. I look in https://www.reg.ru/support/help/api2 (russian) https://www.reg.com/support/help/api2 (english). There, requests are executed based on the domain name. 

I can try to ask a question in tech support of REGRU company how to do this.

Issue #2970, opened 2020-05-30
Understanding the acme.sh data model wrt account email address


pini-gh:
Hi,

This is not a bug report but a question to @Neilpang.

I'm trying to follow up on the [initial work by @buchdag](https://github.com/nginx-proxy/docker-letsencrypt-nginx-proxy-companion/issues/510) to use `acme.sh` instead of `simp_le` for `letsencrypt-nginx-proxy-companion`.
As I undertand it:
* An `acme.sh` configuration directory is tied to one and only one email address
* An `acme.sh` configuration directory can hold several accounts for different ACME service providers. These accounts are configured under `<acme.sh_conf_dir>/ca/<service_provider_fqdn>`.
Then, say I want to update the email address for a given `acme.sh` configuration directory, I'd need to issue an `acme.sh --update-account` command for each configured ACME service provider. This command would look like:
```
acme.sh --update-account --server <service_provider_directory_url> --accountemail <new_email>
```
Is that correct? In this case it would be convenient to have the service provider directory url stored somewhere into the account configuration,to be able to script the above command. Would you mind me submitting a PR to this end?

In case my understanding stated above is not correct, please tell me.

Thanks in advance,

_g.

Neilpang:
Yes, correct.
But I won't accept the PR for just the case that you want to enumerate all the  CA to update email.

pini-gh:
Ack. Closing this issue as the question was answered.

PR #2971, opened 2020-05-30
Store ACME_DIRECTORY into ca.conf

pini-gh:
See issue #2970.
The use case is when one wants to script an account email address change.
Something like:
```
  for ca_conf in <acme.sh_conf_home>/ca/*/ca.conf; do
    eval $(grep '^ACME_DIRECTORY=' "$ca_conf")
    acme.sh --update-account --config_home <acme.sh_conf_home> --server "$ACME_DIRECTORY" --accountemail <new_email>
  done
```

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3464,6 +3464,7 @@ _regAccount() {
     return 1
   fi
   _savecaconf "ACCOUNT_URL" "$_accUri"
+  _savecaconf "ACME_DIRECTORY" "$ACME_DIRECTORY"
   export ACCOUNT_URL="$_accUri"
 
   CA_KEY_HASH="$(__calcAccountKeyHash)"


pini-gh:
Neil (in #2970):
> But I won't accept the PR for just the case that you want to enumerate all the CA to update email.

This is not just about my use case. This is about fixing a currently broken data model.

Neilpang:
what "data model" ?

pini-gh:
> what "data model" ?

Rephrasing #2970:
* An acme.sh configuration directory is tied to one and only one email address
* An acme.sh configuration directory can hold several accounts for different ACME service providers. These accounts are configured under <acme.sh_conf_dir>/ca/<service_provider_fqdn>.
* In case one need to update the default configuration email, there is no way to retrieve the acme_directory setting tied to each account to accomplish this task.

I need this information for [my fork of docker-letsencrypt-nginx-proxy-companion](https://github.com/pini-gh/docker-letsencrypt-nginx-proxy-companion/tree/acme.sh) to work with acme.sh as Let's Encrypt client. Because this image allows default email changes and is expected to update the accounts under the default configuration directory automatically.

For now I patch acme.sh when building the docker image, but this is not ideal.

BTW this is a one line patch with no possible side effect at all.

pini-gh:
OK, I've found another way to workaround this issue: use 'default' config for empty account email only.

Keeping this PR open still, since other people could run into the same issue.

pini-gh:
Closing this PR now, as I'm deleting my fork.

PR #2979, opened 2020-06-04
make sure the private key is never world-readable

vszakats:
1. Set the private key to chmod 0600 on creation.
2. Avoid possible race condition by setting the private key backup to 0600 _first_ and then saving the key.
3. Reset existing backup key to chmod 0600 before saving the new key.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1076,6 +1076,8 @@ _createkey() {
     fi
   fi
 
+  chmod 600 "$f"
+
   if _isEccKey "$length"; then
     _debug "Using ec name: $eccname"
     if _opkey="$(${ACME_OPENSSL_BIN:-openssl} ecparam -name "$eccname" -genkey 2>/dev/null)"; then
@@ -5265,10 +5267,12 @@ _installcert() {
       cp "$_real_key" "$_backup_path/key.bak"
     fi
     if [ -f "$_real_key" ]; then
+      chmod 600 "$_real_key"
       cat "$CERT_KEY_PATH" >"$_real_key" || return 1
     else
-      cat "$CERT_KEY_PATH" >"$_real_key" || return 1
+      touch "$_real_key"
       chmod 600 "$_real_key"
+      cat "$CERT_KEY_PATH" >"$_real_key" || return 1
     fi
   fi
 


vszakats:
Is there any interest in (or concern with) this?

logankaser:
Just another user, but this seems in line with best practice for storing keys on Unix, AWS even enforces this.
I also don't see any downsides with this change, and your implementation seems to correctly handle the possible race condition.

acme.sh
```diff
@@ -5265,10 +5267,12 @@ _installcert() {
       cp "$_real_key" "$_backup_path/key.bak"
     fi
     if [ -f "$_real_key" ]; then
+      chmod 600 "$_real_key"
```
Neilpang, line 5270: don't  chmod if the key file is already existing.

acme.sh
```diff
@@ -5265,10 +5267,12 @@ _installcert() {
...
       cat "$CERT_KEY_PATH" >"$_real_key" || return 1
     else
-      cat "$CERT_KEY_PATH" >"$_real_key" || return 1
+      touch "$_real_key"
```
Neilpang, line 5273: why do we need  to "touch" it first ?

Neilpang: I don't think it necessary.

vszakats: Touch is used to create a zero-length file. Without this zero-length file, the subsequent `chmod` command would fail.

acme.sh
vszakats: One of the goal of this patch is to avoid leaking the private key via wrongly set file mode. Without this command, the new private key saved in the subsequent `cat` command may end up in a file that's world readable (if someone set them that way between two `acme.sh` runs). If somebody is intentionally willing to leak the private key, the "best" solution is to copy it somewhere else, but IMO `acme.sh`'s correct behaviour is to always make sure that a saved private keys don't end up in a world-readable file.

Neilpang: >> (if someone set them that way between two acme.sh runs). If somebody is intentionally willing to leak the private key, the "best" solution is to copy it somewhere else,

No, we should chmod only if the file doesn't exist.  If the user changed the chmod, we should respect it.

acme.sh
Neilpang: the "cat" command can create the "$_real_key" file.  No need to "touch"

logankaser: The issue is that cat by itself cannot create a file with a specific mask, you can, of course, use `umask`
but touching then modifying the mask before writing to the file is another somewhat common way to handle this that avoids the problems with `umask`.
The reason this is done is to avoid the race condition where if you write the key to the file and then set the mask, an attacker could read the file after it has had the key written to it, but before the mask has been changed.

vszakats: The point is to set the mask _before_ writing the secret content. If you set it after finishing writing the file, there will always be a moment when the file content is world-readable. So, e.g. if someone knows the file location and hammers that path with reads in a loop, this can be exploited. This patch solves this by first creating the file with the secure mask and only then start outputting the secret content into it.

acme.sh
vszakats: You assume that the user does this intentionally and that the user is an actual person. I'd argue this is easier done by mistake by a human user or done by buggy tools. `acme.sh` is a tool designed to create a secure and unattended way to update and handle private keys, so I think it's rightly expected that is enforces security around the files it manages. Also note that the user is not expected to tamper with files managed by `acme.sh` (and similar automated tools.)

Neilpang: If the key file doesn't exist,  we create it and chmod to 600. That's ok.
But if the key file already exists,  we should not chmod to the file.  We MUST respect the user's choice.

vszakats: Okay, I won't argue just closing this and move on to a secure client.

vszakats: Some final thoughs in case somebody is interested in securing this ACME client:

1. If this project keeps insisting on "respecting user's choice" for private key file modes, a subsequent fix for the world-readable issue will NEVER be fixed for existing installations, only for new ones, or for newly created certs. The reason is that at this moment `acme.sh` creates files with the wrong (insecure) file mode, which will be kept as-is, even though it was never a user's choice, but a result to buggy (insecure) `acme.sh` behavior. Also, any time in the future there will be a bug in `acme.sh` that results in a wrong file mode or similar, it will similarly be never fixed by subsequent updates, if this "rule" is followed.

2. Insisting on "respecting user's choice" for a file that is a) not created by the user b) managed automatically by this tool c) not supposed to be touched by the user: makes no sense.

3. Respecting user's choice makes absolute sense for files created, edited and managed by users. `acme.sh`-managed private keys aren't such files.

4. If there are `acme.sh` users who require their private keys to be world readable, or relied on the existing buggy behavior, either those environments should be fixed, or a wrapper script created that recreates the insecure state each time after running `acme.sh`. But, the default should be secure, not insecure as it is now.

Neilpang: I agree that the private keys file could be  "600",  but only if the file is created by acme.sh.

After we generate the key file, if the user wants to change the mod other than 600,  he could do "chmod" by hand.

When the cert is renewed in the future,  the user's "chmod" will be kept.


So, there is NO secure issue unless the user change the mod by his hand.  
If the user chmod on the key file, he takes responsibility.

All we do is to  keep the cert working when renewal without any damage to the user's system.

Neilpang: private key files are recommended to be 600,  but remember that it's just a normal file, it can be any mod.
What the mod is MUST meet the user's needs.
What meets the needs is the best choice.

Neilpang: what acme.sh does is never to make/change choices for the user. Instead, we respect the user's choice MOST.

vszakats: You're repeating this notion of respecting the user's choice, but you're ignoring all the issues I've mentioned above that arises from this, in this particular case, including the fact that this isn't a user created file, not is it a "normal file". Note that e.g. ssh will even fail with an error unless the file is 0600. `acme.sh` is lucky that most web servers aren't such picky about this.

To me it looks like `acme.sh` is purposefully designed to be insecure and for some unknown reason maintainers want to keep it that way.

Neilpang: >>  but you're ignoring all the issues I've mentioned above

Not true.  I didn't ignore the issues.  And I said I agree with you that the key file can be set to 600 when we first create the file.
The only argument between us is that:  whether to set the key file to 600 every time we renew the cert.

You insist to set to 600 every time, no matter the file is already existing or not.
My point is not to do so if the file is already existing.

And I don't agree with you:  acme.sh is secure enough.


>> acme.sh is lucky that most web servers aren't such picky about this.

you know why is this?  why most web servers aren't such picky?
There must be a balance between Secure and Convenience.  
100% secure is not always the best choice in all the cases.

vszakats: Okay, so I understand you confirmed that `acme.sh` is purposefully designed to be insecure, and there is no desire to fix this issue for existing installations and existing private keys.

vszakats:
Since this patch was rejected by maintainers, my solution was to disable further `acme.sh` automatic updates (via removing `AUTO_UPGRADE='1'` from `account.conf`, then apply this patch locally, till migrating to a different ACME client.

skyscooby:
Honestly.. this is the craziest thing I have read on GitHub.. A certificate management tool is unwilling to enforce the protection of the sacred private key file it owns? I just tested master out and it still creates brand new keys with world read access.

@vszakats I think you were wrong to close and delete this, you should have at least left it open as there was willingness to at least have 'new' keys being created correctly.. which I totally agree is not enough but it is also a improvement for new users of the tool.

@Neilpang based on your questions in the review it's pretty clear that you are not qualified to be reviewing a security oriented project.

vszakats:
@skyscoopy Unclosed issues were stacking up on one's profile page (at least back then, not sure if still true), so I used to close things that don't get interest/traction (or outright rejected) to declutter. The problem still seems valid indeed though. Definitely not a happy resolution, because even though you can choose a different client (certbot does support ECC by now), but acme.sh is being used as a dependency by other projects, so it's a security issue one can fall into without being aware.

I'd gladly reopen, but the option isn't available anymore.

vszakats:
@skyscooby Unclosed issues were stacking up on one's profile page (at least back then, not sure if still true), so I used to close things that don't get interest/traction (or outright rejected) to declutter. The problem still seems valid indeed though. Definitely not a happy resolution, because even though you can choose a different client (certbot does support ECC by now), but acme.sh is being used as a dependency by other projects, so it's a security issue one can fall into without being aware.

I'd gladly reopen, but the option isn't available anymore.

vszakats:
Somebody reported this again a few weeks after this PR and this was then merged for the most part (but without referencing or crediting this one). Ref: https://github.com/acmesh-official/acme.sh/issues/3127 af193291faf5b9765feedaae12f237b0f16fa518

Neilpang:
here is one of my test cert,  the key file is 600:
image

vmmello:
This issue was fixed on release 3.0.4, by commit https://github.com/acmesh-official/acme.sh/commit/af193291faf5b9765feedaae12f237b0f16fa518, as mentioned above by @vszakats.

(this comment is almost unnecessary, but as this fix was never explicitly acknowledged anywhere, users of older versions keep reporting it, so it's confusing whether it was ever fixed or not).

Issue #3008, opened 2020-06-27
Add --key-chain option


Delicates:
To avoid race conditions, Postfix (one of the most popular email servers) requires certificates to be provided in a single unencrypted PEM file that contains both the private key and full certificate chain.

See: http://www.postfix.org/TLS_README.html

Can we please add a new --key-chain option that behaves similarly to --key-file (root ownership with 0600 permissions), but outputs an unencrypted PEM file with (key, cert, chain) sequence.

**Note:** --toPkcs option may be related, but it requires a password, see issue: #2681

Neilpang:
you can specify the same for "cert-file" and "key-file" and "ca-file":

```
acme.sh --installcert  --key-file  /path/to/all-in-one.pem  --cert-file  /path/to/all-in-one.pem   --ca-file  /path/to/all-in-one.pem
```

Delicates:
@honzahommer Thanks a lot for dropping the patch!

It needs some minor corrections.

As per the Postfix documentation linked above it requires:

> One or more chain files that **start with a key** that is immediately followed by the corresponding certificate and any additional issuer certificates.

So the output needs to be reordered:

```
--- acme.sh     2020-07-04 13:18:45.111533348 +1000
+++ acme.sh.new 2020-07-04 13:39:35.740547101 +1000
@@ -5303,7 +5303,7 @@
     if [ ! -f "$_real_fullchain_key" ]; then
       install -c -m600 /dev/null "$_real_fullchain_key"
     fi
-    cat "$CERT_FULLCHAIN_PATH" "$CERT_KEY_PATH" >"$_real_fullchain_key" || return 1
+    cat "$CERT_KEY_PATH" "$CERT_FULLCHAIN_PATH" >"$_real_fullchain_key" || return 1
   fi
 
   if [ "$_reload_cmd" ]; then
```

The `$_real_key` uses `chmod`, while `$_real_fullchain_key` uses `install` to set file permissions.
I'd suggest to make them consistent.

Also why do we have `$CERT_FULLCHAIN_PATH` file but don't have `$CERT_FULLCHAIN_KEY_PATH` file?
I'd suggest creating it for completeness.

@Neilpang Different server software requires different cert files provided to them, and we need to be careful around how they are managed. After specifying the same cert output file, I don't want some other server to serve full chain with private key to clients. Similarly I don't want to have to execute `acme.sh` twice to generate different cert files required by different server software.

Neilpang:
[...]

PR #3011, opened 2020-06-27
Add `--fullchain-key-file` option to `--install-cert` command

honzahommer:
As requested in https://github.com/acmesh-official/acme.sh/issues/3008, `--fullchain-key-file` option (contains fullchain cert and private key with chmod 600) to `--install-cert` command added.


Neilpang:
no,  we won't need this.  use `--reloadcmd` to append the key to fullchain.

Issue #2850, opened 2020-04-10
Trivial grammar error "doesn't exists"


TonyGravagno:
When singular items are missing we say "does not exist" or doesn't exist".
When singular items are present we say "does exist" or "exists".
Yes, English can be difficult.
https://github.com/acmesh-official/acme.sh/search?q=%22doesn%27t+exists%22&unscoped_q=%22doesn%27t+exists%22
Text says "doesn't exists".

Neilpang:
would you mind sending a PR for this ?

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

neilpang:
Fixed in eca57bee: "doesn't exists"/"does not exists" grammar corrected in acme.sh.

Issue #2849, opened 2020-04-10
Trivial spelling error: MajEr to MajOr


TonyGravagno:
https://github.com/acmesh-official/acme.sh/search?q=majer&unscoped_q=majer
apacheMajer variable should be apacheMajor

Neilpang:
it's a typo. would you mind a PR ?

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

neilpang:
Fixed in 94787d53: apacheMajer renamed to apacheMajor.

PR #3015, opened 2020-06-29
Trivial text changes

TonyGravagno:
Ref #2850. Text-only changes in literals and comments. There is one example where text expected from a remote query should not be corrected.
There might be confusion about how I forked this. I forked from base 'master' to my repo 'dev', and GH did its own update from base 'dev' into my dev. So this should all be based on 'dev'. From there my commitments went into 'dev'. I believe this is all correct, but please check anyway. Thanks.


TonyGravagno:
I botched the commits and PR. This PR now also contains a simple change for #2849.

PR #3037, opened 2020-07-10
adding Acme2certifier to supported CA list

flosch-dev:
May you consider to add [Acme2certifier ](https://github.com/grindsa/acme2certifier) to list of supported CA. We use acme.sh to run tests against our acme-server implementation. Also within PR #3033 we fixed an issue with ECC account keys we found during such tests.


Neilpang:
Sorry,  this is not a valid CA.

PR #3050, opened 2020-07-15
dns_dnsmasq

rodvlopes:
My use case is that I am running my own dns server, it doesn't have an API, but I can edit its zone file. I tested it for generating a certificate for my server.


summershrimp:
+1

mxvin:
Thanks :+1:

PR #3064, opened 2020-07-18
adjust to changes in Selectel's API

VadimDor:
Hi! For some reasons it works with Selectel.RU CDN only with changes provided. Without these changes it terminates with errors or hangs. May be due to some changes in Selectel's API since last commit in official repo 3 years ago.  Please consider to include changes in master branch to enable auto-upgrade feature in configurations files. Thanks.

### DIFF
--- a/dnsapi/dns_selectel.sh
+++ b/dnsapi/dns_selectel.sh
@@ -104,7 +104,7 @@ dns_selectel_rm() {
 # _domain=domain.com
 # _domain_id=sdjkglgdfewsdfg
 _get_root() {
-  domain=$1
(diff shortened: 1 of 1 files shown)


dnsapi/dns_selectel.sh
```diff
@@ -104,7 +104,7 @@ dns_selectel_rm() {
 # _domain=domain.com
 # _domain_id=sdjkglgdfewsdfg
 _get_root() {
-  domain=$1
+  domain="$(echo "$1" | idn -u)"
```
Neilpang, line 107: use `_idn`

PR #3069, opened 2020-07-23
Fixes #3068 - Adds docker multi-arch support

### LINKED ISSUE #3068: Adding Docker multi-arch support
I haven't found any reference to this but as I currently require an arm64 Docker container of this, I noticed that currently this is not creating a Docker multi-arch images.

@Neilpang if this something that interests you, I'm willing to try to put this working via Github Actions (as they do allow to create and push multi-arch Docker images)

PedroLamas:
This adds multi-arch docker build support.

The whole thing is based on [this article](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/), but please be aware that some changes need to happen in order to make this work:

- You should disable the automatic builds in Docker Hub
- You need to get create an authentication token in Docker Hub and add it to the GitHub repo as a secret (this is explained in the above article)

After that everything else should just work fine!

### DIFF
--- a/.github/workflows/dockerhub.yml
+++ b/.github/workflows/dockerhub.yml
@@ -8,9 +8,19 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - name: trigger
-      run: curl -X POST https://hub.docker.com/api/build/v1/source/1813a660-2ee5-4583-a238-dd54e9a6ebac/trigger/c8cd9f1f-f269-45bc-9750-a08327257f62/call/
-    
-    
-
-
+      - name: checkout code
+        uses: actions/checkout@v2
+      - name: install buildx
+        id: buildx
+        uses: crazy-max/ghaction-docker-buildx@v1
+        with:
+          version: latest
+      - name: login to docker hub
+        run: |
+          echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
+      - name: build the image
+        run: |
+          docker buildx build \
+            --push \
+            --tag neilpang/acme.sh:latest \
+            --platform linux/arm64/v8,linux/amd64,linux/arm/v6,linux/arm/v7,linux/386 .


.github/workflows/dockerhub.yml
```diff
@@ -8,9 +8,19 @@ jobs:
...
+        with:
+          version: latest
+      - name: login to docker hub
+        run: |
+          echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
+      - name: build the image
+        run: |
+          docker buildx build \
+            --push \
+            --tag neilpang/acme.sh:latest \
```
Neilpang, line 25: is it possible to change the tag "latest" ?
If this is the master branch, we use "latest",  otherwise, we use the branch name("dev").

PedroLamas: I've now added that by having 2 separate build steps, one for "master" and one for "dev": the only difference between them is the label used in Docker that follows your specification!

Neilpang: I think GITHUB_REF env can do this job.
https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions

https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

PedroLamas: Indeed it can, however as you said you want "dev" -> "dev" and "master" -> "latest" so this was just an easier way to achive it!

If you prefer to do it via bash shell before, that can be achieved also, only reason I picked this is that it was easier and totally compounded (it guarantees that won't deploy any other branch than "dev" and "master")

PedroLamas: @Neilpang as it stands, this PR will work for "latest" and "dev" tagging of the Docker image, but if you want I can make the changes necessary to use the bash style setting of the tag instead... what do you say?

PedroLamas:
@Neilpang is there something else I can help in regards to this PR?

.github/workflows/dockerhub.yml
```diff
@@ -8,9 +8,27 @@ jobs:
   build:
```
Neilpang, line 8: please add:

```
on:
  push:
    branches: [ master, dev ]
```

.github/workflows/dockerhub.yml
```diff
@@ -8,9 +8,27 @@ jobs:
...
+        with:
+          version: latest
+      - name: login to docker hub
+        run: |
+          echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
+      - name: build and push the image (master branch)
+        run: |
+          docker buildx build \
+            --push \
+            --tag neilpang/acme.sh:latest \
```
Neilpang, line 25: please use `${{github.ref}}`  instead of `latest` here.

.github/workflows/dockerhub.yml
```diff
@@ -8,9 +8,27 @@ jobs:
...
+            --platform linux/arm64/v8,linux/amd64,linux/arm/v6,linux/arm/v7,linux/386 .
+        if: ${{ github.ref == 'refs/heads/master' }}
```
Neilpang, line 27: remove all the code after this line.

.github/workflows/dockerhub.yml
PedroLamas: That's already above this line, I didn't change that on this PR!

.github/workflows/dockerhub.yml
PedroLamas: `${{github.ref}}` will have something like `refs/master` are you sure that is what you want here?

Neilpang: sorry, it was my mistake. it's ok.

.github/workflows/dockerhub.yml
Neilpang: forget this

.github/workflows/dockerhub.yml
```diff
@@ -8,9 +8,27 @@ jobs:
...
+      - name: build and push the image (dev branch)
+        run: |
+          docker buildx build \
+            --push \
+            --tag neilpang/acme.sh:dev \
```
Neilpang, line 32: please do not hardcode `dev` here.
I want to make it work for any branches other than "master".
I also need it work for a "Tag" for a "Release".

PedroLamas: in that case, would it make sense to just tag "latest" for master branch, and for other branches to tag with the branch name? Maybe even double tagging and adding the git sha?

Neilpang:
please also change the Dockerfile, to make `AUTO_UPGRADE` as an Build_Arg.

When we build for master branch(latest),  we set "AUTO_UPGRADE=1".    When we build a Tag or a Release, "AUTO_UPGRADE=" is empty.

.github/workflows/dockerhub.yml
Neilpang: Yes,   Only the master is tagged as "latest".
Any other branches will be tagged as branch name.
And for any release,  it will be tagged as release number, like "neilpang/acme.sh:2.8.7"

Neilpang: If it's easier, I'm ok to split the workflow files in to multiple ones.
One for master, one for branches, one for releases, etc.   As long as it's easier for you.

PedroLamas: I think it should be fine doing it on a single workflow, but I'm gonna need to try this on my fork first so I can test it out and then update this PR! I hope to have the changes done in the next hour or so!

Neilpang: Cool, take your time.   I hope it won't become complicated in just one single file.

PR #3080, opened 2020-08-01
Add deploy script for HP iLO

imgrant:
Use in conjunction with, e.g. https://gist.github.com/imgrant/ea576d21c7e849a8d39378bcfec99307, to generate a CSR directly from the iLO for use with `--signcsr` command.


imgrant:
Will open a new one after rebasing onto current main, it's been a while.

Issue #2091, opened 2019-02-13
Report bugs to Rackspace API


ianw:
For reporting bugs in the Rackspace DNS API

jjamfd:
The RackSpace DNS API only returns 100 domains at a time when you use the "list domains" endpoint.   This can cause the _get_root_zone() function to falsely return the "invalid domain" error. 

I recommend using the "domain search" endpoint instead: https://developer.rackspace.com/docs/cloud-dns/v1/api-reference/domains/#search-domains 

If you change line 97 as follows it works nicely for accounts with hundreds of domains:
if ! _rackspace_rest GET "$RACKSPACE_Tenant/domains/search?name=$h"; then

wilson18:
Rackspace uses different endpoints for different regions. E.g. UK uses https://lon.dns.api.rackspacecloud.com. This means that the current code does not work for people in the UK.

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

Issue #799, opened 2017-04-18
DNS API env variables are not per domain so you cant use different accounts for different domains


gitwer:
DNS API env variables are not able to be set per domain, meaning you can only use a single account for all domains. IE: you can't have 2 Cloudflare accounts one for example.com and a different account for other.com

Steps to reproduce
------------------
1) set variables for Cloudflare:
export CF_Key="sdfdxxxxxxxosdfgje"
export CF_Email="email@example.com"
2) issue a cert for example.com
acme.sh --issue -d example.com --dns dns_cf
3) install cert
acme.sh --install-cert -d example.com (etc etc etc)

4) set variables for DIFFERENT Cloudflare account:
export CF_Key="wgdsyyyyyyyyyykghtysd"
export CF_Email="email@other.com"
5) issue a cert for other.com
acme.sh --issue -d other.com --dns dns_cf
6) install cert
acme.sh --install-cert -d other.com (etc etc etc)

7) the .acme.sh/account.conf file structure does not work with/allow different DNS API variables for the same DNS provider for different domains.

Can the required DNA API variables (currently saved using "_saveaccountconf") be saved to the .acme.sh/example.com/example.com.conf file instead? This would allow different domains to use different accounts.

Debug log
-----------------
n/a

Neilpang:
yes, that is known limitation.

for most of the users, there is only one cloudflare account,   they just need to save the api key one time.  All the the domains will get benefit.

As a workaround,  you can run acme.sh with different `--config-home`.

```
acme.sh  --config-home "/root/myacme.sh2/"   --issue -d example.com  --dns dns_cf
```

you will need to add another cronjob to auto-renew the certs:
```
7 0 * * * "/home/neil/.acme.sh"/acme.sh --cron --home "/root/.acme.sh" --config-home "/root/myacme.sh2/" > /dev/null

```

FernandoMiguel:
I've made a Cloudflare PM promise me that they would have keys per domain, not full accounts.
I wish we could go as far as per subdomain / app.

so in the future, this is something @Neilpang  might have to address

Neilpang:
@FernandoMiguel 
Yes,  I will think about it.

gitwer:
Thanks @Neilpang and good to know @FernandoMiguel 
The different --config-home's will work for me (save me creating a different account to run it from or hacking the script). DIdn't think of that (then again, it is bed time!)

xBytez:
[...]

PR #3085, opened 2020-08-04
DNSAPI per-domain conf wrapper funcs

senseisimple:
This is meant as an example to address Per-domain api auth setting #799 and related improvement to #2091

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2229,7 +2229,41 @@ _readaccountconf_mutable() {
 _clearaccountconf() {
   _clear_conf "$ACCOUNT_CONF_PATH" "$1"
 }
-
+#_readdnsapiconf key
+_readdnsapiconf() {
+  local acctmutcnf=$(_readaccountconf_mutable $1)
+  local acctcnf=$(_readaccountconf $1)
+  local domcnf=$(_readdomainconf "SAVED_$1")
+  if [ -n "$domcnf" ]; then
+    dnsapi_auth_conf_loc=domainconf
+    echo -ne "$domcnf"
+  elif [ -n "$acctmutcnf" ]; then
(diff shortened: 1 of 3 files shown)


senseisimple:
COMMIT 1  https://github.com/acmesh-official/acme.sh/pull/3085/commits/3df276c25e23cf7e2e270cecb3f076cb9a9c83df (dns_rackspace.sh) Shows a simplistic and functional update to a single dnsapi script using existing read/save domainconf functions in acme.sh

COMMIT 2   https://github.com/acmesh-official/acme.sh/pull/3085/commits/716b9b8aad3581809de0f532ef988237ad785fb3 Provides a solution and migration path for existing dnsapi scripts to eliminate issues switching from accountconf storage to domainconf storage allowing for backward compatibility with existing configurations. It updates acme.sh and the dns_cf.sh scripts, which would also necessitate an update to the wiki page at https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide

hknet:
Would you consider an option to override domainconf by accountconf in case user/key/password/secret has to be changed on a bigger scale (account hacked/...)?

Neilpang:
Yes, we need a unique way to do that.

We should save them into account.conf by default to keep compatibility.

```sh
export xxxx=xxx
export cccc=ddddd

acme.sh --issue xxxxx   --dns  dns_cf
```

We need another new option to save it in domain conf

```sh

export xxx=xxx
export ccc=ddd

acme.sh --issue  xxxx    --dns dns_cf    --dns-save-per-cert
```

When renewal,  we should check domain conf first, then fallback to accont.conf.

I think we need too functions:

```
#  name   value
_savednsconf () {

}

# name
_readdnsconf() {

}
```

They should work similarly to the deplohhook functions.

PR #3097, opened 2020-08-09
change shebang if env is not in the default location.  fix https://gi…

kadirselcuk:
…thub.com/Neilpang/acmetest/issues/3


kadirselcuk:
Acme test

PR #3184, opened 2020-09-17
add dns api for websupport

trgosk:
Hello,
please merge new DNS API for Slovak DNS provider WEBSUPPORT. Its tested on trgo.sk domain + spellchecked as suggested.

I'm going to add to WIKI page about DNS APIs.

Thanks for the acme.sh project, its very nice.
trgo.sk

### DIFF
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -170,5 +170,3 @@ jobs:
         cd ../acmetest
         ./letest.sh
 
-
-        
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -152,21 +152,11 @@ jobs:
     - name: Run acmetest
       shell: bash
       run: |
-        if [ "${{ secrets.TokenName1}}" ] ; then
-          export ${{ secrets.TokenName1}}=${{ secrets.TokenValue1}}
-        fi
-        if [ "${{ secrets.TokenName2}}" ] ; then
-          export ${{ secrets.TokenName2}}=${{ secrets.TokenValue2}}
-        fi
-        if [ "${{ secrets.TokenName3}}" ] ; then
-          export ${{ secrets.TokenName3}}=${{ secrets.TokenValue3}}
-        fi
-        if [ "${{ secrets.TokenName4}}" ] ; then
-          export ${{ secrets.TokenName4}}=${{ secrets.TokenValue4}}
-        fi
-        if [ "${{ secrets.TokenName5}}" ] ; then
-          export ${{ secrets.TokenName5}}=${{ secrets.TokenValue5}}
-        fi
+      if NOT [${{secrets.TokenName1}}] == [] set ${{secrets.TokenName1}}=${{secrets.TokenValue1}}
+      if NOT [${{secrets.TokenName2}}] == [] set ${{secrets.TokenName2}}=${{secrets.TokenValue2}}
+      if NOT [${{secrets.TokenName3}}] == [] set ${{secrets.TokenName3}}=${{secrets.TokenValue3}}
+      if NOT [${{secrets.TokenName4}}] == [] set ${{secrets.TokenName4}}=${{secrets.TokenValue4}}
+      if NOT [${{secrets.TokenName5}}] == [] set ${{secrets.TokenName5}}=${{secrets.TokenValue5}}
         cd ../acmetest
         ./letest.sh
 
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -152,12 +152,23 @@ jobs:
     - name: Run acmetest
       shell: bash
       run: |
-      if NOT [${{secrets.TokenName1}}] == [] set ${{secrets.TokenName1}}=${{secrets.TokenValue1}}
-      if NOT [${{secrets.TokenName2}}] == [] set ${{secrets.TokenName2}}=${{secrets.TokenValue2}}
-      if NOT [${{secrets.TokenName3}}] == [] set ${{secrets.TokenName3}}=${{secrets.TokenValue3}}
-      if NOT [${{secrets.TokenName4}}] == [] set ${{secrets.TokenName4}}=${{secrets.TokenValue4}}
-      if NOT [${{secrets.TokenName5}}] == [] set ${{secrets.TokenName5}}=${{secrets.TokenValue5}}
+        if [ "${{ secrets.TokenName1}}" ] ; then
+          export ${{ secrets.TokenName1}}=${{ secrets.TokenValue1}}
+        fi
+        if [ "${{ secrets.TokenName2}}" ] ; then
+          export ${{ secrets.TokenName2}}=${{ secrets.TokenValue2}}
+        fi
+        if [ "${{ secrets.TokenName3}}" ] ; then
+          export ${{ secrets.TokenName3}}=${{ secrets.TokenValue3}}
+        fi
+        if [ "${{ secrets.TokenName4}}" ] ; then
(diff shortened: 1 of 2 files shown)


.github/workflows/DNS.yml
```diff
@@ -152,21 +152,11 @@ jobs:
...
-          export ${{ secrets.TokenName4}}=${{ secrets.TokenValue4}}
-        fi
-        if [ "${{ secrets.TokenName5}}" ] ; then
-          export ${{ secrets.TokenName5}}=${{ secrets.TokenValue5}}
-        fi
+      if NOT [${{secrets.TokenName1}}] == [] set ${{secrets.TokenName1}}=${{secrets.TokenValue1}}
+      if NOT [${{secrets.TokenName2}}] == [] set ${{secrets.TokenName2}}=${{secrets.TokenValue2}}
+      if NOT [${{secrets.TokenName3}}] == [] set ${{secrets.TokenName3}}=${{secrets.TokenValue3}}
+      if NOT [${{secrets.TokenName4}}] == [] set ${{secrets.TokenName4}}=${{secrets.TokenValue4}}
+      if NOT [${{secrets.TokenName5}}] == [] set ${{secrets.TokenName5}}=${{secrets.TokenValue5}}
```
Neilpang, line 159: don't change dns.yml.
It's in cygwin bash, not in cmd.

Neilpang:
please enable github actions on your fork.
And read: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,255 @@
...
+    _debug2 "GET $WS_Api${pa}"
+    response="$(_get "$WS_Api${pa}")"
+  fi
+
+  _debug2 response "$response"
+  return "$?"
+}
+
+_hmac_value() {
+  alg="$1"
```
Neilpang, line 200: don't copy the implementation here.
you can just use `_hmac()`  function instead.
remove this function.

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,255 @@
...
+  for i in $(echo "$va" | sed "s/{/ /g"); do
+    if _contains "$i" "$fi"; then
+      echo "$i"
+      break
+    fi
+  done
+}
+
+_time2str_iso() {
+  #BSD
```
Neilpang, line 235: don't copy the implementation.

dnsapi/dns_websupport.sh
trgosk: my __hmac_value()_ function isnt a simple copy of __hmac()_ funcion,
my __hmac_value()_ function accepts stdin and passes it to openssl.

I made a simple test with these parameters:
key="2c7f93ed-8661-42dc-a1b8-61ea685462d9"
value="GET /v1/user/self 1600100313"

verified result by calling my __hmac_value()_ `"$(printf "%s" "$value" | _hmac_value sha1 "$key" hex)"` is 0a460c49bd0fd3c5854862e35554da072f6a8ea4

by calling the original __hmac()_ `"$(printf "%s" "$value" | _hmac sha1 "$key" hex)"` result is different 112222775ba571715038fa13bc696712b7999164 and the API call will fail

trgosk: hopefully doing the right thing i removed the _hmac_value() function by replacing by a inline call

dnsapi/dns_websupport.sh
trgosk: hopefully doing the right thing i removed the _time2str_iso() function by replacing by a inline call

akulumbeg:
Hello, fellow Websupport user here (and I think there are more than trgosk and I). 
I would like to see this merged. I was desperately looking for a solution to hook up my Synology NAS with a DNS-challenge Let's Encrypt renewal and trgosk's implementation worked like a charm.

Just out of curiosity, after seeing that the checks did not pass in this particular Issue, I made a fork and replicated everything with my Repo Secrets and the only check it fails is Solaris. No idea how to debug stuff there and also I am not a shell expert, but since it passes everywhere else, couldn't that be just Solaris problem? 

It would really be a pity to let this go, Websupport is the largest webhosting company in Slovakia, and prominent also in neighboring countries, so I believe, there are definitely some more people who would benefit from this.

Neilpang:
you can get a solaris shell here to debug:

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,208 @@
...
+
+_ws_rest() {
+  me=$1
+  pa="$2"
+  da="$3"
+
+  _debug2 api_key "$WS_ApiKey"
+  _debug2 api_secret "$WS_ApiSecret"
+
+  timestamp="$(date +%s)"
```
Neilpang, line 159: please use `_time()` function instead.

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,208 @@
...
+  datez=$(date -u -r "$timestamp" +%Y-%m-%dT%H:%M:%S%z 2>/dev/null || date -u -d@"$timestamp" +%Y-%m-%dT%H:%M:%S%z)
```
Neilpang, line 160: please use `_utc_date()` function instead

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,208 @@
...
+  canonical_request="${me} ${pa} ${timestamp}"
+  alg="sha1"
+  signature_hash=$( (printf "%s" "$canonical_request" | ${ACME_OPENSSL_BIN:-openssl} dgst -"$alg" -mac HMAC -macopt "key:$WS_ApiSecret" 2>/dev/null || printf "%s" "$canonical_request" | ${ACME_OPENSSL_BIN:-openssl} dgst -"$alg" -hmac "$(printf "%s" "$WS_ApiSecret" | _h2b)") | cut -d = -f 2 | tr -d ' ')
```
Neilpang, line 163: don't use `ACME_OPENSSL_BIN`.
please use `_hmac()` function instead

dnsapi/dns_websupport.sh
akulumbeg: Fixed, works as supposed

dnsapi/dns_websupport.sh
akulumbeg: Fixed (hopefully right), seems to be working at my end, passing the checks.

dnsapi/dns_websupport.sh
akulumbeg: A question here - the _utc_date() function provides the "+%Y-%m-%d %H:%M:%S" output, but the ISO8601 output is needed for the API, created by "+%Y-%m-%dT%H:%M:%S%z" (mind the "T" and "%Z"). Unless I change the whole function in the acme.sh file, there is no other way around this... I think?

akulumbeg: Also, I believe I figured out the issue with Solaris - the "datez" variable is not created, so this happens:

[Sunday, January 31, 2021 at 10:40:04 PM UTC] timestamp='1612132804'
[Sunday, January 31, 2021 at 10:40:04 PM UTC] datez
...
[Sunday, January 31, 2021 at 10:40:04 PM UTC] H4='Date: '
...
[Sunday, January 31, 2021 at 10:40:05 PM UTC] response='{"message":"Missing date header.","code":400}'


of course, the H4 should look for example like 'Date: 2021-01-31T19:24:24+00:00'
But for some reason, Solaris is very picky about this. Do you have any idea, from experience, why this is happening?

PR #3198, opened 2020-09-30
dsn_nsd: Increment serial number

snimmagadda:
Let's Encrypt could validate against any authoritative server and without
a bumped serial number, secondary servers never get a chance to update
TXT records.


Neilpang:
don't use awk

snimmagadda:
Thanks, closing this PR and will think of solving it without awk.

PR #3205, opened 2020-10-02
add deploy for PKCS#12/PFX certificate

SunMar:
Because one of my applications needs the certificate in PKCS#12 format I created a deploy script so also that application and automatically be updated with the new certificate.


CyrosX (approved):
Working, tested with HP printer

PR #3223, opened 2020-10-16
Use `hostedzonesbyname` Route 53 API endpoint instead of `hostedzones` endpoint.

michael-roger:
The `hostedzones` endpoint returns all hosted zones for a given Route 53 account in groups of 100.  For AWS Route 53 accounts with many domains, this could mean a large number of requests to the `hostedzones` endpoint as it progresses through each page of 100 results.  This will often result in a "Rate exceeded" API error from Route 53.

Instead of using `hostedzones` endpoint, we can use `hostedzonesbyname` and then filter by the specific domain we are looking for and ask for a `max-items` of 1.

The while loop in _get_root() starts with a given domain and removes parts from the front of the given domain if no match is found.

For example, when requesting a certificate for `test.www.domain.co.uk`, the while loop will check for Route 53 hosted zones for:

1st: test.www.domain.co.uk
2nd: www.domain.co.uk
3rd: domain.co.uk
4th: co.uk
5th: uk

The first two checks will result in no matches, while the third check should be successful (if, of course, domain.co.uk is actually a hosted zone in the given AWS account).

Now imagine that the given AWS account owns 2500 domains and, therefore, has 2500 hosted zones.

Using the `hostedzones` endpoint would result in:

1st: 25 GET requests to the Route 53 API looking for a match to test.www.domain.co.uk
2nd: 25 GET requests to the Route 53 API looking for a match to www.domain.co.uk
3rd: 25 GET requests to the Route 53 API looking for a match to domain.co.uk
4th: 25 GET requests to the Route 53 API looking for a match to co.uk
5th: 25 GET requests to the Route 53 API looking for a match to uk

This would far exceed the Route 53 limit of five requests per second.

Using `hostedzonesbyname` results in a dramatic reduction in Route 53 API GET requests for AWS accounts with large numbers of hosted zones.

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -12,7 +12,7 @@
 AWS_HOST="route53.amazonaws.com"
 AWS_URL="https://$AWS_HOST"
 
-AWS_WIKI="https://github.com/acmesh-official/acme.sh/wiki/How-to-use-Amazon-Route53-API"
+AWS_WIKI="https://github.com/Neilpang/acme.sh/wiki/How-to-use-Amazon-Route53-API"
 
 ########  Public functions #####################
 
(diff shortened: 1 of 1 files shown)


dnsapi/dns_aws.sh
```diff
@@ -152,50 +152,58 @@ dns_aws_rm() {
...
-            _err "Truncated request error."
-          fi
-        fi
-        #not valid
-        _err "Invalid domain"
-        return 1
-      fi
+  # Start with field 2 since each domain starts with _acme-challenge
+  # Example: _acme-challenge.www.domain.com
+  field=2
```
Neilpang, line 157: don't start with field 2. 
In dns alias mode,  there is no _acme-challenge prefix at all.

Neilpang:
please fix the CI errors.

mroger-vnn:
Will work on fixing the CI errors!

PR #3227, opened 2020-10-19
fix dnsapi/dns_1984hosting

phedoreanu:
This PR removes two private functions and the need to check their return codes with `$?`, which is bad practice.


phedoreanu:
The build passed in my fork, but I can't find a way of refreshing its status here. Any ideas @Neilpang ? 😀

phedoreanu:
Thanks!

PR #1690, opened 2018-06-23
Refactoring amcedns api

dampfklon:
Account data moved from account to domain config as this is the indented mode of operation for acmedns [ref ](https://github.com/joohoi/acme-dns/issues/76#issuecomment-385149688)

acmedns accounts are created on creation so no unnecessary manual work, but support for the old mode of operation.

PR #3231, opened 2020-10-24
Refactoring amcedns api (second try)

0xFelix:
See https://github.com/acmesh-official/acme.sh/pull/1690

I picked up @dampfklon's patches and adjusted them so ACMEDNS_BASE_URL is saved in domainconf.

Hopefully this can get merged now. :)

### DIFF
--- a/dnsapi/dns_acmedns.sh
+++ b/dnsapi/dns_acmedns.sh
@@ -1,31 +1,62 @@
 #!/usr/bin/env sh
 #
 #Author: Wolfgang Ebner
-#Report Bugs here: https://github.com/webner/acme.sh
+#Author: Sven Neubuaer
+#Report Bugs here: https://github.com/dampfklon/acme.sh
+#
+# Usage:
+# export ACMEDNS_BASE_URL="https://auth.acme-dns.io"
+#
+# You can optionally define an already existing account:
+#
+# export ACMEDNS_USERNAME="<username>"
+# export ACMEDNS_PASSWORD="<password>"
+# export ACMEDNS_SUBDOMAIN="<subdomain>"
 #
 ########  Public functions #####################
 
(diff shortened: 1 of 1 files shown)


0xFelix:
Here are the results of the successful DNS GitHub action: https://github.com/0xFelix/acme.sh/actions/runs/333224683

0xFelix:
https://github.com/0xFelix/acme.sh/runs/1320852829

The successful results for commit https://github.com/0xFelix/acme.sh/commit/23ef7d20e2489ac93962241ef7c15f52c42a975a

0xFelix:
Is there anything preventing the merge of this? :)

0xFelix:
@Neilpang Another shy ping

melkypie:
@Neilpang any news on this? Current acme-dns implementation restricts it's use to only one subdomain for which it was built to prevent against.

Neilpang:
moving the credentials to domain conf is a big change.
I'm not sure if its necessary.  cause we have 100+ dns apis to consider.

melkypie:
The problem here is if the acme-dns keeps using account conf we can only assign one subdomain to all domains that will be getting certificates while the actual use for acme-dns is to not use a single subdomain for all of them. As this is a breaking change for people who are already using acme-dns, maybe renaming the hook to acmedns_v2 would be a good idea?

Neilpang:
>> maybe renaming the hook to acmedns_v2 would be a good idea?

no,  we need another unique way for all the dns apis to use.

jonmchan:
I want to confirm this is a breaking change that I only discovered after my website domain expired.

My original config only had `ACMEDNS_UPDATE_URL=`. Adding `ACMEDNS_BASE_URL=` is required now.

tonyfinn:
Also adding that this was breaking for me also.

Issue #2564, opened 2019-10-27
Report bugs to Variomedia dns api


peterkelm:
This is the place to report bugs in the Variomedia DNS API.

If you experienced a bug, please report it in this issue.

Thanks!

morlac:
Hi,

suddenly, when renewing my certs, for one of my subdomains I see the following response in the logs:
----
response='{"errors":[{"id":"invalid_api_data","status":"422","title":"Invalid API data.","detail":"Not ACE-encodable.","source":{"pointer":"/"}}],"links":{"self":"https://api.variomedia.de/domains/_acme-challenge.<SUBDOMAIN_HERE>"}}'
----
followed by: 
----
response='{"errors":[{"id":"resource_not_found","status":"404","title":"Domain with ID '<SUBDOMAIN_HERE>' not found."}],"links":{"self":"https://api.variomedia.de/domains/<SUBDOMAIN_HERE>"}}'
----

I'm running on an all-updated 'Ubuntu 16.04.7 LTS'-Linode

peterkelm:
@morlac, does that affect only one domain name of many? And if so, is there anything "special" about it?

morlac:
@peterkelm it happens on adding the TXT entry for the 2nd sub domain.
My config looks like:
```
Le_Domain='<DOMAIN>'
Le_Alt='*.<DOMAIN>,*.<SUB1>.<DOMAIN>,*.<SUB2>.<DOMAIN>'
```
there is nothing special with this domain at all.

Also any cleanup afterwards (removing of so far added TXT entries _acme-challange.<DOMAIN> and _acme-challange.<SUB1>.<DOMAIN>) fails and must be done manually.

peterkelm:
@morlac, I do not use the API with multiple subdomains, so there's a possibility that this isn't fully tested. However, your initial note seems to indicate that it worked okay at one point. Correct? 

Writing the DNS entry happens in line 33 of `dns_variomedia.sh` - all other DNS API calls use only very simple data, so if something goes wrong this is most likely the place...

One of my certs renewed successfully yesterday... so it doesn't seem to be a general issue with the Variomedia API.
However, the Variomedia API has proven to be picky in the past when it comes to spaces in the JSON data.

Lastly, is there any further info in the `.acme.sh/acme.sh.log` logfile regarding this issue? It should contain a part like this:

```
[Sat Sep  5 00:23:11 CEST 2020] _sub_domain='_acme-challenge.subdomain'
[...]

PR #3244, opened 2020-11-01
Improved determination of root domain for the Variomedia DNS API

peterkelm:
Rewrite of `_get_root()` to address an issue with the handling of subdomains in the `/domains` call of the Variomedia DNS API. See #2564.


dnsapi/dns_variomedia.sh
```diff
@@ -93,11 +89,12 @@ dns_variomedia_rm() {
 # _sub_domain=_acme-challenge.www
 # _domain=domain.com
 _get_root() {
-  fulldomain=$1
-  i=1
+  domain=$1
+  i=2
```
Neilpang, line 93: the value of "i" should be begging from `1`, not `2`.

In dns alias mode,  the fulldomain can be no  `_acme-challenge`  prefix

morlac: it's been tested:
- successfully with i=2
- unsuccessfully with i=1

fraenki:
@peterkelm Using this patch, I can confirm that it fixes the *creation* of TXT records with acme.sh. However, in my case it failed to *remove* the TXT records afterwards. Some unusual cURL output is shown and then a failure is reported:

```
[Mon Feb 21 00:21:46 CET 2022] Here is the curl dump log:
[Mon Feb 21 00:21:46 CET 2022] == Info:   Trying 81.28.224.32:443...
== Info: Connected to api.variomedia.de (81.28.224.32) port 443 (#0)
== Info: ALPN, offering h2
== Info: ALPN, offering http/1.1
== Info:  CAfile: /usr/local/etc/ssl/cert.pem
== Info:  CApath: none
=> Send SSL data, 5 bytes (0x5)
...
[Mon Feb 21 00:21:46 CET 2022] ret='3'
[Mon Feb 21 00:21:46 CET 2022] Error dns-records?filter[domain]=example.com
[Mon Feb 21 00:21:46 CET 2022] Error
[Mon Feb 21 00:21:46 CET 2022] Error removing txt for domain:_acme-challenge.example.com
```

TobiasGrave:
> @peterkelm Using this patch, I can confirm that it fixes the _creation_ of TXT records with acme.sh. However, in my case it failed to _remove_ the TXT records afterwards. Some unusual cURL output is shown and then a failure is reported:
> 
> ```
> [Mon Feb 21 00:21:46 CET 2022] Here is the curl dump log:
> [Mon Feb 21 00:21:46 CET 2022] == Info:   Trying 81.28.224.32:443...
> == Info: Connected to api.variomedia.de (81.28.224.32) port 443 (#0)
> == Info: ALPN, offering h2
> == Info: ALPN, offering http/1.1
> == Info:  CAfile: /usr/local/etc/ssl/cert.pem
> == Info:  CApath: none
> => Send SSL data, 5 bytes (0x5)
> ...
> [Mon Feb 21 00:21:46 CET 2022] ret='3'
> [Mon Feb 21 00:21:46 CET 2022] Error dns-records?filter[domain]=example.com
> [Mon Feb 21 00:21:46 CET 2022] Error
> [Mon Feb 21 00:21:46 CET 2022] Error removing txt for domain:_acme-challenge.example.com
> ```

The removal of the the TXT record fails when its content starts with a `-`. 

This problem is caused by `grep "$txtvalue"` in the following line:
`record_id="$(echo "$response" | cut -d '[' -f2 | cut -d']' -f1 | sed 's/},[ \t]*{/\},§\{/g' | tr § '\n' | grep "$_sub_domain" | grep "$txtvalue" | sed 's/^{//;s/}[,]?$//' | tr , '\n' | tr -d '\"' | grep ^id | cut -d : -f2 | tr -d ' ')"`

This is due to grep interpreting everything starting with a `-` as an option and not as a search term.
In order to fix this, the option `--` has to be added before the search term (https://www.gnu.org/software/grep/manual/grep.html#index-_002d_002d):
`_record_id="$(echo "$response" | cut -d '[' -f2 | cut -d']' -f1 | sed 's/},[ \t]*{/\},§\{/g' | tr § '\n' | grep "$_sub_domain" | grep -- "$txtvalue" | sed 's/^{//;s/}[,]?$//' | tr , '\n' | tr -d '\"' | grep ^id | cut -d : -f2 | tr -d ' ')"`

PR #3270, opened 2020-11-26
Update mailcow.sh

christianbur:
I have modified the following things:
- Originally, "/data/assets/ssl/" is always appended to the varialbe ${_mailcow_path}.  Since I use acme.sh as docker container, I only want to include the mailcow-ssl directory in the acem.sh container and not the complete mailcow directory. So now it is checked if the file generate_config.sh is in the directory (then it is the mailcow root directory, see https://github.com/mailcow/mailcow-dockerized) and only then "/data/assets/ssl/" is appended, in all other cases the passed variable is taken over unchanged. 

- Because of the RP https://github.com/mailcow/mailcow-dockerized/pull/2443 I have extended the script with ECC certificates. 

- I adapted the reboot commands as described in the mailcow manual (https://mailcow.github.io/mailcow-dockerized-docs/firststeps-ssl/#how-to-use-your-own-certificate).

### DIFF
--- a/deploy/mailcow.sh
+++ b/deploy/mailcow.sh
@@ -27,26 +27,47 @@ mailcow_deploy() {
     return 1
   fi
 
-  _ssl_path="${_mailcow_path}/data/assets/ssl/"
+  #Tests if _ssl_path is the mailcow root directory. 
+  if [ -f "${_mailcow_path}/generate_config.sh" ]; then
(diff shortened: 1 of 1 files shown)


Neilpang:
please change  send this PR to `dev` branch.

christianbur:
#3280

PR #3288, opened 2020-12-06
Make domain names available to pre hook

Sp1l:
Export Le_Domains and Le_Alt so your pre-hook script can run additional checks.

Allows running checks on the domain names before the first call to the ACME API. Thereby not counting against the rate-limit when an issue is going to be problematic.


Sp1l:
That commit history is a mess, will re-open a new PR

PR #3305, opened 2020-12-15
Update dns_pdns.sh

funzoneq:
Missing application/json header is causing this PATCH to fail, because it gets processed as form data.

### DIFF
--- a/dnsapi/dns_pdns.sh
+++ b/dnsapi/dns_pdns.sh
@@ -207,7 +207,7 @@ _pdns_rest() {
   export _H1="X-API-Key: $PDNS_Token"
 
   if [ ! "$method" = "GET" ]; then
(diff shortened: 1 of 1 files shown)


NaPs:
I got the same problem and was about to provide a PR when I found this one.

I fixed it by changing the [L211](https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_pdns.sh#L211) by this:

```
    response="$(_post "$data" "$PDNS_Url$ep" "" "$method" "application/json")"
```

(5th argument of `_post` is used to provide content type of request body)

Neilpang:
@NaPs  please show me your PR.

Thanks

NaPs:
@Neilpang Here it is: https://github.com/acmesh-official/acme.sh/pull/3454

PR #3308, opened 2020-12-16
add a dns api dns_ls.sh

ls-hk:
a dns api


recolic:
Hi, you forgot to modify the author name in your shell script

PR #3320, opened 2020-12-24
enables defining pre and post hooks in account.conf.

mkkeffeler:
This enables users to define 
```
ACMESH_PRE_HOOK="command/script"
ACMESH_POST_HOOK="command/script"
```
in account.conf, and then acme.sh leverage them. Great for use cases where a user has default hook behavior that they don't want to define each time they run acme.sh.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4098,10 +4098,9 @@ issue() {
   _savedomainconf "Le_Domain" "$_main_domain"
   _savedomainconf "Le_Alt" "$_alt_domains"
   _savedomainconf "Le_Webroot" "$_web_roots"
-
-  _savedomainconf "Le_PreHook" "$_pre_hook" "base64"
-  _savedomainconf "Le_PostHook" "$_post_hook" "base64"
-  _savedomainconf "Le_RenewHook" "$_renew_hook" "base64"
+  _savedomainconf "Le_PreHook" "$ACMESH_PRE_HOOK" "base64"
+  _savedomainconf "Le_PostHook" "$ACMESH_POST_HOOK" "base64"
+  _savedomainconf "Le_RenewHook" "$ACMESH_RENEW_HOOK" "base64"
 
   if [ "$_local_addr" ]; then
     _savedomainconf "Le_LocalAddress" "$_local_addr"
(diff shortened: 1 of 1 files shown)


mkkeffeler:
I tested defining using the existing ```--pre-hook``` and ```--post-hook```, those work as expected. but now also, if account.conf has 
```
ACMESH_PRE_HOOK="command/script"
ACMESH_POST_HOOK="command/script"

``` 

the script will still invoke those hooks, and I can visibly see output on the logs.

Neilpang:
no,   pre-hook and post-hook  are per domain.

mkkeffeler:
That seems like it can differ by use case though? For example,

User has the same pre hook for all certs on a server because the hook configuration doesn't change based on the domain. 

Can I get more info as to why we would not make this optional functionality for those with use cases like the above?

mkkeffeler:
Another use case would be passing the parameter of the domain as an env variable or some other means to a generic hook which uses the domain as a variable.

Neilpang:
use  --pre-hook  for each domain to issue.

mkkeffeler:
Based on these docs, https://github.com/acmesh-official/acme.sh/wiki/Using-'--pre-hook',--'--post-hook',-'---renew-hook'-and-'---reloadcmd'

There was envisioning of a use case for notifying of updates/issuances, yes? Why should I specify that on each command, when, if that is my use case, I could have it invoked on all attempts automatically? 

I'm trying to understand how I can run pre hooks without having to remember to run them each time. If there is another pattern I would love to hear it, otherwise I don't understand why we would not want to enable this optional behavior for those with use cases like have been discussed.

dwchurch:
I must have this feature, please help

saxx0n:
This would be a very helpful feature for my end users

amahesh98:
This is a feature that I've been looking for

Neilpang:
>> I'm trying to understand how I can run pre hooks without having to remember to run them each time. 

It's easy to define an alias to do that:

```
alias  myissue="acme.sh --issue --pre-hook ....  --post-hook .... "
```

Next time you can use your alias to issue certs:

```
myissue  -d example.com --dns dns_cf 
```



>>  I don't understand why we would not want to enable this optional behavior for those with use cases like have been discussed.

acme.sh is a multi-cert system.  Each cert can define different hooks and commands.  I don't want to make it complicated.

PR #3327, opened 2020-12-28
Fix: Unifi deploy hook support Unifi Cloud Key

### LINKED ISSUE #3326: unifi deploy hook doesn't fully support Unifi Cloud Key
When `--deploy-hook unifi` is used on a Unifi Cloud Key, it only deploys the certificate to one of the two locations needed. After a successful deploy, opening the Cloud Key management page shows a browser certificate error.

Problem
-------

A Unifi Cloud Key runs two https servers: the Unifi Controller app on port 8443 and the Cloud Key management app on port 443. The v2.8.8 unifi deploy hook only updates the keystore used for the Unifi Controller. This means browsing to the Cloud Key management app will show a certificate error, because the default self-signed cert is still being used.

(The unifi deploy hook works just fine for running Unifi Controller on your own machine; this is only a problem on Unifi's Cloud Key device.)

Steps to reproduce
------------------
[Substitute your own domain for "unifi.example.com"]
1. On a Unifi Cloud Key device, install acme.sh and run it to issue a certificate for unifi.example.com (using whatever options you want).
2. Run `./acme.sh --deploy -d "unifi.example.com" --deploy-hook unifi`. Wait for it to complete successfully.
3. Open a browser to the Unifi Controller app at https://unifi.example.com:8443. Observe that it is using your newly-issued, valid Let's Encrypt certificate, and there are no browser certificate errors. (The unifi deploy-hook works correctly for the Unifi Controller app.)
4. Open a browser to the Unifi Cloud Key management app at https://unifi.example.com.

Results: browser error about untrusted certificate

Expected: should be using your new certificate just like the Controller app does on port 8443

Debug log
-----------------
<details>
<summary>[probably not helpful; click to show]</summary>

```
root@UniFi-CloudKey:~/.acme.sh# ./acme.sh --version
v2.8.8
root@UniFi-CloudKey:~/.acme.sh# ./acme.sh --deploy -d "unifi.example.com" --deploy-hook unifi --debug 2
[Mon Dec 28 14:46:21 PST 2020] Lets find script dir.
[Mon Dec 28 14:46:21 PST 2020] _SCRIPT_='./acme.sh'
[Mon Dec 28 14:46:21 PST 2020] _script='/root/.acme.sh/acme.sh'
[Mon Dec 28 14:46:21 PST 2020] _script_home='/root/.acme.sh'
[Mon Dec 28 14:46:21 PST 2020] Using default home:/root/.acme.sh
[Mon Dec 28 14:46:21 PST 2020] Using config home:/root/.acme.sh
[Mon Dec 28 14:46:21 PST 2020] LE_WORKING_DIR='/root/.acme.sh'
v2.8.8
[Mon Dec 28 14:46:21 PST 2020] Running cmd: deploy
[Mon Dec 28 14:46:21 PST 2020] Using config home:/root/.acme.sh
[Mon Dec 28 14:46:21 PST 2020] default_acme_server
[Mon Dec 28 14:46:21 PST 2020] ACME_DIRECTORY='https://acme-v02.api.letsencrypt.org/directory'
[Mon Dec 28 14:46:21 PST 2020] _ACME_SERVER_HOST='acme-v02.api.letsencrypt.org'
[Mon Dec 28 14:46:21 PST 2020] DOMAIN_PATH='/root/.acme.sh/unifi.example.com'
[Mon Dec 28 14:46:21 PST 2020] _deployApi='/root/.acme.sh/deploy/unifi.sh'
[Mon Dec 28 14:46:22 PST 2020] _cdomain='unifi.example.com'
[Mon Dec 28 14:46:22 PST 2020] _ckey='/root/.acme.sh/unifi.example.com/unifi.example.com.key'
[Mon Dec 28 14:46:22 PST 2020] _ccert='/root/.acme.sh/unifi.example.com/unifi.example.com.cer'
[Mon Dec 28 14:46:22 PST 2020] _cca='/root/.acme.sh/unifi.example.com/ca.cer'
[Mon Dec 28 14:46:22 PST 2020] _cfullchain='/root/.acme.sh/unifi.example.com/fullchain.cer'
[Mon Dec 28 14:46:22 PST 2020] _unifi_keystore='/usr/lib/unifi/data/keystore'
[Mon Dec 28 14:46:22 PST 2020] Generate import pkcs12
[Mon Dec 28 14:46:22 PST 2020] Modify unifi keystore: /usr/lib/unifi/data/keystore
Importing keystore /tmp/tmp.e3ZAglrpHS to /usr/lib/unifi/data/keystore...
Warning: Overwriting existing alias unifi in destination keystore
[Mon Dec 28 14:46:29 PST 2020] Import keystore success!
[Mon Dec 28 14:46:29 PST 2020] Run reload: service unifi restart
[Mon Dec 28 14:48:17 PST 2020] Reload success!
[Mon Dec 28 14:48:17 PST 2020] Success
```
</details>

## Suggested fix

When running on a Cloud Key, the unifi deploy hook also needs to:
* install the certificate to /etc/ssl/private/cloudkey.{crt,key}
* regenerate cert.tar in that directory
* run `service nginx restart` to reload the management app

PR follows.

medmunds:
When running on a Unifi Cloud Key device, also deploy to
/etc/ssl/private/cloudkey.{crt,key} and reload nginx.

When running on a newer UnifiOS device, also deploy to
/data/unifi-core/config/unifi-core.{crt,key} and reload unifi-core.

This makes the new cert available for the Cloud Key or UnifiOS
management app running on port 443 (as well as the port 8443
Unifi Controller app the deploy hook already supported, on devices
where a separate Controller is still used).

Fixes #3326

### DIFF
--- a/deploy/unifi.sh
+++ b/deploy/unifi.sh
@@ -55,6 +55,18 @@ unifi_deploy() {
   _debug _cca "$_cca"
   _debug _cfullchain "$_cfullchain"
 
+  _getdeployconf DEPLOY_UNIFI_KEYSTORE
+  _getdeployconf DEPLOY_UNIFI_KEYPASS
+  _getdeployconf DEPLOY_UNIFI_CLOUDKEY_CERTDIR
+  _getdeployconf DEPLOY_UNIFI_CORE_CONFIG
+  _getdeployconf DEPLOY_UNIFI_RELOAD
+
+  _debug2 DEPLOY_UNIFI_KEYSTORE "$DEPLOY_UNIFI_KEYSTORE"
+  _debug2 DEPLOY_UNIFI_KEYPASS "$DEPLOY_UNIFI_KEYPASS"
+  _debug2 DEPLOY_UNIFI_CLOUDKEY_CERTDIR "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR"
+  _debug2 DEPLOY_UNIFI_CORE_CONFIG "$DEPLOY_UNIFI_CORE_CONFIG"
+  _debug2 DEPLOY_UNIFI_RELOAD "$DEPLOY_UNIFI_RELOAD"
+
   # Space-separated list of environments detected and installed:
   _services_updated=""
 
@@ -191,32 +203,12 @@ unifi_deploy() {
     return 1
   fi
 
-  # Successful, so save all config:
-  if [ "$DEPLOY_UNIFI_KEYSTORE" ]; then
-    _savedomainconf DEPLOY_UNIFI_KEYSTORE "$DEPLOY_UNIFI_KEYSTORE"
-  else
-    _cleardomainconf DEPLOY_UNIFI_KEYSTORE
-  fi
-  if [ "$DEPLOY_UNIFI_KEYPASS" ]; then
-    _savedomainconf DEPLOY_UNIFI_KEYPASS "$DEPLOY_UNIFI_KEYPASS"
-  else
-    _cleardomainconf DEPLOY_UNIFI_KEYPASS
-  fi
-  if [ "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR" ]; then
-    _savedomainconf DEPLOY_UNIFI_CLOUDKEY_CERTDIR "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR"
-  else
-    _cleardomainconf DEPLOY_UNIFI_CLOUDKEY_CERTDIR
-  fi
-  if [ "$DEPLOY_UNIFI_CORE_CONFIG" ]; then
-    _savedomainconf DEPLOY_UNIFI_CORE_CONFIG "$DEPLOY_UNIFI_CORE_CONFIG"
-  else
-    _cleardomainconf DEPLOY_UNIFI_CORE_CONFIG
-  fi
-  if [ "$DEPLOY_UNIFI_RELOAD" ]; then
-    _savedomainconf DEPLOY_UNIFI_RELOAD "$DEPLOY_UNIFI_RELOAD"
-  else
-    _cleardomainconf DEPLOY_UNIFI_RELOAD
-  fi
+  # Successful, so save all (non-default) config:
+  _savedeployconf DEPLOY_UNIFI_KEYSTORE "$DEPLOY_UNIFI_KEYSTORE"
+  _savedeployconf DEPLOY_UNIFI_KEYPASS "$DEPLOY_UNIFI_KEYPASS"
+  _savedeployconf DEPLOY_UNIFI_CLOUDKEY_CERTDIR "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR"
+  _savedeployconf DEPLOY_UNIFI_CORE_CONFIG "$DEPLOY_UNIFI_CORE_CONFIG"
+  _savedeployconf DEPLOY_UNIFI_RELOAD "$DEPLOY_UNIFI_RELOAD"
 
   return 0
 }
--- a/deploy/unifi.sh
+++ b/deploy/unifi.sh
@@ -55,11 +55,14 @@ unifi_deploy() {
   _debug _cca "$_cca"
   _debug _cfullchain "$_cfullchain"
 
+  # Space-separated list of environments detected and installed:
+  _services_updated=""
+
   # Default reload commands are accumulated in an &&-separated string
   # as we auto-detect environments:
   DEFAULT_UNIFI_RELOAD=""
 
-  # Unifi Controller (self-hosted or Cloud Key Gen1) environment --
+  # Unifi Controller environment (self hosted or any Cloud Key) --
   # auto-detect by file /usr/lib/unifi/data/keystore:
   DEFAULT_UNIFI_KEYSTORE="/usr/lib/unifi/data/keystore"
   _unifi_keystore="${DEPLOY_UNIFI_KEYSTORE:-$DEFAULT_UNIFI_KEYSTORE}"
@@ -101,7 +104,10 @@ unifi_deploy() {
       return 1
     fi
 
-    DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD} ${DEFAULT_UNIFI_RELOAD:+&&} service unifi restart"
+    if systemctl -q is-active unifi; then
+      DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD}${DEFAULT_UNIFI_RELOAD:+ && }service unifi restart"
+    fi
+    _services_updated="${_services_updated} unifi"
     _info "Install Unifi Controller certificate success!"
   elif [ "$DEPLOY_UNIFI_KEYSTORE" ]; then
     _err "The specified DEPLOY_UNIFI_KEYSTORE='$DEPLOY_UNIFI_KEYSTORE' is not valid, please check."
@@ -130,9 +136,12 @@ unifi_deploy() {
     cp "$_cfullchain" "${_cloudkey_certdir}/cloudkey.crt"
     cp "$_ckey" "${_cloudkey_certdir}/cloudkey.key"
     (cd "$_cloudkey_certdir" && tar -cf cert.tar cloudkey.crt cloudkey.key unifi.keystore.jks)
-    _info "Install Cloud Key Gen1 certificate success!"
 
-    DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD} ${DEFAULT_UNIFI_RELOAD:+&&} service nginx restart"
+    if systemctl -q is-active nginx; then
+      DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD}${DEFAULT_UNIFI_RELOAD:+ && }service nginx restart"
+    fi
+    _info "Install Cloud Key Gen1 certificate success!"
+    _services_updated="${_services_updated} nginx"
   elif [ "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR" ]; then
     _err "The specified DEPLOY_UNIFI_CLOUDKEY_CERTDIR='$DEPLOY_UNIFI_CLOUDKEY_CERTDIR' is not valid, please check."
     return 1
@@ -151,15 +160,18 @@ unifi_deploy() {
 
     cp "$_cfullchain" "${_unifi_core_config}/unifi-core.crt"
     cp "$_ckey" "${_unifi_core_config}/unifi-core.key"
-    _info "Install UnifiOS certificate success!"
 
-    DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD} ${DEFAULT_UNIFI_RELOAD:+&&} systemctl restart unifi-core"
+    if systemctl -q is-active unifi-core; then
+      DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD}${DEFAULT_UNIFI_RELOAD:+ && }systemctl restart unifi-core"
+    fi
+    _info "Install UnifiOS certificate success!"
+    _services_updated="${_services_updated} unifi-core"
   elif [ "$DEPLOY_UNIFI_CORE_CONFIG" ]; then
     _err "The specified DEPLOY_UNIFI_CORE_CONFIG='$DEPLOY_UNIFI_CORE_CONFIG' is not valid, please check."
     return 1
   fi
 
-  if [ -z "$DEFAULT_UNIFI_RELOAD" ]; then
+  if [ -z "$_services_updated" ]; then
     # None of the Unifi environments were auto-detected, so no deployment has occurred
     # (and none of DEPLOY_UNIFI_{KEYSTORE,CLOUDKEY_CERTDIR,CORE_CONFIG} were set).
     _err "Unable to detect Unifi environment in standard location."
@@ -170,6 +182,12 @@ unifi_deploy() {
   fi
 
   _reload="${DEPLOY_UNIFI_RELOAD:-$DEFAULT_UNIFI_RELOAD}"
+  if [ -z "$_reload" ]; then
+    _err "Certificates were installed for services:${_services_updated},"
+    _err "but none appear to be active. Please set DEPLOY_UNIFI_RELOAD"
+    _err "to a command that will restart the necessary services."
+    return 1
+  fi
   _info "Reload services (this may take some time): $_reload"
   if eval "$_reload"; then
     _info "Reload success!"
--- a/deploy/unifi.sh
+++ b/deploy/unifi.sh
@@ -119,6 +119,13 @@ unifi_deploy() {
       _err "The directory $_cloudkey_certdir is not writable; please check permissions."
       return 1
     fi
+    # Cloud Key expects to load the keystore from /etc/ssl/private/unifi.keystore.jks.
+    # Normally /usr/lib/unifi/data/keystore is a symlink there (so the keystore was
+    # updated above), but if not, we don't know how to handle this installation:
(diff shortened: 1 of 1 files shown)


medmunds:
The new settings here just follow the approach of the existing unifi.sh code. But should all of the settings be re-written to use `_getdeployconf` and `_savedeployconf`? (And if so, would you want that in this PR or separately?)

deploy/unifi.sh
```diff
@@ -72,29 +102,47 @@ unifi_deploy() {
     return 1
   fi
 
+  if [ "$_cloudkey_deploy" = "yes" ]; then
+    _info "Install Cloud Key certificate: $_cloudkey_certdir"
+    cp "$_cfullchain" "${_cloudkey_certdir}/cloudkey.crt"
+    cp "$_ckey" "${_cloudkey_certdir}/cloudkey.key"
+    (cd "$_cloudkey_certdir" && tar -cf cert.tar cloudkey.crt cloudkey.key unifi.keystore.jks)
```
medmunds, line 109: Do I need to manually check exit status of the `cp` and `tar` commands, or will any failures get caught and end the script? (There's already a preflight check that `$_cloudkey_certdir` is a writable directory, so I'm having trouble thinking of reasonable cases where these commands could fail.)

medmunds:
- [x] Updated [unifi deploy-hook docs](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#23-deploy-the-cert-on-a-unifi-controller-or-cloud-key) in the wiki.
- [x] Opened new issue for future [unifi deploy-hook bugs](https://github.com/acmesh-official/acme.sh/issues/3359). (I didn't originally create this deploy hook, but there didn't seem to be one.)

deploy/unifi.sh
```diff
@@ -42,102 +55,123 @@ unifi_deploy() {
...
-    -deststorepass "$_unifi_keypass" -destkeypass "$_unifi_keypass" -destkeystore "$_unifi_keystore" \
-    -srckeystore "$_import_pkcs12" -srcstoretype PKCS12 -srcstorepass "$_unifi_keypass" \
-    -alias unifi -noprompt; then
-    _info "Import keystore success!"
-    rm "$_import_pkcs12"
-  else
-    _err "Import unifi keystore error, please report bug to us."
-    rm "$_import_pkcs12"
+  # UnifiOS environment -- auto-detect by /data/unifi-core/config/unifi-core.key:
+  DEFAULT_DEPLOY_UNIFI_CORE_CONFIG="/etc/ssl/private"
```
mjbnz, line 135: Looks like you'll need to change this to `/data/unifi-core/config` ?

deploy/unifi.sh
```diff
@@ -42,102 +55,123 @@ unifi_deploy() {
...
-  fi
 
-  _info "Generate import pkcs12"
-  _import_pkcs12="$(_mktemp)"
-  _toPkcs "$_import_pkcs12" "$_ckey" "$_ccert" "$_cca" "$_unifi_keypass" unifi root
-  if [ "$?" != "0" ]; then
-    _err "Oops, error creating import pkcs12, please report bug to us."
+    cp "$_cfullchain" "${_cloudkey_certdir}/cloudkey.crt"
+    cp "$_ckey" "${_cloudkey_certdir}/cloudkey.key"
+    (cd "$_cloudkey_certdir" && tar -cf cert.tar cloudkey.crt cloudkey.key unifi.keystore.jks)
```
mjbnz, line 125: This `tar` command is adding the keystore file - but it's not copied or generated in place like it is above for a self-hosted controller. I wonder if the java keystore should be generated regardless in the `$domain` dir, and just copied into place when required? (either into `$DEFAULT_UNIFI_KEYSTORE`, or `$DEFAULT_DEPLOY_UNIFI_CLOUDKEY_CERTDIR/unifi.keystore.jks` ?

mjbnz: (this location on the CK is where the SDN controller loads its cert from)

deploy/unifi.sh
```diff
@@ -1,30 +1,43 @@
 #!/usr/bin/env sh
 
-#Here is a script to deploy cert to unifi server.
+# Here is a script to deploy cert on a Unifi Controller or Cloud Key device.
+# It supports:
+#   - self-hosted Unifi Controller
+#   - Unifi Cloud Key Gen1
+#   - UnifiOS (Cloud Key Gen2)
```
mjbnz, line 7: For clarity, I'd adjust documentation to this:
```
#  - Unifi Cloud Key (Gen1/2/2+)
#  - Unifi Cloud Key running UnifiOS (v2.0.0+, Gen2/2+ only)
```

deploy/unifi.sh
medmunds: On my CK Gen1, /usr/lib/unifi/data/keystore is a symlink to /etc/ssl/private/unifi.keystore.jks. I'm assuming that's the standard setup? (And that this is how the current released version of the unifi deploy-hook is able to partially work on a Cloud Key.)

Is there a valid CK config where /usr/lib/unifi/data/keystore is *different* from /etc/ssl/private/unifi.keystore.jks, that this deploy hook should support? If not, it might be easiest to just check before generating the tar that those files have the same contents, and error with "unsupported CloudKey config" if they don't.

(My concern with generating the java keystore in $domain and copying it into place is that the released unifi hook *imports* into the existing keystore. Copying over it would blow away any other certs that might be in there, right? Not a concern on a CK, but I don't know what people might be doing with self-hosted Controller configs.)

deploy/unifi.sh
medmunds: Done

deploy/unifi.sh
medmunds: Ugh, yes of course, thanks for the catch.

deploy/unifi.sh
medmunds: Added this before the tar, see if you think that's sufficient:

```sh
    # Cloud Key expects to load the keystore from /etc/ssl/private/unifi.keystore.jks.
    # Normally /usr/lib/unifi/data/keystore is a symlink there (so the keystore was
    # updated above), but if not, we don't know how to handle this installation:
    if ! cmp -s "$_unifi_keystore" "${_cloudkey_certdir}/unifi.keystore.jks"; then
      _err "Unsupported Cloud Key configuration: keystore not found at '${_cloudkey_certdir}/unifi.keystore.jks'"
      return 1
    fi
```

mjbnz: OK, a bit to disect here.

> On my CK Gen1, /usr/lib/unifi/data/keystore is a symlink to /etc/ssl/private/unifi.keystore.jks. I'm assuming that's the standard setup? (And that this is how the current released version of the unifi deploy-hook is able to partially work on a Cloud Key.)

Well, look at that - it is too. I was aware of `/srv/unifi/data` on the CK, but did not know that's where `/usr/lib/unifi/data` is linked to:

```
root@CK:~# ls -l /usr/lib/unifi/data
lrwxrwxrwx 1 root root 15 Jan  4 20:19 /usr/lib/unifi/data -> /srv/unifi/data
root@CK:~# ls -l /srv/unifi/data/keystore
lrwxrwxrwx 1 unifi unifi 35 Apr 16  2020 /srv/unifi/data/keystore -> /etc/ssl/private/unifi.keystore.jks
```

> Is there a valid CK config where /usr/lib/unifi/data/keystore is _different_ from /etc/ssl/private/unifi.keystore.jks, that this deploy hook should support? If not, it might be easiest to just check before generating the tar that those files have the same contents, and error with "unsupported CloudKey config" if they don't.

I am not familiar enough with the internals of the SDN controller to say for sure, but you are probably more than safe doing that.

> (My concern with generating the java keystore in $domain and copying it into place is that the released unifi hook _imports_ into the existing keystore. Copying over it would blow away any other certs that might be in there, right? Not a concern on a CK, but I don't know what people might be doing with self-hosted Controller configs.)

Fair point. I'd actually completely missed that it was importing - with my previous certificate management (before upgrading to UnifiOS), I was actually generating a new keystore on a remote host, and scp'ing it into place. What you're doing seems sane.

> Added this before the tar, see if you think that's sufficient:

Yes, more than sufficient. Because you're going to be running on a debian system, or a CK, you could also assume the existence of `realpath`, so you could check its output instead (or as well as):

```
root@CK:~# realpath /usr/lib/unifi/data/keystore
/etc/ssl/private/unifi.keystore.jks
root@CK:~# [ "$(realpath "$_unifi_keystore")" = "${_cloudkey_certdir}/unifi.keystore.jks" ] && echo true
true
root@CK:~#
```

But, at a guess, I'd say that the controller only uses `/usr/lib/unifi/data/keystore` at startup, so what's in `/etc/ssl/private` is probably irrelevant if it's not the ultimate location of the symlink.

In summary, what you have is great. good stuff.

deploy/unifi.sh
```diff
@@ -24,77 +55,155 @@ unifi_deploy() {
...
+      _err "Unsupported Cloud Key configuration: keystore not found at '${_cloudkey_certdir}/unifi.keystore.jks'"
+      return 1
+    fi
+
+    cp "$_cfullchain" "${_cloudkey_certdir}/cloudkey.crt"
+    cp "$_ckey" "${_cloudkey_certdir}/cloudkey.key"
+    (cd "$_cloudkey_certdir" && tar -cf cert.tar cloudkey.crt cloudkey.key unifi.keystore.jks)
+    _info "Install Cloud Key Gen1 certificate success!"
+
+    DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD} ${DEFAULT_UNIFI_RELOAD:+&&} service nginx restart"
```
mjbnz, line 135: Ahh... this is where it could fall apart though. An upgraded (v1 -> v2) CKG2/2+ will still have these files present (`cloudkey.{key,crt}`), but no longer runs `nginx`. This attempted restart will cause a failure.

I'm not sure what the right solution is... I'll put my thinking cap on.

mjbnz: perhaps `systemctl -q is-active nginx && systemctl restart nginx`

deploy/unifi.sh
medmunds: OK. FWIW, there's a [forum comment from UI-Team](https://community.ui.com/questions/Unifi-Cloud-key-certificate-installation/d3483712-e44c-4d94-a0e0-c50093b5bd09#answer/03a79172-0b78-4539-be29-7ff1a39e883d) about the files in /etc/ssl/private, that the CK "boot up process will check if they are out of sync" with the cert.tar contents. (But it doesn't say what happens if they are.)

deploy/unifi.sh
medmunds: `systemctl -q is-active nginx` looks good on my CKG1.

Maybe a good idea to guard each of the three restart commands with this. (Would an upgraded CKG2 have a leftover /usr/lib/unifi/data/keystore that would trigger an unwanted `service unifi restart`?)

mjbnz: Also, on CKG2+, which runs `unifi-protect`, it might make sense to restart that as well, when `cloudkey.crt` is replaced. I cannot verify anymore, but I suspect something like this addition would work:

```sh
if [ -L /srv/unifi-protect/data/unifi-protect.crt ];then
  if [ "$(realpath /srv/unifi-protect/data/unifi-protect.crt)" = "${_cloudkey_certdir}/cloudkey.crt" ]; then
    DEFAULT_UNIFI_RELOAD="${DEFAULT_UNIFI_RELOAD} ${DEFAULT_UNIFI_RELOAD:+&&} systemctl -q is-active unifi-protect && systemctl restart unifi-protect"
  fi
fi
```

I have asked a friend with a CKG2+ still on v1 to check if that file is a symlink or not - it no longer is on v2.

mjbnz: > `systemctl -q is-active nginx` looks good on my CKG1.

Sweet, that's handy

> Maybe a good idea to guard each of the three restart commands with this.

Yes, it would make things safer for sure.

> (Would an upgraded CKG2 have a leftover /usr/lib/unifi/data/keystore that would trigger an unwanted `service unifi restart`?)

Very, very good question. This is something I've actually posted about recently on the UBNT community forums. Technically, it's still used - not by an end user using the controller admin interface, but it is used as the cert for a Guest network captive portal... I suspect it's still worthwhile to leave as you have it - being updated and `unifi` restarted.

medmunds: OK, for each of the services unifi, nginx, and unifi-core, I believe the script will now try to:
* If the cert files for that service are present (whether or not that service is still in use), update with the new cert
* If the service is active, restart it to pick up the updated certs

mjbnz: Your patch looks great, nice work.

medmunds: Thanks for all the help.

mjbnz: No problem - would love to see a restart of `unifi-protect` added if I can confirm the cert link state for you.. just an extra restart to add when replacing `/etc/ssl/private/cloudkey.crt` when conditions are met. I'll keep you updated.

Neilpang:
Is this change compatible with the current version?

I mean if the current existing cert can be deployed successfully when it's renewed ?

medmunds:
> Is this change compatible with the current version?

I tried to make sure it is compatible. It supports all of the configuration variables of the current version, and uses them in the same way as the current version. And the new code should only have effect in environments that weren't supported by the current version.

deploy/unifi.sh
```diff
@@ -24,77 +55,173 @@ unifi_deploy() {
...
+  else
+    _cleardomainconf DEPLOY_UNIFI_KEYSTORE
+  fi
+  if [ "$DEPLOY_UNIFI_KEYPASS" ]; then
+    _savedomainconf DEPLOY_UNIFI_KEYPASS "$DEPLOY_UNIFI_KEYPASS"
+  else
+    _cleardomainconf DEPLOY_UNIFI_KEYPASS
+  fi
+  if [ "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR" ]; then
+    _savedomainconf DEPLOY_UNIFI_CLOUDKEY_CERTDIR "$DEPLOY_UNIFI_CLOUDKEY_CERTDIR"
```
Neilpang, line 211: please never use `_savedomainconf` or `_readdomainconf`.
please use `_getdeployconf ` and `_savedeployconf ` instead.
https://github.com/acmesh-official/acme.sh/blob/deac3fc918952e5af92645ac98024b2effabf7fb/deploy/docker.sh#L29

medmunds: @Neilpang the current released version of this deploy-hook [uses `_savedomainconf`](https://github.com/acmesh-official/acme.sh/blob/2.8.8/deploy/unifi.sh#L78-L92) (and doesn't call `_readdomainconf`—it just saved expects config to be in the environment). If I switch all config to `_savedeployconf` will it remain compatible? (And work when existing certs are renewed?)

Neilpang: >> will it remain compatible?

Yes, of course.  but you can also try it.

medmunds: Done

medmunds:
@Neilpang I believe this unifi deploy hook update is ready to merge, unless you see other problems

deploy/unifi.sh
```diff
@@ -24,77 +55,160 @@ unifi_deploy() {
...
+    fi
+    # Cloud Key expects to load the keystore from /etc/ssl/private/unifi.keystore.jks.
+    # Normally /usr/lib/unifi/data/keystore is a symlink there (so the keystore was
+    # updated above), but if not, we don't know how to handle this installation:
+    if ! cmp -s "$_unifi_keystore" "${_cloudkey_certdir}/unifi.keystore.jks"; then
+      _err "Unsupported Cloud Key configuration: keystore not found at '${_cloudkey_certdir}/unifi.keystore.jks'"
+      return 1
+    fi
+
+    cp "$_cfullchain" "${_cloudkey_certdir}/cloudkey.crt"
```
Neilpang, line 144: don't use `cp`.   use `cat .....  > .....` instead.  `cp` will override the file permissions, while `cat` can keep it.

medmunds: Done

PR #3330, opened 2020-12-30
Implement smtp notify hook

medmunds:
Add support for notifications via direct SMTP server connection, including SMTP authentication and secure connections with TLS (explicit TLS, STARTTLS) or SSL (implicit TLS, TLS wrapper).

In addition to your (or your company's) own SMTP server, nearly every commercial Email Service Provider supports SMTP, including Amazon SES, Mailjet, and SparkPost. (Mailgun, Postmark and SendGrid also support it, but they already have dedicated notify hooks that send via their HTTP APIs.)

This uses Python to communicate with SMTP server. Python includes an SMTP client in its standard distribution, and is often preinstalled on platforms/devices that don't include one of the executables required by the "mail" notify hook (sendmail, mutt, mail, mstp). Although I've tried to swear off writing Python 2 code in general, I made this compatible with both Python 3 and Python 2.7, as some older devices may never get Python 3. It may even work with Python 2.6, but I don't plan to test that.

(Of course a pure shell SMTP implementation would be more portable, but I'm not aware of any that support TLS, authentication, and error handling—and that's certainly beyond my shell skills. There is [this start](https://stackoverflow.com/a/9408945/647002) if anyone wants to try.)

Tested on a Unifi Cloud Key with both Python 3.4.2 and Python 2.7.9, sending through Amazon SES with either STARTTLS or implicit TLS.

### DIFF
--- a/notify/smtp.sh
+++ b/notify/smtp.sh
@@ -2,6 +2,8 @@
 
 # support smtp
 
+# Please report bugs to https://github.com/acmesh-official/acme.sh/issues/3358
+
 # This implementation uses Python (2 or 3), which is available in many environments.
 # If you don't have Python, try "mail" notification instead of "smtp".
 
--- a/notify/smtp.sh
+++ b/notify/smtp.sh
@@ -63,13 +63,13 @@ smtp_send() {
   SMTP_SECURE="${SMTP_SECURE:-$(_readaccountconf_mutable SMTP_SECURE)}"
   SMTP_SECURE="${SMTP_SECURE:-none}"
   case "$SMTP_SECURE" in
-    "none") SMTP_DEFAULT_PORT="25";;
-    "ssl")  SMTP_DEFAULT_PORT="465";;
-    "tls")  SMTP_DEFAULT_PORT="587";;
-    *)
-      _err "Invalid SMTP_SECURE='$SMTP_SECURE'. It must be 'ssl', 'tls' or 'none'."
-      return 1
-      ;;
+  "none") SMTP_DEFAULT_PORT="25" ;;
+  "ssl") SMTP_DEFAULT_PORT="465" ;;
+  "tls") SMTP_DEFAULT_PORT="587" ;;
+  *)
+    _err "Invalid SMTP_SECURE='$SMTP_SECURE'. It must be 'ssl', 'tls' or 'none'."
+    return 1
+    ;;
   esac
 
   SMTP_USERNAME="${SMTP_USERNAME:-$(_readaccountconf_mutable SMTP_USERNAME)}"
@@ -125,7 +125,8 @@ _smtp_send() {
   fi
 
   # language=Python
-  smtp_send_output="$($SMTP_PYTHON <<EOF
+  smtp_send_output="$(
+    $SMTP_PYTHON <<EOF
 # This code is meant to work with either Python 2.7.x or Python 3.4+.
 try:
     try:
@@ -189,7 +190,7 @@ finally:
     if smtp is not None:
         smtp.quit()
 EOF
-)"
+  )"
   _ret=$?
   _debug "smtp_send_output" "$smtp_send_output"
   return "$_ret"
--- a/notify/smtp.sh
+++ b/notify/smtp.sh
@@ -287,6 +287,7 @@ try:
         from email.message import EmailMessage
     except ImportError:
         from email.mime.text import MIMEText as EmailMessage  # Python 2
+    from email.utils import formatdate as rfc2822_date
     from smtplib import SMTP, SMTP_SSL, SMTPException
     from socket import error as SocketError
 except ImportError as err:
@@ -318,6 +319,7 @@ except (AttributeError, TypeError):
 msg["Subject"] = subject
 msg["From"] = from_email
 msg["To"] = to_emails
+msg["Date"] = rfc2822_date(localtime=True)
 msg["X-Mailer"] = x_mailer
 
 smtp = None
--- a/notify/smtp.sh
+++ b/notify/smtp.sh
@@ -2,7 +2,103 @@
 
 # support smtp
 
+# This implementation uses Python (2 or 3), which is available in many environments.
+# If you don't have Python, try "mail" notification instead of "smtp".
+
+# SMTP_FROM="from@example.com"  # required
+# SMTP_TO="to@example.com"  # required
+# SMTP_HOST="smtp.example.com"  # required
+# SMTP_PORT="25"  # defaults to 25, 465 or 587 depending on SMTP_SECURE
+# SMTP_SECURE="none"  # one of "none", "ssl" (implicit TLS, TLS Wrapper), "tls" (explicit TLS, STARTTLS)
+# SMTP_USERNAME=""  # set if SMTP server requires login
+# SMTP_PASSWORD=""  # set if SMTP server requires login
+# SMTP_TIMEOUT="15"  # seconds for SMTP operations to timeout
+# SMTP_PYTHON="/path/to/python"  # defaults to system python3 or python
+
 smtp_send() {
+  # Find a Python interpreter:
+  SMTP_PYTHON="${SMTP_PYTHON:-$(_readaccountconf_mutable SMTP_PYTHON)}"
+  if [ "$SMTP_PYTHON" ]; then
+    if _exists "$SMTP_PYTHON"; then
+      _saveaccountconf_mutable SMTP_PYTHON "$SMTP_PYTHON"
+    else
+      _err "SMTP_PYTHON '$SMTP_PYTHON' does not exist."
+      return 1
+    fi
+  else
+    # No SMTP_PYTHON setting; try to run default Python.
+    # (This is not saved with the conf.)
+    if _exists python3; then
+      SMTP_PYTHON="python3"
+    elif _exists python; then
+      SMTP_PYTHON="python"
+    else
+      _err "Can't locate Python interpreter; please define SMTP_PYTHON."
+      return 1
+    fi
+  fi
+  _debug "SMTP_PYTHON" "$SMTP_PYTHON"
+  _debug "Python version" "$($SMTP_PYTHON --version 2>&1)"
+
+  # Validate other settings:
+  SMTP_FROM="${SMTP_FROM:-$(_readaccountconf_mutable SMTP_FROM)}"
+  if [ -z "$SMTP_FROM" ]; then
+    _err "You must define SMTP_FROM as the sender email address."
+    return 1
+  fi
+
+  SMTP_TO="${SMTP_TO:-$(_readaccountconf_mutable SMTP_TO)}"
+  if [ -z "$SMTP_TO" ]; then
+    _err "You must define SMTP_TO as the recipient email address."
+    return 1
+  fi
+
+  SMTP_HOST="${SMTP_HOST:-$(_readaccountconf_mutable SMTP_HOST)}"
+  if [ -z "$SMTP_HOST" ]; then
+    _err "You must define SMTP_HOST as the SMTP server hostname."
+    return 1
+  fi
+  SMTP_PORT="${SMTP_PORT:-$(_readaccountconf_mutable SMTP_PORT)}"
+
+  SMTP_SECURE="${SMTP_SECURE:-$(_readaccountconf_mutable SMTP_SECURE)}"
+  SMTP_SECURE="${SMTP_SECURE:-none}"
+  case "$SMTP_SECURE" in
+    "none") SMTP_DEFAULT_PORT="25";;
+    "ssl")  SMTP_DEFAULT_PORT="465";;
+    "tls")  SMTP_DEFAULT_PORT="587";;
+    *)
+      _err "Invalid SMTP_SECURE='$SMTP_SECURE'. It must be 'ssl', 'tls' or 'none'."
+      return 1
+      ;;
+  esac
+
+  SMTP_USERNAME="${SMTP_USERNAME:-$(_readaccountconf_mutable SMTP_USERNAME)}"
+  SMTP_PASSWORD="${SMTP_PASSWORD:-$(_readaccountconf_mutable SMTP_PASSWORD)}"
+
+  SMTP_TIMEOUT="${SMTP_TIMEOUT:-$(_readaccountconf_mutable SMTP_TIMEOUT)}"
+  SMTP_DEFAULT_TIMEOUT="15"
+
+  _saveaccountconf_mutable SMTP_FROM "$SMTP_FROM"
+  _saveaccountconf_mutable SMTP_TO "$SMTP_TO"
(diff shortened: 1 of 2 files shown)


tiberios1:
Any idea when this is likely to be merged? I'm looking to use this function in the docker image...

medmunds:
@tiberios1 if you need to use this before it's merged, you could copy [acme.sh/notify/smtp.sh](https://github.com/acmesh-official/acme.sh/blob/ef53b0f7fa57894445e5deeea7ffa4dcff0281ef/notify/smtp.sh) from this PR into your acme.sh installation in Docker, overwriting the "unimplemented" notify/smtp.sh stub in the current acme.sh distribution. (SMTP support is entirely self-contained in notify/smtp.sh, and no other files need to be changed.)

tiberios1:
@medmunds The one problem with that is that I don't think the docker image has python installed on it so I guess that would need to be added to the dockerfile

medmunds:
> I don't think the docker image has python installed on it

Ah, I'm guessing that will still be a problem even if this PR is merged. The problem with sending SMTP from sh is it pretty much requires some sort of helper app. I chose Python for this PR, but am open to other suggestions.

For your use case, if the docker image already includes sendmail or one of the other apps usable for the ["mail" notify hook](https://github.com/acmesh-official/acme.sh/wiki/notify#3-set-notification-for-mail), you might be able to use that instead. Otherwise it seems like you'd probably need to extend the docker image (with either Python or sendmail) in either case.

medmunds:
- [x] Added [SMTP notify-hook docs](https://github.com/acmesh-official/acme.sh/wiki/notify#12-set-notification-for-smtp) in the wiki.
- [x] Opened new issue for future [smtp notify-hook bugs](https://github.com/acmesh-official/acme.sh/issues/3358).
- [x] Rebased to current dev

CharlieYu4994:
Why not just use the smtp「mailx 」command.
```
#!/usr/bin/env sh

smtp_send() {
  SMTP_FROM="${SMTP_FROM:-$(_readaccountconf_mutable SMTP_FROM)}"
  if [ -z "$SMTP_FROM" ]; then
    _err "You must define SMTP_FROM as the sender email address."
    return 1
  fi

  SMTP_TO="${SMTP_TO:-$(_readaccountconf_mutable SMTP_TO)}"
  if [ -z "$SMTP_TO" ]; then
    _err "You must define SMTP_TO as the recipient email address."
    return 1
  fi

  SMTP_HOST="${SMTP_HOST:-$(_readaccountconf_mutable SMTP_HOST)}"
  if [ -z "$SMTP_HOST" ]; then
    _err "You must define SMTP_HOST as the SMTP server hostname."
    return 1
  fi

  SMTP_USERNAME="${SMTP_USERNAME:-$(_readaccountconf_mutable SMTP_USERNAME)}"
  SMTP_PASSWORD="${SMTP_PASSWORD:-$(_readaccountconf_mutable SMTP_PASSWORD)}"
  SMTP_CONTEXT="${SMTP_CONTEXT:-$(_readaccountconf_mutable SMTP_CONTEXT)}"

  _saveaccountconf_mutable SMTP_FROM "$SMTP_FROM"
  _saveaccountconf_mutable SMTP_TO "$SMTP_TO"
  _saveaccountconf_mutable SMTP_HOST "$SMTP_HOST"
  _saveaccountconf_mutable SMTP_USERNAME "$SMTP_USERNAME"
  _saveaccountconf_mutable SMTP_PASSWORD "$SMTP_PASSWORD"
  _saveaccountconf_mutable SMTP_CONTEXT "$SMTP_CONTEXT" "base64"

  if ! _smtp_send "$@"; then
    _err "$smtp_send_output"
    return 1
  fi

  return 0
}

_smtp_send() {
  _subject="${1}"
  _content="${2}"
  _statusCode="${3}"
  _debug "_subject" "$_subject"
  _debug "_content" "$_content"
  _debug "_statusCode" "$_statusCode"

  _debug "SMTP_FROM" "$SMTP_FROM"
  _debug "SMTP_TO" "$SMTP_TO"
  _debug "SMTP_HOST" "$SMTP_HOST"
  _debug "SMTP_USERNAME" "$SMTP_USERNAME"
  _debug "SMTP_PASSWORD" "$SMTP_PASSWORD"

  if [ "${DEBUG:-$DEBUG_LEVEL_NONE}" -ge "$DEBUG_LEVEL_2" ]; then
    smtp_debug="True"
  else
    smtp_debug=""
  fi

  smtp_send_output="$(echo "${_context}" | mailx -v \
                      -s "${_subject}" \
                      -S ssl-verify=ignore\
                      -S smtp="${SMTP_HOST}" \
                      -S smtp-auth=login \
                      -S smtp-auth-user="${SMTP_USERNAME}" \
                      -S smtp-auth-password="${SMTP_PASSWORD}" \
                      -S from="${SMTP_FROM}" \
                      "${SMTP_TO}")"
  _ret=$?
  _debug "smtp_send_output" "$smtp_send_output"
  return "$_ret"
}
```

medmunds:
@CharlieYu4994 I was looking for a way to send email from a hardware device where I can't control what packages are installed. Most constrained devices don't include full email send/receive clients like mail, mailx, or mutt (which are supported by the existing [mail notify hook](https://github.com/acmesh-official/acme.sh/blob/master/notify/mail.sh)—see [docs for mail notify hook](https://github.com/acmesh-official/acme.sh/wiki/notify#3-set-notification-for-mail)). But they sometimes have Python. (Perl could be another possibility.)

Maybe it would make sense to rename the hook in this PR "smtp-python" to clarify it's for platforms that already have Python, and leave the existing "smtp" hook as unimplemented. (Probably add a note to the unimplemented smtp hook suggesting "mail" hook if you have mail, mutt, ssmtp, or msmtp, and "smtp-python" if you only have Python).

medmunds:
Ah, so [`curl` supports sending through SMTP](https://ec.haxx.se/usingcurl/usingcurl-smtp), and seems even more likely than Python to be available on a constrained device (that can run acme.sh). 

I'm going to close this PR and open a new one either adding curl support to the existing "mail" notify hook, or implementing this smtp hook using curl, whichever seems cleaner.

CharlieYu4994:
Yesterday I tried to use the bash feature,  the `/dev/tcp/${host}/${port}`, but I failed to enable the SSL protocol.

CharlieYu4994:
> Yesterday I tried to use the bash feature, the `/dev/tcp/${host}/${port}`, but I failed to enable the SSL protocol.

It's too complex so it may not be the best way

Neilpang:
@medmunds  @CharlieYu4994 

I think  python approach can be one of the options.

You can add many alternatives:

1. if  python exists, try use python way.
2. else if curl exists,  use curl
3. else if  current script is running in bash, you can use `/dev/tcp/${host}/${port}`

you can also add other alternatives based on simple tools.

Thanks.

medmunds:
@Neilpang that sounds good. I'm working on the curl implementation, will add that in the next couple of days.

medmunds:
@Neilpang I think this smtp notify-hook is ready for review. I've added curl support (preferred where available) and cleaned up the Python support a little.

medmunds:
@tiberios1 just updated the smtp notify-hook to support SMTP sends using curl, so it would probably work in an unmodified Docker image. (Once it's merged. You'll need to copy in [notify/smtp.sh](https://github.com/acmesh-official/acme.sh/blob/ffe7ef476439df04e1878ea08f0c6b3eb91653c6/notify/smtp.sh) from this PR before that.)

notify/smtp.sh
```diff
@@ -2,14 +2,330 @@
...
+  if ! smtp_output="$($_smtp_send)"; then
+    _err "Error sending message with $_SMTP_BIN."
+    if [ -n "$smtp_output" ]; then
+      _err "$smtp_output"
+    fi
+    return 1
+  fi
+
+  # Save config only if send was successful:
+  _saveaccountconf_mutable SMTP_BIN "$SMTP_BIN"
```
Neilpang, line 144: we don't need to save all these variables.

please save them at the begging, only when the user specifies a non-default value.

notify/smtp.sh
```diff
@@ -2,14 +2,330 @@
...
+  echo "$raw_message" | "$_SMTP_BIN" "$@"
+}
+
+# Output an RFC-822 / RFC-5322 email message using _SMTP_* variables
+_smtp_raw_message() {
+  echo "From: $_SMTP_FROM"
+  echo "To: $_SMTP_TO"
+  echo "Subject: $(_mime_encoded_word "$_SMTP_SUBJECT")"
+  if _exists date; then
+    echo "Date: $(date +'%a, %-d %b %Y %H:%M:%S %z')"
```
Neilpang, line 232: please don't use `date` command.

please use the `_utc_date()` function and `sed/tr` instead.

notify/smtp.sh
```diff
@@ -2,14 +2,330 @@
...
+  fi
+  echo "Content-Type: text/plain; charset=utf-8"
+  echo "X-Mailer: acme.sh --notify-hook smtp"
```
Neilpang, line 235: you don't need to use hardcode `acme.sh`,  you can use `${PROJECT_NAME}` instead.

notify/smtp.sh
medmunds: The email Date header must be in the format specified by [RFC 2822 Section 3.3](https://tools.ietf.org/html/rfc2822#section-3.3):
> Date: Mon, 15 Feb 2021 00:03:12 -0800

1. The month name ("Feb") must be a three-character English abbreviation; numeric months are not permitted
2. The weekday ("Mon") is optional, but if present must be a three-character English abbreviation; as a practical matter weekday almost always seems to be included, and omitting it could break some (poorly implemented) email clients or scripts
3. The time zone "SHOULD express local time" (though UTC is permitted)

Some options:
A. Ignore recommendations 2 and 3, and try to hack something together with `sed` on `_utc_date()` to generate month abbreviations for requirement 1. (This seems really complicated compared to just including `%b` in a `date` format specifier.)
B. Add a new `_rfc2822_date()` function to abstract this, let it call `date` like here. (Also, I notice it should override `LC_TIME=C` to force required English abbreviations.)
C. Don't add a Date header to the generated message, and hope that the SMTP server adds one for us. (Many will.)

Please let me know which you'd prefer. I can implement option B or C; if option A is preferred, perhaps you could suggest the sed expression.

notify/smtp.sh
medmunds: Done. Also added `${VER}`, and added X-Mailer header to Python version.

notify/smtp.sh
medmunds: I initially had coded it that way, calling `_saveaccountconf_mutable()` only for non-default values. But then there was no way to switch back to the default value, other than manually editing `account.conf` to remove the overrides.

It seems like we would need to call `_clearaccountconf_mutable()` if the user has specified the default value. That function doesn't exist. (The current mail notify hook tries to [implement this logic](https://github.com/acmesh-official/acme.sh/blob/2.8.8/notify/mail.sh#L19-L24):

```sh
  MAIL_NOVALIDATE="${MAIL_NOVALIDATE:-$(_readaccountconf_mutable MAIL_NOVALIDATE)}"
  if [ -n "$MAIL_NOVALIDATE" ]; then
    _saveaccountconf_mutable MAIL_NOVALIDATE 1
  else
    _clearaccountconf "MAIL_NOVALIDATE"
  fi
```

... but it calls `_clearaccountconf()`, which I think clears the wrong variable name.)

Should I add `_clearaccountconf_mutable()` to the main `acme.sh`?
```sh
#key
_clearaccountconf_mutable() {
  _clearaccountconf "SAVED_$1"
}
```

Or another option could be to add a "default" parameter to the existing `_saveaccountconf_mutable()`, and make it save the value if non-default, otherwise clear the value. That avoids duplicating the "if non-default save otherwise clear" logic in lots of other places:
```sh
#key  value base64encode default
_saveaccountconf_mutable() {
  if [ "$2" = "$4" ]; then
    _clear_conf "$ACCOUNT_CONF_PATH" "SAVED_$1"
  else
    _save_conf "$ACCOUNT_CONF_PATH" "SAVED_$1" "$2" "$3"
  fi
  #remove later
  _clearaccountconf "$1"
}
```

Neilpang: >> Should I add _clearaccountconf_mutable() to the main acme.sh?

Yes, go ahead.

But I think it should clear both names:   `XXX` and `SAVED_XXXX`

what do you think?

notify/smtp.sh
Neilpang: I'd like option B more.

please make sure `_rfc2822_date` can return exact the same format on Linux, Mac, BSD, and Solaris.


If you need BSD vm and Solaris VM, see:

https://github.com/vmactions/shell-freebsd


you can also implement `_rfc2822_date ` based on the output of `_utc_date`,  just  `sed` and `tr`it.

medmunds: > implement `_rfc2822_date` based on the output of `_utc_date`, just `sed` and `tr` it

Sorry, this is a little beyond my skills. How would I use `sed` and `tr` to transform `2021-02-16 17:14:43` into `Tue, 16 Feb 2021 17:14:43 +0000`? (I know how to rearrange pieces of text, but not how to change `02` into `Feb` and `2021-02-16` into `Tue`.)

notify/smtp.sh
medmunds: Agreed, `_clearaccountconf_mutable()` should clear both names. Added to main acme.sh.

I've reworked the code to save the variables at the beginning (as they're validated), and only if they are not the default values, as requested. 

I added a couple of helpers in notify/smtp.sh (at the bottom) to reduce some of the repetitive code around handling default values. Let me know if those belong somewhere else.

notify/smtp.sh
medmunds: (I've added `_rfc2822_date` and tested on the various platforms; would still need help with `sed` if you want that changed.)

notify/smtp.sh
```diff
@@ -2,14 +2,393 @@
...
+  SMTP_BIN="$(_readaccountconf_mutable_default SMTP_BIN)"
+  if [ -n "$SMTP_BIN" ] && ! _exists "$SMTP_BIN"; then
+    _err "SMTP_BIN '$SMTP_BIN' does not exist."
+    return 1
+  fi
+  if [ -z "$SMTP_BIN" ]; then
+    # Look for a command that can communicate with an SMTP server.
+    # (Please don't add sendmail, ssmtp, mutt, mail, or msmtp here.
+    # Those are already handled by the "mail" notify hook.)
+    for cmd in curl python3 python2.7 python pypy3 pypy; do
```
Neilpang, line 39: please move `curl` to the last one.
I'd prefer python code better.

medmunds: Done

Neilpang:
please update the wiki page again, and if you are finished, I will merge it.

thanks.

medmunds:
> please update the wiki page again

Done. I believe this is ready to merge.

Thanks for your help. (And thanks for acme.sh!)

PR #3333, opened 2021-01-04
Add Telegram notify hook

mjbnz:
Created as a modifed copy of `slack.sh`

Requires:
- API Token for a bot created with the Telegram Bot Father.
- A Chat ID for a user/group that the bot has permission to post to.

### DIFF
--- a/notify/telegram.sh
+++ b/notify/telegram.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 #Support Telegram Bots
 
--- a/notify/telegram.sh
+++ b/notify/telegram.sh
@@ -37,7 +37,7 @@ telegram_send() {
   _telegram_bot_url="https://api.telegram.org/bot${TELEGRAM_BOT_APITOKEN}/sendMessage"
   if _post "$_data" "$_telegram_bot_url"; then
     # shellcheck disable=SC2154
-    _message=$(printf "%s\n" "$response" | sed -ne 's/.*"ok":\([^,]*\).*/\1/p')
+    _message=$(printf "%s\n" "$response" | sed -n 's/.*"ok":\([^,]*\).*/\1/p')
     if [ "$_message" = "true" ]; then
       _info "telegram send success."
       return 0
--- a/notify/telegram.sh
+++ b/notify/telegram.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+#Support Telegram Bots
+
+#TELEGRAM_BOT_APITOKEN=""
+#TELEGRAM_BOT_CHATID=""
+
+telegram_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
+  _debug "_statusCode" "$_statusCode"
+
+  TELEGRAM_BOT_APITOKEN="${TELEGRAM_BOT_APITOKEN:-$(_readaccountconf_mutable TELEGRAM_BOT_APITOKEN)}"
+  if [ -z "$TELEGRAM_BOT_APITOKEN" ]; then
+    TELEGRAM_BOT_APITOKEN=""
+    _err "You didn't specify a Telegram BOT API Token TELEGRAM_BOT_APITOKEN yet."
+    return 1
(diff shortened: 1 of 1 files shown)


Neilpang:
please read this guide first to follow our common rules:

Neilpang:
add your usage here:

mjbnz:
> please read this guide first to follow our common rules:
> https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide

Because this is a notification hook, and not a DNS API hook, I'd mostly ignored that page - my apologies. I've fixed the couple of relevant things mentioned there - shebang (I wasn't aware that the install edited shebangs, I'd copied `slack.sh` from my installed copy), and `sed -e`.

> add your usage here:
> https://github.com/acmesh-official/acme.sh/wiki/notify

Done.

notify/telegram.sh
```diff
@@ -0,0 +1,49 @@
...
+#TELEGRAM_BOT_APITOKEN=""
+#TELEGRAM_BOT_CHATID=""
+
+telegram_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
+  _debug "_statusCode" "$_statusCode"
+
+  TELEGRAM_BOT_APITOKEN="${TELEGRAM_BOT_APITOKEN:-$(_readaccountconf_mutable TELEGRAM_BOT_APITOKEN)}"
```
Neilpang, line 14: see : https://github.com/acmesh-official/acme.sh/blob/master/deploy/docker.sh#L22

please use "_getdeployconf" and "_savedeployconf "

mjbnz: No problem, I can do that - but, to be clear, that's different to all of the other [existing notify hooks](https://github.com/acmesh-official/acme.sh/tree/dev/notify), is that intentional? (should they be converted to use `_{save,get}deployconf` too? so this would just be the first notify hook to use them?)

Neilpang: Yes,  the `_{save,get}deployconf` is the newer way to get and set deploy options.

Most of the existing hooks are not updated yet.

mjbnz: After reading the main shell script yesterday, I now see the intention of the various `_{save,get}*conf` functions... however, is a new set of functions required for notify hooks, since they're more of a global setting, not a per-domain setting? As it stands, the `deployconf` variants save to the `$domain/$domain.conf` file. Unless I'm mistaken, notification hooks are called for all domains after `--cron`, no?

mjbnz: (I can also see a world where each domain might have different notification settings, which are processed after a renewal attempt, instead of one global notification.. while more complicated, that might have merit?)

mjbnz: > Unless I'm mistaken, notification hooks are called for all domains after `--cron`, no?

[sorry for the multiple messages, I should stop and finish reading before posting!]

Turns out I am mistaken, I'd completely missed the `$NOTIFY_MODE` setting... however, that does indicate that perhaps there would be some unintended behaviour if notify hook specific settings are saved in a domain file. When `$NOTIFY_MODE` is set to `0` for bulk notifications (which is the default), then I think the settings of the _last_ domain renewed would be used for the bulk notification? The notification hooks would need to check `$NOTIFY_MODE` as well, and choose to load from the domain config, or the account config.

hopefully this discussion is helpful, and not unwelcome. I would be keen to put together a PR that addresses these issues once this one is merged if you like.

Neilpang: I'm so sorry. I missed up when I was reviewing multiple PRs.  please forget the `_{save,get}deployconf`, it's for deploy hooks only.

It's ok for you notification hook to use account options.

mjbnz: No problem - however, I do think that there might actually be merit in having per domain notifications, and I can see it being relatively easy to implement.

Thanks for the merge!

Neilpang: >>  No problem - however, I do think that there might actually be merit in having per domain notifications, and I can see it being relatively easy to implement.

I don't think so.   it will make the notifications more complicated for non-expert users.
We can think about it if there are too many users require. but for now, I don't hear anyone need that.

PR #3334, opened 2021-01-04
fix arvan cloud api problem

fvahid:


fvahid:
Hi all,
Sorry for question I don't know why DNS /Fail section call and I don't have any idea how can fix this problem so DNS action in my repository successfully run without any problem

PR #3340, opened 2021-01-06
Update dns_duckdns.sh to work with subdomains

figurcoe:
_duckdns_get_domain:
Adjusted sed-command to accept wildcard-domains and sub-sub-domains. Now 
- *.DOMAIN.duckdns.org
- *.sub1.sub2.DOMAIN.duckdns.org
- sub1.sub2.DOMAIN.duckdns.org
will yield "DOMAIN". This includes the versions preceeded by "_acme-challenge."

Tested for quite some examples to an extend that I'm quite confident, that this should work under most circumstances. But you never know...

### DIFF
--- a/dnsapi/dns_duckdns.sh
+++ b/dnsapi/dns_duckdns.sh
@@ -91,12 +91,13 @@ dns_duckdns_rm() {
 
 ####################  Private functions below ##################################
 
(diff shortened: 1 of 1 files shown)


figurcoe:
solved well enough under #3307

PR #3347, opened 2021-01-08
Update README to reflect the fact that this is a general-purpose ACME client

JamesTheAwesomeDude:
This software _is_ compatible with most-any CA software that's RFC8555-compliant, right? Even one internal, non-public, etc.?


Neilpang:
thanks.

PR #3350, opened 2021-01-10
Deploy gcore_cdn fix

temoffey:
1. Deploy gcore_cdn fix api auth
2. Deploy gcore_cdn fix api parse resource list


bkfon:
Is everything ok in my request? Can I do something to speed up its adoption?

PR #3364, opened 2021-01-15
Allow DNSAPI Folder to be defined using FullPath

bcsanford:
Allow _hookcat aka _SUB_FOLDER_DNSAPI to be defined using FullPath


Neilpang:
why do we need this ?

bcsanford:
I can only answer for myself by i personally do not store the dnsapi folder within the acme home or my home directory for that matter, without this addition it will not allow the folder to be found, now that i know what to add i could easily redo it each time a new version comes out but this is a small mod to the bigger project that hurts nothing and may benefit others doing similar things

PR #3369, opened 2021-01-15
Implement smtp notify hook using mailx

CharlieYu4994:


medmunds:
I think `mailx` is a version of the `mail` command, which is already supported by the existing [mail notify nook](https://github.com/acmesh-official/acme.sh/blob/master/notify/mail.sh). (See [docs for mail notify hook](https://github.com/acmesh-official/acme.sh/wiki/notify#3-set-notification-for-mail).)

CharlieYu4994:
> I think `mailx` is a version of the `mail` command, which is already supported by the existing [mail notify nook](https://github.com/acmesh-official/acme.sh/blob/master/notify/mail.sh). (See [docs for mail notify hook](https://github.com/acmesh-official/acme.sh/wiki/notify#3-set-notification-for-mail).)

You are right. 
And I wondered if there is a methods more basic to implement SMTP

PR #3375, opened 2021-01-20
add  aliyun cdn deploy hook

acooler15:


shoyuf:
this method is no longer working

PR #3376, opened 2021-01-20
Docker non root

fradev:
removed root as the user for docker container
using acmesh (UID/GID can be modified at build time via ENV Variable) 
Added supercronic fo scheduling a cron job


ibotty:
This is two things at once.  Supercronic is 14MB.  And I personally can't envision why one should run cron inside this container.  The container can be run periodically from outside (with kubernetes cronjobs, cron on bare docker hosts, etc.).

And this image is now only usable on amd64.

PR #3392, opened 2021-02-04
Adding DNSAPI wrapper for Websupport webhosting (Slovakia)

akulumbeg:
Hello @Neilpang, I managed to fix everything and I am opening a new pull request for easier organization. Please close the original #3184 .

A couple of things that I fixed and improved from the original script by @trgosk (again, great job!).

- the timestamp is now generated by the _time() function as requested by @Neilpang 
- the authorization is now done via the _hmac() function
- Solaris check passes (see the [actions in my repo](https://github.com/akulumbeg/acme.sh/actions) )

For the future reference, anyone struggling with Solaris as much as I did, read this:
Solaris shell **cannot** convert dates from epoch properly. Neither "date -u -d @"timestamp"", nor "date -u -r "timestamp"" work.
To mend this, after some experimenting, did another 2>/dev/null condition resulting in using the _utc_date() function modified by sed to achieve the ISO8601 format (line 161).

**Please do not change this.** The Websupport API is particularly sensitive to a match of the epoch timestamp and the utc date generated from the timestamp. Every other system can convert the timestamp variable into a date and work with that, so that is prioritized. If that fails, the _utc_date()/sed kicks in as a failsafe. 

However, there is a (virtually nonexistent) possibility of a bug occurence, when there is a 1 (or more) second delay between execution of _time() and _utc_date(), then they won't match. Again this should never happen because we are in the 21st century and have powerful processors, but the possibility exists. In Solaris there is no other way that I know of to mend this, perhaps using some long perl scripts etc., but I don't think it's worth the time finding this solution out.

Other than that, everything works perfectly, I tested it on my Synology NAS and believe it can be pushed as-is. Once it gets pushed into the dev branch, I will update the Wiki entry.


akulumbeg:
Obviously, the DNS check will fail in the pull request due to different secrets etc, but again, the actions in my repo and my real-life test on NAS show that it is working properly.

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,207 @@
...
+_ws_rest() {
+  me=$1
+  pa="$2"
+  da="$3"
+
+  _debug2 api_key "$WS_ApiKey"
+  _debug2 api_secret "$WS_ApiSecret"
+
+  timestamp=$(_time)
+  datez="$(date -u -d @"$timestamp" +%Y-%m-%dT%H:%M:%S%z 2>/dev/null || date -u -r "$timestamp" +%Y-%m-%dT%H:%M:%S%z 2>/dev/null || _utc_date | sed "s/ /T/" | sed "s/$/+0000/")"
```
Neilpang, line 161: don't use `date` command here.
I think you can just use `_utc_date()` and `sed` or `tr`. that's enough.

dnsapi/dns_websupport.sh
```diff
@@ -0,0 +1,207 @@
...
+  canonical_request="${me} ${pa} ${timestamp}"
+  signature_hash=$(printf "%s" "$canonical_request" | _hmac sha1 "$(printf "%s" "$WS_ApiSecret" | _hex_dump | tr -d " ")" hex)
```
Neilpang, line 163: I think you should use `_hmac sha1 "$WS_ApiSecret"`

dnsapi/dns_websupport.sh
akulumbeg: Alright, I will do that (it works this way, tested three commits ago), even though the Websupport API really relies on the epoch timestamp conversion to ISO8601 date.

I'd personally argue that using this conversion completely prevents the possibility of any future bugs caused by a possibly delayed execution of timestamp=$(_time) and datez=(_utc_date | sed ...)  with the exception of Solaris.

dnsapi/dns_websupport.sh
akulumbeg: I need to test this, let's see.

akulumbeg: Unfortunately, the way you suggest it does not work, it works the way I wrote it earlier.

akulumbeg:
Tried all sorts of combinations for the _hmac authorization but none worked, I reverted back to what worked originally.

dnsapi/dns_websupport.sh
akulumbeg: Marking this as resolved, nothing else seems to be passing the checks...

Neilpang:
Fix your CI checks: 
https://github.com/akulumbeg/acme.sh/runs/1832357550?check_suite_focus=true

Neilpang:
please rebace to the latest dev code and fix your CI errors.

akulumbeg:
The code is exactly as it was when it initially worked, passing all checks, in February. Now Oraclelinux:8 fails and the log does not give me any hint what might be wrong...

Neilpang:
you can debug  oraclelinux docker locally.

It's easy to run acmetest locally: https://github.com/acmesh-official/acmetest

All the CI checks must be passing before it can be merged.

Thanks.

akulumbeg:
@Neilpang after some time, I got back to this. I tested everything locally and it worked flawlessly even on oracle linux. I also re-ran the checks and they all passed (see [here](https://github.com/akulumbeg/acme.sh/actions), commit 33 String update). 

I came to the conclusion that when I was testing this a month ago, there was actually a short-term issue on the webhosting provider server, because in fact it gave me a 400 error only one one platform, without any clashes on the others.

This being said, I am confident that this implementation works, it passes all the checks, it has been tested locally on at least three platforms in DNS and Standalone modes and I think it is ready to be pushed to the dev repo, and then to master.

Neilpang:
did you create an issue for bug reporting ?

akulumbeg:
Thank you for reminding me, I did now here #3486

PR #3397, opened 2021-02-07
duckdns - fix "integer expression expected" errors

jerrm:


jerrm:
This is a trivial fix to duckdns to remove error messages when the DEBUG var is not set.

PR #3428, opened 2021-02-25
Workaround for DogtagPKI - Challenge is already valid

johnruble:
Dogtag was returning "<h1>HTTP Status 500 – Internal Server Error</h1><hr class="line" /><p><b>Type</b> Exception Report</p><p><b>Message</b> java.lang.Exception: Challenge is already valid</p>" instead of a nice clean "status":"valid".

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4705,6 +4705,15 @@ $_authorizations_map"
       response="$(echo "$response" | _normalizeJson)"
       _debug2 response "$response"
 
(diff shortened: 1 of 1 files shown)


Neilpang:
Is that a bug of dogtagPKI ?
It's better if they can fix it to follow the RFC8555.

johnruble:
> Is that a bug of dogtagPKI ?
> It's better if they can fix it to follow the RFC8555.

Indeed it would be.  I just needed a fix now to get around a hurdle at work and could not wait.

PR #3430, opened 2021-02-26
dnsapi/ionos: Update API to use POST instead of PATCH for adding TXT records

lbrocke:
The IONOS DNS API now supports a POST route for adding records. Therefore checking for already existing records and including them in a PATCH request is no longer necessary. This does not affect functionality but rather simplifies the `dns_ionos_add()` function.

[Docker/MacOS and Windows DNS tests are passing](https://github.com/lbrocke/acme.sh/actions/runs/593583016), FreeBSD fails due to an package installation error unrelated to this code change:
```
[..]
/usr/bin/ssh -t freebsd
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: no access to tty (Bad file descriptor).
Thus no job control in this shell.
Installing pkg-1.16.3...
Newer FreeBSD version for package pkg:
To ignore this error set IGNORE_OSVERSION=yes
- package: ***2000
- running kernel: ***1000
Ignore the mismatch and continue? [y/N]: 
Failed to install the following 1 package(s): /tmp//pkg.txz.6vPgxn
Bootstrapping pkg from pkg+http://pkg.FreeBSD.org/FreeBSD:12:amd64/quarterly, please wait...
Verifying signature with trusted certificate pkg.freebsd.org.2013102301... done
Error: The process '/usr/bin/ssh' failed with exit code 1
```

### DIFF
--- a/dnsapi/dns_ionos.sh
+++ b/dnsapi/dns_ionos.sh
@@ -24,20 +24,9 @@ dns_ionos_add() {
     return 1
   fi
 
-  _new_record="{\"name\":\"$_sub_domain.$_domain\",\"type\":\"TXT\",\"content\":\"$txtvalue\",\"ttl\":$IONOS_TXT_TTL,\"prio\":$IONOS_TXT_PRIO,\"disabled\":false}"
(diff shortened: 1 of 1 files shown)


Neilpang:
rebase your code to latest dev branch first.  the freebsd is fixed.

lbrocke:
Done, [tests are passing now](https://github.com/lbrocke/acme.sh/actions/runs/603010256).

PR #3431, opened 2021-02-27
Add http webhook support

ashokgelal:
I understand I need to update the [Notify wiki page](https://github.com/acmesh-official/acme.sh/wiki/notify) as well but I wanted to send this PR first to make sure if there is no opposition to merging this. Please let me know if this might get approved or not so that I can start working on updating the wiki. Either way, this is a great project. Thank you for your hard work!

### DIFF
--- a/notify/http.sh
+++ b/notify/http.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/bash
+
+#Support http webhooks api
+
+#HTTP_WEBHOOK="xxxx"
+
+# subject content statusCode
+http_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
+  _debug "_subject" "$_subject"
+  _debug "_content" "$_content"
+  _debug "_statusCode" "$_statusCode"
+
+  HTTP_WEBHOOK="${HTTP_WEBHOOK:-$(_readaccountconf_mutable HTTP_WEBHOOK)}"
+  if [ -z "$HTTP_WEBHOOK" ]; then
+    HTTP_WEBHOOK=""
+    _err "You didn't specify a http webhook HTTP_WEBHOOK yet."
+    return 1
+  fi
+  _saveaccountconf_mutable HTTP_WEBHOOK "$HTTP_WEBHOOK"
+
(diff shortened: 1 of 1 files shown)


Neilpang:
thanks for your contribute, but do you think this is much reusable for other people?  because the POST format may be vary for different users .

ashokgelal:
Thanks for having a look. I needed this for my own purpose so I thought I would contribute back. I believe that supporting a http hook opens up a lot of possibilities for advanced use cases and also I feel like http is so fundamental supporting it doesn't hurt.

As far as the formatting goes, like I said, this is probably for advanced use case and whoever needs this might know how to digest an incoming request with a json payload. In future, we can always extend this to support more formats and maybe whoever needs another format might send another PR just like what I did :)

I totally get your concern though so I won't feel bad just closing this. I could just copy it and make it work for my use case. Although, having it official would be great :)

Neilpang:
let's keep this PR open to see if there will be more users who need it.

Thanks.

christianbur:
Some time ago I also had the problem that I had to access my own backend with acme.sh. 
I used the dnsapi "[dns_acmeproxy](https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_acmeproxy.sh)" for my purposes.
This plugin is very easy to customize for own backends and the project "lego" even uses the api of acmeproxy as a generic plugin (https://go-acme.github.io/lego/dns/httpreq/).

[acmeproxy](https://github.com/mdbraber/acmeproxy/)

>    Works with the httpreq DNS challenge provider in lego and with the acmeproxy provider in acme.sh (currently in the dev branch).

Therefore acme.sh doesn't need a new generic dnsapi (in my opinion), rather the description for dns_acmeproxy should be adapted, that this api can also be used well for own projects.

ashokgelal:
@christianbur Correct me if I’m wrong but I’m not sure if we are talking about the same thing. My PR is for adding a http webhook for sending notifications and you pointed something to dns api? What am I missing?

christianbur:
Sorry, I was wrong :-)

notify/http.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/bash
```
liuyangc3, line 1: shouldn't be `env sh`?

PR #3435, opened 2021-03-02
Add --always-force-reuse-domain-key option (useful for pinning)

marcoarment:
For applications that use pinning against known-valid private keys, this option forces the reuse of the existing private key and fails if it cannot be used, rather than automatically generating a new one when needed.

This is the behavior of certbot's `--reuse-key` option, and effectively the opposite of acme.sh's existing `--always-force-new-domain-key` option (from which I derived the name — I frankly don't care what it's called, as long as it exists).


Neilpang:
>> and effectively the opposite of acme.sh's existing --always-force-new-domain-key option 

Yes,  so we do we need another opposite option "--always-force-reuse-domain-key"  ?

Issue #3143, opened 2020-09-02
dns_constellix fails for wildcard domains


resaperk:
When trying to issue a cert for example.com and *.example.com, using dns-01 with constellix, dns_constellix.sh attempts to create the same TXT record for "_acme-challenge.example.com" twice, and fails "already exists" on the second

Steps to reproduce
------------------
```
acme.sh --issue \
    --dns dns_constellix
    -d example.com
    -d *.example.com
```

Debug log
-----------------

```
Lets find script dir.
[Wed Sep  2 16:21:34 CDT 2020] _SCRIPT_='/usr/local/sbin/acme.sh'
[Wed Sep  2 16:21:34 CDT 2020] _script='/usr/local/sbin/acme.sh'
[Wed Sep  2 16:21:34 CDT 2020] _script_home='/usr/local/sbin'
[Wed Sep  2 16:21:34 CDT 2020] Using config home:/var/db/acme/.acme.sh
[Wed Sep  2 16:21:34 CDT 2020] Running cmd: issue
[Wed Sep  2 16:21:34 CDT 2020] _main_domain='example.com'
[Wed Sep  2 16:21:34 CDT 2020] _alt_domains='*.example.com'
[Wed Sep  2 16:21:34 CDT 2020] Using config home:/var/db/acme/.acme.sh
[Wed Sep  2 16:21:34 CDT 2020] ACME_DIRECTORY='https://acme-v02.api.letsencrypt.org/directory'
[Wed Sep  2 16:21:34 CDT 2020] DOMAIN_PATH='/var/db/acme/certs/example.com'
[Wed Sep  2 16:21:34 CDT 2020] Using ACME_DIRECTORY: https://acme-v02.api.letsencrypt.org/directory
[Wed Sep  2 16:21:34 CDT 2020] _init api for server: https://acme-v02.api.letsencrypt.org/directory
[Wed Sep  2 16:21:34 CDT 2020] GET
[Wed Sep  2 16:21:34 CDT 2020] url='https://acme-v02.api.letsencrypt.org/directory'
[Wed Sep  2 16:21:34 CDT 2020] timeout=
[Wed Sep  2 16:21:34 CDT 2020] _CURL='curl -L --silent --dump-header /var/db/acme/.acme.sh/http.header  -g '
[Wed Sep  2 16:21:35 CDT 2020] ret='0'
[...]

PR #3445, opened 2021-03-10
Add wildcard certificate support for dns_constellix

woutd:
With these changes, the plugin will detect if the record already exists and update it instead of adding a new one (which fails). This adds support for certain wildcard certificates (#3143) where multiple TXT values should be added for the same `_sub_domain`.

I also decreased the TTL from 120 to 60 seconds and some minor cleanup.

I am using it in production already, but extra testing is always welcome.

### DIFF
--- a/dnsapi/dns_constellix.sh
+++ b/dnsapi/dns_constellix.sh
@@ -30,7 +30,7 @@ dns_constellix_add() {
     return 1
   fi
 
-  # The TXT record might already exist when working with wilcard certificates. In that case, update the record by adding the new value.
+  # The TXT record might already exist when working with wildcard certificates. In that case, update the record by adding the new value.
   _debug "Search TXT record"
   if _constellix_rest GET "domains/${_domain_id}/records/TXT/search?exact=${_sub_domain}"; then
     if printf -- "%s" "$response" | grep "{\"errors\":\[\"Requested record was not found\"\]}" >/dev/null; then
--- a/dnsapi/dns_constellix.sh
+++ b/dnsapi/dns_constellix.sh
@@ -44,7 +44,7 @@ dns_constellix_add() {
         fi
       fi
     else
-      _record_id=$(printf "%s\n" "$response" | _egrep_o "\"id\":[0-9]+" | cut -d ':' -f 2)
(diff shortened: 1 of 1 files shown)


Neilpang:
can you please fix the CI checks first ?

https://github.com/woutd/acme.sh/actions/workflows/DNS.yml

woutd:
I am following the instructions from https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

Repository Actions seemed to be enabled already on the fork, so I did not change anything there.

Next, in the "Secrets" menu I can choose "Environment" or "Action" secrets, the wiki does not specify this so I went with "Actions" as that seems to be the kind of default when clicking on the "New repository secret". So I created the following "Action secrets":

```
TokenName1 = CONSTELLIX_Key
TokenValue1 = XXX (hidden)
TokenName2 = CONSTELLIX_Secret
TokenValue2 = XXX (hidden)
TEST_DNS = dns_constellix
TestingDomain = foobar.rambla.eu
TEST_DNS_SLEEP = 120
```

Then I tried to retrigger the checks with an empty commit but it keeps failing with the same error.
Do I need to make them "Environments secrets" or do you know what I am doing wrong?

Screenshot_2021-03-14_16-23-13

woutd:
I managed to add the required secrets to the forked repository, and re-run the workflow jobs. But it seems something goes wrong with the docker image:

`centos:latest [FAIL]`

Not sure how I can fix this.

woutd:
I got it working, but it seems something goes wrong in the `_get_root` function on Solaris. Not sure what is the problem here.

Neilpang:
you can use this project to debug on Solaris.

woutd:
I managed to find the issue on Solaris. The problem is here:
`_domain_id=$(printf "%s\n" "$response" | _egrep_o "\"id\":[0-9]+" | cut -d ':' -f 2)`
While `[0-9]+` is a valid regex it does not seem to work on Solaris. I can replace it with `[0-9]*`.
(I could also use `[0-9]\{1,\}`, which is a bit more correct, but harder to read. Also, this one does not seem to work on the other operating systems.)

I suspect this to be an issue in a lot of other plugins as well:
`
grep "\[0-9\]+" dnsapi/* | wc -l
34
`

woutd:
Needed some more changes... but finally got it fully working on Solaris.
Screenshot_2021-03-24_15-44-42

PR #3455, opened 2021-03-22
fix _exists error message if MAIL_BIN env variable is not set

ecm75:
This PR fixes the behavior of mail.sh notify.
Before it returned an "error" message from the function _exists if the MAIL_BIN environment variable was not set, because it would call _exists with an empty string resulting in crontab mails stating
`
Usage: _exists cmd
`

### DIFF
--- a/notify/mail.sh
+++ b/notify/mail.sh
@@ -78,8 +78,13 @@ mail_send() {
 
 _mail_bin() {
   _MAIL_BIN=""
+	_MAIL_BINS="sendmail ssmtp mutt mail msmtp"
 
-  for b in "$MAIL_BIN" sendmail ssmtp mutt mail msmtp; do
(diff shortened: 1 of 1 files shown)


notify/mail.sh
```diff
@@ -78,8 +78,13 @@ mail_send() {
 
 _mail_bin() {
   _MAIL_BIN=""
+  _MAIL_BINS="sendmail ssmtp mutt mail msmtp"
 
-  for b in "$MAIL_BIN" sendmail ssmtp mutt mail msmtp; do
```
Neilpang, line 82: why not just remove the double quotation marks here?

```
for b in $MAIL_BIN sendmail ssmtp mutt mail msmtp; do
```

ecm75: :facepalm: So true... it's so simple, yet I failed to see that. Thank you.

PR #3491, opened 2021-04-28
Consul deploy hooks

bgarret:
Consul API and CLI deploy hooks, inspired by the Vault deploy hooks.

### DIFF
--- a/deploy/consul.sh
+++ b/deploy/consul.sh
@@ -18,81 +18,81 @@
 #domain keyfile certfile cafile fullchain
 consul_deploy() {
 
-	_cdomain="$1"
-	_ckey="$2"
-	_ccert="$3"
-	_cca="$4"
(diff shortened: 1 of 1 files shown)


Neilpang:
why are there 2 hooks?
why not make them as one?

Default behavior: find and use cli first if cli exists, otherwise,  try to us API.
There can also be a variable  : `USE_CLI` to force to use cli.

bgarret:
There are 2 hooks because the Vault integration is structured this way, I used it as a reference and did not think too much on it.

Your way is indeed better, I'll update the pull request with the modifications, and update the wiki docs once the pull request is merged.

Thanks!

bgarret:
Changes done

bgarret:
Do you need something else to get this merged? I'm happy to help in any way I can.

PR #3539, opened 2021-06-05
Add DNS API plugin for Oracle Cloud Infrastructure DNS Service

Djelibeybi:
This plugin is has noticeably more required fields than most
other plugins due to the requirement that all requests to
the OCI REST API must be cryptographically signed by the client
using the draft standard proposed in <https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-08>

The OCI specific implementation details of the draft standard are
documented in the Developer Guide: <https://docs.oracle.com/en-us/iaas/Content/API/Concepts/signingrequests.htm#five>

> **NOTE**: there is maximum allowed clock skew of five minutes between the
client and the API endpoint. Requests will be denied if the skew is
greater.

This PR also includes a minor tweak to the Solaris job in the DNS
workflow so that it uses the pre-installed GNU tools, curl and OpenSSL 1.1.1.
Without these changes, the signature generation function does not
work on Solaris.

Signed-off-by: Avi Miller <avi.miller@oracle.com>

### DIFF
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -59,7 +59,7 @@ jobs:
       run: cd .. && git clone https://github.com/acmesh-official/acmetest.git  && cp -r acme.sh acmetest/
     - name: Set env file
       run: |
-        cd ../acmetest 
+        cd ../acmetest
         if [ "${{ secrets.TokenName1}}" ] ; then
           echo "${{ secrets.TokenName1}}=${{ secrets.TokenValue1}}" >> env.list
         fi
@@ -75,7 +75,7 @@ jobs:
         if [ "${{ secrets.TokenName5}}" ] ; then
           echo "${{ secrets.TokenName5}}=${{ secrets.TokenValue5}}" >> env.list
         fi
-        echo "TEST_DNS_NO_WILDCARD" >> env.list 
+        echo "TEST_DNS_NO_WILDCARD" >> env.list
         echo "TEST_DNS_SLEEP" >> env.list
     - name: Run acmetest
       run: cd ../acmetest && ./rundocker.sh  testall
@@ -226,8 +226,10 @@ jobs:
     - uses: vmactions/solaris-vm@v0.0.3
       with:
         envs: 'TEST_DNS TestingDomain TEST_DNS_NO_WILDCARD TEST_DNS_SLEEP CASE TEST_LOCAL DEBUG ${{ secrets.TokenName1}} ${{ secrets.TokenName2}} ${{ secrets.TokenName3}} ${{ secrets.TokenName4}} ${{ secrets.TokenName5}}'
-        prepare: pkgutil -y -i socat curl
+        prepare: pkgutil -y -i socat
         run: |
+          pkg set-mediator -v -I default@1.1 openssl
+          export PATH=/usr/gnu/bin:$PATH
           if [ "${{ secrets.TokenName1}}" ] ; then
             export ${{ secrets.TokenName1}}=${{ secrets.TokenValue1}}
           fi
@@ -245,5 +247,3 @@ jobs:
           fi
           cd ../acmetest
           ./letest.sh
-
-        
--- a/dnsapi/dns_oci.sh
+++ b/dnsapi/dns_oci.sh
@@ -166,7 +166,6 @@ _oci_config() {
 
 }
 
-
 # _get_zone(): retrieves the Zone name and OCID
 #
 # _sub_domain=_acme-challenge.www
@@ -321,6 +320,5 @@ _readini() {
   _inivalue="$(printf "%b" "$(eval "echo $_lineini | sed \"s/^ *${_key} *= *//g\"")")"
   _debug2 _inivalue "$_inivalue"
   echo "$_inivalue"
-
 
 }
--- a/dnsapi/dns_oci.sh
+++ b/dnsapi/dns_oci.sh
@@ -3,7 +3,10 @@
 # Acme.sh DNS API plugin for Oracle Cloud Infrastructure
 # Copyright (c) 2021, Oracle and/or its affiliates
 #
-# Required OCI CLI environment variables:
+# The plugin will automatically use the default profile from an OCI SDK and CLI
+# configuration file, if it exists.
+#
+# Alternatively, set the following environment variables:
 # - OCI_CLI_TENANCY : OCID of tenancy that contains the target DNS zone
 # - OCI_CLI_USER    : OCID of user with permission to add/remove records from zones
 # - OCI_CLI_REGION  : Should point to the tenancy home region
@@ -19,18 +22,15 @@ dns_oci_add() {
   _fqdn="$1"
   _rdata="$2"
 
-  if _oci_config; then
+  if _get_oci_zone; then
 
-    if [ "$_sub_domain" ] && [ "$_domain" ]; then
-      _add_record_body="{\"items\":[{\"domain\":\"${_sub_domain}.${_domain}\",\"rdata\":\"$_rdata\",\"rtype\":\"TXT\",\"ttl\": 30,\"operation\":\"ADD\"}]}"
-      response=$(_signed_request "PATCH" "/20180115/zones/${_domain}/records" "$_add_record_body")
-      if [ "$response" ]; then
-        _info "Success: added TXT record for ${_sub_domain}.${_domain}."
-      else
-        _err "Error: failed to add TXT record for ${_sub_domain}.${_domain}."
-        return 1
-      fi
+    _add_record_body="{\"items\":[{\"domain\":\"${_sub_domain}.${_domain}\",\"rdata\":\"$_rdata\",\"rtype\":\"TXT\",\"ttl\": 30,\"operation\":\"ADD\"}]}"
+    response=$(_signed_request "PATCH" "/20180115/zones/${_domain}/records" "$_add_record_body")
+    if [ "$response" ]; then
+      _info "Success: added TXT record for ${_sub_domain}.${_domain}."
     else
+      _err "Error: failed to add TXT record for ${_sub_domain}.${_domain}."
+      _err "Check that the user has permission to add records to this zone."
       return 1
     fi
 
@@ -44,18 +44,15 @@ dns_oci_rm() {
   _fqdn="$1"
   _rdata="$2"
 
-  if _oci_config; then
+  if _get_oci_zone; then
 
-    if [ "$_sub_domain" ] && [ "$_domain" ]; then
-      _remove_record_body="{\"items\":[{\"domain\":\"${_sub_domain}.${_domain}\",\"rdata\":\"$_rdata\",\"rtype\":\"TXT\",\"operation\":\"REMOVE\"}]}"
-      response=$(_signed_request "PATCH" "/20180115/zones/${_domain}/records" "$_remove_record_body")
-      if [ "$response" ]; then
-        _info "Success: removed TXT record for ${_sub_domain}.${_domain}."
-      else
-        _err "Error: failed to remove TXT record for ${_sub_domain}.${_domain}."
-        return 1
-      fi
+    _remove_record_body="{\"items\":[{\"domain\":\"${_sub_domain}.${_domain}\",\"rdata\":\"$_rdata\",\"rtype\":\"TXT\",\"operation\":\"REMOVE\"}]}"
+    response=$(_signed_request "PATCH" "/20180115/zones/${_domain}/records" "$_remove_record_body")
+    if [ "$response" ]; then
+      _info "Success: removed TXT record for ${_sub_domain}.${_domain}."
     else
+      _err "Error: failed to remove TXT record for ${_sub_domain}.${_domain}."
+      _err "Check that the user has permission to remove records from this zone."
       return 1
     fi
 
@@ -66,12 +63,41 @@ dns_oci_rm() {
 }
 
 ####################  Private functions below ##################################
+_get_oci_zone() {
+
+  if ! _oci_config; then
+    return 1
+  fi
+
+  if ! _get_zone "$_fqdn"; then
+    _err "Error: DNS Zone not found for $_fqdn in $OCI_CLI_TENANCY"
+    return 1
+  fi
+
+  return 0
+
+}
+
 _oci_config() {
 
-  OCI_CLI_TENANCY="${OCI_CLI_TENANCY:-$(_readaccountconf_mutable OCI_CLI_TENANCY)}"
-  OCI_CLI_USER="${OCI_CLI_USER:-$(_readaccountconf_mutable OCI_CLI_USER)}"
-  OCI_CLI_KEY="${OCI_CLI_KEY:-$(_readaccountconf_mutable OCI_CLI_KEY)}"
-  OCI_CLI_REGION="${OCI_CLI_REGION:-$(_readaccountconf_mutable OCI_CLI_REGION)}"
+  OCI_CLI_CONFIG_FILE="${OCI_CLI_CONFIG_FILE:-$HOME/.oci/config}"
+  OCI_CLI_PROFILE="${OCI_CLI_PROFILE:-DEFAULT}"
+
+  # Let's try and find the values automagically first
+  # But still let any environment variables take precendence
+  if [ -f "$OCI_CLI_CONFIG_FILE" ]; then
+    _info "Reading OCI configuration file: $(_green "$OCI_CLI_CONFIG_FILE")"
+    OCI_CLI_TENANCY="${OCI_CLI_TENANCY:-$(_read_oci_config tenancy)}"
+    OCI_CLI_USER="${OCI_CLI_USER:-$(_read_oci_config user)}"
+    OCI_CLI_KEY_FILE="${OCI_CLI_KEY_FILE:-$(_read_oci_config key_file)}"
+    OCI_CLI_REGION="${OCI_CLI_REGION:-$(_read_oci_config region)}"
+  else
+    OCI_CLI_TENANCY="${OCI_CLI_TENANCY:-$(_readaccountconf_mutable OCI_CLI_TENANCY)}"
+    OCI_CLI_USER="${OCI_CLI_USER:-$(_readaccountconf_mutable OCI_CLI_USER)}"
+    OCI_CLI_KEY="${OCI_CLI_KEY:-$(_readaccountconf_mutable OCI_CLI_KEY)}"
+    OCI_CLI_REGION="${OCI_CLI_REGION:-$(_readaccountconf_mutable OCI_CLI_REGION)}"
+    _save_config="true"
+  fi
 
   _not_set=""
   _ret=0
@@ -85,7 +111,7 @@ _oci_config() {
     if [ -f "$OCI_CLI_KEY_FILE" ]; then
       OCI_CLI_KEY=$(_base64 <"$OCI_CLI_KEY_FILE")
     else
-      _err "Fatal: unable to read $OCI_CLI_KEY_FILE."
+      _err "Fatal: unable to read key file: $OCI_CLI_KEY_FILE"
       return 1
     fi
   fi
@@ -106,19 +132,18 @@ _oci_config() {
     _err "Fatal: required environment variable(s): ${_not_set} not set."
     _ret=1
   else
-    _saveaccountconf_mutable OCI_CLI_TENANCY "$OCI_CLI_TENANCY"
-    _saveaccountconf_mutable OCI_CLI_USER "$OCI_CLI_USER"
-    _saveaccountconf_mutable OCI_CLI_KEY "$OCI_CLI_KEY"
-    _saveaccountconf_mutable OCI_CLI_REGION "$OCI_CLI_REGION"
+    if [ "$_save_config" ]; then
+      _saveaccountconf_mutable OCI_CLI_TENANCY "$OCI_CLI_TENANCY"
+      _saveaccountconf_mutable OCI_CLI_USER "$OCI_CLI_USER"
+      _saveaccountconf_mutable OCI_CLI_KEY "$OCI_CLI_KEY"
+      _saveaccountconf_mutable OCI_CLI_REGION "$OCI_CLI_REGION"
+    else
+      _info "Success: OCI configuration retrieved from $OCI_CLI_CONFIG_FILE."
+    fi
   fi
 
   if ! _contains "PRIVATE KEY" "$OCI_CLI_KEY"; then
     OCI_CLI_KEY=$(printf "%s" "$OCI_CLI_KEY" | _dbase64 multiline)
-  fi
-
-  if ! _get_zone "$_fqdn"; then
-    _err "Error: DNS Zone not found for $_fqdn."
-    _ret=1
   fi
 
   return $_ret
@@ -242,5 +267,39 @@ _signed_request() {
 
   printf "%s" "$_return"
   return $_ret
+
+}
+
+# file  key  [section]
+_read_oci_config() {
+  _key="$1"
+
+  _start_n=$(grep -n '\['"$OCI_CLI_PROFILE"']' "$OCI_CLI_CONFIG_FILE" | cut -d : -f 1)
+  _debug2 _start_n "$_start_n"
+  if [ -z "$_start_n" ]; then
+    _err "Can not find section: $OCI_CLI_PROFILE"
+    return 1
+  fi
+
+  _start_nn=$(_math "$_start_n" + 1)
+  _debug2 "_start_nn" "$_start_nn"
+
+  _left="$(sed -n "${_start_nn},99999p" "$OCI_CLI_CONFIG_FILE")"
+  _debug2 _left "$_left"
+  _end="$(echo "$_left" | grep -n "^\[" | _head_n 1)"
+  _debug2 "_end" "$_end"
+  if [ "$_end" ]; then
+    _end_n=$(echo "$_end" | cut -d : -f 1)
+    _debug "_end_n" "$_end_n"
+    _seg_n=$(echo "$_left" | sed -n "1,${_end_n}p")
+  else
+    _seg_n="$_left"
+  fi
+
+  _debug2 "_seg_n" "$_seg_n"
+  _lineini="$(echo "$_seg_n" | grep "^ *$_key *= *")"
+
+  _debug2 "_lineini" "$_lineini"
+  printf "%b" "$(eval "echo $_lineini | sed -e \"s/${_key}[[:space:]]*=[[:space:]]*//g\"")"
 
 }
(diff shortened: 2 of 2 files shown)


Djelibeybi:
Here is the documentation that will be added to the [dnsapi](https://github.com/acmesh-official/acme.sh/wiki/dnsapi) wiki page:

## Using Oracle Cloud Infrastructure (OCI) DNS Services to automatically issue certificates

The Oracle Cloud Infrastructure Domain Name System (DNS) service lets you [create and manage DNS zones](https://docs.oracle.com/en-us/iaas/Content/DNS/Tasks/managingdnszones.htm#Managing_DNS_Service_Zones) in your tenancy. The DNS Service is limited to 1000 zones per account and 25,000  records per zone and is [charged per 1M queries](https://www.oracle.com/cloud/price-list.html#dns).

To use the OCI DNS Service, please follow the [OCI Developer Guide](https://docs.oracle.com/en-us/iaas/Content/API/Concepts/devtoolslanding.htm) to:

* [generate the required API signing key](https://docs.oracle.com/en-us/iaas/Content/API/Concepts/apisigningkey.htm)
* get your [tenancy and service account OCIDs](https://docs.oracle.com/en-us/iaas/Content/API/Concepts/apisigningkey.htm#five)

We recommend [installing the OCI CLI](https://docs.oracle.com/en-us/iaas/Content/API/Concepts/cliconcepts.htm) so that you can make ad-hoc changes as well. Once the CLI is installed and configured, the required values can be found in `~/.oci/config`.

The following four variables are mandatory:

```shell
export OCI_TENANCY="ocid1.tenancy.oc1..aaaaaaaa..."
export OCI_USER="ocid1.user.oc1..aaaaaaaa..."
export OCI_FINGERPRINT="xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx"
export OCI_PRIVATE_KEY="$(<~/.oci/oci_api_key.pem)"
```

Note that the `OCI_PRIVATE_KEY` variable will contain line breaks. If necesary, the plugin supports an alternative single-line string created by by using `openssl` to base64-encode the key, e.g.

```shell
export OCI_PRIVATE_KEY="$(cat ~/.oci/oci_api_key.pem | openssl enc -d -a -A)"
```

To issue a certificate, run:

```shell
acme.sh --issue --dns dns_oci -d example.com -d www.example.com
```

To issue a wildcard certificate, run:

```shell
acme.sh --issue --dns dns_oci -d example.com -d *.example.com
```

Please report any issues to `https://github.com/acmesh-official/acme.sh/issues/XXX`

Djelibeybi:
Could a maintainer please approve the workflows for this pull request?

Djelibeybi:
All the checks have passed. Is there anything else I need to do to get this merged?

Neilpang:
can you please check this guide first ?

please show me the usage.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,246 @@
...
+# - OCI_FINGERPRINT: fingerprint of the public key for the user
+# - OCI_PRIVATE_KEY: Path to private API signing key file in PEM format
+#
+# Optional environment variables:
+# - OCI_KEY_PASSPHRASE: if the private key above s encrypted, the passphrase is required
+# - OCI_REGION: Your home region will probably response the fastest
+#
+
+dns_oci_add() {
+  _fqdn="$1"
```
Neilpang, line 18: You can check if  `oci` is already installed on the machine,  if installed, you can prefer to use oci instead of curl.

And you can also another optional env variable:  `OCI_FORCE_GET` to force to use `curl`, default is empty,

Djelibeybi: I specifically don't use `oci` so that it's completely portable to platforms on which `oci` has not been tested/ported. There's no real benefit to using `oci` now, as it's actually slightly slower than the method I'm using (because it has to load in all of the Python SDK).

Djelibeybi:
> please show me the usage.

My intention is for the content in this issue: https://github.com/acmesh-official/acme.sh/issues/3540 to become the usage guide in the wiki as well.

dnsapi/dns_oci.sh
Neilpang: In case the `oci` is already installed and configured,  the user can just install acme.sh and use it, he doesn't need to configure for acme.sh again.

It's possible that the user will change the private key in futuren, he won't want to update twice.

And I believe `oci` is more safe to use.

Neilpang: this is the approach we use in the dns aws plugin.

Djelibeybi: Fair enough. I'll add `oci` support at the same time as removing the need to specify a fingerprint.

Neilpang:
can you please the env names in thie page:

https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/clienvironmentvariables.htm

Djelibeybi:
Huh, I totally _,meant_ to use the same names. Not sure why I didn't. Will fix.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,246 @@
...
+  OCI_FINGERPRINT="${OCI_FINGERPRINT:-$(_readaccountconf_mutable OCI_FINGERPRINT)}"
+  OCI_PRIVATE_KEY="${OCI_PRIVATE_KEY:-$(_readaccountconf_mutable OCI_PRIVATE_KEY)}"
+  OCI_KEY_PASSPHRASE="${OCI_KEY_PASSPHRASE:-$(_readaccountconf_mutable OCI_KEY_PASSPHRASE)}"
+  OCI_REGION="${OCI_REGION:-$(_readaccountconf_mutable OCI_REGION)}"
+
+  _not_set=""
+  _ret=0
+
+  if [ -f "$OCI_PRIVATE_KEY" ]; then
+    OCI_PRIVATE_KEY="$(openssl enc -a -A <"$OCI_PRIVATE_KEY")"
```
Neilpang, line 87: what is this for ?

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,246 @@
...
+  _request_target="(request-target): $_request_method $_sig_target"
+  _date_header="date: $_sig_now"
+  _host_header="host: $_sig_host"
+
+  _string_to_sign="$_request_target\n$_date_header\n$_host_header"
+  _sig_headers="(request-target) date host"
+
+  if [ "$_sig_body" ]; then
+    _secure_debug3 _sig_body "$_sig_body"
+    _sig_body_sha256="x-content-sha256: $(printf %s "$_sig_body" | openssl dgst -binary -sha256 | openssl enc -e -base64)"
```
Neilpang, line 201: dont use openssl here.

please use `_digest()` function.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,246 @@
...
+    _string_to_sign="$_string_to_sign\n$_sig_body_sha256\n$_sig_body_type\n$_sig_body_length"
+    _sig_headers="$_sig_headers x-content-sha256 content-type content-length"
+  fi
+
+  _tmp_file=$(_mktemp)
+  if [ -f "$_tmp_file" ]; then
+    printf '%s' "$OCI_PRIVATE_KEY" >"$_tmp_file"
+    # Double quoting the file and passphrase breaks openssl
+    # shellcheck disable=SC2086
+    _signature=$(printf '%b' "$_string_to_sign" | openssl dgst -sha256 -sign $_tmp_file $_sig_passinArg | openssl enc -e -base64 | tr -d '\r\n')
```
Neilpang, line 213: use `_sign` function

Neilpang:
don't use `openssl` in your code, please use the function instead.

dnsapi/dns_oci.sh
Djelibeybi: It's to encode the private key into a single line so that it could be passed via `env.list` for the Docker-based GitHub Actions tests. Using an `--env-file` with Docker doesn't support multi-line variiables (like a private key).

dnsapi/dns_oci.sh
Djelibeybi: Ok, have added this to my list of things to research and replace.

dnsapi/dns_oci.sh
Djelibeybi: Ok, have added this to my list of things to research and replace.

dnsapi/dns_oci.sh
Neilpang: can you please not use `openssl` here?  please find another solution.

Djelibeybi:
> don't use `openssl` in your code, please use the function instead.

Will do. May I ask the reasoning behind it? Just so i can understand it in context.

Neilpang:
> > don't use `openssl` in your code, please use the function instead.
> 
> Will do. May I ask the reasoning behind it? Just so i can understand it in context.


`openssl` has different options in different versions  and different OS.  There is also a libreSSL that is emulating as `openssl`.

If everyone is using `openssl` everywhere, it will be out of control.

We need to restrict the use of openssl.    We can easily implement the functions without openssl.

Djelibeybi:
That's good information to know. Thanks!

dnsapi/dns_oci.sh
Djelibeybi: I just want to double-check this because I've reviewed `dns_aws.sh` and as far as I can tell, it only uses `_get()` and `_post()`. I'm happy to add support for using the OCI CLI, I just wanted to be sure that's what you meant.

Djelibeybi:
I've made (almost) all the requested changes except implement support for using the `oci` client directly, pending feedback on how that should actually be done.

Neilpang:
please update the usage.
And make it a single page in the wiki, and then add it in to this page: https://github.com/acmesh-official/acme.sh/wiki/dnsapi

Thanks.

Djelibeybi:
Wiki page added: <https://github.com/acmesh-official/acme.sh/wiki/Using-Oracle-Cloud-Infrastructure-DNS> and [dnsapi](https://github.com/acmesh-official/acme.sh/wiki/dnsapi) updated.

.github/workflows/DNS.yml
```diff
@@ -226,8 +226,10 @@ jobs:
     - uses: vmactions/solaris-vm@v0.0.3
       with:
         envs: 'TEST_DNS TestingDomain TEST_DNS_NO_WILDCARD TEST_DNS_SLEEP CASE TEST_LOCAL DEBUG ${{ secrets.TokenName1}} ${{ secrets.TokenName2}} ${{ secrets.TokenName3}} ${{ secrets.TokenName4}} ${{ secrets.TokenName5}}'
-        prepare: pkgutil -y -i socat curl
+        prepare: pkgutil -y -i socat
         run: |
+          pkg set-mediator -v -I default@1.1 openssl
```
Neilpang, line 231: Is this change a must ?

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,242 @@
...
+    if [ "$_sub_domain" ] && [ "$_domain" ]; then
+      _add_record_body="{\"items\":[{\"domain\":\"${_sub_domain}.${_domain}\",\"rdata\":\"$_rdata\",\"rtype\":\"TXT\",\"ttl\": 30,\"operation\":\"ADD\"}]}"
+      response=$(_signed_request "PATCH" "/20180115/zones/${_domain}/records" "$_add_record_body")
+      if [ "$response" ]; then
+        _info "Success: added TXT record for ${_sub_domain}.${_domain}."
+      else
+        _err "Error: failed to add TXT record for ${_sub_domain}.${_domain}."
+        return 1
+      fi
+    fi
```
Neilpang, line 33: there is no "else" for this "if".

.github/workflows/DNS.yml
Djelibeybi: Yes, we need OpenSSL 1.1 on Solaris to correctly sign our requests. And it's installed by default. This command just switches to using it .

dnsapi/dns_oci.sh
Djelibeybi: Correct.  If there is no _sub_domain and _domain, we can't add a record, so we just return.

Neilpang: I mean you should add an `else` to `return 1`, it's missing.

Neilpang:
According to this page: https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/clienvironmentvariables.htm

The `OCI_CLI_*` variables can also be saved in the config file: `~/.oci/config`.

Can we try to read them from the config file when the user doesn't set env variables?
And in this case, we don't need to save the variables to the account conf.

dnsapi/dns_oci.sh
Djelibeybi: Ah, ok. Will fix

Djelibeybi:
> According to this page: https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/clienvironmentvariables.htm
> 
> The `OCI_CLI_*` variables can also be saved in the config file: `~/.oci/config`.
> 
> Can we try to read them from the config file when the user doesn't set env variables?
> And in this case, we don't need to save the variables to the account conf.

Not quite. The settings in `~/.oci/config` can be overriden by the environment variables, but they're not the same format at all. The `~/.oci/config` file is in INI format and the field names are different. You can't (easily) source it from a shell.

Neilpang:
> > According to this page: https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/clienvironmentvariables.htm
> > The `OCI_CLI_*` variables can also be saved in the config file: `~/.oci/config`.
> > Can we try to read them from the config file when the user doesn't set env variables?
> > And in this case, we don't need to save the variables to the account conf.
> 
> Not quite. The settings in `~/.oci/config` can be _overriden_ by the environment variables, but they're not the same format at all. The `~/.oci/config` file is in INI format and the field names are different. You can't (easily) source it from a shell.
> 
> It looks like this:
> 
> ```
> [DEFAULT]
> user=ocid1.user.oc1..aaaaaaaas....
> fingerprint=b0:e1:58:b6:de:3.,...
> key_file=/Users/username/.oci/oci_api_key.pem
> tenancy=ocid1.tenancy.oc1..aaaaaaaawdfda...
> region=ap-melbourne-1
> ```


Yes,  I knew it's in ini format, and the name is different.

If a user already configured oci config on his machine,  acme.sh can work with Zero configurations.

Djelibeybi:
> If a user already configured oci config on his machine, acme.sh can work with Zero configurations.

Only if there is a single config stanza, otherwise the user would have to provide the profile name, which means we'd have to parse for that, which makes it even more compliicatied. Likewise, this config is not fixed in any way, while the values provided/required by the environment variables. are. To be honest, I'd rather stick to a single method of configuring the plugiun regardless of whether the user has installed our CLI or not.

Neilpang:
> > If a user already configured oci config on his machine, acme.sh can work with Zero configurations.
> 
> Only if there is a single config stanza, otherwise the user would have to provide the profile name, which means we'd have to parse for that, which makes it even more complicated. Likewise, this config is not fixed in any way, while the values provided/required by the environment variables. are. To be honest, I'd rather stick to a single method of configuring the plugiun regardless of whether the user has installed our CLI or not.



You can use thie function  to read ini file:
```
# file  key  [section]
_readini() {
  _file="$1"
  _key="$2"
  _section="${3:-DEFAULT}"

  _start_n=$(grep -n '\['$_section']' "$_file" | cut -d : -f 1)
  _debug2 _start_n "$_start_n"
  if [ -z "$_start_n" ]; then
  	_err "Can not find section: $_section"
  	return 1
  fi

  _start_nn=$(_math $_start_n + 1)
  _debug2 "_start_nn" "$_start_nn"

  _left="$(sed -n "${_start_nn},99999p" "$_file")"
  _debug2 _left "$_left"
  _end="$(echo "$_left" | grep -n "^\[" | _head_n 1)"
  _debug2 "_end" "$_end"
  if [ "$_end" ]; then
    _end_n=$(echo "$_end" | cut -d : -f 1)
    _debug "_end_n" "$_end_n"
    _seg_n=$(echo "$_left" | sed -n "1,${_end_n}p")
  else
    _seg_n="$_left"
  fi

  _debug2 "_seg_n" "$_seg_n"
  _lineini="$(echo "$_seg_n" | grep "^ *$_key *= *")"

  _debug2 "_lineini" "$_lineini"
  echo "$(
      eval "$_lineini"
      eval "printf \"%b\" \"\$$_key\""
  )"

}
```

The usage is simple:

```

_readini   /path/to/oci.config     compartment-id

or:  

_readini   /path/to/oci.config     compartment-id   DEFAULT


```

Djelibeybi:
In the case where the config is automatically pulled from the configuration file, does it still need to be saved into `account.conf` or should it not be saved so that it's re-read at renewal time?

Neilpang:
> In the case where the config is automatically pulled from the configuration file, does it still need to be saved into `account.conf` or should it not be saved so that it's re-read at renewal time?

No,  don't save it in this case.    it will be read again from the configuration file when renewall.

Djelibeybi:
I tweaked your function slightly and I've updated the wiki to reflect that the configuration is now automatic if an OCI SDK configuration file exists.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,305 @@
...
+    _seg_n=$(echo "$_left" | sed -n "1,${_end_n}p")
+  else
+    _seg_n="$_left"
+  fi
+
+  _debug2 "_seg_n" "$_seg_n"
+  _lineini="$(echo "$_seg_n" | grep "^ *$_key *= *")"
+
+  _debug2 "_lineini" "$_lineini"
+  printf "%b" "$(eval "echo $_lineini | sed -e \"s/${_key}[[:space:]]*=[[:space:]]*//g\"")"
```
Neilpang, line 303: please revert back.  the "[[:space:]]" tag is not available on all the platforms,  that's why I used the literal space key ` `.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,305 @@
...
+  printf "%s" "$_return"
+  return $_ret
+
+}
+
+# file  key  [section]
+_read_oci_config() {
+  _key="$1"
+
+  _start_n=$(grep -n '\['"$OCI_CLI_PROFILE"']' "$OCI_CLI_CONFIG_FILE" | cut -d : -f 1)
```
Neilpang, line 277: please also revert back, and don't use `OCI_CLI_PROFILE` variable here.
In future, if another dns apid needs to read ini file, I can easily just move this function into acme.sh, so that anyone can use it.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,305 @@
...
+
+}
+
+_oci_config() {
+
+  OCI_CLI_CONFIG_FILE="${OCI_CLI_CONFIG_FILE:-$HOME/.oci/config}"
+  OCI_CLI_PROFILE="${OCI_CLI_PROFILE:-DEFAULT}"
+
+  # Let's try and find the values automagically first
+  # But still let any environment variables take precendence
```
Neilpang, line 87: I think we should try `OCI_CLI_*` env variables first,  if the env variables are not found, then we try to find and use the config file.

Djelibeybi: Sorry, no. This is the opposite to what the OCI CLI and all the SDKs do, so it would be unexpected behaviour. Config is always least to most specific, i.e. default -> disk -> environment -> instantiantion (command-line)

dnsapi/dns_oci.sh
Djelibeybi: Righto.

dnsapi/dns_oci.sh
Djelibeybi: Your sample function didn't work as written, so I had to modify it to return the right result. I'll tweak it so that it accepts zero or more spaces either side of the equals sign.

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,307 @@
...
+  if ! _get_zone "$_fqdn"; then
+    _err "Error: DNS Zone not found for $_fqdn in $OCI_CLI_TENANCY"
+    return 1
+  fi
+
+  return 0
+
+}
+
+_oci_config() {
```
Neilpang, line 81: please use this function instead:
```
_oci_config() {

  _DEFAULT_OCI_CLI_CONFIG_FILE="$HOME/.oci/config"
  OCI_CLI_CONFIG_FILE="${OCI_CLI_CONFIG_FILE:-$(_readaccountconf_mutable OCI_CLI_CONFIG_FILE)}"

  if [ "$_DEFAULT_OCI_CLI_CONFIG_FILE" != "$OCI_CLI_CONFIG_FILE" ]; then
    _saveaccountconf_mutable OCI_CLI_CONFIG_FILE "$OCI_CLI_CONFIG_FILE"
  else
    OCI_CLI_CONFIG_FILE="$_DEFAULT_OCI_CLI_CONFIG_FILE"
    _clearaccountconf_mutable OCI_CLI_CONFIG_FILE
  fi

  OCI_CLI_PROFILE="${OCI_CLI_PROFILE:-$(_readaccountconf_mutable OCI_CLI_PROFILE)}"
  if [ "$OCI_CLI_PROFILE" ]; then
    _saveaccountconf_mutable OCI_CLI_PROFILE "$OCI_CLI_PROFILE"
  else
    _clearaccountconf_mutable OCI_CLI_PROFILE
  fi


  OCI_CLI_TENANCY="${OCI_CLI_TENANCY:-$(_readaccountconf_mutable OCI_CLI_TENANCY)}"
  if [ "$OCI_CLI_TENANCY" ]; then
    _saveaccountconf_mutable OCI_CLI_TENANCY "$OCI_CLI_TENANCY"
  elif && [ -f "$OCI_CLI_CONFIG_FILE" ]; then
    _debug "Try read tenancy from $OCI_CLI_CONFIG_FILE"
    OCI_CLI_TENANCY="${OCI_CLI_TENANCY:-$(_readini "$OCI_CLI_CONFIG_FILE" tenancy "$OCI_CLI_PROFILE")}"
  fi
  if [ -z "$OCI_CLI_TENANCY" ]; then
    _err "Can not read OCI_CLI_TENANCY"
    return 1
  fi

  OCI_CLI_USER="${OCI_CLI_USER:-$(_readaccountconf_mutable OCI_CLI_USER)}"
  if [ "$OCI_CLI_USER" ]; then
    _saveaccountconf_mutable OCI_CLI_USER "$OCI_CLI_USER"
  elif && [ -f "$OCI_CLI_CONFIG_FILE" ]; then
    _debug "Try read tenancy from $OCI_CLI_CONFIG_FILE"
    OCI_CLI_USER="${OCI_CLI_USER:-$(_readini "$OCI_CLI_CONFIG_FILE" user "$OCI_CLI_USER")}"
  fi
  if [ -z "$OCI_CLI_USER" ]; then
    _err "Can not read OCI_CLI_USER"
    return 1
  fi

  OCI_CLI_REGION="${OCI_CLI_REGION:-$(_readaccountconf_mutable OCI_CLI_REGION)}"
  if [ "$OCI_CLI_REGION" ]; then
    _saveaccountconf_mutable OCI_CLI_REGION "$OCI_CLI_REGION"
  elif && [ -f "$OCI_CLI_CONFIG_FILE" ]; then
    _debug "Try read region from $OCI_CLI_CONFIG_FILE"
    OCI_CLI_REGION="${OCI_CLI_REGION:-$(_readini "$OCI_CLI_CONFIG_FILE" region "$OCI_CLI_REGION")}"
  fi
  if [ -z "$OCI_CLI_REGION" ]; then
    _err "Can not read OCI_CLI_REGION"
    return 1
  fi


  OCI_CLI_KEY="${OCI_CLI_KEY:-$(_readaccountconf_mutable OCI_CLI_KEY)}"
  if [ "$OCI_CLI_KEY" ]; then
    _saveaccountconf_mutable OCI_CLI_KEY "$OCI_CLI_KEY"
  else
    _clearaccountconf_mutable OCI_CLI_KEY
    OCI_CLI_KEY_FILE="${OCI_CLI_KEY_FILE:-$(_readaccountconf_mutable OCI_CLI_KEY_FILE)}"
    if [ "$OCI_CLI_KEY_FILE" ]; then
      _saveaccountconf_mutable OCI_CLI_KEY_FILE "$OCI_CLI_KEY_FILE"
    elif && [ -f "$OCI_CLI_CONFIG_FILE" ]; then
      _debug "Try read region from $OCI_CLI_CONFIG_FILE"
      OCI_CLI_KEY_FILE="${OCI_CLI_KEY_FILE:-$(_readini "$OCI_CLI_CONFIG_FILE" key_file "$OCI_CLI_KEY_FILE")}"
    fi
  fi

  if [ -z "$OCI_CLI_KEY_FILE" ] && [ -z "$OCI_CLI_KEY" ]; then
    _err "Fatal: you must provide a value for either OCI_CLI_KEY_FILE or OCI_CLI_KEY."
    return 1
  fi

  if ! _contains "PRIVATE KEY" "$OCI_CLI_KEY"; then
    OCI_CLI_KEY=$(printf "%s" "$OCI_CLI_KEY" | _dbase64 multiline)
  fi


}
```

dnsapi/dns_oci.sh
```diff
@@ -0,0 +1,307 @@
...
+    _seg_n=$(echo "$_left" | sed -n "1,${_end_n}p")
+  else
+    _seg_n="$_left"
+  fi
+
+  _debug2 "_seg_n" "$_seg_n"
+  _lineini="$(echo "$_seg_n" | grep "^ *$_key *= *")"
+
+  _debug2 "_lineini" "$_lineini"
+  printf "%b" "$(eval "echo $_lineini | sed -e \"s/^ *${_key} *= *//g\"")"
```
Neilpang, line 305: `sed -e` option is not supported on all the platforms. please find another solutions

dnsapi/dns_oci.sh
Djelibeybi: `shellcheck` throws an error for each instance of `elif &&` so this will break the tests. Is there another option for this?

Neilpang: please remove ` && `

dnsapi/dns_oci.sh
Djelibeybi: `sed -e` is a requirement for IEEE/POSIX compliance: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

Djelibeybi: it works fine without the `-e` so I've removed it.

Neilpang: please just remove the `-e` option.

Neilpang:
please update the usage wiki and re-test all the cases from the config file.

Djelibeybi:
> please update the usage wiki and re-test all the cases from the config file.

The wiki was updated before I pushed the code.  I'm not sure what you mean about re-testing all the cases from the config file, but I've already tested with/without the config file and overriding each of the values.

Neilpang:
thanks

Issue #1722, opened 2018-07-10
Unclear documentation about different "run modes" of acme.sh (with / without config files)


wowcut:
From what I understand acme.sh seems to have at least two **different run modes** that seem to be:
* the first run mode **expects** some environment variables to be set and **writes** config files, but does **not read** config files
* the second run mode **reads config files** - but it is not clear if it ignores environment variables.

This is a strange behaviour for a shell script and leads to several problems - e.g. what do I need to do if I want to move / reinstall acme.sh on a new machine? Can I just backup the config files? Also it is not clear what happens with env vars when acme.sh is run a second time. And what about cron? 

It seems like you want to keep the configuration files "private", like in a private java class - I do not think this is a good idea, as it makes handling of acme.sh complicated. Also there is a very well established way of doing things in bash scripts regarding config files and environment variables - usually environment variables **override**  configuration values found in a config file - it is not clear how acme.sh handles this.

It would be very helpful to have more documentation about the relation of acme.sh config files / env vars / "run modes" as a first help. In the long term it seems like that handling should be adopted to the way this might be expected on a unix system, like described above. Thank you!

Neilpang:
sorry about missing doc.

There are a few rules:

1.  acme.sh  respects environment variables first.  acme.sh assumes a default value for each variables, but if you defined a environment variable with the same name, we will use the environment variable instead.
Take the "ACCOUNT_CONF_PATH" variable as an example.
And you can check the `_initpath()` function for more details.
2. If you defined a environment variable rather than the default value,  the variable will be remembered in the config file, and will be reused next time.
3. If you define a new value for the environment variable for the second time,  usually, the old value will be overwritten in the config file.


Yes, acme.sh has 2 running modes:
1. The user mode:   acme.sh is invoked by the user, with command parameters to issue cert or revoke or etc.
2. The cron mode:  acme.sh is running as a cron job, in which, acme.sh expects to load all the variable from the config file(the account conf or the domain conf).

wowcut:
The problem with env vars is that it is not possible to define multiple values - e.g. it is not possible to define multiple target servers for the ssh deploy command. Please see https://github.com/Neilpang/acme.sh/issues/1714 and https://github.com/Neilpang/acme.sh/issues/1406

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

Issue #1155, opened 2017-12-26
DNSimple: Allow for multiple accounts


Hermanverschooten:
Currently it is not possible to have multiple accounts when using DNSimple.
You can only specify one OAUTH_TOKEN.

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

PR #3578, opened 2021-06-28
Support account id configuration for dnsimple

vin01:
Hi, 

Currently dnsimple  uses `/whoami`  API endpoint which does not always contain account ids. `whoami` only responds with account info when `account tokens` are used during authentication.

> If you use an account token and you want to get the ID programmatically, the simplest way is to perform a whoami API call. The response will contain the account identifier as part of the account payload.

- https://developer.dnsimple.com/v2/

However, `user tokens` are also useful with dnsimple in certain cases like:

1.  When a domain from account A is shared with account B. In this case, B account has to use `user tokens` to be able to manage domain(s) shared with it.

2. Using a single user token to manage multiple accounts.

In such cases it would be great to be able to specify account id for which the dns records are going to be managed. These IDs can also be retrieved with `/accounts` endpoint but that might not be a good idea since multiple IDs might be present in the response.

- https://developer.dnsimple.com/v2/accounts/

This PR adds support to dynamically specify account id as env var `DNSimple_ACCOUNT_ID` (similar to what [terraform provider for dnsimple ](https://registry.terraform.io/providers/dnsimple/dnsimple/latest/docs)does). I am not sure if saving this env var is a good idea as that would make it hard to manage multiple accounts, being able to just specify it as env var at runtime seems better for this reason. wdyt?

Please review.

Thanks!

### DIFF
--- a/dnsapi/dns_dnsimple.sh
+++ b/dnsapi/dns_dnsimple.sh
@@ -123,21 +123,26 @@ _get_root() {
 # returns _account_id
 _get_account_id() {
   _debug "retrive account id"
-  if ! _dnsimple_rest GET "whoami"; then
-    return 1
+
+  if [ -n "$DNSimple_ACCOUNT_ID" ]; then
+    _account_id="$DNSimple_ACCOUNT_ID"
+  else
+    if ! _dnsimple_rest GET "whoami"; then
+      return 1
+    fi
+
+    if _contains "$response" "\"account\":null"; then
+      _err "no account associated with this token"
+      return 1
+    fi
+
+    if _contains "$response" "timeout"; then
+      _err "timeout retrieving account id"
+      return 1
+    fi
+    _account_id=$(printf "%s" "$response" | _egrep_o "\"id\":[^,]*,\"email\":" | cut -d: -f2 | cut -d, -f1)
   fi
 
-  if _contains "$response" "\"account\":null"; then
-    _err "no account associated with this token"
-    return 1
-  fi
-
-  if _contains "$response" "timeout"; then
-    _err "timeout retrieving account id"
-    return 1
-  fi
-
-  _account_id=$(printf "%s" "$response" | _egrep_o "\"id\":[^,]*,\"email\":" | cut -d: -f2 | cut -d, -f1)
   _debug _account_id "$_account_id"
 
   return 0
--- a/dnsapi/dns_dnsimple.sh
+++ b/dnsapi/dns_dnsimple.sh
@@ -124,6 +124,7 @@ _get_root() {
 _get_account_id() {
   _debug "retrive account id"
 
+  # shellcheck disable=SC2154
   if [ -n "$DNSimple_ACCOUNT_ID" ]; then
     _account_id="$DNSimple_ACCOUNT_ID"
   else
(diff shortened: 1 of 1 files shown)


Neilpang:
thanks,  but please enable actions in your fork, and fix all the CI errors.

vin01:
I have just disabled that specific shellcheck warning similar to what [notify scripts do](https://github.com/acmesh-official/acme.sh/blob/master/notify/slack.sh#L46).

Another way around this one (`SC2154`) would be to use all caps for exported env vars (`DNSIMPLE_ACCOUNT_ID` instead of `DNSimple_ACCOUNT_ID`) but I thought it best to keep it similar to `DNSimple_OAUTH_TOKEN` to not introduce this subtle breaking change, let me know if this makes sense.

vin01:
@Neilpang  Any feedback on this? I will really like to move forward with this.

Neilpang:
please use `_readaccountconf_mutable` and `_saveaccountconf_mutable` to read and write your env variables.
example is here:  
https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_cf.sh#L21

Neilpang:
>>  just specify it as env var at runtime seems better for this reason

Think about the renewal.   Who is to define the env variable in the cronjob?

So, any variables you must save.

vin01:
In my use case it would be set up as a systemd timer and environment variable will be passed accordingly on each run, be it for fresh issue or renewal (there could be multiple timers for different domains/accounts). Same can be used in cron jobs giving it more flexibility and dynamic configuration.

However I understand that this is a bit different from current design where dnsimple API token is persisted in the config file (which was a bit counter-intutive since once persisted, defining a new token value in environment variable does not override existing value from config file), would you prefer to discuss this aspect in a dedicated issue? 

I think configuration management overall in such a case is very tricky since multiple accounts and multiple tokens are the primary requirement to support that. 

A couple of related issues I could find: https://github.com/acmesh-official/acme.sh/issues/1155, https://github.com/acmesh-official/acme.sh/issues/1722#issuecomment-403860142

Neilpang:
>>  defining a new token value in environment variable does not override existing value from config file

That's why I asked you to use `_readaccountconf_mutable` and `_saveaccountconf_mutable `

vin01:
Alright, I think that is great to support multi-account use case while also preserving current design. I just added it.

I think the same can also be done for `DNSimple_OAUTH_TOKEN` so that multi-account use case with different tokens is also covered, leaving that out for the user to ba managed/overridden via environment variables might be good enough to close issues like https://github.com/acmesh-official/acme.sh/issues/1155

wdyt? I can create a different PR for that

vin01:
@Neilpang do you have any further feedback on this?

Neilpang:
The CI checks are not passed: 
https://github.com/vin01/acme.sh/runs/2932435639?check_suite_focus=true

vin01:
Unfortunately I am no longer using DNSimple, so don't really have a subscription and secrets to use for the DNS test. I guess this is also a problem when using a corporate account as using those credentials in Github actions will not be so nice (perhaps temporary tokens can be used to make the tests go through and revoked later).

Doees the project itself have some secrets configured which can be utilized in the workflow here itself?

Please feel free to close it otherwise.

PR #3591, opened 2021-07-02
Add NextDNS as a DOH option.

jonwltn:


Neilpang:
why do you need this ?

jonwltn:
No need. Options are nice to have. I started looking into DOH options when Cloudflare didn't update their cache before the DNS check timed out. I was really glad to see Google as a secondary option.

Neilpang:
Are both cf and google not working for you ?

jonwltn:
I have not had an instance of when both CF and Google were not working for me. Options are nice to have, but it's not like this PR is fixing a bug or anything. Reject it if you don't want it.

PR #3651, opened 2021-08-06
ISPConfig remote user needs more permissions

RBCOMGmbH:
Added line to first comment, DNS txt functions were not enough on my new ISPConfig installation. 
Maybe ISPConfig behaviour changed, using debug I found out that the remote user did not have permission to execute client_get_id
I also have DNS zone function permissions, though I do not know if that is actually necessary


Neilpang:
can you please update the comments into the usage:

https://github.com/acmesh-official/acme.sh/wiki/dnsapi#12-use-ispconfig-31-api

PR #3679, opened 2021-08-27
fix nginx mode

MrThanlon:
The plus symbol (+) is not standard regular expressions, use extended regular expressions(`grep -E`) instead.


Neilpang:
`grep -E`  is not crossing all the platforms.

MrThanlon:
While the master branch veision does not work properly, it can't detect nginx configuration.

PR #3394, opened 2021-02-06
lighttpd deploy hook

michaelw:
* verbatim copy from haproxy.sh, s/haproxy/lighttpd
* enable issuer

PR #3687, opened 2021-09-01
use _getdeployconf instead of sourcing DOMAIN_CONF

gstrauss:
use _getdeployconf instead of sourcing DOMAIN_CONF
    
(requested by @Neilpang in #3394)

### DIFF
--- a/deploy/lighttpd.sh
+++ b/deploy/lighttpd.sh
@@ -1,0 +1,275 @@
+#!/usr/bin/env sh
+
+# Script for acme.sh to deploy certificates to lighttpd
+#
+# The following variables can be exported:
+#
+# export DEPLOY_LIGHTTPD_PEM_NAME="${domain}.pem"
+#
+# Defines the name of the PEM file.
+# Defaults to "<domain>.pem"
+#
+# export DEPLOY_LIGHTTPD_PEM_PATH="/etc/lighttpd"
+#
+# Defines location of PEM file for Lighttpd.
+# Defaults to /etc/lighttpd
+#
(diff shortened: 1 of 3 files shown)


deploy/lighttpd.sh
```diff
@@ -0,0 +1,280 @@
...
+    if [ "${_ocsp_url}" != "" ]; then
+      # Extract the hostname from the OCSP URL
+      _info "Extracting OCSP URL"
+      _ocsp_host=$(echo "${_ocsp_url}" | cut -d/ -f3)
+      _debug _ocsp_host "${_ocsp_host}"
+
+      # Only process the certificate if we have a .issuer file
+      if [ -r "${_issuer}" ]; then
+        # Check if issuer cert is also a root CA cert
+        _subjectdn=$(openssl x509 -in "${_issuer}" -subject -noout | cut -d'/' -f2,3,4,5,6,7,8,9,10)
```
Neilpang, line 211: don't use `openssl` command directly.

please use `${ACME_OPENSSL_BIN:-openssl}` instead.

gstrauss:
@Neilpang I have added a commit to address your latest request.

Please note that the usage you have asked to change was pre-existing in haproxy.sh, from which lighttpd.sh is derived.
Your code review in #3394 from Feb 27 did not include this latest request.

If you find additional trivial string substitutions that you would like, please know that you may also edit this PR to make such changes.

deploy/lighttpd.sh
gstrauss: additional commit added with this newly requested change

PR #3697, opened 2021-09-09
1984 hosting cookie fix

DerVerruckteFuchs:
It seems that the HTTP header format for 1984 Hosting has changed somewhat, along with the website's UI. Here is a current example of the HTTP header with sensitive data removed:

````
HTTP/2 200                                                                      
server: nginx                                                                   
date: Thu, 09 Sep 2021 02:40:29 GMT                                             
content-type: application/json                                                  
content-length: 30                                                              
vary: Accept-Language, Cookie                                                   
content-language: en                                                            
set-cookie: sessionid=$SESSIONID; expires=Thu, 09 Sep 2021 03:40:29 GMT; HttpOnly; Max-Age=3600; Path=/; SameSite=Lax
set-cookie: csrftoken=$CSRFTOKEN; expires=Thu, 08 Sep 2022 02:40:29 GMT; Max-Age=31449600; Path=/; SameSite=Lax; Secure
strict-transport-security: max-age=31536000                                     
x-frame-options: SAMEORIGIN
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
````
Currently, the script uses `tail` to get the session ID cookie. Because the CSRF token is in the last `set-cookie` line, it gets printed instead of the session ID cookie. The CSRF line then gets `egrep`ped away as it does not match the `sessionid=[^;]*;` pattern we are trying to `egrep` for, which results in `One984HOSTING_COOKIE` being assigned an empty string. This causes the `dns_1984hosting.sh` script to fail when trying to log into an account.

Either the CSRF token cookie is a recent addition, or the order of the `set-cookie` lines in the HTTP header have been reversed. Using `head` instead of `tail` gets us the session ID cookie as intended.

### CODE REVIEW

File: dnsapi/dns_1984hosting.sh [hunk 1]
```diff
@@ -150,7 +150,7 @@ _1984hosting_login() {
   _debug2 response "$response"
 
   if _contains "$response" '"loggedin": true'; then
-    One984HOSTING_COOKIE="$(grep -i '^set-cookie:' "$HTTP_HEADER" | _tail_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
+    One984HOSTING_COOKIE="$(grep -i '^set-cookie:' "$HTTP_HEADER" | _head_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
```
Neilpang, line 153: why not grep for "sessionid" instead?

DerVerruckteFuchs: I got preoccupied with fixing the bug while preserving how the script currently behaves. While my change currently fixes things, removing the unnecessary step of using head/tail and grepping directly for the `sessionid` is more "future-proof" if the `set-cookie` lines change order and/or increase/decrease again. To go a step further, I used cat on a test cookie text file using the example I posted above to see if removing the first grep does not break anything, since it seems just as unnecessary as I'm looking at it again. It seems to work fine from typing it into the terminal. I'm going to test it with the script and commit the change it if it works.

DerVerruckteFuchs: > I got preoccupied with fixing the bug while preserving how the script currently behaves. While my change currently fixes things, removing the unnecessary step of using head/tail and grepping directly for the `sessionid` is more "future-proof" if the `set-cookie` lines change order and/or increase/decrease again. To go a step further, I used cat on a test cookie text file using the example I posted above to see if removing the first grep does not break anything, since it seems just as unnecessary as I'm looking at it again. It seems to work fine from typing it into the terminal. I'm going to test it with the script and commit the change it if it works.

So I must need coffee or something, since that first grep is what is getting the HTTP header piped into egrep and tr in the first place. It is still possible to slim it down to one grep/egrep though since we can skip straight to the `sessionid` and not bother with the `set-cookie:` pattern.

DerVerruckteFuchs: I committed the simplified grepping. I tested it and it works just fine for me on pfsense.

dnsapi/dns_1984hosting.sh
```diff
@@ -150,7 +150,7 @@ _1984hosting_login() {
   _debug2 response "$response"
 
   if _contains "$response" '"loggedin": true'; then
-    One984HOSTING_COOKIE="$(grep -i '^set-cookie:' "$HTTP_HEADER" | _tail_n 1 | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
+    One984HOSTING_COOKIE="$(grep -io 'sessionid=[^;]*;' "$HTTP_HEADER"  | tr -d ';')"
```
Neilpang, line 153: don't use `grep -io` options.

please just grep "sessionid=" first, and then grep "set-cookie"

DerVerruckteFuchs: Is `grep -i` okay? The last time I did any fixes for the `set-cookie` business I used `grep -i`. If there are any DNS API changes that do not permit `grep -i` I can work with that. I checked the [guide](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide#12-cross-platform-compatibility-guide), and just `grep -o` is listed as a cross-compatibility issue.

Neilpang:
Looks better now.

please fix the CI Checks: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

DerVerruckteFuchs:
> Looks better now.
> 
> please fix the CI Checks: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

I need to add the Referer and X-CSRFToken headers before everything works correctly. I want to repurpose the code that gets the zone ID by turning it into a function. I have a proof of concept that seems to be working, but I need to clean it up a bit first.

DerVerruckteFuchs:
I managed to get the script to set and delete the `_acme_challenge.domain.com` TXT records. The problem I'm running into is the Docker action is adding two TXT entries. This causes the letsencrypt verification to fail since the TXT record is going to be wrong when it gets checked. I think `dns_1984hosting_add()` gets called twice, then `dns_1984hosting_rm()` gets called twice by the Docker action.

I am going to try setting `TEST_DNS_NO_WILDCARD` to `1` to disable the wildcard since I am only testing with a subdomain at the moment.

DerVerruckteFuchs:
I managed to get all of my changes to work. However, I'm having DNS timeouts that cause failure toward the very end. The results are hit-and-miss, so testing a staging cert on pfsense worked okay but everywhere else failed. Increasing `TEST_DNS_SLEEP` does help, but I don't remember needing more than the default setting. I set it to 3600 seconds, since that is what the TTL is set to. It should be a few minutes to see if all goes well.

DerVerruckteFuchs:
I set `TEST_DNS_SLEEP` to 900 to match the new TTL value I set. The tests took over 5 hours, but they all seem to have passed successfully.

phedoreanu:
I can confirm it works. Well done @DerVerruckteFuchs ! 👍 🥳

PR #3708, opened 2021-09-21
knot plugin: find responsible DNS zone by SOA queries

nils-wisiol:
This PR changes the knot plugin such that it determines the DNS zone responsible for serving `_acme-challenge.example.com` via a number of DNS queries using `dig`.

The reason this is useful is because the current implementation is broken in some settings, as the zone responsible is just determined by using the two right-most labels of the DNS name. For example, `_acme-challenge.www.example.co.uk` currently results in an attempted update on `co.uk`. This breaks the publication of the acme challenge for everyone using knot and a domain under `co.uk` (or any other domain where the delegation doesn't happen at the first label).

This PR updates the implementation to determine the responsible zone by the means of queries for the SOA record. It resolves above problem but requires (1) the `dig` binary and (2) query access to the DNS.

### DIFF
--- a/dnsapi/dns_knot.sh
+++ b/dnsapi/dns_knot.sh
@@ -71,20 +71,31 @@ EOF
 # returns
 # _domain=domain.com
 _get_root() {
-  domain=$1
-  i="$(echo "$fulldomain" | tr '.' ' ' | wc -w)"
-  i=$(_math "$i" - 1)
+  ancestor="${1}."
 
-  while true; do
-    h=$(printf "%s" "$domain" | cut -d . -f "$i"-100)
-    if [ -z "$h" ]; then
+  while true; do # loops over all ancestors of $1
+
+    # count labels
+    num_labels="$(echo "$ancestor" | tr '.' ' ' | wc -w)"
+
+    # abort if empty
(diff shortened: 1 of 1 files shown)


Neilpang:
please enable the DNS CI checks:

nils-wisiol:
I've fixed the errors that appeared the last logs

Neilpang:
please enable the CI checks on your forked repo, and make it pass.

nils-wisiol:
GitHub actions on my fork are enabled (and have been all along). I'm unaware of any publicly available DNS service that uses this update method, so for further testing, a **local** knot auth NS would be required. Therefore, I did not add any secrets to my forked repo.

Neilpang:
you can use the `knotc`  to check the zone:
https://www.knot-dns.cz/docs/2.5/html/man_knotc.html

Your should use the api to detect the zones:
https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide#3-detect-which-part-is-your-root-zone

Neilpang:
the `dig` deosn't exist on every system. 
If you have to use it,  check with `_exists()` first.  
It's strongly recommended to use the `_ns_lookup()` function to dig dns records with doh.

PR #3710, opened 2021-09-23
Rackspace changed their API response, fixed the sed matching

tcocca:
Rackspace recently changed the API response on the Domains Search API endpoint (https://docs.rackspace.com/docs/cloud-dns/v1/api-reference/domains#search-domains)

We used to get a response like the following:

```
{
  "domains": [ 
    {
      "ttl": 3600,
      "accountId": 12345,    
      "id": 2725257,
      "name" : "example.com",
      "emailAddress" : "sample@rackspace.com",
      "updated" : "2021-06-23T03:09:34.000+0000",
      "created" : "2021-06-23T03:09:33.000+0000"
    }
  ],
  "totalEntries" : 1
}
```

We are now getting this format:

```
{
  "totalEntries": 1,
  "domains": [
    {
      "id": "2725257",
      "accountId": "12345",
      "name": "example.com",
      "ttl": 3600,
      "emailAddress": "sample@rackspace.com",
      "updated": "2021-09-21T03:25:53.146Z",
      "created": "2009-12-11T01:01:13.000Z"
    }
  ]
}
```

They have switched the order of the `id` and `accountId` fields and also they have made those fields strings instead of integers in the response.   I updated the sed call for parsing out the correct domain id.

### DIFF
--- a/dnsapi/dns_rackspace.sh
+++ b/dnsapi/dns_rackspace.sh
@@ -80,7 +80,7 @@ _get_root_zone() {
     _debug2 response "$response"
     if _contains "$response" "\"name\":\"$h\"" >/dev/null; then
       # Response looks like:
-      #   {"id": "12345","accountId": "1111111","name": "example.com","ttl": 3600,"emailAddress": ...<and so on>
+      #   {"id": "12345","accountId": "1111111","name": "example.com","ttl": 3600,"emailAddress": ... <and so on>
       _domain_id=$(echo "$response" | sed -n "s/^.*\"id\":\"\([^,]*\)\",\"accountId\":\"[0-9]*\",\"name\":\"$h\",.*/\1/p")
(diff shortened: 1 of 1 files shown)


Neilpang:
can you fix the CI checks ?
https://github.com/tcocca/acme.sh/actions/runs/1265769189

tcocca:
@Neilpang I'm trying to fix that. I setup the secrets in the actions on my fork and pushed (a whitespace change) to trigger them again but its still failing.  

Does case matter? Github keeps uppercasing the names, so its turning `TokenName1` into `TOKENNAME1`:

image

tcocca:
Hmm, looks like case doesn't matter so now I'm not sure why I am failing https://github.blog/changelog/2020-05-29-ui-and-api-updates-for-secrets/

PR #3757, opened 2021-10-13
deploy hook for openmediavault 5

eddul-h:


sjtuross:
Want to share some feedback per my test of this PR.

1. Need to run "nginx -s reload" after applying changes, otherwise the old cert is still being served.
2. When applying changes, just leave modules empty to apply all changes, otherwise web gui still prompts to apply changes.
3. Need to call _getdeployconf and _savedeployconf to persist env variables.
4. Would be good to support remote deploy through ssh.

Neilpang:
@sjtuross  Please send a new pr.

PR #3760, opened 2021-10-13
[dns_aws] Allow use of a secure subzone

feld:
Allow use of a more secure _acme-challenge.$h subzone in Route53 to limit scope of access if an attacker compromises your system running acme.sh

Example:

1. Setup a zone `foo.com.` in Route53
2. Setup a zone `_acme-challenge.foo.com.` in Route53
3. Take the **NS** from subzone `_acme-challenge.foo.com.` and create an **NS** record in parent zone `foo.com.` for `_acme-challenge.foo.com.`
4. Now create IAM role/keys for access to the `_acme-challenge.foo.com.` subzone
5. Use acme.sh

Now if your server running acme.sh gets compromised the attacker can only tamper with your DNS within the `*._acme-challenge.foo.com.` zone

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -178,7 +178,20 @@ _get_root() {
         return 1
       fi
 
-      if _contains "$response" "<Name>$h.</Name>"; then
+      if _contains "$response" "<Name>_acme-challenge.$h.</Name>"; then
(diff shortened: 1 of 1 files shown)


backeby:
This makes sense to me, it would browse down through the subzones (If they exist) and hit the root zone last.

Issue #3735, opened 2021-10-03
Problem with domain validation - DNS-01, opnsense with bind


wacki4:
Hi!
I'am trying to validate with DNS-01 my subdomain using opnsense acme plugin, and bind.

My DNS works without a problem - it is avaiable from outside, and returns correct IP addresses for entrances which i made.

But i cannot generate certificate cause of error: "invalid domain".

I found out that it is returned by  acme.sh/dnsapi/dns_opnsense.sh .

Running /api/bind/domain/get i got in return:

`{
  "domain": {
    "domains": {
      "domain": {
        "19b51657-5206-4a1f-a0a6-98b7eab8eb51": {
          "enabled": "1",
          "type": {
            "master": {
              "value": "master",
              "selected": 1
            },
            "slave": {
              "value": "slave",
              "selected": 0
            }
          },
          "masterip": {
            "": {
              "value": "",
              "selected": 1
            }
          },
          "allownotifyslave": {
            "": {
              "value": "",
              "selected": 1
            }
          },
          "domainname": "test.pl",
          "allowtransfer": {
            "": {
              "value": "none",
              "selected": 0
            },
            "7ff76356-1da2-45bd-a733-52fe959f2518": {
              "value": "all",
              "selected": 1
            }
          },
          "allowquery": {
            "": {
              "value": "none",
              "selected": 0
            },
            "7ff76356-1da2-45bd-a733-52fe959f2518": {
              "value": "all",
              "selected": 1
            }
          },
          "serial": "2110031633",
          "ttl": "86400",
          "refresh": "21600",
          "retry": "3600",
          "expire": "3542400",
          "negative": "3600",
          "mailadmin": "admin.test.pl",
          "dnsserver": "ns.test.pl"
        }
      }
    }
  }
}`

And this is my log where problem occurs:

<html><body>

2021-10-03T16:44:56 | acme.sh[91474] | ] Please check log file for more details: /var/log/acme.sh.log
-- | -- | --
2021-10-03T16:44:56 | acme.sh[7671] | ] _on_issue_err
2021-10-03T16:44:56 | acme.sh[44902] | ] Error add txt for domain:_acme-challenge.aaa.test.pl
2021-10-03T16:44:56 | acme.sh[90579] | ] invalid domain
2021-10-03T16:44:56 | acme.sh[49488] | ] h='pl'
2021-10-03T16:44:56 | acme.sh[79755] | ] h='test.pl'
[...]

PR #3764, opened 2021-10-16
Update dns_opnsense.sh

wacki4:
Update for opnsense regards to error in #3735


Neilpang:
please fix the CI checks:

blablup:
How do I set the domain-alias for DNS Test? I need to test with "domain-alias" option.

Daan-:
I manually applied this patch on my OpnSense firewall and can validate that it works.
A better (long term) solution would be to use the JSON object, maybe parsing with jq. Regexes on JSON data are brittle.
I also don't have a public DNS available for testing purposes.

PR #3794, opened 2021-11-05
Removed newline at the end of subject, Added MIME-Version header

Scre13:
Minor fixes to be more RFC comliant.
Some spam filters (e.g. rspamd) do trigger on the nofitications an put spam scores on the mail.

image

### DIFF
--- a/notify/mail.sh
+++ b/notify/mail.sh
@@ -62,7 +62,7 @@ mail_send() {
   fi
 
   contenttype="text/plain; charset=utf-8"
-  subject="=?UTF-8?B?$(echo "$_subject" | _base64)?="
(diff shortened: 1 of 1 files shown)


notify/mail.sh
```diff
@@ -62,7 +62,7 @@ mail_send() {
   fi
 
   contenttype="text/plain; charset=utf-8"
-  subject="=?UTF-8?B?$(echo "$_subject" | _base64)?="
+  subject="=?UTF-8?B?$(echo -n "$_subject" | _base64)?="
```
Neilpang, line 65: don't use  `echo -n` option, it's not supported on all the platforms.
please use  `printf "%b"` instead.

PR #3798, opened 2021-11-09
removed -- at beginning of subject

Scre13:
Minor fix to #3797, removed -- at the beginning of the subject.

### DIFF
--- a/notify/mail.sh
+++ b/notify/mail.sh
@@ -62,7 +62,7 @@ mail_send() {
   fi
 
   contenttype="text/plain; charset=utf-8"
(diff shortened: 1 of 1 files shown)


Neilpang:
why

Scre13:
> why

You will not be able to group the notification emails by subject. I personally prefer such notifications in the format "subject on hostname". This helps me a lot, as I'm receiving a lot of notifications a day from many systems in my job.

Perhaps later I'll start a new pull to add "hostname" to the notifications, in the hope that it will also benefit others.

PR #3807, opened 2021-11-12
Add SYNO_TOTP_SECRET for user with two-factor authentication enabled

dacrystal:
Added ability to auto generate OTP code by setting SYNO_TOTP_SECRET environment variable.  OTP secret can be retrieved from 2-factor authentication setup via link(can't scan it) below QR. 

Tested on DSM 7 (7.0.1-42218) , DS920+


alidaf:
Finding the OTP secret seems to have changed in DSM. Does anyone have another method?

PR #3834, opened 2021-11-29
Create dns_jamotech.sh

Jamo-Tech:


Jamo-Tech:
Please note that this helper script does not only create the _acme-challenge TXT record but will also create an A record pointing to an IP passed as $JTECH_ENDIP and this is only valid and working for our customer base subdomains which are subdomains on the root domain jamo.tech, this script will force that the jamo.tech domain is used as the root domain and this is meant for use on our customer systems to enable our customerbase to renew and create SSL certificates without the need of our support staff. We are an MSP type company and not a domain registrar or similar and therefor it should only be our clientbase that use this helper script.

Jamo-Tech:
Closed pull request due to missing tests.

PR #3853, opened 2021-12-13
Allow optional "NEW" in CSR header and footer

pearj:
When generating a CSR in Windows it seems to create a CSR header that looks like `-----BEGIN NEW CERTIFICATE REQUEST-----`, but the addition of `NEW` breaks the parsing of the CSR. Making `NEW ` optional fixes the problem.

Apparently [certbot](https://community.letsencrypt.org/t/error-parsing-certificate-request-resolved/40039/6) is tolerant of both forms.

I tested this locally with grep, and it matches both with and without `NEW`.


pearj:
I rebased this code onto the dev branch (despite my fork saying master)

PR #3859, opened 2021-12-20
Deploy hook ssh support scp

Hawthorn2013:
Deploy hook ssh put file content in ssh cmd. This usage causes the cmd too long to execute on some remote server, even if DEPLOY_SSH_MULTI_CALL is set to yes.

This pull request implements deployment of file by scp, avoiding the problem of excessively long cmd.


Neilpang:
don't change the default behavior.  It was intended not to use scp.

PMExtra:
This will be solved with #4091

PR #3863, opened 2021-12-22
Go dns_regru To IDN!

vitaly80:
`service/get_list` returns domains in utf. But if use utf-domains in acme.sh, then appears error `Error parsing certificate request: x509: SAN dNSName is malformed`.

Lets use idn-domains in acme.sh command string! Then we need `service/get_list` will be in IDN!

### DIFF
--- a/dnsapi/dns_regru.sh
+++ b/dnsapi/dns_regru.sh
@@ -92,9 +92,10 @@ _get_root() {
   domains_list=$(echo "${response}" | grep dname | sed -r "s/.*dname=\"([^\"]+)\".*/\\1/g")
 
   for ITEM in ${domains_list}; do
+    IDN_ITEM="$(echo "${ITEM}" | idn)"
     case "${domain}" in
-    *${ITEM}*)
-      _domain=${ITEM}
+    *${IDN_ITEM}*)
+      _domain=${IDN_ITEM}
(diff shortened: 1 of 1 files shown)


dnsapi/dns_regru.sh
```diff
@@ -92,9 +92,10 @@ _get_root() {
   domains_list=$(echo "${response}" | grep dname | sed -r "s/.*dname=\"([^\"]+)\".*/\\1/g")
 
   for ITEM in ${domains_list}; do
+    IDN_ITEM="$(echo "${ITEM}" | idn)"
```
Neilpang, line 95: don't use `idn`,  please use `_idn()` function instead.

vitaly80: Hello! Cant use this now, can't check dns record, because it is valid, dns verification expires 27 december) I Check it later.

vitaly80: > 

How to use `_idn()`? 
```suggestion
    IDN_ITEM="$(echo "${ITEM}" | idn)"
```

vitaly80: Hello, I did it!

`    IDN_ITEM="$(_idn "${ITEM}")"
`

PR #3876, opened 2022-01-06
deploy/fritzbox: allow hook to be used with multiple fritzboxes

fraenki:
Previously the deploy hook config was stored in the account config. This seems odd and adds unnecessary limitations to the hook. For example, the hook could not be used to deploy the same (SAN) cert to multiple fritzboxes.

Now we're using the correct `_*deployconf()` functions to read and write the deploy hook config. This allows to use the deploy hook for multiple fritzboxes, as demonstrated by my own testing:

```
# deploy on fritzbox 1
acme.sh[64899]: [Thu Jan  6 16:17:51 CET 2022] Running cmd: deploy
...
acme.sh[80739]: [Thu Jan  6 16:17:51 CET 2022] DEPLOY_FRITZBOX_URL='http://fritz1.example.com'
acme.sh[36765]: [Thu Jan  6 16:17:51 CET 2022] DEPLOY_FRITZBOX_USERNAME='fritzuser1'
...

# deploy on fritzbox 2
acme.sh[28397]: [Thu Jan  6 16:18:12 CET 2022] Running cmd: deploy
...
acme.sh[97335]: [Thu Jan  6 16:18:12 CET 2022] DEPLOY_FRITZBOX_URL='http://fritz2.example.com'
acme.sh[97335]: [Thu Jan  6 16:18:12 CET 2022] DEPLOY_FRITZBOX_USERNAME='fritzuser2'
...
```

Without this patch, deploying on "fritzbox 2" would use the same values for `DEPLOY_FRITZBOX_*` as the first run for "fritzbox 1".

This patch also ensure to remove the old config values from account config.

### DIFF
--- a/deploy/fritzbox.sh
+++ b/deploy/fritzbox.sh
@@ -36,43 +36,51 @@ fritzbox_deploy() {
     fi
   fi
 
-  _fritzbox_username="${DEPLOY_FRITZBOX_USERNAME}"
-  _fritzbox_password="${DEPLOY_FRITZBOX_PASSWORD}"
-  _fritzbox_url="${DEPLOY_FRITZBOX_URL}"
+  # Clear traces of incorrectly stored values
+  _clearaccountconf DEPLOY_FRITZBOX_USERNAME
+  _clearaccountconf DEPLOY_FRITZBOX_PASSWORD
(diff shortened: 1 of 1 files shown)


fraenki:
@fritteli, FYI, since you're the original author of the fritzbox deply hook :)

PR #3673, opened 2021-08-24
Fix simply.com API

tomsommer:
- Normalize JSON and fix not handling return code correctly
- Add some information to comments
- Fix trailing slash on URIs
- Add 60 second sleep for zone to be written

PR #3877, opened 2022-01-06
Cannot wait for PR #3673 to be completed

### LINKED ISSUE #2906: UnoEuro changed its name to Simply.com
Hi

I would like to update [dns_unoeuro.sh](https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_unoeuro.sh) as UnoEuro recently changed its name to Simply.com. New API hostname is api.simply.com.

I don't want to break any existing use of the UnoEuro/Simply.com DNS plugin – what is recommended way? Create a new files named dns_simplycom.sh with updated content and keep the dns_unoeuro.sh file and just update the name and hostname in the existing file? Or can I rename the file without breaking current use of the DNS plugin?

### LINKED ISSUE #3831: Failing success check in dns_simply.sh
Steps to reproduce
When renewing for simply.com it cannot find the root. Is is using the REST api and calls
https://api.simply.com/1/[account name]/[key]/my/products/appvice.dk/dns

And checks the result for:
"status": 200

The problem is that simply.com seems to have changed their output (json, so it is acceptable) to be:
{"zone":{"serial":[serial],"name":"[account name]","origin":"ns1.simply.com"},"status":200,"message":"success"}

Note that there is NO space after "status":

A quick workaround is to remove the space in dns_simply.sh
SIMPLY_SUCCESS_CODE='"status":200'

But the correct solution would be to decode the JSON and look for the domain. I however do not know enough shell scripting to be able to do this well:-)

jvandborg:
PR #3673 Fix simply.com API seems abandoned by maintainer and I need this fixed asap

Changes implemented
* Normalize JSON and fix not handling return code correctly
* Add some information to comments
* Fix trailing slash on URIs
* Fix parsing record_data and record_type

Fixes #3831
Closes #2906

### DIFF
--- a/dnsapi/dns_simply.sh
+++ b/dnsapi/dns_simply.sh
@@ -204,9 +204,6 @@ _simply_add_record() {
     return 1
   fi
 
-  _info "Waiting 60 seconds for DNS changes to be written"
-  _sleep 60
-
   return 0
 }
 
--- a/dnsapi/dns_simply.sh
+++ b/dnsapi/dns_simply.sh
@@ -203,7 +203,7 @@ _simply_add_record() {
     _err "$response"
     return 1
   fi
-  
+
   _info "Waiting 60 seconds for DNS changes to be written"
   _sleep 60
 
@@ -252,7 +252,7 @@ _simply_rest() {
     _err "error $ep"
     return 1
   fi
-  
+
   response="$(echo "$response" | _normalizeJson)"
 
   _debug2 response "$response"
--- a/dnsapi/dns_simply.sh
+++ b/dnsapi/dns_simply.sh
@@ -1,15 +1,15 @@
 #!/usr/bin/env sh
 
-#
+# API-integration for Simply.com (https://www.simply.com)
+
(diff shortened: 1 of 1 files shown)


dnsapi/dns_simply.sh
```diff
@@ -204,6 +204,9 @@ _simply_add_record() {
     return 1
   fi
 
+  _info "Waiting 60 seconds for DNS changes to be written"
+  _sleep 60
```
Neilpang, line 208: why do we need to sleep 60 here?
If it's for the dns record to propagate,  we should ask the user to use `--dnssleep 60` command line parameter to wait.


Imagine you have 10 domains in one cert,   if you sleep here,  you will sleep 10 minutes.   
but if you use "--dnssleep 60",  you will just have to sleep 60 seconds.

jvandborg: You are absolutely right. I must admit I just adopted the code from the original contributer of PR #3673 without much review.

I've removed the unnecessary sleep now.

PR #2413, opened 2019-07-29
deploy hook: routeros: store the env vars within the domainconf

herbetom:
I edited the deploy hook for RouterOS so that the enviroment variables for the routeros deploy hook are now stored inside the domain configuration file, so that it is not necessary to provide them each time. 
Also the option `ROUTER_OS_WEB_SERVICE` with the default value `yes` was added for the ability to disable updating the certificate for the www-ssl service, useful for example when a special certificate only for the hotspot server is needed.

This changes shouldn't cause an issue for current users of the deploy hook.

Issue #2344, opened 2019-06-18
Report bugs to RouterOS deploy API


cngarrison:
This is the place to report bugs in the RouterOS deploy API.

If you experience a bug, please report it in this issue.

Thanks!

herbetom:
Not really a bug, more a feature request, but only having a single RouterOS installation is probably quite uncommon, any posssibility to deploy to multiple instances of RouterOS?

cngarrison:
I like the idea, but let's flesh it out a bit more. You want a wildcard cert that is deployed to multiple routers? Or one cert per router?

The first should be easy to add by passing a list for `ROUTER_OS_HOST` (would assume same value across all routers for `ROUTER_OS_USERNAME` and `ROUTER_OS_ADDITIONAL_SERVICES`) and looping over the list. I'll consider adding that (PR's are also welcomed).

For the second case it should be doable now; just do another `acme.sh` run with different values for `ROUTER_OS_*` vars.

herbetom:
Thanks for the quick reply. Well, usually i don't use Wildcard Certificates, but the idea to specify a list of hosts would be definitely interesting. 

I have to admit, that while i wrote the first post, i only had tried the [fritzbox](https://github.com/Neilpang/acme.sh/blob/master/deploy/fritzbox.sh) deployhook. That hook saves the host, username and password within the `account.conf`. That's in my opinion not so nice, especially if you have multiple targets to deploy to and the reason for my comment. I will try to get that changed.

But now, after i loocked into the diffrent deploy script's, i noticed, that the handling of how and if at all such configurations are saved is diffrent from script to script. 

Since i now know that `ROUTER_OS_HOST`, `ROUTER_OS_USERNAME` and `ROUTER_OS_ADDITIONAL_SERVICES` aren't saved at all, i see that theire is not really a problem deploying to multiple routeros installations (like you said), since the vars need anyway be set before running the deploy script. 
I would really like it, if the vars would be saved in the domain configuration (like in the [ssh-deploy-hook](https://github.com/Neilpang/acme.sh/blob/master/deploy/ssh.sh)). That seems to me to be the best solution to me.

After i now tried to use the script for the first time i now got the first bug. Yeah ;)

I used the command `acme.sh --deploy -d router.example.com --deploy-hook routeros --debug` on a raspbian installation and got the following output:
```
[Do 25. Jul 04:57:41 CEST 2019] Lets find script dir.
[...]

PR #3892, opened 2022-01-16
routeros deploy hook: store the env vars within the domainconf

sjtuross:
This is actually a bug fix. Without getting the env vars stored, renew by cron won't work but it's supposed to work per design.

Related to #2344 and #2413


sjtuross:
@Neilpang thanks. I should create a separate branch for the routeros bug fix. I'm going to close this PR.

PR #3896, opened 2022-01-19
add remote deploy hook for openmediavault 5

sjtuross:
This is based on #3757 with the below enhancements.

1. Run "nginx -s reload" after applying changes, otherwise the old cert is still being served.
2. When applying changes, leave modules empty to apply all changes, otherwise web gui still prompts to apply changes.
3. Call _getdeployconf and _savedeployconf to persist env variables.
4. Support remote deploy through ssh. By default, OMV web gui admin doesn't have ssh access but it is required as the user parameter for the omv-rpc command, so need 2 separate env variables: DEPLOY_OMV_WEBUI_ADMIN (default is admin) and DEPLOY_OMV_SSH_USER (default is root).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,8 @@ RUN apk --no-cache add -f \
   tzdata \
   oath-toolkit-oathtool \
   tar \
-  libidn
+  libidn \
+  jq
 
 ENV LE_CONFIG_HOME /acme.sh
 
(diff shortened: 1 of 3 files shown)


Neilpang:
write the usage here: 

sjtuross:
@Neilpang usage has been added.

PR #3915, opened 2022-01-29
Dockerfile: run crond foreground

pexcn:
crond can be run foreground, use `-f` param.


Neilpang:
I knew it can run with `-f`.  but the `crond -f` can not handle  `Ctrl-c` interrupt.

PR #3934, opened 2022-02-10
Add environment variables ROUTER_OS_PORT

mac-zhou:
tested on ros v7.1.1


mac-zhou:
> please update the usage here too: https://github.com/acmesh-official/acme.sh/wiki/deployhooks#17-deploy-the-cert-to-remote-routeros
> 
> otherwise, nobody will know this option.

ok, wiki updated

PR #3940, opened 2022-02-12
Use short User agent options

stokito:
The shorter makes the command line shorter and easier to understood. Other options already uses short form when possible.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1893,29 +1893,29 @@ _post() {
     if [ "$needbase64" ]; then
       if [ "$body" ]; then
         if [ "$_postContentType" ]; then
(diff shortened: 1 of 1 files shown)


Neilpang:
the langer form is easier to understand than the short form.

stokito:
The `-U "$USER_AGENT"` is self explanatory

Neilpang:
I don't think '-U'  is easier to read than `--user-agent`.
And the source code is not for the end user, so, nobody would care about that.
When we change the code, there is a risk that the short option is not supported by some of the platforms, on which it used to run.
So, it brings almost nothing benefit to us,  just more risk. It's not worth that.

stokito:
Well, if you carry about other platforms support then it's better to use short options because BusyBox wget may be compiled without long options (but still --header doesn't have a short form).
Anyway, thank you for responding

PR #3947, opened 2022-02-19
deploy/routeros.sh fixes

abiessmann:
Changes to deploy/routeros.sh:

 * have `scp` and `ssh` configurable to set specific parameters to those commands
 * delete all three certificates of full chain
 * set owner from ssh user
 * set commented line in script as RouterOS comment


abiessmann:
Just saw another change in dev ... @mac-zhou introduced ROUTER_OS_PORT which was not in my base ... need to discuss if we stick with ROUTER_OS_PORT and ROUTER_OS_SSH_CMD/ROUTER_OS_SCP_CMD variables or to kick ROUTER_OS_PORT

PR #3978, opened 2022-03-10
Fixes Simply.com to use REST API version 2 with Basic Auth

nikolajbrinch:


leifnel:
The txt-records does not get deleted after verification. 
This also happened for the gratisdns plugin.

nikolajbrinch:
I will look into that when I get the time. At least it was merged 😊

PR #4007, opened 2022-03-30
[dns_acmedns] Handle many domain cert with many accounts

gaby64:
My solution is to use dynamic variables and append the domain, with the dot replaced by an underscore, to the variable; so you can export a set [username, password, subdomain] for each domain.

### DIFF
--- a/dnsapi/dns_acmedns.sh
+++ b/dnsapi/dns_acmedns.sh
@@ -20,8 +20,15 @@
 # Used to add txt record
 dns_acmedns_add() {
   fulldomain=$1
-  i=2
-  d=$(printf "%s" "$fulldomain" | cut -d . -f $i-100)
+  i=1
+  while [ -z "$d" ]; do
+    _d=$(printf "%s" "$fulldomain" | cut -d . -f $i-100)
+    c=$(echo $_d | awk -F'.' '{ print NF }')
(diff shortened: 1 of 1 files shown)


gaby64:
I learned some shell scripting like 3 days ago, I wont be able to rewrite it to be posix compliant.

LoganDark:
This is terrifying

gaby64:
> This is terrifying

What is terrifying? I have it working well for me.

LoganDark:
Just the extremely long chain of `sed`, `tr`, and `awk` commands. Never said it doesn't *work*, but...

gaby64:
> Just the extremely long chain of `sed`, `tr`, and `awk` commands. Never said it doesn't _work_, but...

I didnt write that part: https://www.reddit.com/r/learnprogramming/comments/7sdcdl/comment/dt45446/

gaby64:
would be nice if this could be merged

gaby64:
closed but no reason given, there many looking for a solution like this

#4221

maddes-b:
I wrote my own acme-dns-client that works with acme.sh and certbot with multiple domains in a single certificate.
The script uses Python, so it does not fully fit the acme.sh approach of "shell only".
Please have a look here: https://github.com/maddes-b/acme-dns-client-2
Feedback welcome.

PR #4008, opened 2022-03-30
_get_root was not giving me the root domain

gaby64:
changed "i" to 2, now I get the root domain and the easydns request works correctly.


gaby64:
> why does it have to start from `2` instead of `1`?
> 
> We must start from `1` because the domain can be a root zone in the alias mode: https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode

Where this is being used, $h must be the actual domain for which the easysns service is used, otherwise it will not produce a valid api request.

`_EASYDNS_rest GET "zones/records/all/$h"`

gaby64:
it was giving me _acme-challenge.root.com as the result

requesting "zones/records/all/_acme-challenge.root.com" is invalid


I took the 2 from other dnsapi scripts

Issue #3088, opened 2020-08-05
Report bugs to Netlify API


draevin:
Please report any bugs related to the Netlify API here!

thekevinbrown:
I'm having some trouble with the Netlify API via pfSense. I'm trying to use it but I keep getting this type of log:

(Note: I've replaced references to my actual domain with `mydomain.com` and the challenge value with `[challenge value]`.

```
[Tue Apr  5 20:46:04 AEST 2022] Checking domain: _acme-challenge.firewall.mydomain.com
[Tue Apr  5 20:46:04 AEST 2022] Checking domain: firewall.mydomain.com
[Tue Apr  5 20:46:04 AEST 2022] Checking domain: mydomain.com
[Tue Apr  5 20:46:04 AEST 2022] _domain_id='609dddf4b9337b4899666f24
609dd9c1410efd3c571ecfa2'
[Tue Apr  5 20:46:04 AEST 2022] _sub_domain='_acme-challenge.firewall'
[Tue Apr  5 20:46:04 AEST 2022] _domain='mydomain.com'
[Tue Apr  5 20:46:04 AEST 2022] dns_zones/609dddf4b9337b4899666f24
609dd9c1410efd3c571ecfa2/dns_records
[Tue Apr  5 20:46:04 AEST 2022] data='{"type":"TXT", "hostname":"_acme-challenge.firewall", "value":"[challenge value]", "ttl":"10"}'
[Tue Apr  5 20:46:04 AEST 2022] POST
[Tue Apr  5 20:46:04 AEST 2022] _post_url='https://api.netlify.com/api/v1/dns_zones/609dddf4b9337b4899666f24
609dd9c1410efd3c571ecfa2/dns_records'
[Tue Apr  5 20:46:04 AEST 2022] body='{"type":"TXT", "hostname":"_acme-challenge.firewall", "value":"[challenge value]", "ttl":"10"}'
[Tue Apr  5 20:46:04 AEST 2022] _postContentType
[Tue Apr  5 20:46:04 AEST 2022] Http already initialized.
[Tue Apr  5 20:46:04 AEST 2022] _CURL='curl --silent --dump-header /tmp/acme/firewall.mydomain.com//http.header  -L '
[Tue Apr  5 20:46:04 AEST 2022] Please refer to https://curl.haxx.se/libcurl/c/libcurl-errors.html for error code: 3
[Tue Apr  5 20:46:04 AEST 2022] _ret='3'
[...]

PR #4013, opened 2022-04-05
Fix / Netlify API script is unable to parse response containing similar domains

thekevinbrown:
When using Netlify integration while having both of these domains configured in Netlify:

- example.com
- example.com.au

The regex in the current script fails to parse the response. [See details and example log here](https://github.com/acmesh-official/acme.sh/issues/3088#issuecomment-1088653961)

This PR updates the regex to also check for the closing quote after the domain name so that only exact matches are returned.

I've tested on my mac. Let me know if there's anything else needed to merge this PR.

### DIFF
--- a/dnsapi/dns_netlify.sh
+++ b/dnsapi/dns_netlify.sh
@@ -114,7 +114,7 @@ _get_root() {
     fi
 
     if _contains "$response" "\"name\":\"$h\"" >/dev/null; then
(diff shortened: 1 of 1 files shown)


thekevinbrown:
I've enabled Actions on the fork, but I'm not sure what to do from this point based on that Wiki page. I haven't added any new secrets (so I assume I don't need to set them?) and there doesn't seem to be a button anywhere to actually run the tests. What should I do next?

Neilpang:
follow the wiki: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

PR #4017, opened 2022-04-06
Fix / Netlify API should only match exact domain matches.

thekevinbrown:
Also check for the closing quote so that only exact domain matches are found.


thekevinbrown:
I did the following:
- I deleted my previous fork
- I forked the repository again
- I enabled GitHub actions before pushing my commit.
- I then followed all instructions on this page exactly https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

PR #4034, opened 2022-04-12
LoopiaAPI error handling isn't compatible with FreeBSD

ccope-netgate:
The error handling in dns_loopia.sh uses grep -P, which isn't compatible with FreeBSD. I made this pull request to instead use sed. It's tested and working on both Linux and FreeBSD now.


Neilpang:
please enable the DNS ci checks:  https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

PR #4035, opened 2022-04-13
LoopiaAPI error handling isn't compatible with FreeBSD

ccope-netgate:
The error handling in dns_loopia.sh uses grep -P, which isn't compatible with FreeBSD. I made this pull request to instead use sed. It's tested and working on both Linux and FreeBSD now.


ccope-netgate:
> can you please fix the DNS ci checks?
> 
> https://github.com/ccope-netgate/acme.sh/actions/runs/2163397488

I don't have an account for LoopiaAPI. This simply fixes the last pull request that was merged that should have failed.

These are the lines I used to test. The first, which is in the current code works on Linux, but fails on FreeBSD. The second works on both, which is in this pull request.
```
echo "<methodResponse><fault><value><struct><member><name>faultCode</name><value><int>429</int></value></member><member><name>faultString</name><value><string>Rate Limit Exceeded</string></value></member></struct></value></fault></methodResponse>" | grep -oPm1 "(?<=<string>)[^<]+"
```
```
echo "<methodResponse><fault><value><struct><member><name>faultCode</name><value><int>429</int></value></member><member><name>faultString</name><value><string>Rate Limit Exceeded</string></value></member></struct></value></fault></methodResponse>" | sed 's/.*<string>\(.*\)<\/string>.*/\1/'
```

PR #4037, opened 2022-04-15
Update dns_mydevil.sh

ahwayakchih:
MyDevil's tool now requires y|n confirmation to remove record. This updated passes "y" to the tool, so it keeps working automatically.


Neilpang:
please fix the CI check in your fork:
https://github.com/ahwayakchih/acme.sh/actions/runs/2171714081

ahwayakchih:
@Neilpang i'm not sure how to handle this, because mydevil's api is local only, through SSH. I would have to implement special handling for the tests only, to ssh to some account and call the specific commands there.

PR #4038, opened 2022-04-15
docker: use `sleep infinity` instead `sleep 1`

pexcn:
参考：https://github.com/linuxserver/docker-wireguard/blob/master/root/etc/services.d/wireguard/run#L12-L14

经测试过可以接收 Ctrl+C 信号

相关PR: https://github.com/acmesh-official/acme.sh/pull/3915


Neilpang:
merged into dev, thanks.

PR #4042, opened 2022-04-19
added: ipv6 identifier support

xiaohuilam:
The original code before my patch, has problem to process with IPv6 addresses(such as `2402:4e00:1a10:1500:0:9557:d561:34ed`, which'll be seperated into `2402`, first part before the `:`), and causes authz error in later logic(can't grep domain challenge datas).

Currently, there's no Public ACME CA issuing ipv6 address certificate, but [will be soon](https://crt.sh/?id=6574918840).


Neilpang:
merged to dev

PR #4045, opened 2022-04-20
handle challenge-alias "false"

axelhahn:
allows --challenge-alias (fqdn|false) to enable/ disable dns auth per domain.

./acme.sh/acme.sh --issue \
    -d host1.example.com --challenge-alias false \
    -d extern1.example.net --challenge-alias example.com \
    -d host2.example.com --challenge-alias false \
    -d extern2.example.net --challenge-alias example.com \
    --dns dns_infoblox --staging

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4680,6 +4680,7 @@ $_authorizations_map"
           _dns_root_d="$(echo "$_dns_root_d" | sed 's/*.//')"
         fi
         _d_alias="$(_getfield "$_challenge_alias" "$_alias_index")"
+        test "$_d_alias" = "false" && _d_alias=""
         _alias_index="$(_math "$_alias_index" + 1)"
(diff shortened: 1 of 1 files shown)


axelhahn (commented):
Yep, that's better. Using $NO_VALUE is a clean solution compared to a hard coded string.

Neilpang:
please update the usage here:

https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode

axelhahn:
I added a section "Mix dns alias and default dns auth" in https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode
But it works only if the new code line was published. Shall we add a "coming soon"?

Neilpang:
Thanks,  I will publish the code soon.

Issue #3007, opened 2020-06-25
Reload command not executed during cron when encoded length exceeds base64 single line limit


jshafton:
Steps to reproduce
------------------

Issue a cert with a reload command that, when base64 encoded, exceeds the limit for single line base64 encoding.

Example:

```
/opt/acme.sh/acme.sh --debug --home /etc/acme.sh --server https:********@z****.com --dns dns_aws --issue -d 'staging.network360.com' -d '*.staging.network360.com' --reloadcmd "/etc/acme.sh/reload-command /etc/certs/network360/combined.pem root" --debug --force --log initial-cert-issue.log --dnssleep 30
```

Debug log
-----------------

```
$ /opt/acme.sh/acme.sh --debug --home   /etc/acme.sh --server https://acme-v02.api.letsencrypt.org/directory --syslog 6 --cron
[Thu Jun 25 22:09:10 UTC 2020] Lets find script dir.
[Thu Jun 25 22:09:10 UTC 2020] _SCRIPT_='/opt/acme.sh/acme.sh'
[Thu Jun 25 22:09:10 UTC 2020] _script='/opt/acme.sh/acme.sh'
[Thu Jun 25 22:09:10 UTC 2020] _script_home='/opt/acme.sh'
[Thu Jun 25 22:09:10 UTC 2020] Using config home:/etc/acme.sh
https://github.com/Neilpang/acme.sh
v2.8.5
[Thu Jun 25 22:09:10 UTC 2020] Using server: https://acme-v02.api.letsencrypt.org/directory
[Thu Jun 25 22:09:10 UTC 2020] Running cmd: cron
[Thu Jun 25 22:09:10 UTC 2020] Using config home:/etc/acme.sh
[Thu Jun 25 22:09:10 UTC 2020] ACME_DIRECTORY='https://acme-v02.api.letsencrypt.org/directory'
[Thu Jun 25 22:09:10 UTC 2020] ===Starting cron===
[Thu Jun 25 22:09:10 UTC 2020] Using config home:/etc/acme.sh
[Thu Jun 25 22:09:10 UTC 2020] ACME_DIRECTORY='https://acme-v02.api.letsencrypt.org/directory'
[Thu Jun 25 22:09:10 UTC 2020] _stopRenewOnError
[Thu Jun 25 22:09:10 UTC 2020] _set_level='2'
[Thu Jun 25 22:09:10 UTC 2020] di='/etc/acme.sh/staging.network360.com/'
[...]

PR #4057, opened 2022-04-28
issue 3007 - fix base64 decoding logic for single- and multi-line encoded string

mrakopes:
#3007 

The parameters for `openssl base64 -d` in _dbase64() are wrong. When used with current version of openssl, its actualy working fine as far as i know but the parameters are wrong and in could cause other unexpected bugs or issues.

More info in my comment: https://github.com/acmesh-official/acme.sh/issues/3007#issuecomment-1112084766

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -983,9 +983,9 @@ _base64() {
 #Usage: multiline
 _dbase64() {
   if [ "$1" ]; then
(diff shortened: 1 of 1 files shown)


Neilpang:
thanks

annieoxe:
I have Openssl 1.0.1 and during registration it fails because `openssl base64 -d` returns nothing.

PR #4089, opened 2022-05-14
Support for update bind9 via "nsupdate -y"

moose-kazan:
Hi!

Existing "dns_nsupdate.sh" used key from file, but sometimes it not possible, and i want to provide key as cmdline param (via "-y" option). I wrote script "dns_nsupdatey.sh", based on "dns_nsupdate.sh", but uses "-y" param, instead of "-k".


Neilpang:
is this pr working?
https://github.com/acmesh-official/acme.sh/pull/4058

moose-kazan:
#4058 looks like good solution

PR #4096, opened 2022-05-16
Fix dns_nederhost to work correctly with wget instead of curl.

sebastiaanhoogeveen:
The dns_nederhost DNS API relies on the exact HTTP status code to be
returned (e.g.  204); however, the _get function always returns 200 for a
succesful call when using wget instead of curl.  This patch fixes this by
using the _post function for all requests done by dns_nederhost.


Neilpang:

sebastiaanhoogeveen:
The DNS.yml workflow has run for the (trivial) second commit; all tests passed. Thanks!

PR #4108, opened 2022-05-27
fix response parsing errors

egorogl:
error 1: server's replies is not contains spaces between name and value (in json)
error 2: the "id" in the response is before the "content" attribute


egorogl:
Screenshot_20220528_202113

Neilpang:
merged,  thanks.

PR #4115, opened 2022-05-31
make ip-whitelisting configurable for DNS TransIP

koter84:
i need to use the TransIP API without a whitelisted IP because of dynamic IP.
it wasn't possible to configure this without changing the dns_transip.sh script, so i made the existing variable configurable, yet retain the default of false

### DIFF
--- a/dnsapi/dns_transip.sh
+++ b/dnsapi/dns_transip.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env sh
 TRANSIP_Api_Url="https://api.transip.nl/v6"
 TRANSIP_Token_Read_Only="false"
-TRANSIP_Token_Global_Key="false"
 TRANSIP_Token_Expiration="30 minutes"
 # You can't reuse a label token, so we leave this empty normally
 TRANSIP_Token_Label=""
@@ -96,7 +95,11 @@ _transip_get_token() {
   nonce=$(echo "TRANSIP$(_time)" | _digest sha1 hex | cut -c 1-32)
   _debug nonce "$nonce"
 
-  data="{\"login\":\"${TRANSIP_Username}\",\"nonce\":\"${nonce}\",\"read_only\":\"${TRANSIP_Token_Read_Only}\",\"expiration_time\":\"${TRANSIP_Token_Expiration}\",\"label\":\"${TRANSIP_Token_Label}\",\"global_key\":\"${TRANSIP_Token_Global_Key}\"}"
+  # make IP whitelisting configurable
+  TRANSIP_Token_Global_Key="${TRANSIP_Token_Global_Key:-$(_readaccountconf_mutable TRANSIP_Token_Global_Key)}"
+  _saveaccountconf_mutable TRANSIP_Token_Global_Key "$TRANSIP_Token_Global_Key"
+
+  data="{\"login\":\"${TRANSIP_Username}\",\"nonce\":\"${nonce}\",\"read_only\":\"${TRANSIP_Token_Read_Only}\",\"expiration_time\":\"${TRANSIP_Token_Expiration}\",\"label\":\"${TRANSIP_Token_Label}\",\"global_key\":\"${TRANSIP_Token_Global_Key:-false}\"}"
   _debug data "$data"
 
   #_signature=$(printf "%s" "$data" | openssl dgst -sha512 -sign "$TRANSIP_Key_File" | _base64)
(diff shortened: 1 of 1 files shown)


Neilpang:
fix the ci checks first:  https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

koter84:
getting the CI to run correctly took more work then the actual change, since the TransIP API requires a keyfile, and not a variable, so i added some code to create a keyfile from the base64 encoded keyfile in the secrets..

i tried to decode this with the base64 -d command, but that isn't available by default on the BSD OS-es, but then i found the _dbase64 function in acme.sh which uses openssl and works as expected.

the workflow is currently running, and i expect it to complete successfully, since it managed to get through the whole workflow before removing a bunch of _debug messages, and the last rebase to upstream/dev of course..

koter84:
@Neilpang just did a new rebase, could you approve the workflow?

koter84:
@Neilpang CI checks are all fixed

dnsapi/dns_transip.sh
```diff
@@ -139,6 +141,17 @@ _transip_setup() {
   _saveaccountconf_mutable TRANSIP_Username "$TRANSIP_Username"
   _saveaccountconf_mutable TRANSIP_Key_File "$TRANSIP_Key_File"
 
+  # for automated testing, put 'base64 -w0 keyfile' output into github secret for TRANSIP_Key_File variable
+  if [ -n "$TEST_DNS" ]; then
```
Neilpang, line 145: why do we need this ?

koter84: Because the TransIP api requires an RSA key-file for authentication, but I didn't find a way to add a file to the secrets.
Also couldn't find how others before me did that with the CI, so chose this way to create a key file from the github secrets.

Neilpang: upload your key file to somewhere, like:  https://transfer.sh/

And set `TRANSIP_Key_File=https://xxxx.download.url/xxxxxx`

when you use `TRANSIP_Key_File`,  check if it exists as a local file first.
otherwise, check if it starts with  "https://...", then just `get()` it and use it.

dnsapi/dns_transip.sh
```diff
@@ -139,6 +141,14 @@ _transip_setup() {
   _saveaccountconf_mutable TRANSIP_Username "$TRANSIP_Username"
   _saveaccountconf_mutable TRANSIP_Key_File "$TRANSIP_Key_File"
 
+  # download keyfile if it's an URL
+  if [ "$(echo "$TRANSIP_Key_File" | head -c8)" = "https://" ]; then
```
Neilpang, line 145: use the `_startswith()` function

dnsapi/dns_transip.sh
```diff
@@ -139,6 +141,14 @@ _transip_setup() {
...
+    _debug "download transip keyfile"
+    temp_key_file=$(mktemp /tmp/transip.XXXXXXXXXX)
```
Neilpang, line 147: Use the `_mktemp()`  function

dnsapi/dns_transip.sh
```diff
@@ -139,6 +141,14 @@ _transip_setup() {
...
+    _get "$TRANSIP_Key_File" >temp_key_file
```
Neilpang, line 148: check the return value of `_get()` function.

```
if ! _get xxxxxxx ; then
  _err "Error getting key file from : $TRANSIP_Key_File"
  return 1
fi
```

dnsapi/dns_transip.sh
koter84: i was searching for beginswith and couldn't find it... thanks for the pointer

koter84:
it's a bit of a learning curve, i use bash quite a lot for scripting, but not so much the POSIX variant, and mostly on Debian, Ubuntu and Fedora.
thanks for pointing me in the right direction!

dnsapi/dns_transip.sh
```diff
@@ -139,6 +141,17 @@ _transip_setup() {
   _saveaccountconf_mutable TRANSIP_Username "$TRANSIP_Username"
   _saveaccountconf_mutable TRANSIP_Key_File "$TRANSIP_Key_File"
 
+  # download keyfile if it's an URL
+  if _startswith "$TRANSIP_Key_File" "https://"; then
+    _debug "download transip keyfile"
+    temp_key_file="$(_mktemp)"
+    if ! _get "$TRANSIP_Key_File" >"$temp_key_file"; then
```
Neilpang, line 148: add  `chmod 600 "$temp_key_file" ` before  puting the key in it.

Neilpang:
can we delete the temp key file ?

koter84:
Well, I was thinking about that, but it should only delete the file during the test run, otherwise people lose their private key, so I figured since it's a short lived VM, and the files are rather small, I could just let the destruction of the VM handle that.
But now that I think of it, it's actually a new feature where you could have the private key on a local webserver, so then it should be cleaned, and documented that you could also use a URL for the key.

You think I should check for http:// as well then? Or just http?

koter84:
I'll add the changes tomorrow probably..

Neilpang:
delete the temp key file only when it's downloaded  from web.
If the file is not a URL,  don't delete it.

koter84:
Apparently Solaris doesn't accept the CA for transfer.sh so i used PasteBin instead, also had to change the delete function a bit to make sure the URL gets restored after deleting the temp-file.

Right now the checks are all on green.. :-D

dnsapi/dns_transip.sh
```diff
@@ -139,6 +141,18 @@ _transip_setup() {
   _saveaccountconf_mutable TRANSIP_Username "$TRANSIP_Username"
   _saveaccountconf_mutable TRANSIP_Key_File "$TRANSIP_Key_File"
```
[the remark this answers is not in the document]
koter84: i thought i did that after the get, in the get_token function, but it seems i didn't

fixed now

PR #4139, opened 2022-06-11
deploy api script to upload certs to proxmox using proxmox api

wsellitti:

### DIFF
--- a/deploy/proxmoxve.sh
+++ b/deploy/proxmoxve.sh
@@ -107,7 +107,7 @@ proxmoxve_deploy(){
   _json_payload=$(cat << HEREDOC
 {
   "certificates": "$(tr '\n' ':' < "$_cfullchain" | sed 's/:/\\n/g')",
-  "key": "$(tr '\n' ':' < "$_ckey" |sed 's/:/\\\n/g')",
+  "key": "$(tr '\n' ':' < "$_ckey" |sed 's/:/\\n/g')",
   "node":"$_node_name",
   "restart":"1",
   "force":"1"
@@ -115,7 +115,7 @@ proxmoxve_deploy(){
 HEREDOC
 )
   _debug2 Payload "$_json_payload"
-
+  
   # Push certificates to server.
   export _HTTPS_INSECURE=1
   export _H1="Authorization: PVEAPIToken=${_proxmoxve_header_api_token}"
--- a/deploy/proxmoxve.sh
+++ b/deploy/proxmoxve.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+
+# Deploy certificates to a proxmox virtual environment node using the API.
+#
+# Environment variables that can be set are:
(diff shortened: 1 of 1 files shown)


deploy/proxmoxve.sh
```diff
@@ -0,0 +1,123 @@
...
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
```
Neilpang, line 32: use `_getdeployconf ` and `_savedeployconf`

see example:

https://github.com/acmesh-official/acme.sh/blob/f897ab4eb85acd3cb80a261381fa43be0f25fd62/deploy/docker.sh#L22

deploy/proxmoxve.sh
```diff
@@ -0,0 +1,123 @@
...
+  "key": "$(tr '\n' ':' < "$_ckey" |sed 's/:/\\n/g')",
+  "node":"$_node_name",
+  "restart":"1",
+  "force":"1"
+}
+HEREDOC
+
+  # Push certificates to server.
+  #
+  # --insecure is to ignore certificate errors.
```
Neilpang, line 106: don't use `curl`.

use `_post()` function.

define `export HTTPS_INSECURE=1`

deploy/proxmoxve.sh
Neilpang: use `_savedeployconf` if the vlaue is not empty.

Neilpang:
please write the usage here: 

wsellitti:
Done. Thank you for your patience.

Neilpang:
the usage is not detail enough.
please discribe the usage step by step from zero,  as if the node is just setup as blank.

wsellitti:
Thanks. Does that work better?

deploy/proxmoxve.sh
```diff
@@ -0,0 +1,132 @@
...
+  "node":"$_node_name",
+  "restart":"1",
+  "force":"1"
+}
+HEREDOC
+  )
+  _debug2 Payload "$_json_payload"
+
+  # Push certificates to server.
+  export _HTTPS_INSECURE=1
```
ohartl, line 128: This should have been without the underscore: `export HTTPS_INSECURE=1`. Currently this causes the deployment to fail, when the proxmox cert is not valid anymore or self signed (initial deployment).

See https://github.com/acmesh-official/acme.sh/blob/377a37e4c9c23bb6988fe5f8863f21b19d3e3a40/acme.sh#L1966
and https://github.com/acmesh-official/acme.sh/blob/377a37e4c9c23bb6988fe5f8863f21b19d3e3a40/acme.sh#L1967

ohartl: This was also mentioned by @Neilpang in https://github.com/acmesh-official/acme.sh/pull/4139#discussion_r900708439

PR #4150, opened 2022-06-22
fix: Challenge not skipped for pre-validated wildcard domain orders

defnull:
Some CAs auto-validate orders based on account-level rules and do not require a challenge at all. Sectigo also introduced a non-standard challenges named `sectigo-dns-01`, presumably to work around this issue in certbot. This works fine for non-wildcard domains in acme.sh, but wildcard domains are rejected because acme.sh hard-codes `dns-01` as the only allowed challenge for wildcard domains, which is not offered by Sectigo.

This change simply moves the `"status":"valid"` check up a bit and ignores challenge type mismatches or missing tokens if no challenge is needed anyway.

With this patch, I successfully issued a wildcard domain certificate with sectigo.com using either `-w /tmp/not/required/anyway` or `--dns --yes-I-know-dns-manual-mode-enough-go-ahead-please`. The challenge is completely skipped for pre-validated orders.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4600,28 +4600,26 @@ $_authorizations_map"
         thumbprint="$(__calc_account_thumbprint)"
       fi
 
+      keyauthorization=""
+
+      if echo "$response" | grep '"status":"valid"' >/dev/null 2>&1; then
+        _debug "$d is already valid."
+        keyauthorization="$STATE_VERIFIED"
+        _debug keyauthorization "$keyauthorization"
+      fi
+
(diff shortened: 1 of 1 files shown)


defnull:
The failing check seems to be caused by `"The server could not connect to validation target"`. I can not see how this might be caused by this change. It looks unrelated to me.

homer2012:
It is possible to integrate this patch ? https://gitlab-ce.gwdg.de/gwdg/acme.sh/-/commit/f4dd4ba8c2c3bd87479c03c8a73460d83e924b61 please

PR #4161, opened 2022-06-27
remove dns_do as it does not work anymore

seidler2547:
The API that it uses was shut down in May 2022


Neilpang:
please do not remove the dns_do from the wiki page.  please revert it.
And change its content to "Removed".
Example:

image

henrikalves:
@Neilpang can you merge this please?

PR #4172, opened 2022-07-08
Fix for issue #3735

SBado:
This PR fix issue #3735, but it adds a (optional) dependency to [jq](https://stedolan.github.io/jq/) in OPNsense.
Thanks to @koelle25 for his improvements to the code.

### DIFF
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -150,7 +150,8 @@ _get_root() {
       return 1
     fi
     _debug h "$h"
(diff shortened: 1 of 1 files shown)


koelle25:
Thanks @SBado for opening a PR with this and mentioning me :) Nevertheless I think @blablup is right regarding not adding the `jq` dependency.

Also, his/her PR #4189 seems to do a good job in fixing this issue, too. The new RegEx seems to be a lot more robust.

I'd vote for closing this PR in favor of #4189.

PR #4174, opened 2022-07-09
Also restart Postfix container for Mailcow

Ry3nlNaToR:
Hi, The current deploy script for Mailcow only restarts Nginx and Dovecot but not Postfix resulting in Postfix not using the current ssl cert, Postfix did get restarted in the past but got removed in commit 201673ca8aabcd4becd90119b3d0118078daedee.

This pull request fixes that issue by also restarting Postfix.

Ryan

### DIFF
--- a/deploy/mailcow.sh
+++ b/deploy/mailcow.sh
@@ -67,7 +67,7 @@ mailcow_deploy() {
     return 1
   fi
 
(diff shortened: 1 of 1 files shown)


Neilpang:
@Ry3nlNaToR  @quthla 

Please double check if this pr is necessary and reply here.
otherwise, I will revert this pr.

Thanks

PR #4182, opened 2022-07-14
Update dns_world4you

NerLOR:
Fix small issues in the dns_world4you script.


dnsapi/dns_world4you.sh
```diff
@@ -12,7 +12,7 @@ RECORD=''
 
 # Usage: dns_world4you_add <fqdn> <value>
 dns_world4you_add() {
-  fqdn="$1"
+  fqdn=$(echo "$1" | tr '[:upper:]' '[:lower:]')
```
Neilpang, line 15: please use `_lower_case()` function

PR #4184, opened 2022-07-14
dns_ovh: save OVH_CK in all cases

### LINKED ISSUE #4183: dns_ovh: consumer key not saved if provided via environment variable
When we create an application with https://api.ovh.com/createToken ([wiki advanced usage](https://github.com/acmesh-official/acme.sh/wiki/How-to-use-OVH-domain-api#advanced-usage)),
we provide consumer key via `export OVH_CK="your consumer key"` before calling acme.sh.

It obviously works but it will probably fail on certificate renewal, because `OVH_CK` isn't saved in config (`SAVED_OVH_CK`)
=> it's only saved if `OVH_CK` is acquired through dns_ovh.

I'll do a PR.

Maxime-J:
Fixes #4183

### DIFF
--- a/dnsapi/dns_ovh.sh
+++ b/dnsapi/dns_ovh.sh
@@ -118,6 +118,7 @@ _initAuth() {
     #return and wait for retry.
     return 1
   fi
+  _saveaccountconf OVH_CK "$OVH_CK"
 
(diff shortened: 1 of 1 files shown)


Neilpang:
thank you.

PR #4189, opened 2022-07-16
Fix for issue #3735

blablup:
Fix for issue #3735

Sry for this late fix, I had problems with Solaris tests several month ago. Hadn't have the Time to look into it.

@Neilpang the main Problem is that the `egrep -o` regex is not 100% compatible with `sed -n` maybe `sed -En` would be better, but I didn't test. I had trouble to parse json with this incompatibility on all Platforms. 

Change to searchDomain reqeust, because it was a shorter answer and easier to parse. This still checks that the Domain is managed on that Device (type:master) and is enabled.

### DIFF
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -150,7 +150,7 @@ _get_root() {
       return 1
     fi
     _debug h "$h"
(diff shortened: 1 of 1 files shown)


koelle25:
@Neilpang Can you elaborate why this fix hasn't been integrated in the new 3.0.5 release?

Neilpang:
@koelle25   Once merged, it  must be in the new release.

PR #4193, opened 2022-07-17
Create notify hook for a custom shell script

ioqy:
Notify via a custom shell script.

The script will be called with the following parameters:

1. Subject
2. Content
3. Status Code

To set the notification hook:

```sh
export CUSTOMSCRIPT_PATH="/usr/local/bin/acme-notification.sh"

acme.sh --set-notify --notify-hook customscript
```

The custom script can look like this:

```sh
#!/usr/bin/env sh

subject="$1"
message="$2"
status="$3"

do-something "$subject ($status): $message"
```

Changes to the Wiki: https://github.com/acmesh-official/acme.sh/wiki/notify/24df66b68f4c2daa2d5c0cfbb9fbfca5e644bff9

### DIFF
--- a/notify/customscript.sh
+++ b/notify/customscript.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env sh
+
+# Support custom notification script
+
+# CUSTOMSCRIPT_PATH="/usr/local/bin/acme-notification.sh"
+
+customscript_send() {
+  _subject="$1"
+  _content="$2"
+  _statusCode="$3" #0: success, 1: error 2($RENEW_SKIP): skipped
+  _debug "_subject" "$_subject"
+  _debug "_content" "$_content"
+  _debug "_statusCode" "$_statusCode"
+
+  CUSTOMSCRIPT_PATH="${CUSTOMSCRIPT_PATH:-$(_readaccountconf_mutable CUSTOMSCRIPT_PATH)}"
+  if [ -n "$CUSTOMSCRIPT_PATH" ] && ! _exists "$CUSTOMSCRIPT_PATH"; then
+    _err "It seems that the command $CUSTOMSCRIPT_PATH is not in path."
+    return 1
+  fi
+
+  if [ -n "$CUSTOMSCRIPT_PATH" ]; then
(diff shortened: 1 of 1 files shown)


lscorcia:
+1 for this please. It's already in the wiki but it's really confusing as the supplied instructions don't work without this PR. Had to Google why and found this issue...

cerealconyogurt:
I am having an issue with this instructions:

```
$ acme.sh --set-notify --notify-hook customscript
[Sat Jul  6 22:33:17 BRT 2024] Set notify hook to: customscript
[Sat Jul  6 22:33:17 BRT 2024] Sending via: customscript
[Sat Jul  6 22:33:17 BRT 2024] Can not find the hook file for customscript
```

The current instruction is not working :-(

neilpang:
Thanks -- the feature is needed and the wiki already points at it. One blocking issue, a few minor:

**Blocking -- `eval` allows command injection and breaks on quotes:**

```sh
result=$(eval "'$CUSTOMSCRIPT_PATH' '$_subject' '$_content' '$_statusCode'" 2>&1)
```

A single quote in the message (e.g. `can't`) breaks it, and crafted content runs arbitrary commands. Drop `eval` and call directly:

```sh
_result="$("$CUSTOMSCRIPT_PATH" "$_subject" "$_content" "$_statusCode" 2>&1)"
```

Minor:
- Add an early `[ -z "$CUSTOMSCRIPT_PATH" ]` error+return, like the other hooks (currently an unset path falls through to a confusing error).
- Use `_clearaccountconf_mutable` to match `_saveaccountconf_mutable`.
- For an absolute path, `[ -f "$CUSTOMSCRIPT_PATH" ]` is more appropriate than `_exists` (which is a PATH lookup); reword the "not in path" message.
- Prefix `result` -> `_result`.

neilpang:
Thanks @ioqy -- this landed in `dev` as 44c045b0, based on your hook. One change from this PR: the target script is invoked directly instead of through `eval`, because the subject/content contain domain names and CA messages and the quoted-eval form allowed command injection through them. Closing this PR; credit to you for the hook.

PR #4194, opened 2022-07-18
Add Slack App notification hook

velimir:
Slack Incoming webhooks is a legacy custom integration - an outdated
way for teams to integrate with Slack. These integrations lack newer
features and they will be deprecated and possibly removed in the
future. Slack team do not recommend their use. Instead, it's suggested
to use Slack apps.


Neilpang:
please update the usage here:

velimir:
@Neilpang should be done now: https://github.com/acmesh-official/acme.sh/wiki/notify#5-set-notification-for-slack-app

Issue #4197, opened 2022-07-20
PANOS deploy-hook Commits Other Pending Changes


Hossy:
Steps to reproduce
------------------
1. Log into PANOS as an admin and make changes, but don't commit.
2. Use deploy-hook as a DIFFERENT admin user to deploy cert to Palo.

**Result:** Pending change from admin in step 1 are committed to the firewall.

sg1888:
@Hossy @Neilpang - Please close this issue.  It has been resolved In the latest release.  Thanks!

PR #4198, opened 2022-07-20
Fix Palo Alto commit to be a partial admin-focused commit

Hossy:
Issue #4197 -- Commit currently commits all pending changes to the firewall, which could be very, very bad.  This changes to commit to a partial commit of just the admin user acme.sh is using, which should be a dedicated admin user and not shared by anything else.

### DIFF
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -11,6 +11,9 @@ on:
       - 'dnsapi/*.sh'
       - '.github/workflows/DNS.yml'
 
+concurrency: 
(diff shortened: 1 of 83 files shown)


Hossy:
Additional reference: https://docs.paloaltonetworks.com/pan-os/9-1/pan-os-panorama-api/pan-os-xml-api-request-types/commit-configuration-api/commit#id4e36ab51-cce0-4bd1-8953-2413189ab1c6

Hossy:

PR #4200, opened 2022-07-20
Fix default Syno port to match documentation

Hossy:
Fix default Syno port to match documentation that says it defaults to 5001 when scheme is https

Reference: https://github.com/acmesh-official/acme.sh/wiki/deployhooks#20-deploy-the-cert-into-synology-dsm


Hossy:

Eagle3386:
This PR is probably outdated - nowadays, the linked Wiki page states both, the HTTP way with port 5000 as well as its HTTPS alternative with port 5001.
Additionally, it urges to use the HTTPS way instead of the unencrypted one. However, it still defaults to HTTP, because most people tend to run their NAS within their local network where `https://localhost:5001/` (or a fixed, private IP address instead) is simply invalid.

PR #4202, opened 2022-07-20
Encode username, password, did, and totp for security

Hossy:
Maybe not "security" but at least for sensitive information :-)


Hossy:

Eagle3386:
Since #4646 was merged, there's no need to store the TOTP secret anymore - in fact, any new setup won't even store it.
So, IMHO you should adapt to those changes.

Other than that, I do see your point, but wonder how a simple Base64 encoding should help - probably most if not all attackers spot the encoding & copy-pastes the string for decoding it, e.g. via Notepad++'s built-in function, right away…

PR #4217, opened 2022-08-02
Changes for the new KASAPI by all-inkl.com

Hobby-Student:
Hi,

I'm using the KASAPI by all-inkl.com. Unfortunately they changed the URL and some details in the response. I tested those modifications and they do work as expected.

If there are any questions, feel free to ask.

Kind regards
Hobby


Hobby-Student:
at first it worked. deployed to several machines and it stopped working. the respone seems to be too large and therefore does not get handled right. will come back with a better pull request.

PR #4219, opened 2022-08-02
Fix for (broken) KASAPI by all-inkl.com

Hobby-Student:
Hi,

I'm using the KASAPI by all-inkl.com. Unfortunately the current version is not working anyomore since a few weeks. I tested those modifications and they do work as expected.

This pull request is a complete overhaul in regards of the API calls. It's not using the test form anymore, it now depends on the official API documentation.

If there are any questions, feel free to ask.

Kind regards
Hobby

PR #4222, opened 2022-08-04
Fix and Upgrade KAS API Call. 

Marco4223:
Switching from formula.php to SOAP
Now session-based login 
Only record entries with corresponding values will now be deleted

### DIFF
--- a/dnsapi/dns_kas.sh
+++ b/dnsapi/dns_kas.sh
@@ -14,39 +14,43 @@
 # Git repo: https://github.com/phlegx/acme.sh
 # TODO: Better Error handling
 ########################################################################
-KAS_Api="https://kasapi.kasserver.com/dokumentation/formular.php"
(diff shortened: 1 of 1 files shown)


Marco4223:
As already known. Dockers failes with SSL connection timeout on random requests. 
I'm now in contact with ALL Inkl to check if this is a problem on there side or on Docker.
Currently its running on my two Synology NAS, PFSense an a also few users confirmend that its working on there side.

Hobby-Student:
just for completion #4219 
I was reading the documentation and talking to staff at all-inkl.com. My approach is working without a session, like the documentation describes. Don't know which way is "better"...

I'm currently extending things to catch failures and handle them accordingly. For now I did some kind of a hack to pass the github actions, which is not for production purpose. The version for production use is running on serveral opnsense boxes and ubuntu/debian servers. After passing the github actions I'll update my pull request accordingly.

PR #4224, opened 2022-08-06
Localhost deploy hook mimicking certbot behavior.

Github-Citizen:
Deploys cert files to centralized cert directory mimicking certbot behavior, allowing multiple services to share certs.

### DIFF
--- a/deploy/localhost.sh
+++ b/deploy/localhost.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env sh
 #
 #    Deploy cert to localhost similar to certbot behavior
 #
--- a/deploy/localhost.sh
+++ b/deploy/localhost.sh
@@ -2,7 +2,7 @@
 #
 #    Deploy cert to localhost similar to certbot behavior
 #
-#    export DEPLOY_LOCALHOST_PATH="/path/to/certs"
+#    export DEPLOY_LOCALHOST_ROOT_PATH="/path/to/certs"
 #
 #    Deploys as:
 #        /path/to/certs/domain.tld/privkey.pem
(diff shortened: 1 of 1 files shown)


Neilpang:
why not use the `--install-cert ` command instead ?

https://github.com/acmesh-official/acme.sh#3-install-the-cert-to-apachenginx-etc

Github-Citizen:
> why not use the `--install-cert ` command instead ?
> https://github.com/acmesh-official/acme.sh#3-install-the-cert-to-apachenginx-etc

Because using `--install-cert` requires hard coding four full paths for each cert. If you are setting up multiple domains you have then change the paths for each domain needing lots of manual editing.

The localhost hook does not require manual editing per certificate/domain. You just set the root path once when you install acme.sh, for example `/etc/letsencrypt/live`, and not have to worry about it anymore or change it when you issue new certs. Just like certbot it will auto create sub-directories for each domain and place the certificates accordingly. For example:
```
/etc/letsencrypt/live/example.com/*.pem
/etc/letsencrypt/live/domain.com/*.pem
```
No skin off my back if no one wants to use this, i can just save it for myself and drop it in. I was just sharing with the community because i assumed there must be others like me who use certs for multiple services such as httpd, postfix, dovecot and i didn't want to manually man-handle copies of the same certs for each service.

deploy/localhost.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/bash
```
Neilpang, line 1: use `sh` shebang,  see other files.

deploy/localhost.sh
```diff
@@ -0,0 +1,103 @@
...
+#
+#    Deploy cert to localhost similar to certbot behavior
+#
+#    export DEPLOY_LOCALHOST_PATH="/path/to/certs"
```
Neilpang, line 5: the variable name should be something like :  `DEPLOY_LOCALHOST_ROOT_PATH`

Neilpang:
just copying the certs is ok,  but there is no way to  restart/reload  the services when the cert is renewed.

The `install-cert` has a `--reload-cmd` to restart all the services.

Anyway, please fix the ci checks first.  we will see who need this.

deploy/localhost.sh
Github-Citizen: I changed the bash line and renamed DEPLOY_LOCALHOST_ROOT_PATH. What else do i need to do?

PR #4259, opened 2022-08-20
fix a issue, when profile not end with newline

Mon-ius:


Neilpang:
please check if the profile has an empty line first.
if there is no empty line at the end,  just `echo "" >> profile`.

PR #4275, opened 2022-08-28
Add removal deploy script

sestegra:
In my case, I'm using `vault-cli` and `removal` as deploy hooks.
By this way, no generated data are kept on the acme.sh running host.

### DIFF
--- a/deploy/removal.sh
+++ b/deploy/removal.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env sh
+
+#Here is a script to remove certs and private key.
+
+#returns 0 means success, otherwise error.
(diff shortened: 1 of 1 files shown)


Neilpang:
why do you need this behavior?

sestegra:
To reduce the attack surface, I need to remove the sensitive data (private key) from the server (where acme.sh runs) and store them on Vault.

My workflow is the following:
issue cert (or renew cert) -> deploy-hook vault-cli -> deploy-hook removal

sestegra:
@Neilpang Ok, I will use `--signcsr` instead.

PR #4277, opened 2022-08-29
Just resort _get_credential_token and _check_and_save for a bester perf.

Marco4223:
Hi @Neilpang , like last time the checks are not working tigether with the KAS API from All inkl. 
Thistime I only changed the order of _check_and_save and _get_credential_token because of a user reported that this is fixing there problem on there distribution. So please can you merge this PR?
Cheers
MarcO


Marco4223:
Sorry, Im unable to merg in github

Issue #4287, opened 2022-09-06
DNS API Tests fail on Solaris


nosilver4u:
I'm working on an integration for the new Bunny.net DNS API, and I've got it working great. But I thought I'd see if I could run the test suite to contribute the script for others to use and ran into an issue.

The Solaris VM appears to have trouble validating the SSL certs for api.bunny.net, as evidenced by a cURL error 60:

```
  [Tuesday, September  6, 2022 at 12:23:51 AM UTC] url='https://api.bunny.net/dnszone'
  [Tuesday, September  6, 2022 at 12:23:51 AM UTC] timeout=
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] Http already initialized.
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] _CURL='curl --silent --dump-header /root/.acme.sh/http.header  -L  -g '
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] Please refer to https://curl.haxx.se/libcurl/c/libcurl-errors.html for error code: 60
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] ret='60'
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] error in domain_list response: 
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] domain not found in your account for addition
  [Tuesday, September  6, 2022 at 12:23:52 AM UTC] Error add txt for domain:_acme-challenge.***
```

I don't think there's anything I can do about that, is that something you can resolve, or should I just submit pull request regardless?

nosilver4u:
Thanks bot, I'm running the latest, and already provided the relevant bit from the log :)

Neilpang:
please send PR first, so that I could check the solaris issues.

Thanks

nosilver4u:
I used your Solaris vm/shell and it complains that the cert is expired. The shell timed out before I could investigate too much more. The cert for api.bunny.net is issued by LE and expires in November, but I wonder if it's hitting an issue with the "cross-signed" root from the LE root CA transition a while back?
If memory serves correctly, it can be fixed with newer versions of openssl, but this says there might be some workarounds: https://community.letsencrypt.org/t/production-chain-changes/150739/3

At any rate, I'll submit a PR and we'll take it from there!

bbruun:
I have the same issue in the dns_dnsservices.sh script - it fails with 
[...]

PR #4288, opened 2022-09-06
add Bunny DNS API support

nosilver4u:
Adds DNS verification for domains using Bunny DNS.


nosilver4u:
As mentioned in #4287, the Solaris test isn't working due to some issue with an "expiration" issue, but I was able to get all the other tests to pass.

PR #4294, opened 2022-09-07
Fix OPNsense DNS plugin (again)

koelle25:
The latest OPNsense Bind plugin (_os-bind_ 1.24 on _OPNsense_ 22.7.2 in my case) now differentiates between master and slave zones:
https://github.com/opnsense/plugins/blob/master/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/DomainController.php#L42

The API Endpoint is now `/domain/searchMasterDomain` (instead of just `/domain/searchDomain`) and also the regular expression did not match anymore, because the `"domainname"` key now directly follows after the `"master"` key (without anything else in between).

This fixes [opnsense/plugins#3103](https://github.com/opnsense/plugins/issues/3103).

### DIFF
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -137,7 +137,7 @@ _get_root() {
   domain=$1
   i=2
   p=1
-  if _opns_rest "GET" "/domain/searchDomain"; then
+  if _opns_rest "GET" "/domain/searchMasterDomain"; then
     _domain_response="$response"
   else
     return 1
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -150,7 +150,7 @@ _get_root() {
       return 1
     fi
     _debug h "$h"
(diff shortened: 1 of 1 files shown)


koelle25:
> **Welcome** Please make sure you're read our [DNS API Dev Guide](../wiki/DNS-API-Dev-Guide) and [DNS-API-Test](../wiki/DNS-API-Test). Then reply on this message, otherwise, your code will not be reviewed or merged. We look forward to reviewing your Pull request shortly sparkles

I enabled the Actions workflows in my fork and re-pushed my changes. The workflows ran, but failed at the OpenBSD check (FreeBSD and all others before worked flawlessly). The error in the log is:
```
  [Wed Sep  7 ***2:5***:***7 UTC 2022] Register account Error: {
    "type": "urn:ietf:params:acme:error:serverInternal",
    "detail": "failed check for existing account",
    "status": 500
  }
```

I have no idea how to resolve this, as this has nothing to do with my changes.

Neilpang:
looks like a temp error from the CA,  just rerun your CI action.

koelle25:
Thanks, that did the trick :smile: Now all checks passed on my fork :+1:

PR #4297, opened 2022-09-08
add domain alias support for Digital Ocean API

nosilver4u:
Start 'i' at 1 to support "plain" domains in domain alias mode, like _acme-challenge.A.com => validate.B.com.
Also fixed an issue with 'found' not being (re)initialized each time _get_base_domain() is run.


nosilver4u:
I verified that dns_dgon.sh would not pass the tests previously and does now indeed pass all tests. Additionally, I specifically tested dns alias mode with a domain from another provider and an "alias" domain in my DO account to confirm that it works properly.

PR #4299, opened 2022-09-08
fixed 1984.hosting see issue #2851

iuliandita:
This should fix 1984.hosting not working (#2851).


Neilpang:
https://github.com/iuliandita/acme.sh/actions/runs/3013944999

PR #4318, opened 2022-09-25
Add dns_he_dyntxt, a simpler dns.he.net API.

pmarks-net:
These are the pros and cons of dns_he_dyntxt, compared to dns_he:

**Pros:**
- No need to store a dns.he.net account password on your server
- Uses a very simple write-only API

**Cons:**
- You must manually create placeholder _acme-challenge TXT records, and generate/copy the same DDNS key across all records.
- This script WILL FAIL to issue both a domain and its wildcard, because '-d example.com -d *.example.com' requires multiple TXT records. Switch to 'dns_he' if you need this feature.

### DIFF
--- a/dnsapi/dns_he_dyntxt.sh
+++ b/dnsapi/dns_he_dyntxt.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env sh
+
+########################################################################
+# Hurricane Electric hook script for acme.sh (dynamic TXT API)
+#
+# These are the pros and cons of dns_he_dyntxt, compared to dns_he:
+# Pros:
+# - No need to store a dns.he.net account password on your server
+# - Uses a very simple write-only API
+# Cons:
+# - You must manually create placeholder _acme-challenge TXT records,
+#   and generate/copy the same DDNS key across all records.
+# - This script WILL FAIL to issue both a domain and its wildcard, because
+#   '-d example.com -d *.example.com' requires multiple TXT records.
+#   Switch to 'dns_he' if you need this feature.
+#
+# Environment variable:
+#   HE_DynTXT_Key - DDNS key for all _acme-challenge TXT records
+########################################################################
+
+# Cheat sheet for passing the DNS.yml API test:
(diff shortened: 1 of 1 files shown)


pmarks-net:
This code passes DNS-API-Test, but only with `TEST_DNS_NO_WILDCARD=1`.

I realize that this API cannot support multiple TXT records, but people may find it useful to avoid storing a full account password on their server.

I think I could emit an obvious error message when adding multiple TXT records, by stashing a PID in _saveaccountconf_mutable, but that seems too janky to be worth the effort.

kraygy (commented):
I don't have access to approve this but it probably would have been useful to provide the updates to **acme.inc** for the **acme_domain_validation_method** dict with the new key, **dns_he_dyntxt**, and the associated field for the HE_DynTXT_key.

pmarks-net:
@kraygy `acme.inc` and `acme_domain_validation_method` don't exist in this project; looks like something related to pfsense?

kraygy:
> @kraygy `acme.inc` and `acme_domain_validation_method` don't exist in this project; looks like something related to pfsense?

@pmarks-net 
You are correct.  Sorry for the confusion.

myzhang1029:
Another alternative could be asking the user to append the FQDN to a common passphrase when setting the HE DDNS key, then we can simply POST `password=$HE_DynTXT_Key$hostname` (although it slightly degrades security). For example,
if `HE_DynTXT_Key=abc123` and the domain is `example.com`, set the DDNS key to `abc123_acme-challenge.example.com`.

If we are not constrained to support `sh`, we may also use variable indirection to allow baking the hostname into the variable name.

coreybrett:
Any chance this will get merged ?

I have been looking for a DNS provider that supports a per record API token, this is the only one I have found so far.

Neilpang:
the limitations are not acceptable.

markkuleinio:
FYI:

#5237 has been merged in for HE DDNS API support (with its limitations).

Waiting for the next acme.sh release to get it out in Proxmox VE for example.

PR #4329, opened 2022-09-29
[Deploy: synology_dsm] Make usage of DID with 2FA working for DSM 7

lippertmarkus:
Under DSM 7 the device ID (which is stored as a cookie when remembering a device and logging in with 2FA - see https://github.com/acmesh-official/acme.sh/wiki/Synology-NAS-Guide#deploy-the-default-certificate) also needs to be passed via URL and not just as a Cookie (see Example 4 under https://global.download.synology.com/download/Document/Software/DeveloperGuide/Os/DSM/All/enu/DSM_Login_Web_API_Guide_enu.pdf#page=17) to make 2FA login without OTP generation work.

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -108,7 +108,7 @@ synology_dsm_deploy() {
     _debug3 H1 "${_H1}"
   fi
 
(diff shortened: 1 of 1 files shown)


tresni:
What happens if you pass the parameter but aren't using TFA? Will having the parameter be there but empty still work in those situations?  I am running DSM 7 now so I'll try to give this a test over the weekend.

lippertmarkus:
In my tests it did work when those parameters were empty but please validate on your side as well :)

PR #4348, opened 2022-10-10
AzureDNS: Add support for AzureUSGovernment endpoints.

### LINKED ISSUE #4347: acme.sh Azure DNS Does not support USGovernment endpoints
The Azure DNS script (dnsapi/dns_azure.sh) does not support USGovernment endpoints and cannot work with anything but the Azure Cloud (commercial) endpoints.  This precludes use of Azure DNS services in other environments.  See https://learn.microsoft.com/en-us/powershell/module/az.accounts/get-azenvironment for powershell commands describing the available environments.

Steps to reproduce
------------------
Fill in your Tenant ID, Subscription ID, Application ID and Secret, utilizing a US Government environment application/DNS.  The attempt to validate will fail.

Debug log
-----------------
Your mileage may vary, but the first error I saw was:
```
 Confidential Client is not supported in Cross Cloud request.
```

I will have a PR for this shortly, hopefully.  I have a version of this working so the hard-coded endpoints appear to be the issue.

The only significant element of discussion for the PR will be the backwards compatibility.  I would recommend that if no environment is specified, that the AzureCloud environment is assumed, because every current user of this repo is already assuming that for the code to work.  Thus, I would silently set that variable if none is specified, but log an information message to that effect.

justchris1:
Fixes #4347
To preserve backwards compatibility, we will assume that if no environment is given, it defaults silently to AzureCloud.

For the purposes of this PR, I used the term 'environment' to conform to Azure's definition.  See https://learn.microsoft.com/en-us/powershell/module/az.accounts/get-azenvironment  The cmdlet is how to query information about the various environments.  I am unable to test against other environments, so I only added AzureUSGovernment, although the others would be easy to add if there was someone who could test.  

I also realize the three variables I selected are quite long: `active_directory_service_endpoint_resource_id`, `service_management_url`, and `active_directory_authority`.  However, I did so to further conform to the cmdlet referenced above.  If you run `Get-AzEnvironment -Name AzureUSGovernment | Format-List` you will see output which lines up to the endpoints I selected.  Hopefully this will make future modifications for more environments easier to maintain.

I have successfully tested this with a tenant in the AzureUSGovernment environment.

It has admittedly been a very long time since I wrote anything in bare SH, so comments welcomed!

### DIFF
--- a/dnsapi/dns_azure.sh
+++ b/dnsapi/dns_azure.sh
@@ -17,6 +17,7 @@ dns_azure_add() {
   AZUREDNS_SUBSCRIPTIONID="${AZUREDNS_SUBSCRIPTIONID:-$(_readaccountconf_mutable AZUREDNS_SUBSCRIPTIONID)}"
   if [ -z "$AZUREDNS_SUBSCRIPTIONID" ]; then
     AZUREDNS_SUBSCRIPTIONID=""
+    AZUREDNS_ENVIRONMENT=""
     AZUREDNS_TENANTID=""
     AZUREDNS_APPID=""
     AZUREDNS_CLIENTSECRET=""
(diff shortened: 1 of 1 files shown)


justchris1:
I have successfully tested this with a tenant in the AzureUSGovernment environment.

It has admittedly been a very long time since I wrote anything in bare SH, so comments welcomed!

justchris1:
I also realize the wiki will need to be updated, upon merge.  However, I am uncertain how to do that.  Please advise and I can make suggestions for the documentation updates.

justchris1:
Pushed changes to address syntax preference.

justchris1:
@Neilpang Could you please authorize the workflow approval?  I think I addressed the issue in my push a couple of days ago.  Thanks

PR #4350, opened 2022-10-11
dns_edgedns: Replace the ntp.org HTTP GET for retrieving timestamp with builtin _utc_date function.

### LINKED ISSUE #3973: dns_edgedns: Akamai Edge DNS API requests intermittently fail due to use of an unreliable time source
With `dns_edgedns`, API requests to Akamai intermittently fail because in generating an `Authorization` header, a timestamp derived from the `Date` response header returned from a HTTPS GET request for `https://www.ntp.org`. At the time of writing, the `Date` header is sometimes 55 minutes behind real time (taking TZ offsets into account).

Recreation of the issue:
```
$ date; curl -L --insecure -I --user-agent 'acme.sh/3.0.2 (https://github.com/acmesh-official/acme.sh)' -H '' -H '' -H '' -H '' -H '' https://www.ntp.org ; date 
Tue  8 Mar 14:57:51 GMT 2022
HTTP/1.1 200 OK
Date: Tue, 08 Mar 2022 15:02:07 GMT
Server: Apache
Content-Type: text/html
Strict-Transport-Security: max-age-16000000; includeSubDomains; preload;

Tue  8 Mar 14:57:51 GMT 2022
(reverse-i-search)`date': date; curl --resolve www.ntp.org:443:204.93.207.22 --dump-header ~/Desktop/test-dump -L --insecure -I --user-agent 'acme.sh/3.0.2 (https://github.com/acmesh-official/acme.sh)' -H '' -H '' -H '' -H '' -H '' https://www.ntp.org ; ^Cte 
```
NB: since capturing above, testing later on the server is now returning an accurate timestamp but it was incorrect for several hours.

The code which does this is in the `_edgedns_eg_timestamp()` function of `dnsapi/dns_edgedns.sh`.

A partial example output from `acme.sh`:
```
...
Adding txt value: REDACTED for domain:  re.dact.ed'
Bad request presented
Managed zone validation failed. Error response: 1'
Invalid domain
Error add txt for domain:re.dact.ed'
Please add '--debug' or '--log' to check more details.
See: https://github.com/acmesh-official/acme.sh/wiki/How-to-debug-acme.sh'
...
```

I've tested with `amce.sh` versions 3.0.1 and 3.0.2. The above output was to commands using `--issue` and `--domain-alias`. Apologies for not providing verbatim output from a command with `--debug 2` but I was struggling to find the cause of the issue until I instead ran `acme.sh` via `bash -x` although the output from that command would be far too large to post here. Via that approach, I was able to see the verbatim response body from Akamai:

```
{
  "type": "https://problems.luna.akamaiapis.net/-/pep-authn/request-error",
  "title": "Bad request",
  "status": 400,
  "detail": "Invalid timestamp",
  "instance": "https://redacted.luna.akamaiapis.net/config-dns/v2/zones/re.dact.ed/names/re.dact.ed/types/TXT",
  "method": "GET",
  "serverIp": "re.da.t.ed",
  "clientIp": "re.da.t.ed",
  "requestId": "redacted",
  "requestTime": "2022-03-08T14:53:12Z"
}
```

The `Invalid timestamp` message led me to find the issue with the timestamp (`_eg_timestamp`).

## Workaround

Amending `dnsapi/dns_edgedns.sh` to use a timestamps generated from the local system's clock worked for me. A dirty but quick way to implement this workaround, which worked for me, is to find this line:
```
_edgedns_eg_timestamp() {
```
and insert these two lines below it:
```
  _eg_timestamp=$(TZ=UTC date +%Y%m%dT%H:%M:%S%z)
  return
```

## Suggestions

If there's an intention to not trust the system's local clock by default, perhaps `acme.sh` could support an option a user could pass to tell it to use the local system clock.

tweitzel-akamai:
Previously this plugin would request a timestamp for Akamai OPEN API calls by doing an HTTP GET to ntp.org and grepping the headers. This became unreliable due to DDOS protection on that domain.

The plugin now uses acme.sh's own _utc_date function to request the current UTC time and reformats it according to the requirements (yyyyMMddTHH:mm:ss+0000) at https://developer.akamai.com/legacy/introduction/Client_Auth.html.

Fixes #3973

### DIFF
--- a/dnsapi/dns_edgedns.sh
+++ b/dnsapi/dns_edgedns.sh
@@ -359,19 +359,11 @@ _edgedns_rest() {
 }
 
 _edgedns_eg_timestamp() {
-  _debug "Generating signature Timestamp"
-  _debug3 "Retriving ntp time"
-  _timeheaders="$(_get "https://www.ntp.org" "onlyheader")"
-  _debug3 "_timeheaders" "$_timeheaders"
-  _ntpdate="$(echo "$_timeheaders" | grep -i "Date:" | _head_n 1 | cut -d ':' -f 2- | tr -d "\r\n")"
-  _debug3 "_ntpdate" "$_ntpdate"
-  _ntpdate="$(echo "${_ntpdate}" | sed -e 's/^[[:space:]]*//')"
-  _debug3 "_NTPDATE" "$_ntpdate"
-  _ntptime="$(echo "${_ntpdate}" | _head_n 1 | cut -d " " -f 5 | tr -d "\r\n")"
-  _debug3 "_ntptime" "$_ntptime"
-  _eg_timestamp=$(date -u "+%Y%m%dT")
-  _eg_timestamp="$(printf "%s%s+0000" "$_eg_timestamp" "$_ntptime")"
-  _debug "_eg_timestamp" "$_eg_timestamp"
+  _debug "Generating signature timestamp"
+  # Gets the current UTC date and time and reformats it to yyyyMMddTHH:mm:ss+0000.
+  # Reference: https://developer.akamai.com/legacy/introduction/Client_Auth.html
+  _eg_timestamp="$(_utc_date | sed 's/-//g' | sed 's/ /T/' | sed 's/$/+0000/')"
+  _debug3 "_eg_timestamp" "$_eg_timestamp"
 }
 
 _edgedns_new_nonce() {


tweitzel-akamai:
I have read the API dev guide and the API test guide, but I am unable to get github actions working in my forked repo and do not know where to get help for it. I have never used github actions before.

Neilpang:
please push any new changes to your fork to trigger the github actions to run.

tweitzel-akamai:
I got the tests running but the test fails in openbsd for unknown reasons. All the other systems are working.

Neilpang:
It's not passing yet:

https://github.com/tweitzel-akamai/acme.sh/actions/runs/3245182644

tweitzel-akamai:
I'm having a lot of difficulty with openbsd, as can be seen in this run (forgive the massive amounts of printf debugging; I will clean it up before rolling into this PR) https://github.com/tweitzel-akamai/acme.sh/actions/runs/3277730703

Because the openbsd tests do not have iconv or uconv, I have to fall back to perl to encode/decode UTF-8 using code found in the fritzbox deploy plugin.

The output is reporting that diagnostic data is in the "le_test_dnsapi.log" file, but I cannot find this file when I download the logs from the github action. Where is it? Also, is it possible to enable --output-insecure?

jschauma:
The git action logs all 404 by now; is there any way to trigger the action again so we can see what the errors look like?

Issue #4366, opened 2022-10-25
--cert-name


dusan-ivanco:
Hi - can you please please please implement --cert-name as is in certbot?

For example - i have 2 domains only with subdomains (that can change)

a1.domain1.tld
h8.domain1.tld

i3.domain2.tld
bb.domain2.tld

and it will be very nice and script friendly, that this domains path will be for example:
domain1.tld/domain1.tld.cer
domain1.tld/domain1.tld.conf
.
.
.

domain2.tld/domain2.tld.cer
domain2.tld/domain2.tld.conf
.
.
.

thank you

PR #4380, opened 2022-11-03
Add --cert-name option

nichivo:
I have lots of somewhat related domains in SAN certificates but there isn't really a single "primary" domain that indicates what other domains are in the certificate. It would be nice to be able to have certificate filenames that indicate their purpose rather than just use the first domain from the SAN list.

```
# this patch adds --cert-name to be used when issuing certs only
acme.sh --issue --cert-name demo-sites ...
# other functions like renewal still use --domain
acme.sh --renew --domain demo-sites
```

I've been using this patch for a while, and looks like there's at least one other person that wants it too! Should fix https://github.com/acmesh-official/acme.sh/issues/4366


Binokiller911:
This function can be available on production branch?

binokiller:
Why it’s be close without change?

PR #4381, opened 2022-11-04
Fix name.com DNS API for processing IDNs

mikelei8291:
The current name.com API implementation didn't parse the response from the API correctly when the domain name was an IDN.

This PR fixes the issue by converting the `fulldomain` argument of the `dns_namecom_add()` and `dns_namecom_rm()` functions to punycode representation of the IDN when necessary (the `idn` command won't be invoked if the value of `fulldomain` was not an IDN).


mikelei8291:
It was a small change and was tested locally with an IDN which I don't want to disclose to the public. The test was successful so I would assume this small change fixes the problem and won't affect existing functionalities.

If you would like to have it tested thoroughly with GitHub Actions, I will see if I could make this happen, but no guarantees.

Neilpang:
>> If you would like to have it tested thoroughly with GitHub Actions,

yes, please

PR #4382, opened 2022-11-06
Fix dns_transip: Added support for API KEY variable

tweemeterjop:
This PR adds support for supplying the API KEY in an environment variable. Some implementations that use acme do not support keyfiles and only environment variables (like pfsense).

The added code is based on the key + keyfile implementation from dns_oci.sh.

The provided API KEY should be in PEM format. We store the API KEY in account conf as base64 encoded. Also we now use a temp file to store the key and have openssl sign the request.

### DIFF
--- a/dnsapi/dns_transip.sh
+++ b/dnsapi/dns_transip.sh
@@ -5,6 +5,10 @@ TRANSIP_Token_Expiration="30 minutes"
 # You can't reuse a label token, so we leave this empty normally
 TRANSIP_Token_Label=""
 
+# One of the following two evironment variables is required:
+# - TRANSIP_Key_File: Path to a file containing the private key in PEM format; or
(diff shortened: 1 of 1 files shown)


neilpang:
1. Format mismatch: a key read from `TRANSIP_Key_File` is persisted base64-encoded (`_base64 <"$TRANSIP_Key_File"`), while a user-supplied `TRANSIP_Key` env var is raw PEM -- `_transip_get_token` writes whichever it has verbatim to the temp file for `_sign`, so one of the two paths signs with garbage. Store and decode consistently.
2. If `_mktemp` fails, the signing block is silently skipped and the request goes out with an empty `Signature` header -- error out instead.
3. The DNS-API-Test has never run on your fork -- required for any dnsapi change; rebase onto current dev first (this branch is from 2022).

PR #4384, opened 2022-11-07
update documentation for --cert-home parameter

fraenki:
Although the main use-case may be the `--install` command, this command also proves to be useful for the `--signcsr` and `--issue` commands.

This PR updates the CLI help text accordingly.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -6897,7 +6897,7 @@ Parameters:
 
   --accountconf <file>              Specifies a customized account config file.
   --home <directory>                Specifies the home dir for $PROJECT_NAME.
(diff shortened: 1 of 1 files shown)


fraenki:
@Neilpang Any chance to get this merged? Some projects rely on this parameter, so having it "officially" supported would be appreciated. Thanks!

fraenki:
Hi @Neilpang, could you please review this PR? It's a no-op change, it only updates the documentation. I make heavy use of the `--cert-home` option in both the OPNsense [acme.sh plugin](https://github.com/opnsense/plugins/tree/master/security/acme-client) and the [Puppet module for acme.sh](https://github.com/markt-de/puppet-acme), hence it's important for me that this option will not be limited to `--install` in a future release. :blush:

PR #4397, opened 2022-11-16
Openhabian Deployment script

vidia:
This change adds a deployment script that can be used to deploy keys to an install of [Openhabian](https://www.openhab.org/docs/installation/openhabian.html) (which is a way to install [OpenHab](https://www.openhab.org/), a smarthome controller tool). OpenHab uses a Jetty-based webserver to serve its website, so this script installs keys to the existing OpenHab keystore using the proper key names and passwords to make sure it is able to be used properly. This script also restarts the openhab service upon completion in order to make sure the new keys are picked up. 

# Documentation Requirements

This script runs `sudo service openhab restart` after installing the key. This (of course) needs either sudo, or for the user running acme.sh to be able to run that command unprompted. I solved this on my machine by adding this line to my sudoers: `openhabian ALL= NOPASSWD: /usr/sbin/service openhab restart` (for context, openhabian is the default user on openhabian), but I've also added a config option so that users can set their own restart command so they can ensure the command is executable by their user. What is the workflow to edit the wiki for a new PR like this? I'm happy to go in and edit the docs once this is approved, if that's what maintainers prefer. 

# Limitations

I suspect that this script would also be compatible with a plain install of OpenHab, but I don't currently have a way to properly test this so I've over-limited this to "openhabian" instead of OpenHab as a whole. I also suspect that this could be generalized to a generic Jetty deployment, but I am not familiar enough with those tools to be able to test it properly. 

# Testing

I am currently using this script in my own openhabian install so I've only been able to verify in openhabian.

### DIFF
--- a/deploy/openhabian.sh
+++ b/deploy/openhabian.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env sh
+
+# Config variables
+# DEPLOY_OPENHABIAN_KEYPASS : This should be default most of the time since a custom password requires openhab config changes
+# DEPLOY_OPENHABIAN_KEYSTORE : This should generate based on existing openhab env vars.
+
+openhabian_deploy() {
+
+    # Name parameters
+    _cdomain="$1"
+    _ckey="$2"
+    _ccert="$3"
+    _cca="$4"
+    _cfullchain="$5"
+
+    _debug _cdomain "$_cdomain"
+    _debug _ckey "$_ckey"
+    _debug _ccert "$_ccert"
+    _debug _cca "$_cca"
+    _debug _cfullchain "$_cfullchain"
+
+    # TODO: Load from config using _getdeployconf and print with _debug2
+    # Unclear if this is needed in this case.
+
+    # Define configurable options
+    _openhab_keystore=${DEPLOY_OPENHABIAN_KEYSTORE:-${OPENHAB_USERDATA}/etc/keystore}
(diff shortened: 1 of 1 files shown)


neilpang:
Thanks. The happy path works, but a few things to fix before this can be merged:

1. **The backup is taken but never restored.** You `cp` the keystore to `.bak`, but if `keytool -delete` succeeds and `-importkeystore` then fails, the old key is gone and the new one isn't in -- the keystore is left broken with no rollback. Restore from `.bak` on any failure after the backup.
2. **First run: `keytool -delete -alias mykey` fails when the alias doesn't exist yet**, aborting the deploy. Tolerate "alias not found" instead of returning 1.
3. **The keystore password is logged in plaintext** via `_debug _openhab_keypass` / `_debug2 DEPLOY_OPENHABIAN_KEYPASS` -- use `_secure_debug` / `_secure_debug2`.
4. **Private-key temp file leak:** the restart-failure `return 1` path leaves `$_new_pkcs12` (the PKCS12 with the private key) on disk instead of deleting it.
5. **Renewal may lose the keystore path:** `_savedeployconf` saves the raw (possibly empty) `DEPLOY_OPENHABIAN_*` values, not the resolved ones. If the user relied on the `$OPENHAB_USERDATA` default, a cron renewal without that env var resolves to `/etc/keystore`. Save the resolved value, or require the path to be set.

Also please add the usage to the wiki [deployhooks](https://github.com/acmesh-official/acme.sh/wiki/deployhooks) page.

PR #4409, opened 2022-11-26
Update ArvanCloud API URL

hatamiarash7:
The API URL of ArvanCloud changed from `https://napi.arvancloud.com` to `https://napi.arvancloud.ir`

PR #4415, opened 2022-11-28
fix: Update ArvanCloud DNS API domain name

sadeq-n-yazdi:
The ArvanCloud domain name changed from .com to .ir  and .com domain is not active now. so it need to update to work


hatamiarash7:
It's a duplicate. Fixed and merged in #4409

PR #4418, opened 2022-11-29
Updated domain tld of ArvanCloud dns api

RaminMT:
Hi,
This dns api has been broken for a while as of embargo against ArvanCloud company & freezing their common tld (com, net, org, ...) domains. They're now servicing on .ir domain.
I just updated the domains, that's it.


hatamiarash7:
This problem was fixed in #4409

Issue #3899, opened 2022-01-19
After upgrade, acmedns forgets the upgrade-server


mibes-acmesh:
Steps to reproduce
------------------
* Use dns-validation with `acme-dns`
* Use your own acme-dns-server (not the default)
* Upgrade `acme.sh`

The error is in the file `dnsapi/dns_acmedns.sh` on line 27/28, where the compatibility for the `ACMEDNS_UPDATE_URL` is missing.

```
27  #for compatiblity from account conf
28  ACMEDNS_USERNAME="${ACMEDNS_USERNAME:-$(_readaccountconf_mutable ACMEDNS_USERNAME)}"
29  _clearaccountconf_mutable ACMEDNS_USERNAME
```

We propose the following fix, which is UNTESTED, beware of typos etc., to be inserted betwen line 27 and 28

```
  ACMEDNS_TEMP="${ACMEDNS_UPDATE_URL:-$(_readaccountconf_mutable ACMEDNS_UPDATE_URL)}"
  _clearaccountconf_mutable ACMEDNS_UPDATE_URL
  if [ -n "$ACMEDNS_BASE_URL" ]; then
    ACMEDNS_BASE_URL=$(echo $ACMEDNS_TEMP | sed -e 's#/update##')
  fi
```

Neilpang:
sorry, I will fix it later.

mibes-acmesh:
On the other hand, when I think about it once more, how would you set an _initial_ alternative acme-dns server? Ideally you would set it once (i.e. in the account.conf), and use it then for the automatically created domains. But my proposed code would delete it from account.conf. Wouldn't it be better to modify the value (ACMEDNS_UPDATE_URL => ACMEDNS_BASE_URL) and save that in the account.conf for subsequent runs? It would then be used automatically for new subdomains/certificates...

neilpang:
Fixed in dev: when `ACMEDNS_BASE_URL` is not set (env or domain conf), the hook now falls back to the legacy `ACMEDNS_UPDATE_URL` from the account conf (stripping `/update`) and then saves the migrated value to the domain conf as usual. Per your second comment, the legacy key is intentionally NOT cleared, so the remaining domains can still migrate on their next renewal.

PR #4431, opened 2022-12-13
acmedns: Do not clear credentials from account.conf, and use BASE_URL

fulax:
Hello,

This Pull Request aims at taking care of two issues.

First, a minor one (#3899) by adding consistency on how settings are copied from account.conf to domainconf, and removing the now obsolete `ACMEDNS_UPDATE_URL` variable.

Second, a more serious issue. The code change between 3.0.1 and 3.0.2 (#3231) removed credentials from account.conf on first domain renewal. Then, when a second domain has to be renew a few days later, there is no longer any credentials to be copied from the account.conf to the domainconf, as the first renewal process cleared the configuration file. Which breaks compatibility.

As such, I removed the `_clearaccountconf_mutable` directives to keep the credentials around.

### DIFF
--- a/dnsapi/dns_acmedns.sh
+++ b/dnsapi/dns_acmedns.sh
@@ -25,12 +25,11 @@ dns_acmedns_add() {
   _debug "txtvalue $txtvalue"
 
   #for compatiblity from account conf
+  ACMEDNS_BASE_URL="${ACMEDNS_BASE_URL:-$(_readaccountconf_mutable ACMEDNS_BASE_URL)}"
   ACMEDNS_USERNAME="${ACMEDNS_USERNAME:-$(_readaccountconf_mutable ACMEDNS_USERNAME)}"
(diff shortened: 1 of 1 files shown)


fulax:
> **Welcome** Please make sure you're read our [DNS API Dev Guide](../wiki/DNS-API-Dev-Guide) and [DNS-API-Test](../wiki/DNS-API-Test). Then reply on this message, otherwise, your code will not be reviewed or merged. We look forward to reviewing your Pull request shortly sparkles

Hi, I have read both guides, and my changes only concern settings migration from account.conf for an existing api (which is not changed), so I'm not concerned as I'm not adding a new DNS API.

vincib:
Hi. FYI, I like this patch since it keeps the "_global_" acmedns settings in account.conf, which makes possible to get a certificate via acmedns for **new domains** without having to re-enter the acmedns credentials. 
without it, you need to RE-add your acmedns credentials to new domains each time...

PR #4457, opened 2023-01-04
Allow renew time relative to the expiration date

### LINKED ISSUE #3413: Consider to use openssl -enddate to determine certificate renewal
The default renew delay used by acme.sh is 60 days. This makes absolutely sense for 90-day certificates issued by LE. But amce.sh also supports other RFC-8555 compliant CAs. For eg. Sectigo issues 1-year certificates through ACME. In this case I think 60 days is not really appropriate. I know that one can set the delay (`Le_RenewalDays`) in each domain-conf but my idea would be to use `openssl -enddate` to determine the expiration date of a successful issued certificate and then substract the amount of days the certificate should be renewed earlier (eg. expiredate - 30 days).

I already run my own bash script as `--post-hook` to achieve this (I can provide the code if needed), but I think it could be a nice feature to have acme.sh support this out of the box...

AlexanderS:
You can now set a negative value for days, that will be subtracted from the expiration date of the generated certificate.

This will fix #3413.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1867,7 +1867,7 @@ _ssldate2time() {
   if date -j -f "%b %d %T %Y %Z" "$1" +"%s" 2>/dev/null; then
     return
   fi
-  _err "Can not parse _date2time $1"
+  _err "Cannot parse _ssldate2time $1"
   return 1
 }
 
--- a/acme.sh
+++ b/acme.sh
@@ -1018,7 +1018,19 @@ _checkcert() {
 #file
 _enddate() {
   _cf="$1"
-  ${ACME_OPENSSL_BIN:-openssl} x509 -noout -enddate -in "$_cf" | cut -d = -f 2
+  _res="$(${ACME_OPENSSL_BIN:-openssl} x509 -noout -enddate -in "$_cf")"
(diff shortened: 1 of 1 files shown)


AlexanderS:
I do not think, the error of the pipeline is related to the pull request.

saudiqbal:
Any update on this PR?

saudiqbal:
??? Nothing?

saudiqbal:
Looks like someone else needs to take over this project, acmesh-official is not doing anything for pull requests.

lukastribus:
Ultimately #4944 ARI support is going to fulfill this need in most cases, I can see that the work has already began.

The change in this PR would still be useful nonetheless of course.

saudiqbal (approved):
Please approve the PR for relative expiry date.

AlexanderS:
@neilpang Any comments on this? Did you maybe miss this PR by accident?

neilpang:
please also add thie usage in the showhelp() method, so that everyone knows it can be negative numbers.

AlexanderS:
I have updated the PR and included the various suggestions from Copilot (although the extensive error checking does not align with other parts of the code).

PR #4477, opened 2023-01-28
Add docker oathtool option

OrpheeGT:
oathtool binary is not available in DSM6 or DSM7.

It needs to sideload binary and dependencies from debian or anything else to make it works...

As we have docker synology package available on most of Synology products, using a docker container is a good alternative.

This commit replace oathtool binary with docker run commandline. the first time the command is launched, it will take some time to download missing docker image locally.

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -20,7 +20,7 @@
 # Dependencies:
 # -------------
 # - jq and curl
-# - docker (When using 2 Factor Authentication and SYNO_TOTP_SECRET is set)
+# - oathtool or docker (When using 2 Factor Authentication and SYNO_TOTP_SECRET is set)
 #
 #returns 0 means success, otherwise error.
 
@@ -94,10 +94,17 @@ synology_dsm_deploy() {
 
   otp_code=""
   if [ -n "$SYNO_TOTP_SECRET" ]; then
-    if _exists docker; then
-      otp_code="$(docker run --rm -it toolbelt/oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null | cut -b 1-6)"
+    if _exists oathtool; then
+      otp_code="$(oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null)"
+    elif _exists docker; then
+      if [[ "$(docker images -q toolbelt/oathtool:latest 2> /dev/null)" == "" ]]; then
+        read -e -p "docker is available but oathtool docker image must be downloaded, do you want to download it (Y) or abort (N) ? " choice
+        [[ "$choice" == [Yy]* ]] && docker image pull toolbelt/oathtool:latest && otp_code="$(docker run --rm -it toolbelt/oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null | cut -b 1-6)" || { echo "Abort requested or download failed"; return 1; }
+      else
+        otp_code="$(docker run --rm -it toolbelt/oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null | cut -b 1-6)"
+      fi
     else
-      _err "docker could not be found, install docker synology package to use SYNO_TOTP_SECRET"
+      _err "neither oathtool or docker could be found, install oathtool binary or docker synology package to use SYNO_TOTP_SECRET"
       return 1
     fi
   fi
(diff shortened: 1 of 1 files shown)


tresni:
I'm not sure that straight up replacing oathtool with docker is the right choice.

1. There are people that already have everything setup with oathtool and this will break their setup
2. People may not want to install docker
3. This will automatically pull the image without input from the user if they hit that code path

To me, those last 2 things are the major issue.  I would instead leave the oathtool code as is, put a check to see if `docker` is installed, if it is *and* they have either pulled the image already or set some variable in the configuration, only then would I execute a command that can pull down an new image.

Also, bear in mind that this script *does not* have to be run on Synology host, while it is likely the primary use case, I've tried to make sure that it can be run on any networked machine that has access to the synology host.  So getting oathtool may not be that big of an issue for many people.

OrpheeGT:
I restored oathtool binary option
and added a prompt request to download image when docker is available

tresni:
I'm definitely more comfortable with that.  Thanks for making the changes!

Neilpang:
send the pr to the dev branch.  never ever send pr to master branch.

deploy/synology_dsm.sh
```diff
@@ -96,8 +96,15 @@ synology_dsm_deploy() {
   if [ -n "$SYNO_TOTP_SECRET" ]; then
     if _exists oathtool; then
       otp_code="$(oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null)"
+    elif _exists docker; then
+      if [[ "$(docker images -q toolbelt/oathtool:latest 2> /dev/null)" == "" ]]; then
+        read -e -p "docker is available but oathtool docker image must be downloaded, do you want to download it (Y) or abort (N) ? " choice
```
Neilpang, line 101: don't use this interactive anwsers.
just show an info message ,and use the  "docker run"  command.

OrpheeGT: @tresni requested a prompt before pulling the docker image... do I added the interactive answer.

But I can remove the prompt and directly pull the image, it was my first attempt before @tresni's request.

I don't know how to comply with tresni's request but not requesting user interaction...

tresni: I think the answer is to display the info message and tell the user to install the docker image using `docker pull`.  Pulling down a container automatically into an unknown environment is a bad idea. So if @Neilpage is saying no to interactive commands, then the only thing to do is show a message and exit.

tresni: I think we need to put an option around the docker stuff anyway as `docker run` will pull an image if it's not available locally. So this whole block likely needs to change.

```
$ docker run --rm -it toolbelt/oathtool --base32 --totp BADD0G
Unable to find image 'toolbelt/oathtool:latest' locally
latest: Pulling from toolbelt/oathtool
8921db27df28: Pull complete
882084980d3c: Pull complete
Digest: sha256:f4c7fafb00fef5178ffbdf7b6e9f58e234617bb035967c3ba6f879f1f47a4bf3
Status: Downloaded newer image for toolbelt/oathtool:latest
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
oathtool: base32 decoding failed: Base32 string is invalid
```

deploy/synology_dsm.sh
```diff
@@ -96,8 +96,15 @@ synology_dsm_deploy() {
...
+        [[ "$choice" == [Yy]* ]] && docker image pull toolbelt/oathtool:latest && otp_code="$(docker run --rm -it toolbelt/oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null | cut -b 1-6)" || { _err "Abort requested or download failed"; return 1; }
+      else
+        otp_code="$(docker run --rm -it toolbelt/oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null | cut -b 1-6)"
+      fi
```
tresni, line 105:
```suggestion
    elif _exists docker && [ -n "$SYNO_USE_DOCKER" ] ; then
      otp_code="$(docker run --rm -it toolbelt/oathtool --base32 --totp "${SYNO_TOTP_SECRET}" 2>/dev/null | cut -b 1-6)"
```

tresni (changes requested):

deploy/synology_dsm.sh
tresni: Works for me!

Eagle3386:
@OrpheeGT sorry for being late to the party, but since #4646 was merged recently, there's no need for `oathtool` anymore.

While old setups continue to work as before, any new one uses Synology's API to get a `device ID` on first (manual) run, requiring just a one-time input of the TOTP code for the user account in question.
Any subsequent run will then use said `device ID` instead of requiring yet another TOTP code or even storing the TOTP secret _at all_ as the previous implementation did.

Therefore, IMHO, this PR is obsolete as there's neither a requirement for `oathtool` nor Docker at all anymore. Please, correct me if I'm wrong, though. 😅
Otherwise, this PR might be closed, reducing review workload for Neil as well. 😉

Eagle3386:
Well, I can't, because it's your PR & I'm only a contributor like you, not a maintainer.. 😅

PR #4506, opened 2023-02-15
Check validity of authz for wildcard domains, too

AlexanderS:
Authorizations for wildcard domains can also be already valid (esp. when using external account bindings).


AlexanderS:
This was already fixed by: #4150

PR #4509, opened 2023-02-17
Ssh deploy config sets

HRHDaniel:
I had a case where I needed to do ssh deploys to different servers with different configuration options.  Seems like that has also been requested in the forums.  I updated the ssh module so that it can run multiple times with different configuration options using a provided list of suffixes for the configuration variables.  It will also continue to support no suffixes so as to not break any current configuration or those that don't need this option.

Additionally there was a bug in _getdeployconf() not allowing it to read in values that contain spaces - 1 line fix in acme.sh for that bug.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2398,6 +2398,7 @@ _getdeployconf() {
     return 0 # do nothing
   fi
   _saved="$(_readdomainconf "SAVED_$_rac_key")"
+  _saved=$(echo "$_saved" | sed 's/"/\\"/g')
   eval $_rac_key="\"$_saved\""
   export $_rac_key
 }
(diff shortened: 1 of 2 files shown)


PMExtra:
You don't need `_migratedeployconf` with your new options, it's only designed to migrate the old configurations.

HRHDaniel:
@PMExtra Thanks for taking a look.  I believe the only one that is really a new option is "DEPLOY_SSH_CONFIG_SETS" and I don't call migrate for that.  The "old" options were suffixed with "${_config_suffix}" ... I left the call `_migratedeployconf ` for those even though the suffix was added as the config suffix is optional and will be blank for existing configurations that may still need to be migrated.

PMExtra:
Oh, I see. It's awesome.
I suggest you making only one thing with one PR.
Thus we can request to merge the bug fix as soon as possible, then evaluate the new features more carefully.

PMExtra:
Oh, I see. It's awesome.
I suggest you making only one thing with one PR.
Thus we can request to merge the bug fix as soon as possible, then evaluate the new features more carefully.

PR #4521, opened 2023-02-27
Added deploy script for minio

coderonline:
This is useful to have acme.sh running as a cronjob. I am using it in a FreeBSD jail to keep everything together in one place. At the time of writing minio required a certain format for its certificate. While I have not retested if this is really needed any more, the openssl conversion brings the certificates in a format, which still works in my setup. In one of my code comments I have linked the minio documentation, because it has more information which cipher suites are actually supported. I am not sure if I will be able to actively maintain this script, because I am not actively using minio at the moment.

### DIFF
--- a/deploy/minio.sh
+++ b/deploy/minio.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env sh
+#
+# Here is a script to deploy cert to minio server. This script can be called
+# directly to test its configuration and see if its dependencies are installed.
+# It requires the environment variable MINIO_CERTS_PATH to be set to the path
+# where minio stores its certificates (--certs-dir). These must be supported by
+# go. The documentation has recommendations under #supported-tls-cipher-suites,
+# see: https://min.io/docs/minio/linux/operations/network-encryption.html
+#
+#
+# MINIO_CERTS_PATH defaults to:
+# * FreeBSD: /usr/local/etc/minio/certs/
+# * Linux: ${HOME}/.minio/certs
+#
+## public functions ####################
+
+minio_test() {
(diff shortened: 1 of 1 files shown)


mw-cyrano:
Hey, thanks for your script, which just came in handy for me. How can I set the environment variable MINIO_CERTS_PATH in the my-domain.conf? And can you add an option to set the correct user with chmod in the certificates?

neilpang:
Direction is fine, but several blockers:

1. **`minio_test` runs at source time and `kill $$` kills acme.sh.** The bare `minio_test` call at the end executes every time acme.sh sources this hook; with `MINIO_CERTS_PATH` unset it runs `kill $$`, killing the acme.sh process itself. Remove the top-level call, and never `kill $$` -- guard with `_err "..." ; return 1`.

2. **Config isn't persisted, so cron renewals fail** (exactly what @mw-cyrano hit). Read the path with `_getdeployconf MINIO_CERTS_PATH` and save it with `_savedeployconf MINIO_CERTS_PATH "$MINIO_CERTS_PATH"`, so it's stored in the domain conf and available on unattended renewal.

3. **Deploy the fullchain, not the leaf.** `public.crt` should come from `$5` (`_cfullchain`), not `$3` -- clients need the intermediate chain or TLS fails.

4. **`openssl ec` breaks RSA certs.** It errors on an RSA key, and the key is already PEM -- just copy `$2` to `private.key`. If you keep any openssl call, use `${ACME_OPENSSL_BIN:-openssl}` and check with `_exists openssl`, not `which`.

5. Use `_err` / `_info`, not bare `echo`; and the file is tab-indented -- acme.sh is 2-space (shfmt will fail).

Issue #4503, opened 2023-02-12
 The JWS Signature MUST be present


UncaAlby:
Full Error Message:
Register account Error: {"type":"urn:ietf:params:acme:error:malformed","status":400,"detail":"[External Account Binding] The JWS Signature MUST be present"}

There are NO instructions ANYWHERE how to create or provide such a signature.

Steps to reproduce
------------------
acme --issue -d domain.com -d www.domain.com --webroot /var/www/html/domain

Debug log
-----------------
[Sat Feb 11 21:53:15 PST 2023] Running cmd: issue
[Sat Feb 11 21:53:15 PST 2023] _main_domain='uncaalby.com'
[Sat Feb 11 21:53:15 PST 2023] _alt_domains='www.uncaalby.com'
[Sat Feb 11 21:53:15 PST 2023] Using config home:/home/alvin/.acme.sh
[Sat Feb 11 21:53:15 PST 2023] default_acme_server
[Sat Feb 11 21:53:15 PST 2023] ACME_DIRECTORY='https://acme.zerossl.com/v2/DV90'
[Sat Feb 11 21:53:15 PST 2023] DOMAIN_PATH='/home/alvin/.acme.sh/uncaalby.com_ecc'
[Sat Feb 11 21:53:15 PST 2023] Le_NextRenewTime
[Sat Feb 11 21:53:15 PST 2023] Using ACME_DIRECTORY: https://acme.zerossl.com/v2/DV90
[Sat Feb 11 21:53:15 PST 2023] _init api for server: https://acme.zerossl.com/v2/DV90
[Sat Feb 11 21:53:15 PST 2023] GET
[Sat Feb 11 21:53:15 PST 2023] url='https://acme.zerossl.com/v2/DV90'
[Sat Feb 11 21:53:15 PST 2023] timeout=
[Sat Feb 11 21:53:15 PST 2023] _CURL='curl --silent --dump-header /home/alvin/.acme.sh/http.header  -L  -g '
[Sat Feb 11 21:53:15 PST 2023] ret='0'
[Sat Feb 11 21:53:15 PST 2023] ACME_KEY_CHANGE='https://acme.zerossl.com/v2/DV90/keyChange'
[Sat Feb 11 21:53:15 PST 2023] ACME_NEW_AUTHZ
[Sat Feb 11 21:53:15 PST 2023] ACME_NEW_ORDER='https://acme.zerossl.com/v2/DV90/newOrder'
[...]

PR #4527, opened 2023-02-28
Add info msg ZeroSSL missing JWS signature #4503

parawanderer:
Adds some extra error output contextually for default value users as well as a hint of how to more easily switch from the default. I took some liberty in assuming the main team maintaining acme recommends using ZeroSSL for the contextual message.

The idea would be to give lazy people (like me) -- who may have expected a hands-off `Let's Encrypt` style setup and who have no real reason to do anything more complex -- an easy way to accomplish that (think of the 'helpful' git command suggestions you may get during a merge).

Please see discussion in issue report #4503 for the context of this message. 

I apologise in advance for any weirdness in my bash script skills. I do not often implement algorithms in bash scripts 🙈

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -180,6 +180,8 @@ _VALIDITY_WIKI="https://github.com/acmesh-official/acme.sh/wiki/Validity"
 
 _DNSCHECK_WIKI="https://github.com/acmesh-official/acme.sh/wiki/dnscheck"
 
+_ZEROSSL_DEFAULT_CA_WIKI="https://github.com/acmesh-official/acme.sh/wiki/Change-default-CA-to-ZeroSSL"
+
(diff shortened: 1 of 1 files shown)


neilpang:
1. `${_res:0:-2}` is a bashism -- "Bad substitution" under dash/POSIX sh, and acme.sh must run there. Use `${_res%, }`.
2. `_getCAShortNameList` is indented with 4 spaces -- shfmt requires 2.
3. The `_ACME_USE_NONDEFAULT_SERVER` flag is only set by the `--server` option, so renewals that load a saved non-default server from the domain conf would still print the hint on a ZeroSSL-style 400 -- gate on the resolved `ACME_DIRECTORY` instead.
4. This targets a 2023 ZeroSSL quirk (#4503) -- please confirm the "JWS Signature MUST be present" 400 still reproduces on the current ZeroSSL API before this gets a rebase; if it no longer occurs the PR is obsolete.

PR #4528, opened 2023-03-01
SMTP notify: Use grep -E instead of expr

chris03:
expr was printing  `expr: warning: '^.*[<>"]': using '^' as the first character of a basic regular expression is not portable;`

To reproduce the issue:
```
docker run --rm -e SMTP_FROM=alain@example.com alpine:latest sh -c 'apk add curl openssl && curl https://get.acme.sh/ | sh -s email=bob@example.com && /root/.acme.sh/acme.sh --set-notify --notify-hook smtp'
```

### DIFF
--- a/notify/smtp.sh
+++ b/notify/smtp.sh
@@ -169,7 +169,7 @@ _clean_email_header() {
 # email
 _email_has_display_name() {
   _email="$1"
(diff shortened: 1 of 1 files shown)


Neilpang:
fix the shellcheck errors

PR #4577, opened 2023-04-01
prevent whitespace splitting

daschr:
Hi!

After an upgrade to v3.0.6 I get the following errors while deploying with the ssh hook:

`/root/.acme.sh/acme.sh: Zeile 2411: -T: Kommando nicht gefunden.`
`/root/.acme.sh/acme.sh: Zeile 2411: -q: Kommando nicht gefunden.`

Lokking at the saved deploy parameters

` SAVED_DEPLOY_SSH_CMD='ssh -T'`
` SAVED_DEPLOY_SSH_SCP_CMD='scp -q'`

I see that the `eval` at https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L2411 (or https://github.com/acmesh-official/acme.sh/blob/dev/acme.sh#L2415) is splitting at whitespaces.

To prevent that, one should use `eval $_rac_key=\$_saved`

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2412,7 +2412,7 @@ _getdeployconf() {
     return 0 # do nothing
   fi
   _saved="$(_readdomainconf "SAVED_$_rac_key")"
-  eval $_rac_key="$_saved"
+  eval $_rac_key=\$_saved
   export $_rac_key
 }
 


daschr:
Unfortunally, your code is still (again) vulnerable to escaping. F.e.:

```
$ val="ls \" echo pwned \""
$ # your solution
$ eval cmd=\"$val\"
pwned
$ # solution using \$
$ eval cmd=\$val
$ echo $cmd
ls " echo pwned "
```

This is the reason why I escaped the `$`. Since the shell won't evaluate `_saved` before the execution of eval.

Neilpang:
sorry, please make a new pr, I will merge it.

PR #4581, opened 2023-04-04
haproxy deploy hook updates existing certificate over stats socket

wlallemand:
Since version 2.2, HAProxy is able to update dynamically certificates, without a reload.

This patch uses socat to push the certificate into HAProxy in order to achieve hot update. With this method, reloading is not required. This should be used only to update an existing certificate in haproxy.

2 new variables are available:

- DEPLOY_HAPROXY_HOT_UPDATE="yes" update over the stats socket instead of reloading

- DEPLOY_HAPROXY_STATS_SOCKET="UNIX:/run/haproxy/admin.sock" set the path on the stats socket.

### DIFF
--- a/deploy/haproxy.sh
+++ b/deploy/haproxy.sh
@@ -102,6 +102,11 @@ haproxy_deploy() {
     _savedomainconf Le_Deploy_haproxy_pem_name "${Le_Deploy_haproxy_pem_name}"
   elif [ -z "${Le_Deploy_haproxy_pem_name}" ]; then
     Le_Deploy_haproxy_pem_name="${DEPLOY_HAPROXY_PEM_NAME_DEFAULT}"
+    # We better not have '*' as the first character
+    if [ "${Le_Deploy_haproxy_pem_name%%"${Le_Deploy_haproxy_pem_name#?}"}" = '*' ]; then
+      # removes the first characters and add a _ instead
+      Le_Deploy_haproxy_pem_name="_${Le_Deploy_haproxy_pem_name#?}"
+    fi
   fi
 
   # BUNDLE is optional. If not provided then assume "${DEPLOY_HAPROXY_BUNDLE_DEFAULT}"
--- a/deploy/haproxy.sh
+++ b/deploy/haproxy.sh
@@ -44,6 +44,11 @@
 #
 # Require the socat binary. DEPLOY_HAPROXY_STATS_SOCKET variable uses the socat
 # address format.
+#
+# export DEPLOY_HAPROXY_MASTER_CLI="UNIX:/run/haproxy-master.sock"
+#
+# OPTIONAL: To use the master CLI with DEPLOY_HAPROXY_HOT_UPDATE="yes" instead
+# of a stats socket, use this variable.
 
 ########  Public functions #####################
 
@@ -54,6 +59,7 @@ haproxy_deploy() {
   _ccert="$3"
   _cca="$4"
   _cfullchain="$5"
+  _cmdpfx=""
 
   # Some defaults
   DEPLOY_HAPROXY_PEM_PATH_DEFAULT="/etc/haproxy"
@@ -146,6 +152,16 @@ haproxy_deploy() {
     _savedomainconf Le_Deploy_haproxy_stats_socket "${Le_Deploy_haproxy_stats_socket}"
   elif [ -z "${Le_Deploy_haproxy_stats_socket}" ]; then
     Le_Deploy_haproxy_stats_socket="${DEPLOY_HAPROXY_STATS_SOCKET_DEFAULT}"
+  fi
+
+  # MASTER_CLI is optional. No defaults are used. When the master CLI is used,
+  # all commands are sent with a prefix.
+  _getdeployconf DEPLOY_HAPROXY_MASTER_CLI
+  _debug2 DEPLOY_HAPROXY_MASTER_CLI "${DEPLOY_HAPROXY_MASTER_CLI}"
+  if [ -n "${DEPLOY_HAPROXY_MASTER_CLI}" ]; then
+    Le_Deploy_haproxy_stats_socket="${DEPLOY_HAPROXY_MASTER_CLI}"
+    _savedomainconf Le_Deploy_haproxy_stats_socket "${Le_Deploy_haproxy_stats_socket}"
+    _cmdpfx="@1 " # command prefix used for master CLI only.
   fi
 
   # Set the suffix depending if we are creating a bundle or not
@@ -297,18 +313,25 @@ haproxy_deploy() {
   fi
 
   if [ "${Le_Deploy_haproxy_hot_update}" = "yes" ]; then
-    # Update certificate over HAProxy stats socket.
+    # set the socket name for messages
+    if [ -n "${_cmdpfx}" ]; then
+      _socketname="master CLI"
+    else
+      _socketname="stats socket"
+    fi
+
+    # Update certificate over HAProxy stats socket or master CLI.
     if _exists socat; then
       # look for the certificate on the stats socket, to chose between updating or creating one
-      _socat_cert_cmd="echo 'show ssl cert' | socat '${_statssock}' - | grep -q '^${_pem}$'"
+      _socat_cert_cmd="echo '${_cmdpfx}show ssl cert' | socat '${_statssock}' - | grep -q '^${_pem}$'"
       _debug _socat_cert_cmd "${_socat_cert_cmd}"
       eval "${_socat_cert_cmd}"
       _ret=$?
       if [ "${_ret}" != "0" ]; then
         _newcert="1"
-        _info "Creating new certificate '${_pem}' over HAProxy stats socket."
+        _info "Creating new certificate '${_pem}' over HAProxy ${_socketname}."
         # certificate wasn't found, it's a new one. We should check if the crt-list exists and creates/inserts the certificate.
-        _socat_crtlist_show_cmd="echo 'show ssl crt-list' | socat '${_statssock}' - | grep -q '^${Le_Deploy_haproxy_pem_path}$'"
+        _socat_crtlist_show_cmd="echo '${_cmdpfx}show ssl crt-list' | socat '${_statssock}' - | grep -q '^${Le_Deploy_haproxy_pem_path}$'"
         _debug _socat_crtlist_show_cmd "${_socat_crtlist_show_cmd}"
         eval "${_socat_crtlist_show_cmd}"
         _ret=$?
@@ -317,7 +340,7 @@ haproxy_deploy() {
           return "${_ret}"
         fi
         # create a new certificate
-        _socat_new_cmd="echo 'new ssl cert ${_pem}' | socat '${_statssock}' - | grep -q 'New empty'"
+        _socat_new_cmd="echo '${_cmdpfx}new ssl cert ${_pem}' | socat '${_statssock}' - | grep -q 'New empty'"
         _debug _socat_new_cmd "${_socat_new_cmd}"
         eval "${_socat_new_cmd}"
         _ret=$?
@@ -326,9 +349,9 @@ haproxy_deploy() {
           return "${_ret}"
         fi
       else
-        _info "Update existing certificate '${_pem}' over HAProxy stats socket."
+        _info "Update existing certificate '${_pem}' over HAProxy ${_socketname}."
       fi
-      _socat_cert_set_cmd="echo -e 'set ssl cert ${_pem} <<\n$(cat "${_pem}")\n' | socat '${_statssock}' - | grep -q 'Transaction created'"
+      _socat_cert_set_cmd="echo -e '${_cmdpfx}set ssl cert ${_pem} <<\n$(cat "${_pem}")\n' | socat '${_statssock}' - | grep -q 'Transaction created'"
       _debug _socat_cert_set_cmd "${_socat_cert_set_cmd}"
       eval "${_socat_cert_set_cmd}"
       _ret=$?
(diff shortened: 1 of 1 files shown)


wlallemand:
I just updated the PR, it now has the ability to add a new certificate in HAProxy without reloading as well as updating a previous one.

msimerson:
Hey @wlallemand, this is AWESOME, thank you. I was getting ready to write this functionality and decided I should check if it already exists. I ran into a few issues I had to address to get it working for me.

1. my fullchain.pem files have an extra `\n` in them. Apparently the chain cert is coming back that way from Let's Encrypt. Because `<<\n$(cat ...)\n` depends on `\n` as a boundary, the payload fails to send to haproxy correctly. The solution is simple: replace cat with `grep . $path`
2. my haproxy is chroot, so the acme.sh deploy path and the haproxy path to same file are different
    - I store the haproxy installed path in _ipem.
3. I communicate with haproxy admin API via TCP port instead of sock. This is easy to accommodate by moving the UNIX: into the socket variable, so it can be modified to `tcp-connect:192.168.0.1:9000`. Probably best to include an example of each as comments.

wlallemand:
@msimerson Thanks for your feedback, I'm going to fix the script to integrate properly your use case.

> 
>     1. my fullchain.pem files have an extra `\n` in them. Apparently the chain cert is coming back that way from Let's Encrypt. Because `<<\n$(cat $path)\n` (as shown in [haproxy docs](https://www.haproxy.com/blog/dynamic-ssl-certificate-storage-in-haproxy/)) depends on `\n` as a boundary, the payload fails to send to haproxy correctly. The solution is simple: replace cat with `grep . $path`
> 

Indeed. I did not had any problem with the provider I used but that's a valid point, it could contains empty lines or comments between the certificate section.

>     2. I communicate with haproxy admin API via TCP port instead of sock. This is easy to accommodate by moving the UNIX: into `_statssock`, so it can be set to `tcp-connect:192.168.0.1:9000`. Probably best to include an example of each as comments.
> 

I hesitate to include directly the socat format into the variable, maybe I could add a variable to be able to chose any socat prefix. 

>     3. my haproxy is chroot, so the acme.sh deploy path and the haproxy path to same file are different
>        
>        * I store the haproxy installed path in _ipem.

Indeed that could be a problem, I will retest all of this.

Also, I published a documentation on the haproxy wiki: https://github.com/haproxy/wiki/wiki/Letsencrypt-integration-with-HAProxy-and-acme.sh

msimerson:
> I hesitate to include directly the socat format into the variable, maybe I could add a variable to be able to chose any socat prefix.

I'm curious how you think that having 2 socat variables is going to be easier for folks to grok and easier to support than having just one. It's not like the variable contents will ever be used separately. URIs evolved for a reason. :-)

msimerson:
> @msimerson problem is socat and haproxy doesn't have the same "URIs". I don't want the user to need to specify a socat format. Maybe I'll just check if the variable contains any / to select UNIX-CONNECT or TCP-CONNECT.

I see. In looking again, you use `_statssock` (3) times without the `UNIX:` prefix and (3) times with. If you were consistent and used it only without the prefix, then I think you get exactly what you want (user doesn't need to specify), and it'll work as you expect, because socat will assume the right thing:

> Address specifications starting with a number are assumed to be FD (raw file descriptor) addresses; if a ’/’ is found before the first ’:’ or ’,’, GOPEN (generic file open) is assumed.

And anyone needing a different socat formula can specify it.

wlallemand:
> I see. In looking again, you use `_statssock` (3) times without the `UNIX:` prefix and (3) times with. If you were consistent and used it only without the prefix, then I think you get exactly what you want (user doesn't need to specify), and it'll work as you expect, because socat will assume the right thing:

Indeed, my mistake, I should also fix this. My first implementation was a little bit different and I probably forgot it.

> > Address specifications starting with a number are assumed to be FD (raw file descriptor) addresses; if a ’/’ is found before the first ’:’ or ’,’, GOPEN (generic file open) is assumed.

I want to avoid this and find it dangerous, if for some reason the path is wrong and is not a UNIX socket, it would do a GOPEN and the certificate could be dumped on the filesystem. That's exactly why I set UNIX: in the first place.

> And anyone needing a different socat formula can specify it.

That's why I suggested another variable, but to be honest there is not a lot of common setup. It's basically TCP with ipv4 and ipv6 and unix socket, some people could also want to use OPENSSL... I think I'll remove the prefix and adjust the documentation as you suggested in the first place, with a warning about not using a prefix.

msimerson:
>  it would do a GOPEN and the certificate could be dumped on the filesystem

In this case, the certificate is already on the filesystem, right? If you're super worried about it, you can always do a `-S` test on the path (after determining it's a path, by checking that the first character is a `/`).

iarenaza:
>     3. my haproxy is chroot, so the acme.sh deploy path and the haproxy path to same file are different
>        
>        * I store the haproxy installed path in _ipem.

Just wanted to chime in to say that we are not using chroot for haproxy, but running it inside a container. And pretty much the same situation arises there: paths outside the container (where acme.sh stores the certs) and inside the container (where haproxy looks for them) don't match.

Is this something that could be supported?

deploy/haproxy.sh
```diff
@@ -265,15 +295,48 @@ haproxy_deploy() {
...
-  eval "${_reload}"
-  _ret=$?
-  if [ "${_ret}" != "0" ]; then
-    _err "Error code ${_ret} during reload"
-    return ${_ret}
+  if [ "${Le_Deploy_haproxy_hot_update}" = "yes" ]; then
+    # Update certificate over HAProxy stats socket.
+    _info "Update the certificate over HAProxy stats socket."
+    if _exists socat; then
+      _socat_cert_cmd="echo 'show ssl cert' | socat ${_statssock} - | grep -q '^${_pem}$'"
```
dingensundso, line 302: This does not work for wildcard certs since _pem will be something like ```'/etc/haproxy/certs/*.example.com.pem'```.
The * must be escaped.

The first solution that came to mind was changing the grep part to: ```grep -q '^$(printf '%q' "${_pem}")$'```

wlallemand:
Thanks for the feedbacks, I'll rework the PR next week to better handle these cases.

dexter-dopping-ekco:
Hi folks,

To get the script to work right on my setup (Debian 12, HaProxy 2.6.12-1) I had to make a small tweak.

It seems like the stats socket doesn't read the entire input when passing in the file contents, instead stopping after reading an empty line, which means it doesn't read the CA certificate chain. This is what it looks like when running the deploy command with `--debug 2`:

```
...
-----END CERTIFICATE-----\n' | socat 'UNIX:/var/run/haproxy/admin.sock' - | grep -q 'Transaction created''
2023/11/24 20:11:20 socat[118250] E write(1, 0x5588acdfe000, 8192): Broken pipe
[Fri 24 Nov 2023 20:11:20 GMT] _socat_cert_commit_cmd='echo 'commit ssl cert /etc/haproxy/certs/example.com.pem' | socat 'UNIX:/var/run/haproxy/admin.sock' - | grep -q '^Success!$''
[Fri 24 Nov 2023 20:11:20 GMT] Success
```

The deploy script logs it as successful, but socat wasn't able to write everything (Broken pipe). openssl s_client shows that the certificate chain is incomplete:

```
dexter@dexter-dell:~$ openssl s_client -connect example.com:443 -servername example.com </dev/null
CONNECTED(00000003)
depth=0 CN = example.com
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 CN = example.com
verify error:num=21:unable to verify the first certificate
verify return:1
depth=0 CN = example.com
verify return:1
---
Certificate chain
 0 s:CN = example.com
   i:C = US, O = Let's Encrypt, CN = R3
--- ...
```

https://github.com/acmesh-official/acme.sh/issues/4788 mentions the same "Broken pipe" message but the related MR https://github.com/acmesh-official/acme.sh/pull/4841 did not help and even seems to have made things worse, since the key no longers get imported when it's moved to the end of the PEM file.

Only after removing the empty lines does command with hot update work correctly for me.

This is the change I made:

```diff
diff --git a/haproxy-wlallemand.sh b/usr/local/share/acme.sh/deploy/haproxy.sh
index a611ec8..bb2b6b1 100644
--- a/haproxy-wlallemand.sh
+++ b/usr/local/share/acme.sh/deploy/haproxy.sh
@@ -177,7 +177,7 @@ haproxy_deploy() {
   # Create a temporary PEM file
   _temppem="$(_mktemp)"
   _debug _temppem "${_temppem}"
-  cat "${_ckey}" "${_ccert}" "${_cca}" >"${_temppem}"
+  cat "${_ckey}" "${_ccert}" "${_cca}" | grep . >"${_temppem}"
   _ret="$?"

   # Check that we could create the temporary file
```

Now the deploy command runs fine and the openssl s_client command shows the complete chain:

```
dexter@dexter-dell:~$ openssl s_client -connect example.com:443 -servername example.com </dev/null
CONNECTED(00000003)
depth=2 C = US, O = Internet Security Research Group, CN = ISRG Root X1
verify return:1
depth=1 C = US, O = Let's Encrypt, CN = R3
verify return:1
depth=0 CN = example.com
verify return:1
---
Certificate chain
 0 s:CN = example.com
   i:C = US, O = Let's Encrypt, CN = R3
 1 s:C = US, O = Let's Encrypt, CN = R3
   i:C = US, O = Internet Security Research Group, CN = ISRG Root X1
 2 s:C = US, O = Internet Security Research Group, CN = ISRG Root X1
   i:O = Digital Signature Trust Co., CN = DST Root CA X3
--- ...
```

wlallemand:
Small updates:
- sanitize the PEM to remove the '\n'
- shellcheck fixes
- socat format is directly used in the DEPLOY_HAPROXY_STATS_SOCKET variable

chroot and wilcard needs to be handle.

@Neilpang Any chance this can be merged? This is actually asked by a lot of people.

wlallemand:
I'll make more updates for chroot and wilcard as well as the master CLI instead of the stats socket, but probably in a separate PR.

deploy/haproxy.sh
wlallemand: It seems rather dangerous to me, having a filename with an asterisk. Also the '*' in front of a filename in the HAProxy CLI allows to select a file which is a CLI transaction but not yet committed. So there is already a problem without this PR.  I'd rather replace the '*' in the filename by something else, for example it could be something like '__wildcard.foo.bar' or even '@wildcard.foo.bar'

wlallemand: I fixed the issue in  e09d45c844, it will replace the * in the name by a _, so that won't be confusing. Unfortunately '%q' is not POSIX so that can't be used.

wlallemand:
update:
- master CLI support was implemented
- wilcard character is replaced in the filename

deploy/haproxy.sh
dingensundso: I don't think that using asteriks in filenames is dangerous by itself. acme.sh itself saves the certificates and configs with the asterisk in the filename.
But if it's a problem with the HAProxy CLI (haven't looked at the docs whether there is a way of escaping it), replacing it seems like a valid choice.
I'm not sure whether using _ is a good idea. AFAIK _ is not a valid subdomain and Let's Encrypt won't issue certificates for it, but some DNS providers are allowing it as a subdomain (just tested it with one of my domains).

dingensundso (approved):
As mentioned in https://github.com/acmesh-official/acme.sh/pull/4581#discussion_r1425595319 I'm not sure about replacing the asterisk with an underscore, since some DNS servers allow underscore as a subdomain even though RFC1035 doesn't allow it.
Let's Encrypt for one doesn't issue certificates for underscore domains but I don't know about other acme server implementations.

Would love to see input from other people about this.

Otherwise I don't see any problems with the script.

deploy/haproxy.sh
wlallemand: > I don't think that using asteriks in filenames is dangerous by itself. acme.sh itself saves the certificates and configs with the asterisk in the filename.
Not by itself of course, but it could introduce bug more easily with it not escaped correctly.

> But if it's a problem with the HAProxy CLI (haven't looked at the docs whether there is a way of escaping it), replacing it seems like a valid choice.
I'm in fact the maintainer of the SSL subsystem in HAProxy, when this part was written we chose the asterisk because it's uncommon to see this in a file, but I'll work on a way to escape this, but that won't be usable with old versions for now.

> I'm not sure whether using _ is a good idea. AFAIK _ is not a valid subdomain and Let's Encrypt won't issue certificates for it, but some DNS providers are allowing it as a subdomain (just tested it with one of my domains).
I can replace the `_` by `@` or `+` for example, what do you think?

krezovic:
Hi @wlallemand. I just stumbled upon this trying to configure OCSP Stapling in haproxy. Is it possible to use Le_OCSP_Staple info already present in haproxy.sh and add commands from https://www.haproxy.com/documentation/haproxy-runtime-api/reference/set-ssl-ocsp-response/ to the update commands if ocsp is present? Thanks in advance!

wlallemand:
Hello @krezovic, 
1. The .ocsp could be used directly with "set ssl cert basecert.pem.ocsp" before committing. I'll look into this.
2. Regarding the ocsp-update it's not possible with this current design, because the option is only working with a "crt-list" as a file, and the current config uses a directory. However we will make some changes into HAProxy so this could be done as a global option.

krezovic:
Hi @wlallemand! Looking forward to proper support. Thanks for your reply!

wlallemand:
@dingensundso Any chance who could have this merged? This is stuck for a while now.

Neilpang:
please also update the usage here:
https://github.com/acmesh-official/acme.sh/wiki/deployhooks#10-deploy-the-cert-to-haproxy

wlallemand:
Thanks, updated!

ser:
Very nice job, it would be cool to have it available - maybe it's a good time for a new release?

PR #4629, opened 2023-05-09
dnsapi: fix OPNsense script to be compatible with upcoming 23.1.8

fichtner:
The current script is already broken due to Bind 9.16 -> 9.18 changes renaming
master/slave to primary/secondary and do not rely on the compat layer (which
was also broken for other reasons).

The full discussion over at: https://github.com/opnsense/plugins/issues/3420

### DIFF
--- a/dnsapi/dns_opnsense.sh
+++ b/dnsapi/dns_opnsense.sh
@@ -137,7 +137,7 @@ _get_root() {
   domain=$1
   i=2
   p=1
-  if _opns_rest "GET" "/domain/searchMasterDomain"; then
+  if _opns_rest "GET" "/domain/searchPrimaryDomain"; then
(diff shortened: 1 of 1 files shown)


fichtner:
Test was carried out on OPNsense with a modified FreeBSD port version of acme.sh. The script wasn't touched, but rather the API endpoint and return data were adjusted according to the latest code.


Cheers,
Franco

zv0r:
On my instance I need to use
`\"type\":\"Primary\"`
instead of
`\"type\":\"primary\"`
on line 153

fichtner:
It's for 23.1.8 which will be released next week.

fichtner:
For reference see https://github.com/opnsense/plugins/commit/8f8b2f63db

PR #4643, opened 2023-05-27
Add validity parameters for signing CSR

vemireddy-vaibhav:
Signing CSR option currently doesn't have support for validity parameters. This PR aims to add this since _**signcsr**_ function internally calls _**issue**_ function

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -5604,6 +5604,8 @@ signcsr() {
   _local_addr="${11}"
   _challenge_alias="${12}"
   _preferred_chain="${13}"
+  _valid_from="${14}"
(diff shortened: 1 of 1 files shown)


vemireddy-vaibhav:
The above options was used to sign a CSR following validity parameters format as described in Validity wiki and was able to generate certificate with requested valid_to parameter

vemireddy-vaibhav:
@Neilpang Appreciate if you can take a look at this PR

vemireddy-vaibhav:
@Neilpang Do you see any issues/concern with the changes present in the PR. If not, I would appreciate if you can approve the workflows for this PR

PR #4653, opened 2023-06-05
dns_ovh.sh Add ovh-us endpoint

stokito:
The OVH has a new endpoint ovh-us:
https://github.com/ovh/php-ovh/blob/096a7b74132cb6742fc66667df83782bdde45c4a/src/Api.php#L61

After merging the wiki page needs to be updated https://github.com/acmesh-official/acme.sh/wiki/How-to-use-OVH-domain-api

### DIFF
--- a/dnsapi/dns_ovh.sh
+++ b/dnsapi/dns_ovh.sh
@@ -14,6 +14,9 @@
 #'ovh-eu'
 OVH_EU='https://eu.api.ovh.com/1.0'
 
+#'ovh-us'
+OVH_US='https://api.us.ovhcloud.com/1.0'
+
 #'ovh-ca':
 OVH_CA='https://ca.api.ovh.com/1.0'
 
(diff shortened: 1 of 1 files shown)


dnsapi/dns_ovh.sh
```diff
@@ -11,27 +11,14 @@
 
 #OVH_END_POINT=ovh-eu
 
-#'ovh-eu'
 OVH_EU='https://eu.api.ovh.com/1.0'
```
Neilpang, line 14: don't remove the comments here.  just add the new endpoint.

Neilpang:
fix the CI errors, and then update the wiki page first.

dnsapi/dns_ovh.sh
stokito: the envs are listed in _ovh_get_api so this is a duplication and noise

stokito:
The Wiki page updated

PR #4655, opened 2023-06-05
Deploy nutanix

Olen:
Add support for deploying certificates to Nutanix Prism.

Based on the postings here: https://next.nutanix.com/ncm-intelligent-operations-formerly-prism-pro-ultimate-26/using-lets-encrypt-certs-on-prism-central-prism-element-27805

### DIFF
--- a/deploy/nutanix.sh
+++ b/deploy/nutanix.sh
@@ -50,7 +50,7 @@ nutanix_deploy() {
     _nutanix_pass="$NUTANIX_PASS"
     _nutanix_host="$NUTANIX_HOST"
   fi
-  curl --silent --fail --user "$_nutanix_user:$_nutanix_pass" -F caChain=@$_cca -F cert=@$_ccert -F key=@$_ckey -F keyType=RSA_2048 -k https://$_nutanix_host:9440/PrismGateway/services/rest/v1/keys/pem/import >/dev/null
+  curl --silent --fail --user "$_nutanix_user:$_nutanix_pass" -F caChain=@"$_cca" -F cert=@"$_ccert" -F key=@"$_ckey" -F keyType=RSA_2048 -k https://"$_nutanix_host":9440/PrismGateway/services/rest/v1/keys/pem/import >/dev/null
   return $?
 }
 
--- a/deploy/nutanix.sh
+++ b/deploy/nutanix.sh
@@ -8,7 +8,6 @@
 # export NUTANIX_PASS=""  # required
 # export NUTANIX_HOST=""  # required
 
-
 #domain keyfile certfile cafile fullchain
 nutanix_deploy() {
   _cdomain="$1"
@@ -53,5 +52,3 @@ nutanix_deploy() {
   curl --silent --fail --user "$_nutanix_user:$_nutanix_pass" -F caChain=@"$_cca" -F cert=@"$_ccert" -F key=@"$_ckey" -F keyType=RSA_2048 -k https://"$_nutanix_host":9440/PrismGateway/services/rest/v1/keys/pem/import >/dev/null
   return $?
 }
-
-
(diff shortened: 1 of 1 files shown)


neilpang:
1. Raw `curl` is not allowed -- all requests must go through `_post`/`_get` (proxy, CA bundle, wget-only systems). Build the multipart body manually and send it with `_post` and a boundary content-type; auth via `export _H1="Authorization: Basic $(printf "%s" "$NUTANIX_USER:$NUTANIX_PASS" | _base64)"`, which also keeps the password out of the process list (`--user` does not).
2. Hardcoded `-k` disables TLS verification unconditionally -- make it opt-in: `NUTANIX_INSECURE=1` -> `export HTTPS_INSECURE=1` (see `deploy/openwrt.sh`).
3. `keyType=RSA_2048` is hardcoded -- breaks ECC certs and other RSA sizes; derive it from the key or document the limitation.
4. Config flow: call `_getdeployconf` unconditionally first (it already prefers env vars over saved values), validate, then `_savedeployconf` only after a successful deploy. The current branching saves before the deploy is known to work, and the "delete the env vars afterwards" advice conflicts with how cron renewals reload the saved conf.
5. No `_err` on upload failure -- check the response and report.

Also needs a usage entry on the deployhooks wiki page before merge.

Olen:
Thanks for the detailed review. I have reworked the script to address every point (rebased onto the latest `dev`, commit `c7fd5f9`):

1. **No raw `curl`.** The upload now goes through `_post` with a manually built `multipart/form-data` body and boundary (same approach as `deploy/synology_dsm.sh`). Auth is via `export _H1="Authorization: Basic $(printf "%s" "$NUTANIX_USER:$NUTANIX_PASS" | _base64)"`, which also keeps the password out of the process list.
2. **TLS verification is opt-in.** `NUTANIX_INSECURE=1` now sets `export HTTPS_INSECURE=1` (pattern from `deploy/fritzbox.sh`); by default the cert is verified.
3. **`keyType` is derived from the key**, not hardcoded. `_nutanix_key_type()` uses `_isRSA`/`_isEcc` and reports `RSA_<bits>`, `ECDSA_256` or `ECDSA_384`, so ECC and non-2048 RSA certs work (or fail with a clear message).
4. **Config flow fixed.** `_getdeployconf` is now called unconditionally first (env still overrides saved), the values are validated, and `_savedeployconf` runs **only after a successful upload** — so a failed deploy never persists config and cron renewals reload cleanly.
5. **Errors are reported.** The `_post` transport result is checked and the response body is scanned for error indicators; any failure calls `_err` and returns non-zero.

Passes `shellcheck -e SC2181 -e SC2089` and `shfmt -i 2`.

### Usage (for the deployhooks wiki page)

```sh
export NUTANIX_USER="admin"             # required
export NUTANIX_PASS="password"          # required
export NUTANIX_HOST="prism.example.com" # required (hostname or IP, port 9440)
export NUTANIX_INSECURE="1"             # optional, skip TLS verification of Prism
acme.sh --deploy -d example.com --deploy-hook nutanix
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

neilpang:
Rework looks good, two things before merge:

1. Deploy hooks have no CI -- please run the reworked script against a real Prism instance and paste a redacted `--deploy --debug` log. The `keyType` strings (`ECDSA_256`/`ECDSA_384`) and the success/error response shape can only be validated live; ideally test both an RSA and an ECC cert.
2. The failure check can report success on an HTTP error: `_post` returns 0 on 4xx/5xx, and e.g. a 401 body may not contain `error`/`message_list`/`reason`. Check the status line in `"$HTTP_HEADER"` first (e.g. `grep "^HTTP.*200"`), then fall back to the body scan.

Minor: port 9440 is hardcoded -- if `NUTANIX_HOST` ever includes a port it breaks; consider accepting `host:port` or an optional `NUTANIX_PORT`.

PR #4700, opened 2023-07-14
Fix SES region variable

szhu25:
The last version do not save the SES region into the config file, breaking the notification hook.


szhu25:
> **Welcome** Please make sure you're read our [Code-of-conduct](../wiki/Code-of-conduct) and add the usage here: [notify](../wiki/notify). Then reply on this message, otherwise, your code will not be reviewed or merged. We look forward to reviewing your Pull request shortly ✨

Ack

PR #4710, opened 2023-07-26
Fix the API calls that get the list of domains that PLESK can manage

zearan:
I was not getting all domains that are managed within my PLESK accounts when I was using the script. I had to merge the list of domains that is getting pulled via `"<packet><webspace>"`  with a list of domains pulled via `"<packet><site>"`. To ease debugging, I had to adopt the related debug messages and fixed the 'hacked output' in the meantime.


zearan:
I have added changes that should fix the failed workflow tests.

PR #4722, opened 2023-07-31
Added function to clear deployment keys stored in domain conf files

sg1888:
This PR adds the ability to clear keys created by deployment scripts.  It can be useful in situations where a deployment script needs to delete a previously saved key.  To clear saved domain keys now, the script must use `_cleardomainconf "SAVED_$target_key"`.  This change mirrors the feature parity with the `_savedeployconf` and `_migratedeployconf` functions.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2419,6 +2419,11 @@ _savedeployconf() {
   _cleardomainconf "$1"
 }
 
+#_cleardeployconf key
+_cleardeployconf() {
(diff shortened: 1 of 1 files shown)


neilpang:
Applied on dev in 45c0ad41, extended to also clear the legacy unprefixed key (matching `_clearaccountconf_mutable`), and `synology_dsm.sh`'s local copy is removed. Thanks.

PR #4726, opened 2023-08-08
Add renew-tolerance option

mib1185:
Let's assume you have around 50 certificates. you are using the `acme.sh --cron` task within a daily (_at 00:45_) running cronjob.
sometimes this job runs faster than other days. now the job starts at 00:45:03 and tries to renew certificate `host.mydomain` which was last renewed at 00:45:21. it will not be renewed, because it is not on time (_18s left_).

This PR will add a parameter to give an optional tolerance window for this case, so that in the above example the certificate would be renewed even it is still 18s till renewal-time, but without need to `--force` the renewal of all the other certificates which has renewal-time within the next days/weeks

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4385,7 +4385,7 @@ issue() {
   if [ -f "$DOMAIN_CONF" ]; then
     Le_NextRenewTime=$(_readdomainconf Le_NextRenewTime)
     _debug Le_NextRenewTime "$Le_NextRenewTime"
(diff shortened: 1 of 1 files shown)


mib1185:
Hi @Neilpang do you mind to have a look into this? thanks in advanced

Neilpang:
we dont' need this.
the cert should be renewed 1 day before the ending time.  eg, the default renew time is 60 days, but it will try to renew the cert from the 59th day.

Issue #4746, opened 2023-08-20
unrecognized option '--conf-path=[^ ]* ' for nginx config


Nirzak:
Steps to reproduce
------------------
Just try to install a certificate using acme.sh on ubuntu 22.
Debug log
-----------------
[Sun Aug 20 18:52:04 UTC 2023] Nginx mode for domain:zaksb.nirzak.codes
grep: unrecognized option '--conf-path=[^ ]* '
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
[Sun Aug 20 18:52:04 UTC 2023] Can not find nginx conf.
[Sun Aug 20 18:52:04 UTC 2023] Please add '--debug' or '--log' to check more details.
[Sun Aug 20 18:52:04 UTC 2023] See: https://github.com/acmesh-official/acme.sh/wiki/How-to-debug-acme.sh
[Sun Aug 20 18:52:05] ERROR     Certificate could not be issued.

Additional Info
-----------------
The following code may have the issue.
NGINX_CONF="$(nginx -V 2>&1 | _egrep_o "--conf-path=[^ ]* " | tr -d " ")"

No escape parameter is used in the egrep pattern. If I modify the pattern like as the following then the code works.
` _egrep_o "\-\-conf-path=[^ ]* "`

I can make a pull request if you want. Thanks.

```
acme.sh  --issue .....   --debug 2
```

tpedretti:
Same problem in #4725 I would say do a pull request for the fix and submit it, as my issue has been open for 3 weeks now and nothing. The problem started in 3.0.7, if you downgrade to 3.0.6 it works.

Nirzak:
> 

Thanks created a pull request for this. Hope they will merge and fix the issue.

victorxys:
> 

Excuse me may I ask how to downgrade to 3.0.6? I did't find any answer from google

Issue #4725, opened 2023-08-02
letsencrypt - grep: unrecognized option


tpedretti:
I was directed to report this issue upstream from the project that uses acme.sh. The program in question is swizzin, but the problem happens when letsencrypt is ran. The error that is thrown is grep: unrecognized option '--conf-path=[^ ]* ' logs can be found below. 

swizzin issue and post about pushing upstream to acme.sh can be found here: https://github.com/swizzin/swizzin/issues/1055

```plaintext
[Wed Aug 02 17:25:44] LOG       >>>> `box install letsencrypt`
[Wed Aug 02 17:25:44] LOG       git @ 9dc0c50 2>&1
[Wed Aug 02 17:25:44] INFO      Installing letsencrypt
[Wed Aug 02 17:25:44] DOCS      Further reference: https://swizzin.ltd/applications/letsencrypt
[Wed Aug 02 17:25:44] INPUT     Enter domain name to secure with LE
[Wed Aug 02 17:25:50] INPUT     Do you want to apply this certificate to your swizzin default conf?
[Wed Aug 02 17:25:53] INPUT     Is your DNS managed by CloudFlare?
66999b17-21b4-4da8-b61f-27173af290ca
[Wed Aug 02 17:25:54] LOG       Inserted apt logcheck marker
[Wed Aug 02 17:25:54] LOG       Variables unset
[Wed Aug 02 17:25:54] LOG       New variables exported
[Wed Aug  2 17:25:54 UTC 2023] Changed default CA to: https://acme-v02.api.letsencrypt.org/directory
[Wed Aug 02 17:25:54] ...       Registering certificates
[Wed Aug  2 17:25:54 UTC 2023] Using CA: https://acme-v02.api.letsencrypt.org/directory
[Wed Aug  2 17:25:55 UTC 2023] Single domain='fake.website.com'
[Wed Aug  2 17:25:55 UTC 2023] Getting domain auth token for each domain
[Wed Aug  2 17:25:56 UTC 2023] Getting webroot for domain='fake.website.com'
[Wed Aug  2 17:25:56 UTC 2023] Verifying: fake.website.com
[Wed Aug  2 17:25:56 UTC 2023] Nginx mode for domain:fake.website.com
grep: unrecognized option '--conf-path=[^ ]* '
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
[Wed Aug  2 17:25:56 UTC 2023] Can not find nginx conf.
[...]

PR #4749, opened 2023-08-21
Fixed grep pattern regex for nginx conf path

Nirzak:
This pull request will fix `grep :unrecognized option '--conf-path=[^ ]* '` error.


Nirzak:
This pull request will fix issues #4746 and #4725

PR #4766, opened 2023-08-28
AWS Similar Hosted Zone Names

studycom-mrobinson:
Similar to #1628 I experienced an issue with having multiple similar Hosted Zones within AWS, causing acme.sh to try to use the wrong Zone to renew a certificate, and specifically to fail setting the ACME challenge DNS entry. From looking through the code for just this provider, and testing locally, I was able to add just this bit to escape Periods for the domain name part of the string passed into _egrep_o, without affecting anything else.

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -157,7 +157,7 @@ _get_root() {
 
   # iterate over names (a.b.c.d -> b.c.d -> c.d -> d)
   while true; do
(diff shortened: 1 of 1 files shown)


Neilpang:
1. please send pr to the dev branch.
2.  you must finish the dns api test:  https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

PR #4773, opened 2023-09-01
Dns nearlyfreespeech

Travisivart:
Adds support for nearlyfreespeech.


neilpang:
1. Raw `curl` is not allowed -- use `_post`/`_get` (which also handle proxies/CA); and `-k` disables TLS verification entirely.
2. `exit 1` kills the whole acme.sh process -- use `return 1`; report errors with `_err`, not bare `echo`.
3. Credentials are never persisted -- load/save via `_readaccountconf_mutable`/`_saveaccountconf_mutable`, otherwise cron renewal has nothing to use.
4. Root zone is assumed to be the last two labels -- breaks `co.uk`-style domains and any multi-level zone; probe levels like other hooks' `_get_root`.
5. `sha1sum` is not available on all supported platforms -- use `_digest sha1 hex`.
6. Prefix the ALL-CAPS working variables (`FULLDOMAIN`, `PARAMETERS`, `BODY`, `HASH`, ...) with `_nfs_` -- hooks share one shell namespace.
7. Add the standard `dns_nfs_info='...'` block, a dedicated tracking issue (`Issues:`), the dnsapi2 wiki entry (`Docs:`), and a passing DNS-API-Test run in your fork.

PR #4809, opened 2023-09-25
fix: Synology DSM API path regex

winromulus:
This PR fixes the DSM API path regex which currently results in a 404 (since it's looking for a version - probably a bad copy paste from the line below).

Fixes #4808

PR #4810, opened 2023-09-26
Fix deploy script for Synology DSM (wrong RegEx for path extraction)

Eagle3386:
Fixes wrong RegEx as mentioned in corresponding [deploy API issue](https://github.com/acmesh-official/acme.sh/issues/2727#issuecomment-1733927132).

Really don't know, how that'd slipped through, because I wrote the correct RegEx just a couple of comments above the linked one… 🙈😅

Anyway, this fixes the issue once & for all. 😎

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -91,7 +91,7 @@ synology_dsm_deploy() {
 
   _debug "Getting API version"
   response=$(_get "$_base_url/webapi/query.cgi?api=SYNO.API.Info&version=1&method=query&query=SYNO.API.Auth")
-  api_path=$(echo "$response" | grep "SYNO.API.Auth" | sed -n 's/.*"path" *: *"\([0-9]*\)".*/\1/p')
+  api_path=$(echo "$response" | grep "SYNO.API.Auth" | sed -n 's/.*"path" *: *"\([a-z.]*\)".*/\1/p')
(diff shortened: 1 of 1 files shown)


Eagle3386:
@Neilpang I don't know why GitHub states "branch conflict" for simply changing the RegEx range from `0-9` to `a-z.`… 🤔

winromulus:
@Eagle3386  PR #4809 was already merged. Probably that's why you get conflicts.

winromulus:
@Eagle3386 considering Synology's naming conventions, I wouldn't assume that the entry will not contain any numeric or other characters in the path. Ideally this would have been picked up using jq instead of sed.

Eagle3386:
> @Eagle3386 PR #4809 was already merged. Probably that's why you get conflicts.

Probably not,because if it wasn't, my PR would still change the exact same RegEx range.
If it wasn't merged already that'd make sense, but it is, so…

> @Eagle3386 considering Synology's naming conventions, I wouldn't assume that the entry will not contain any numeric or other characters in the path. Ideally this would have been picked up using jq instead of sed.

I've never come across an API endpoint of them that uses numbers. Other chars might be, but again, never seen that up to now.
Using jq would've been an option, yes - especially, since the file header even mentions it as a requirement.
However, neither the original script nor my update actually used it before & so I opted to follow the current way. That's because AFAIK jq isn't available by default on all platforms & Neil's requirement is to write cross platform compatible code, using "standard tools".

PR #4832, opened 2023-10-19
deploy hook for Ruckus ZoneDirector / Unleashed

kchiem:
Tested on my Unleashed install, but according to the code it's adapted form, it should work for ZD installs too.

### DIFF
--- a/deploy/ruckus.sh
+++ b/deploy/ruckus.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+
+# Here is a script to deploy cert to Ruckus Zone Director/Unleashed.
+#
+# Adapted from:
+# https://ms264556.net/pages/PfSenseLetsEncryptToRuckus
+#
+# ```sh
+# acme.sh --deploy -d ruckus.example.com --deploy-hook ruckus
+# ```
+#
+# Then you need to set the environment variables for the
+# deploy script to work.
+#
+# ```sh
+# export RUCKUS_HOST=ruckus.example.com
+# export RUCKUS_USER=myruckususername
+# export RUCKUS_PASS=myruckuspassword
+#
+# acme.sh --deploy -d ruckus.example.com --deploy-hook ruckus
+# ```
+#
+# returns 0 means success, otherwise error.
+
+########  Public functions #####################
+
+#domain keyfile certfile cafile fullchain
+ruckus_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+  _err_code=0
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  _getdeployconf RUCKUS_HOST
+  _getdeployconf RUCKUS_USER
+  _getdeployconf RUCKUS_PASS
+
+  if [ -z "$RUCKUS_HOST" ]; then
+    _debug "Using _cdomain as RUCKUS_HOST, please set if not correct."
+    RUCKUS_HOST="$_cdomain"
+  fi
+
+  if [ -z "$RUCKUS_USER" ]; then
+    _err "Need to set the env variable RUCKUS_USER"
+    return 1
+  fi
+
+  if [ -z "$RUCKUS_PASS" ]; then
+    _err "Need to set the env variable RUCKUS_PASS"
+    return 1
+  fi
+
+  _savedeployconf RUCKUS_HOST "$RUCKUS_HOST"
+  _savedeployconf RUCKUS_USER "$RUCKUS_USER"
+  _savedeployconf RUCKUS_PASS "$RUCKUS_PASS"
+
+  _debug RUCKUS_HOST "$RUCKUS_HOST"
+  _debug RUCKUS_USER "$RUCKUS_USER"
+  _debug RUCKUS_PASS "$RUCKUS_PASS"
+
+  COOKIE_JAR=$(mktemp)
+  cleanup() {
+    rm $COOKIE_JAR
+  }
+  trap cleanup EXIT
+
+  LOGIN_URL=$(curl https://$RUCKUS_HOST -ksSLo /dev/null -w '%{url_effective}')
+  _debug LOGIN_URL "$LOGIN_URL"
+
+  XSS=$(curl -ksSic $COOKIE_JAR $LOGIN_URL -d username=$RUCKUS_USER -d password="$RUCKUS_PASS" -d ok='Log In' | awk '/^HTTP_X_CSRF_TOKEN:/ { print $2 }' | tr -d '\040\011\012\015')
+  _debug XSS "$XSS"
+
+  if [ -n "$XSS" ]; then
+    _info "Authentication successful"
+  else
+    _err "Authentication failed"
+    return 1
+  fi
+
+  BASE_URL=$(dirname $LOGIN_URL)
+  CONF_ARGS="-ksSo /dev/null -b $COOKIE_JAR -c $COOKIE_JAR"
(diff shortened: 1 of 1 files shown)


uberjay:
Just stumbled across this and verified it works for deploying certificate updates from OPNsense to my Ruckus Unleashed controller.

Steps:

1) Place `ruckus.sh` file in `/usr/local/share/examples/acme.sh/deploy/`
2) `chmod 555 ruckus.sh`
3) Set up certificate for ruckus (via the OPNsense GUI). let's call the domain name `$domain`.
4) Run issue/renew certificate action.
5) ssh to your OPNsense system as root and drop to a shell
6) find the corresponding certificate configuration directory under `/var/etc/acme-client/certs/`
    * in my case it was `/var/etc/acme-client/certs/6716ac20201125.16873601`.
    * let's call this full directory path `$certdir`.
7) edit the cert config file (this will be `$certdir/$domain/$domain.conf`) to add the three variables:
    * `RUCKUS_HOST`
    * `RUCKUS_USER`
    * `RUCKUS_PASS`
8) execute the acme.sh deploy action:

```
/usr/local/sbin/acme.sh --deploy --syslog 7 --debug --server 'letsencrypt' \
  --home '/var/etc/acme-client/home' \
  --cert-home "$certdir/$domain" \
  --certpath "$certdir/$domain/cert.pem" \
  --keypath "$certdir/$domain/private.key" \
  --capath "$certdir/$domain/chain.pem" \
  --fullchainpath "$certdir/$domain/fullchain.pem" \
  --domain "$domain" --deploy-hook ruckus
```

So, I vote for including this deployment hook. What else is needed to make it happen? It needs to be upstream (here) before OPNsense will add an option to use it, and it'd be really fantastic to have it "just work". ✨

deploy/ruckus.sh
```diff
@@ -0,0 +1,110 @@
...
+  _debug RUCKUS_USER "$RUCKUS_USER"
+  _debug RUCKUS_PASS "$RUCKUS_PASS"
+
+  COOKIE_JAR=$(mktemp)
+  cleanup() {
+    rm $COOKIE_JAR
+  }
+  trap cleanup EXIT
+
+  LOGIN_URL=$(curl https://$RUCKUS_HOST -ksSLo /dev/null -w '%{url_effective}')
```
Neilpang, line 75: do not use `curl`,  please use `_post()` or `_get()` function instead.

fraenki:
@uberjay I don't know if the original author is still working on this... Maybe you would be interested to finish the job? I'd suggest to apply the proposed fixes and submit a new PR.

ms264556:
> @uberjay I don't know if the original author is still working on this... Maybe you would be interested to finish the job? I'd suggest to apply the proposed fixes and submit a new PR.

I'm the author of the [code this is adapted from](https://ms264556.net/ruckus/PfSenseLetsEncryptToRuckus).

If anyone's already working on a fix then let me know. Otherwise I'm happy to re-implement this correctly: I'm on holiday for a couple more months, so I have the time.

(I had assumed this PR wouldn't be accepted as-is, since it's not idiomatic with the other acme code).

kchiem:
@ms264556 I was planning on looking into it now, but if you're able to, please do go ahead.

uberjay:
@ms264556 That'd be awesome! I don't know if the original PR author is following along here, and I don't personally have the bandwidth. It'd be very much appreciated!

ms264556:
> @ms264556 I was planning on looking into it now, but if you're able to, please do go ahead.

Sure. I'll setup some test APs and have a look this afternoon.

I do remember that the acme code for existing integrations was very low level compared to my lazy 'get curl do all the thinking' code, so I don't think this is a 5 minute job.

deploy/ruckus.sh
ms264556: My reimplemention using _get() and post() has been pushed to this PR now

deploy/ruckus.sh
```diff
@@ -0,0 +1,173 @@
...
+    return 1
+  fi
+
+  _savedeployconf RUCKUS_HOST "$RUCKUS_HOST"
+  _savedeployconf RUCKUS_USER "$RUCKUS_USER"
+  _savedeployconf RUCKUS_PASS "$RUCKUS_PASS"
+
+  _debug RUCKUS_HOST "$RUCKUS_HOST"
+  _debug RUCKUS_USER "$RUCKUS_USER"
+  _debug RUCKUS_PASS "$RUCKUS_PASS"
```
Neilpang, line 66: use `_secure_debug` for password, so that you won't leak it.

deploy/ruckus.sh
```diff
@@ -0,0 +1,173 @@
...
+
+  export HTTPS_INSECURE=1
```
Neilpang, line 68: why do we need "insecure" here?

deploy/ruckus.sh
```diff
@@ -0,0 +1,173 @@
...
+  _post_upload "uploadcert" "$_cfullchain"
+  
+  _info "Uploading private key"
+  _post_upload "uploadprivatekey" "$_ckey"
+
+  _info "Replacing certificate"
+  _replace_cert_ajax='<ajax-request action="docmd" comp="system" updater="rid.0.5" xcmd="replace-cert" checkAbility="6" timeout="-1"><xcmd cmd="replace-cert" cn="'$RUCKUS_HOST'"/></ajax-request>'
+  _post "$_replace_cert_ajax" "$_base_url/_cmdstat.jsp" >/dev/null
+  
+  info "Rebooting"
```
Neilpang, line 130: use `_info`

deploy/ruckus.sh
ms264556: Ruckus devices ship with 1024 bit self-signed server certificates, so curl will fail with error 60 unless I set HTTPS_INSECURE=1.

deploy/ruckus.sh
ms264556: Fixed!

deploy/ruckus.sh
ms264556: Fixed!

ms264556:
Oops. Would help if I quoted all of my _info() args. One more commit incoming

deploy/ruckus.sh
ms264556: Fixed

deploy/ruckus.sh
ms264556: Fixed, and I also fixed the unquoted _info strings

deploy/ruckus.sh
uberjay: Overriding HTTPS_INSECURE to always be enabled seems wrong to me -- for initial deployment one can add `--insecure`: `acme.sh –deploy -d your.domain.here –deploy-hook ruckus –insecure`. On subsequent cert deploy runs, HTTPS_INSECURE=1 should no longer be necessary.

kchiem:
@ms264556 https://github.com/acmesh-official/acme.sh/actions/runs/11811051778/job/32909254460

deploy/ruckus.sh
ms264556: No problem. That's fixed now.

ms264556:
> @ms264556 [acmesh-official/acme.sh/actions/runs/11811051778/job/32909254460](https://github.com/acmesh-official/acme.sh/actions/runs/11811051778/job/32909254460)

Fixed

Neilpang:
please add the usage here:

ms264556:
> please add the usage here: [Wiki: deployhooks](https://github.com/acmesh-official/acme.sh/wiki/deployhooks)

That's done

ms264556:
I noticed in the dev docs that we're not supposed to use awk. Our awk usage isn't doing anything useful, so I've given @kchiem a PR to remove this, which should be pushed here soon

uberjay:
@ms264556 Thanks for taking care of this. Is there anything else that's needed, apart from acme.sh project approval/merge?

I'm excited at the prospect of this being bundled with OPNsense.

PR #4851, opened 2023-11-01
Update 1984hosting auth endpoint

ddvarpdd:
At one point they've changed it to /api/auth/, with old endpoints returning 404.


ddvarpdd:
Superceeded by #4866.

Issue #4852, opened 2023-11-01
Report bug to Samba AD DC DNS API


Max13:

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

PR #4854, opened 2023-11-01
New Samba AD DC DNS API

Max13:


neilpang:
To answer your question: no -- a samba-tool hook (like dns_nsd/dns_knot) cannot run in the hosted DNS test; post a local `acme.sh --issue ... --staging --debug 2` transcript showing add and rm instead. ShellCheck/shfmt must still pass.

1. `exit 1` in `dns_samba_rm` kills the whole acme.sh process -- use `return 1`.
2. `--password="$SAMBA_PASS"` exposes the password to every local user via `ps`. Use `--authentication-file` with a 0600 temp file.
3. `_get_zone` takes the first suffix match -- with both `example.com` and `sub.example.com` zones present, the wrong one can win. Pick the longest match.
4. Convert the header comments to the standard `dns_samba_info='...'` block (`Issues:` -> #4852), add the dnsapi2 wiki entry, quote `_get_zone "$fulldomain"`, and add the missing final newline (shfmt).

PR #4855, opened 2023-11-02
Acme2 similar names

studycom-mrobinson:
Similar to https://github.com/acmesh-official/acme.sh/pull/1628 I experienced an issue with having multiple similar Hosted Zones within AWS, causing acme.sh to try to use the wrong Zone to renew a certificate, and specifically to fail setting the ACME challenge DNS entry. From looking through the code for just this provider, and testing locally, I was able to add just this bit to escape Periods for the domain name part of the string passed into _egrep_o, without affecting anything else.

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -157,7 +157,7 @@ _get_root() {
 
   # iterate over names (a.b.c.d -> b.c.d -> c.d -> d)
   while true; do
-    h=$(printf "%s" "$domain" | cut -d . -f $i-100)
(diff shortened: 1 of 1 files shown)


studycom-mrobinson:
> **Welcome** Please make sure you're read our [DNS API Dev Guide](../wiki/DNS-API-Dev-Guide) and [DNS-API-Test](../wiki/DNS-API-Test). Then reply on this message, otherwise, your code will not be reviewed or merged. We look forward to reviewing your Pull request shortly ✨

I did read the guides, and setup the test to run. As I am making a 1 line change to an existing file, the Dev Guide didn't apply for what I'm trying to do, but I did indeed read it.

PR #4872, opened 2023-11-14
Update dns_gcloud.sh rm logs record added

sandercox:
The logs show record was added twice but the second time was actual the rm command thus the removal of the record!


sandercox:
Just a log message typo fix.

PR #4925, opened 2023-12-22
Logout update for DSM Deploy script (2727 issue)

LordDarkneo:
As per observed in #2727, the actual logout function is not using the variables gathered from the HTML requests above. This results in an error 102 when logging out, as the lougoutt function is called in the wrong API.

+ Enhancement to lougout the Sesson ID (i.e. the CERT user session) not to kill other all connection when renewing the certificate (Domotic issue when killing others users...)

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -223,7 +223,8 @@ synology_dsm_deploy() {
 
 ####################  Private functions below ##################################
 _logout() {
-  # Logout to not occupy a permanent session, e.g. in DSM's "Connected Users" widget
-  response=$(_get "$_base_url/webapi/entry.cgi?api=SYNO.API.Auth&version=$api_version&method=logout")
+  # Logout to not occupy a permanent session, e.g. in DSM's "Connected Users" widget 
+  #Edit Darkneo - reuse previous variables to logout properly only for CERT user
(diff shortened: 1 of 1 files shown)


Eagle3386:
Sorry for being late to the party & with all due respect, but:

> ```sh
> # Logout CERT user only to not occupy a permanent session, e.g. in DSM's "Connected Users" widget (based on previous variables)
> ```

is ridiculous - first, it's _always_ the "cert user" that's logged out, just because it's _that_ user that's doing all stuff prior to the logout & secondly, that "based on previous variables" is way more confusing than it actually helps.

Furthermore, changing `entry.cgi` to `$api_path` in this:

> ```sh
> response=$(_get "$_base_url/webapi/$api_path?api=SYNO.API.Auth&version=$api_version&method=logout&_sid=$sid")`
> ```

not only is ridiculous, but also wrong - any DSM, be it 6.x or 7.x, _always_ used `entry.cgi` & the _only_ exception to that was the login endpoint which was `auth.cgi` on DSM 6.x & was fixed by Synology to be `entry.cgi`, too, since DSM 7.0 was released years ago.

So, the _only_ real change needed & actually good, was adding `&_sid=$sid` - everything else (and I do mean _everything_) should be reverted as it doesn't help - at best!

PR #4970, opened 2024-01-29
fix bug with domain regex

gleam-ru:
If I have similar domains like "**qwe.ru**" and "asd**qwe.ru**" the first found was used.
Now it is fixed :)


Neilpang:

gleam-ru:
> https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test

There is a CI workflow DNS.yml to test your DNS API when you send **PR to add a new DNS API**.

1) I am not adding NEW api, I'm just editing the existing one. Should I follow this instructions?
2) This DNS provider requires whitelisted IP, so how it will be tested?

neilpang:
Fixed in dev by 7def43481a: the match now requires a dot boundary (`case ".${domain}" in *.${IDN_ITEM}*)`), which fixes the `qwe.ru` / `asdqwe.ru` collision and still matches when the record is the zone apex (DNS alias mode) -- which `*\.${IDN_ITEM}*` would have broken.

To your questions: DNS.yml also applies to modifications of an existing API; for providers requiring whitelisted IPs, run acmetest locally and paste the output instead.

Closing as superseded.

Issue #4972, opened 2024-01-30
Error, can not get domain token entry *.domain.tld for dns-01


Hossy:
Steps to reproduce
------------------
Honestly, not quite sure how to get the CA stuck in this pickle, but I can tell you the symptoms.
(note: strings have been randomized to look real and protect security)
```
acme.sh:issue:4671                       response='{"identifier":{"type":"dns","value":"domain.tld"},"status":"invalid","expires":"2024-02-24T07:05:50Z","challenges":[{"type":"dns-01","url":"https://acme.zerossl.com/v2/DV90/chall/Lt2qWPyV5nnHR8xZXClQHr
","status":"invalid","error":{},"token":"pxFDt2QEktWOZsPIJq5r8H_cTasXmLBvUeJoQRRh4Y9D
"}],"wildcard":true}#https://acme.zerossl.com/v2/DV90/authz/X8_yBrTW3uufk6DJb__mlt'+ printf '\n'

...

[Tue Jan 30 00:45:18 CST 2024] entry
[Tue Jan 30 00:45:18 CST 2024] Error, can not get domain token entry *.domain.tld for dns-01
```

The Problem
------------------
The problem that is happening is due to the code on line [4693](https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L4693): `entry="$(echo "$response" | _egrep_o '[^\{]*"type":"'$vtype'"[^\}]*')"`

where this is assuming the end of the challenge array item must be a `}` and the key:value pair of `"error":{}` is throwing a wrench in that.

The Solution (closer anyway, not perfect)
------------------
I have come up with a replacement for line 4693:
`entry="$(echo "$response" | sed -n 's#"challenges":\[\([^]]*\)\]#\1#p' | sed 's#\(\}\),\({\)#\1\n\2#' | grep '"type":"'$vtype'"')"`

It isn't perfect, but without a tool like **jq** and to remain POSIX compliant, it is extremely difficult to cover absolutely every case.  For example, if, for some reason, the challenge array items start containing arrays themselves, we'll be back in the same boat.  At least this should work until then. :-)

Hossy:
> Hi, i'm also hitting this issue when using zerossl.com as CA. @Hossy thank you for the workaround. BR Takalele

I'll pass your thanks onto my wife. lol.  I was up until 4am reading 13.5k lines of bash debugging output. #insanethings

[...]

PR #4973, opened 2024-01-30
fix for issue 4972; error can not get domain token entry

Hossy:
Addresses #4972

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4690,7 +4690,7 @@ $_authorizations_map"
         _debug keyauthorization "$keyauthorization"
       fi
 
-      entry="$(echo "$response" | _egrep_o '[^\{]*"type":"'$vtype'"[^\}]*')"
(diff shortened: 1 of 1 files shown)


arjanhouben:
even with this patch I am unable to renew my certificates, I don't really know what the problem is. How can I tell which command files by looking at acme.sh.log? I am already using --debug 2

Hossy:
> even with this patch I am unable to renew my certificates, I don't really know what the problem is. How can I tell which command fails by looking at acme.sh.log? I am already using --debug 2
> 
> -- seems to be related to my machine now connecting via ipv6 instead of 4, which means the previous whitelisted IP is no longer valid. I found this by reading the acme.sh.log from the start, as the script continuous even after encountering an error, which was surprising to me.

You might need to run acme.sh with bash -x.  Before that, try increasing --debug to 3

Issue #4971, opened 2024-01-29
_isIPv4 vulnerability based on existing files in /acme.sh


Hossy:
https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L4266

Steps to reproduce
------------------
1. touch /acme.sh/1
2. Make request for *.*.*.*

Recommend changing the line from:
for seg in $(echo "$1" | tr '.' ' '); do
to:
for seg in $(echo "$1" | tr -cd '[0-9].' | tr '.' ' '); do

Hossy:
Alternatively, this might be a less vulnerable function and POSIX safe:

```
_isIPv4() {
    # Disable pathname expansion
    set -f

    # Save the current value of IFS
    old_IFS="$IFS"
    IFS='.'

    # Split the IP into octets
    _chk_ipv4="$1"
    # shellcheck disable=SC2086
    set -- $_chk_ipv4

    # Restore the original value of IFS
    IFS="$old_IFS"

    # Re-enable pathname expansion
    set +f

    # Check if the IP has exactly 4 octets
    if [ $# -ne 4 ]; then
        # Invalid IPv4 address
        _debug2 "$_chk_ipv4 does not have 4 octets"
        return 1
    fi

    # Validate each octet
    for octet in "$@"; do
        _debug octet "$octet"
        # Check if octet is numeric
        if ! [ "$octet" -eq "$octet" ] 2>/dev/null; then
            # octet is not numeric
            return 1
        fi

        # Check if octet is in range 0-255
        if [ "$octet" -lt 0 ] || [ "$octet" -gt 255 ]; then
            # octet is out of range
            return 1
        fi
    done

    # If all checks pass, IP is valid
    return 0
}
```

neilpang:
Fixed in dev (988afd0f): splitting now runs under `set -f`, exactly 4 octets required, each a 1-3 digit number <= 255. Regression test `le_test_isIPv4` runs in a directory containing a file named "1" to cover the glob case.

PR #4974, opened 2024-01-30
fix for issue 4971; _isIPv4 vulnerability

Hossy:
Addresses #4971


neilpang:
Fixed in dev by 988afd0f, based on this PR -- thanks. The final version uses a `case` digit check plus a length guard (oversized numbers can't overflow `test`), and adds a regression test (`le_test_isIPv4`) to acmetest. Closing.

PR #4998, opened 2024-02-13
dns_ispconfig.sh client_id not numeric at ispconfig v 3.2.7p1

dagobert:
Getting client_id failed due to incorrect extraction!

At least in version 3.2.7p1 and probably later the plugin is not working any more properly. The result of ```curResult="$(_post "${curData}" "${ISPC_Api}?client_get_id")"``` is something like this ```Result of _ISPC_ClientGetID: '[Tue Jan 23 11:44:57 CET 2024] Retrying post 
{"code":"ok","message":"","response":3}[Tue Jan 23 11:44:57 CET 2024] _hcode 0'```. The parsing code does not work properly and leaves a non numeric value such as ```Client ID: '3[Tue Jan 23 11'```.


dagobert:
Reissued a pull request. Sorry for not changing this one.

PR #5036, opened 2024-03-05
Update dns_regru.sh

koledas:
Fix incorrect select of domain when first is substring of second.

For example:
For domain "test.com.ru" will be selected domain "subtest.com.ru" as root.


neilpang:
Thanks @koledas -- applied your fix to `dev` directly (with a dot-boundary so a shorter registered domain can't match a longer subdomain label). Verified it resolves the substring mis-match you described. Closing as merged.

PR #5043, opened 2024-03-07
Add deploy/zyxel_gs1900.sh

### LINKED ISSUE #5042: feature: deploy to Zyxel GS1900 series swtches
Zyxel has added a web interface for uploading custom certificates to its consumer grade GS1900 series switches as of firmware V2.80.

I've hacked together a script I've been using to deploy from `acme.sh` to my switches and I'm curious if that is something that would be valuable as a PR.

If it is valuable, I will eventually seek to get it included in the list of supported deployments for the [opnsense os-acme-client plugin](https://github.com/opnsense/plugins/blob/master/security/acme-client)

coderjoe:
Add support for deploying to the Zyxel GS1900 line of switches as long as those switches are running at least firmware V2.80.

Tested on a Zyxel GS1900-8 and GS1900-24E

Resolves #5042

### DIFF
--- a/deploy/zyxel_gs1900.sh
+++ b/deploy/zyxel_gs1900.sh
@@ -1,0 +1,500 @@
+#!/usr/bin/env sh
+
+# Deploy certificates to Zyxel GS1900 series switches
+#
+# This script uses the https web administration interface in order
+# to upload updated certificates to Zyxel GS1900 series switches.
+# Only a few models have been tested but untested switches from the
+# same model line may work as well. If you test and confirm a switch
+# as working please submit a pull request updating this compatibility
+# list!
+#
+# Known Issues:
+#   1. This is a consumer grade switch and is a bit underpowered
+#      the longer the RSA key size the slower your switch web UI
+#      will be. RSA 2048 will work, RSA 4096 will work but you may
+#      experience performance problems.
+#   2. You must use RSA certificates. The switch will reject EC-256
+#      and EC-384 certificates in firmware 2.80
+#      See: https://community.zyxel.com/en/discussion/21506/bug-cannot-import-ssl-cert-on-gs1900-8-and-gs1900-24e-firmware-v2-80/
+#
+# Current GS1900 Switch Compatibility:
+#   GS1900-8    - Working as of firmware V2.80
+#   GS1900-8HP  - Untested
+#   GS1900-10HP - Untested
+#   GS1900-16   - Untested
+#   GS1900-24   - Untested
+#   GS1900-24E  - Working as of firmware V2.80
+#   GS1900-24EP - Untested
+#   GS1900-24HP - Untested
+#   GS1900-48   - Untested
+#   GS1900-48HP - Untested
+#
+# Prerequisite Setup Steps:
+#   1. Install at least firmware V2.80 on your switch
+#   2. Enable HTTPS web management on your switch
+#
+# Usage:
+#   1. Ensure the switch has firmware V2.80 or later.
+#   2. Ensure the switch has HTTPS management enabled.
+#   3. Set the appropriate environment variables for your environment.
+#
+#      DEPLOY_ZYXEL_SWITCH          - The switch hostname. (Default: _cdomain)
+#      DEPLOY_ZYXEL_SWITCH_USER     - The webadmin user. (Default: admin)
+#      DEPLOY_ZYXEL_SWITCH_PASSWORD - The webadmin password for the switch.
+#      DEPLOY_ZYXEL_SWITCH_REBOOT   - If "1" reboot after update. (Default: "0")
+#
+#   4. Run the deployment plugin:
+#      acme.sh --deploy --deploy-hook zyxel_gs1900 -d example.com
+#
+# returns 0 means success, otherwise error.
+
+#domain keyfile certfile cafile fullchain
+zyxel_gs1900_deploy() {
+  _zyxel_gs1900_minimum_firmware_version="v2.80"
+
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug2 _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  _getdeployconf DEPLOY_ZYXEL_SWITCH
+  _getdeployconf DEPLOY_ZYXEL_SWITCH_USER
+  _getdeployconf DEPLOY_ZYXEL_SWITCH_PASSWORD
+  _getdeployconf DEPLOY_ZYXEL_SWITCH_REBOOT
+
+  if [ -z "$DEPLOY_ZYXEL_SWITCH" ]; then
+    DEPLOY_ZYXEL_SWITCH="$_cdomain"
+  fi
+
+  if [ -z "$DEPLOY_ZYXEL_SWITCH_USER" ]; then
+    DEPLOY_ZYXEL_SWITCH_USER="admin"
+  fi
+
+  if [ -z "$DEPLOY_ZYXEL_SWITCH_PASSWORD" ]; then
+    DEPLOY_ZYXEL_SWITCH_PASSWORD="1234"
+  fi
+
+  if [ -z "$DEPLOY_ZYXEL_SWITCH_REBOOT" ]; then
+    DEPLOY_ZYXEL_SWITCH_REBOOT="0"
+  fi
+
+  _savedeployconf DEPLOY_ZYXEL_SWITCH "$DEPLOY_ZYXEL_SWITCH"
+  _savedeployconf DEPLOY_ZYXEL_SWITCH_USER "$DEPLOY_ZYXEL_SWITCH_USER"
+  _savedeployconf DEPLOY_ZYXEL_SWITCH_PASSWORD "$DEPLOY_ZYXEL_SWITCH_PASSWORD"
+  _savedeployconf DEPLOY_ZYXEL_SWITCH_REBOOT "$DEPLOY_ZYXEL_SWITCH_REBOOT"
+
+  _debug DEPLOY_ZYXEL_SWITCH "$DEPLOY_ZYXEL_SWITCH"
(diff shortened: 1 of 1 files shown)


coderjoe:
I'm now using this version locally for my switches. Please let me know if there is anything you'd like me to do to improve the PR. No rush. :)

coderjoe:
A bug has occurred on one of my servers with a different version of sed. Working on a fix now.

coderjoe:
Most recent commit should have resolved the issue I found.

deploy/zyxel_gs1900.sh
```diff
@@ -0,0 +1,423 @@
...
+  _CURL="$_ACME_CURL"
+  if [ "$HTTPS_INSECURE" ]; then
+    _CURL="$_CURL --insecure  "
+  fi
+
+  # Unfortunately we cannot rely upon the switch response across switch models
+  # to return a consistent body return - so we cannot inspect the result of this
+  # upload to determine success. We will need to re-query the certificates page
+  # and compare the validity dates to try and identify if they have changed.
+  upload_response_html=$($_CURL \
```
Neilpang, line 289: only for `curl` here?  how about `wget` ?
I would suggest you not use `curl` here,  please use `_post()` method instead.

coderjoe: I originally followed the example in `synology_dsm.sh` and constructed the multpart form request by hand. This did not work in either `_post()` or `wget`. Both of those mangled the binary certificate data.

Specifically some of the shell-unsafe bytes and null bytes seemed to be getting mangled.

I am willing to try again, but I did not have success originally.

coderjoe: I will try, but I'm pretty sure the server is expecting unencoded binary data for the certificate file in the multi part form upload.

coderjoe: Thank you for your suggestion. My computer has broken and I'm now waiting on replacement parts.

Once I'm back up and running I will be trying what you suggest.

coderjoe: I have tried your suggestion but I cannot make it work.
I will am updating the PR to draft, and updating the branch to see if you have any other ideas to avoid the `curl` requirement.

deploy/zyxel_gs1900.sh
```diff
@@ -275,26 +275,41 @@ _zyxel_gs1900_upload_certificate() {
...
+    printf -- "%s\r\n" "${upload_post_boundary}"
+    printf "Content-Disposition: form-data; name=\"cmd\"\r\n\r\n%s\r\n" "31"
+    printf -- "--"
+    printf -- "%s\r\n" "${upload_post_boundary}"
+    printf "Content-Disposition: form-data; name=\"sysSubmit\"\r\n\r\n%s\r\n" "Import"
+    printf -- "--"
+    printf -- "%s\r\n" "${upload_post_boundary}"
+    printf "Content-Disposition: form-data; name=\"http_file\"; filename=\"cert.pfx\"\r\n"
+    printf "Content-Type: application/octet-stream\r\n\r\n"
+    cat "${temp_pkcs12}"
```
coderjoe, line 298: Note: the file located at `temp_pkcs12` contains null `\0` bytes.
As such it cannot be saved to a variable in many shells.

deploy/zyxel_gs1900.sh
```diff
@@ -275,26 +275,41 @@ _zyxel_gs1900_upload_certificate() {
...
-    -X "POST" \
-    -H "$_H1" \
-    --form "XSSID=${form_xss_value}" \
-    --form "http_file=@${temp_pkcs12}" \
-    --form "pwd=${temp_cert_password}" \
-    --form "cmd=31" \
-    --form "sysSubmit=Import" \
-    "${_zyxel_switch_base_uri}/cgi-bin/httpuploadcert.cgi")
+  export _H2="Content-type: multipart/form-data boundary=${upload_post_boundary}"
+  _post "$(cat "${upload_post_request}")" "${_zyxel_switch_base_uri}/cgi-bin/httpuploadcert.cgi"
```
coderjoe, line 311: @Neilpang  this seems to be my problem. The body must be passed as a variable to `_post` but I cannot put null bytes `\0` in a variable in all shells.

This results in the `\0`s being stripped from the body and makes the body invalid.

I cannot url-encode this because the server requires the raw octet byte stream, not an encoded value.

`curl` supports accepting a raw file via the "var=@filename" format.
`wget` does not seem to support accepting raw file input.

Do you have any other ideas that I may be able to try?

deploy/zyxel_gs1900.sh
```diff
@@ -275,26 +275,41 @@ _zyxel_gs1900_upload_certificate() {
...
 
   # Unfortunately we cannot rely upon the switch response across switch models
   # to return a consistent body return - so we cannot inspect the result of this
   # upload to determine success. We will need to re-query the certificates page
   # and compare the validity dates to try and identify if they have changed.
-  upload_response_html=$($_CURL \
-    -X "POST" \
-    -H "$_H1" \
-    --form "XSSID=${form_xss_value}" \
-    --form "http_file=@${temp_pkcs12}" \
```
coderjoe: Note the @ sign in this --form.
Many --form makes it a multipart form
The @ sign makes it read the file for the value of the form data and allows me to have a file with null bytes `\0` in them.

This is what I cannot reproduce.

coderjoe (commented):
These are the differences between the two implementations and why I was using `curl` rather than `_post`. I have not been able to get `_post` to work.

deploy/zyxel_gs1900.sh
coderjoe: Resolved by saving to a file and using --data-binary for curl and --post-file or --body-file for wget.

coderjoe:
I've updated for the shellcheck and shfmt linting in the previous run, have removed the defunct curl errors (now that curl is not a requirement) and have squashed the commit to clean up history a bit.

coderjoe:
I've verified that the updated commits continue to work in all of my deployed environments.
Thank you very much @Neilpang for your help/guidance on wget support.

I don't think I would have figured it out without your suggestion. :)

deploy/zyxel_gs1900.sh
```diff
@@ -0,0 +1,425 @@
...
+
+  return $?
+}
+
+_zyxel_gs1900_upload_certificate() {
+  # Generate a PKCS12 certificate with a temporary password since the web interface
+  # requires a password be present. Then upload that certificate.
+  temp_cert_password=$(head /dev/urandom | tr -dc 'A-Za-z0-9' | head -c 64)
+  temp_cert_password="password"
+  _secure_debug2 "temp_cert_password" "$temp_cert_password"
```
Neilpang, line 243: temp_cert_password  always equals to "password" ?

coderjoe: Ugh... rookie mistake... I thought I reverted that. Left over from testing binary cert upload issues ... 🤦‍♀️

coderjoe: Fixed in latest, as well as removing some unnecessary stdout from the upload that isn't used by the switch.

coderjoe:
I have now been using this in production for a few months. @Neilpang please let me know if there is anything else you would like me to do - or if there's anything I can do to help get this included. :)

Thanks again for your time and help!

coderjoe:
Rebased onto the latest `dev`.
Please let me know what else needs to be done for a review.

Neilpang:
why do you need to "ReadBodyFromFile"?

coderjoe:
> why do you need to "ReadBodyFromFile"?

Great question @Neilpang. I responded in the gitter.im chat while the repo was restricted, but I'll repeat it here just in case it wasn't seen.

I need to read from file because this switch expects a binary exact, not encoded, upload of the certificate. The uploaded certificate contains many \0 NULL bytes which cannot be piped within the shell script without being truncated.

Making `wget` or `curl` read the certificate file directly is the only way I have been able to get the full file uploaded in the manner the switch expects without the file becoming truncated by the null bytes.

I hope this helps clarify.

dark-m0de:
Hi, I would really appreciate if we could add support for Zyxel devices. I'm using a XMG1915 switch, as well as NWA50AX PRO and NWA130BE access points. My assumption is that with a few changes, those devices could be supported as well.
I'd be happy to support the development for this.

coderjoe:
Hello @dark-m0de , I appreciate your interest. I don't think in its current form my PR will be accepted. I think my modifications to the _post method in acme.sh are not of sufficient quality and I'm currently investigating alternatives.

Some notes:

- Web interfaces between different series of Zyxel switches tend to be different enough that they will likely require separate integrations. At least that has been my past experience. 
- I'm entirely unfamiliar with their AP products so I cannot provide any recommendation, but much like the XGS series I would be very hesistant to support them in this script.
- I have an open [feature request on the Zyxel community](https://community.zyxel.com/en/discussion/comment/73153#Comment_73153) to suppor other methods of upload other than just web based upload of binary PKCS12 certificates. That conversation is ongoing at this time and if you'd like to leave a polite "me too" that would be welcome.

For now I think I'm going to put this PR back into draft until I can come up with some alternative methods to get the certificates into the GS1900 series switchs.

dark-m0de:
Thank you @coderjoe, I just prepared my answer for the zyxel community. But somehow I'm not able to post it yet.

Another thought: Do you think it would help if zyxel adds support for uploading certificates to their CLI (e.g. ssh based)? This would add ssh as a dependency to acme in such scenarios. But it would definitely ease the process to manage certificates in a scripted way.

coderjoe:
@Neilpang  I've updated this PR to the latest `dev` and removed all depependence on changes to the acme.sh base functionality. If you manage to find the time I'd appreciate another review.

Thank you in advance. :)

coderjoe:
Updated to correct the failures in the prior commit's ShellCheck and shfmt errors.

Neilpang:
add your usage here:

coderjoe:
Documentation added to the wiki:
https://github.com/acmesh-official/acme.sh/wiki/deployhooks#36-deploying-to-zyxel-gs1900-series-switches

Neilpang:
please create a bug to accept reports:
https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide#9-please-create-a-new-issue-for-future-bugs

coderjoe:
Bug report issue #6370 created and subscribed. :+1:

PR #5049, opened 2024-03-11
fixed handling of key and secret

hknet:
fixed handling of key and secret values read from acme's account.conf


hknet:
previously already stored values in account.conf were not handled in a way to work with those values, this is would be fixed using this PR.

hknet:
could I ask to merge this fix as the current master release does not handle account data already stored correctly.

PR #5053, opened 2024-03-14
Cater for name.com accounts with > 1000 domains

mattgroves:
Max limit from the name.com API is 1000 records - they paginate this, so this patch caters for iterating through the pages, checking the domain exists and then performing the required updates


Neilpang:

PR #5061, opened 2024-03-20
Add deployhook for CacheFly, Edgio and Netlify

b1n23:

### DIFF
--- a/deploy/cachefly.sh
+++ b/deploy/cachefly.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env sh
+
+# Script to deploy certificate to CacheFly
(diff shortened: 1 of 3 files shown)


Neilpang:
fix the ci errors, and add your usage here:

b1n23:
> fix the ci errors, and add your usage here: https://github.com/acmesh-official/acme.sh/wiki/deployhooks

completed

b1n23:
> @b1n23 The script for KeyHelp isn't usable if 2FA is activated. In #6556 I tried to add a solution using KeyHelp's API. Do you see a possibility that the script could use the API as an alternative?

This API appears to be intended for administrators only and isn’t accessible to client user.

I think rename the script to `keyhelp_api.sh` is feasible—perhaps we should check with @Neilpang  to confirm.

PR #5083, opened 2024-04-09
Option for the docker deploy script to execute the reload cmd with the user root

alex1702:
I have a container running with an unprivileged user. As a result, customizing the file permissions via the reload cmd command does not work.
For example, I have configured the following:
```
DEPLOY_DOCKER_CONTAINER_RELOAD_CMD="chmod 664 /opt/emqx/etc/certs/*.pem && /opt/emqx/bin/emqx stop"
```
For the chmod command to work, docker exec must be executed with the parameter -u root. And this is what I have added in this pull request for docker exec and the API call.

The option is called `DEPLOY_DOCKER_CONTAINER_RUN_AS_ROOT` and is false by default.

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,0 +1,10 @@
+PDNS_Url=http://192.168.42.199:8000
+PDNS_ServerId=localhost
+PDNS_Token=aaff153f99761ce9931f6717016ad2f4-0b87b0a7263927744a361008e0c5110b
+DEPLOY_DOCKER_CONTAINER_LABEL=sh.acme.autoload.domain=test.elaon.de
+DEPLOY_DOCKER_CONTAINER_KEY_FILE=/opt/emqx/etc/certs/key.pem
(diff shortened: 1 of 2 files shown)


neilpang:
1. Remove `.test-env` -- it contains your real PDNS token and deploy config; rotate that token.
2. `==` inside `[ ]` is a bashism (3 places) -- use `=`; dash/busybox fail with "unexpected operator".
3. After normalization the variable is always "true"/"false", so `if [ "$DEPLOY_DOCKER_CONTAINER_RUN_AS_ROOT" ]` is always true -- save unconditionally.
4. `_RUN_AS_ROOT` / `_OPTS` are too generic for the shared hook namespace -- use `_DOCKER_RUN_AS_ROOT` etc.
5. `docker exec $_OPTS -i` relies on unquoted word-splitting (SC2086) -- branch explicitly: `docker exec -u root -i ...` vs `docker exec -i ...`.

Closing as stale; a resubmission addressing the above is welcome.

PR #5087, opened 2024-04-11
Allow creation of ACME account with EAB directly from `--issue` command

clement-gouin:
`_regAccount` on `--issue` was not using `--eab-kid` and `--eab-hmac-key` parameters like `--register-account` does

Also, `_eab_kid` was changed to `_eab_id` for no reason along account registration


Klemek:
> don't rename `_eab_kid` if for no reason.

I'ts actually how it's called on the [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555) I don't think that's "no reason"

> [7.3.4](https://datatracker.ietf.org/doc/html/rfc8555#section-7.3.4).  External Account Binding
> [...]
> o The "kid" field MUST contain the key identifier provided by the CA

Issue #3359, opened 2021-01-11
Report bugs to unifi deploy hook


medmunds:
This is the place to report bugs in the [unifi deploy hook](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#23-deploy-the-cert-on-a-unifi-controller-or-cloud-key).

If you experience a bug with `--deploy-hook unifi`, please report it in this issue. (Please include the `--debug` output to assist in diagnosing any problems.)

Thanks!

blackjack4494:
just curious tho this might not be related purely to the unifi deploy hook (first time using acmesh, used certbot before) but i couldn't find something on that matter: does the cronjob that comes with acme is also deploying the certs?  
I wrote a brief 'how-to' [gist](https://gist.github.com/blackjack4494/331e46678c0ea15a61c4cc6756c21969) on how to freshly install, issue and deploy a cert on a unifi controller. And the answer to my question would be nice to know.  
Thanks in advance (:

medmunds:
@blackjack4494 yes, acme.sh saves your deploy-hook settings, and re-deploys with those same settings after a successful cron certificate renewal. (This applies to all deploy hooks, not just unifi.)

SupersixEvo:
I'm just replacing an Ubuntu 16.04 VM running Unifi with a new VM running 22.04.

When trying to use the unifi hook (which worked great on 16.04), I get

```
[Fri 22 Jul 22:49:28 UTC 2022] Installing certificate for Unifi Controller (Java keystore)
[Fri 22 Jul 22:49:28 UTC 2022] _unifi_keystore='/usr/lib/unifi/data/keystore'
[Fri 22 Jul 22:49:28 UTC 2022] Generate import pkcs12
[Fri 22 Jul 22:49:28 UTC 2022] Import into keystore: /usr/lib/unifi/data/keystore
Importing keystore /tmp/tmp.UHja0HRWf4 to /usr/lib/unifi/data/keystore...
keytool error: java.io.IOException: keystore password was incorrect
[Fri 22 Jul 22:49:29 UTC 2022] Error importing into Unifi Java keystore.
[Fri 22 Jul 22:49:29 UTC 2022] Please re-run with --debug and report a bug.
[Fri 22 Jul 22:49:29 UTC 2022] Error deploy for domain:unifi.xxx.me.uk
[Fri 22 Jul 22:49:29 UTC 2022] Deploy error.
```
I can run

[...]

PR #5095, opened 2024-04-17
Unifi OS removed keytool in v3.1+ and other updates to unifi deploy hook

3VAbdAVE:
Ubiquiti removed keytool from UnifiOS ~v3.1, but still uses a JKS keystore with a self-signed for the console application. 
This script probably shouldn't be auto-installing packages on appliances, but this now explains what's up to the user and lets them decide when it fails on a missing keytool.

The Unifi console is not configured to use strong cipher suites required by default letsencrypt RSA certificates, so update the configuration prior to service restart.

In my initial attempts to use this deploy hook, it broke and I had to hunt down original certificates, so backup the existing certificates, which we just assume are working.

### DIFF
--- a/deploy/unifi.sh
+++ b/deploy/unifi.sh
@@ -13,10 +13,10 @@
 #       OS Version v2.0.0+
 #       Network Application version 7.0.0+
 #       OS version ~3.1 removed java and keytool from the UnifiOS. These must be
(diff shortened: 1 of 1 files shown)


Neilpang:
fix the format errors.

medmunds:
Ref #3359

PR #5121, opened 2024-04-29
dns_njalla: fix record_id grep when removing record

alfredfo:
Before remove-record would fail with
response='{"error": {"code": -32000, "message": "Invalid arguments"}, "jsonrpc": "2.0"}'
because record_id would always be empty.

Signed-off-by: Alfred Persson Forsberg <cat@catcream.org>


Emblaze3523:
This is great! Thanks man.

neilpang:
Confirmed by another user and applied in dev with a pattern that accepts both the old numeric and the new string id format: https://github.com/acmesh-official/acme.sh/commit/58cd667d. Credited to you. Closing.

PR #5139, opened 2024-05-07
feat: guide user to run script as root to create temp admin user

scruel:
Message text and comment optimized


scruel:
@Neilpang This PR only contains nit msg correction, can be safely merged, please check this when you have time.

PR #5158, opened 2024-05-28
update dns_doapi

henrikalves:
Hi,
we are moving our API endpoint to a different subdomain (the old one will continue to work for now),
dns_do does not exist anymore so I have removed it from the text.

**The tests are failing due to the acmetestXyzRandomName. subdomain that is being used in the [tests](https://github.com/acmesh-official/acmetest/blob/master/letest.sh#L1506), the api of ours currently only allows subdomains starting with "_acme-challenge.", why is this a requirement now? I can see nowhere in the [RFC 8555](https://datatracker.ietf.org/doc/html/rfc8555#section-8.4) that subdomains not starting with "_acme-challenge." are allowed.**


henrikalves:
Please read the message above why the tests are failing, i have read the guide

Issue #5161, opened 2024-05-30
Report bug to fornex.com dns api v.2 


zer0init1:
This is the place to report bugs in the fornex.com DNS API v.2.2.0.

If you experience a bug, please report it in this issue.

stenri:
New dns_fornex.sh has a number of issues:

* It doesn't correctly handle sub-domains;
* Script removes ALL TXT records from domain regardless of who created them. Bye-bye all marketing google-site-verification, yandex-verification and other such records!
* And script doesn't handle errors well enough.

See my review for PR #5162: https://github.com/acmesh-official/acme.sh/pull/5162#pullrequestreview-2136120864

Also, it was not very wise move for **fornex.com** to kill support for old v0.1 DNS API without waiting for PR to acme.sh with DNS API 2.2.0 support to be accepted and be merged into master. Effectively making all customers who use DNS API to renew domains unable to do that. That's not how a solid company handles their customers.

zer0init1:
Thx @stenri . I have passed it on to my colleague and we will fix it. But the issue with deleting some of the entries should be resolved by now, as far as I know. Perhaps the reason is that the mirroring from our internal git to github didn't work. We will check this case.

Thanks again for your help and feedback.

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

PR #5162, opened 2024-05-30
fornex.com API v.2.2.0 support

zer0init1:
This is official script from Fornex.com Dev-Team https://github.com/fornex-com/

Bugs: https://github.com/acmesh-official/acme.sh/issues/5161


stenri (changes requested):
New `dns_fornex.sh` doesn't correctly handle sub-domains, removes all TXT records from domain, and doesn't handle errors well enough.

zer0init1:
We update soon this project and create new PR.

Thx all for comments.

PR #5187, opened 2024-06-26
improve performance and memory usage for dns_anx

rwese:
when fetching all zones the memory usage can exceede limits and also cause timeouts.

with this change the zone will be searched via the longest to shortest match using the get endpoint.


rwese:
Read and verified the Dev Guide / Api-Test Doc. :)

Issue #5193, opened 2024-07-03
Bug reports for the myLoc/webtropia DNS API


flesniak:
This is the tracker for issues with the `dns_myloc` API backend for domains registered at myloc.de / webtropia.com.

If you encounter any issues, let me know here.

khaledfa2:
Please add/update the dns API hook "dns_myloc" to be included into the list of the dns APIs because the latest software version does not include it.
Using the command "ls /root/.acme.sh/dnsapi" shows that dns_myloc is not included in the list of the various providers. Please see the screenshot attached. and so, cert renewal automation with this dns_api is not possible.
[Screenshot-4  acme.sh list of dns_api.pdf](https://github.com/user-attachments/files/16265095/Screenshot-4.acme.sh.list.of.dns_api.pdf)

flesniak:
@khaledfa2 the `dns_myloc` hook is not merged yet: https://github.com/acmesh-official/acme.sh/pull/5194
If you want to test it early, you can use my fork referenced there.

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

flesniak:
Please reopen the issue reopened in favor of PR #5194. I don't have permission to do so.

PR #5194, opened 2024-07-03
Add dnsapi script for myloc.de/webtropia.com

flesniak:
Hello everyone,
this is a dnsapi script for the API used by myloc.de and webtropia.com.

[Bug tracker](https://github.com/acmesh-official/acme.sh/issues/5193)
[Documentation](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#166-use-mylocde--webtropiacom-dns-api)

The CI does not pass currently, however I believe this is only a temporary issue.

### DIFF
--- a/dnsapi/dns_myloc.sh
+++ b/dnsapi/dns_myloc.sh
@@ -1,0 +1,133 @@
+#!/usr/bin/env sh
+
+# updater for the (experimental) API of myloc.de / webtropia.com
+# usage: acme.sh --issue -d example.com --dns dns_myapi
+# API documentation at https://apidoc.myloc.de/
+# As the API does not support quering available zones yet, the zone for a given
+# domain needs to be guessed. When using a subdomain, e.g. sub1.example.com,
+# the zone needs to be given via the MYLOC_zone environment variable.
+
+# Environment variables:
+# export MYLOC_token=aabbccddeeffgghhiijjkkllmmnnjjkkllmmnnooppqqrrsstt
+# export MYLOC_zone=example.com
+
+myloc="https://zkm.myloc.de/api"
+
+#Usage: add  _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_myloc_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  token="${MYLOC_token:-$(_readaccountconf_mutable MYLOC_token)}"
+  zone="${MYLOC_zone:-$(_readaccountconf_mutable MYLOC_zone)}"
+
+  # the API does not provide an interface to get available zones yet
+  # if no zone is set in the configuration, try to guess it from $fulldomain
+  zone="$(_guess_zone "$fulldomain" "$zone")"
+
+  export _H1="Content-Type: application/json"
+  export _H2="Authorization: Bearer $token"
+
+  _debug "Getting records"
+  response="$(_get "${myloc}/dns/zone/${zone}")"
+
+  if _contains "$response" "error" || ! _contains "$response" "content"; then
+    _err "Failed to query zone records"
+    _debug "$response"
+    return 1
+  fi
+
+  # save token and zone if the previous request was successful
+  _savedomainconf MYLOC_token "$token"
+  _savedomainconf MYLOC_zone "$zone"
+
+  # records="$(_extract_txt_records "$response" "$fulldomain" | head -1)"
+  # _debug "existing record $record"
+  # if [ "$record" ]; then
+  #   _info "Record for $fulldomain already exists, trying to remove it first"
+  #   response="$(_post "$record" "${myloc}/dns/zone/${zone}" "" "DELETE")"
+  #   if [ $? -ne 0 ]; then
+  #     _info "Failed to delete record, continueing anyway"
+  #   fi
+  # fi
+
+  _info "Adding record"
+  record="{\"type\":\"TXT\",\"name\":\"${fulldomain}\",\"content\":\"${txtvalue}\",\"ttl\":60}"
+  _debug "add record $record"
+  response="$(_post "$record" "${myloc}/dns/zone/${zone}" "" "PUT")"
+  _debug "add response $response"
+  if [ $? -eq 0 ]; then
+    if _contains "$response" "error" || _contains "$response" "unexpected"; then
+      _err "Add txt record api error."
+      return 1
+    elif [ -z "$response" ]; then
+      _info "Empty response, OK"
+      return 0
+    else
+      _err "Add txt record unknown response."
+      return 1
+    fi
+  fi
+  _err "Add txt record curl error."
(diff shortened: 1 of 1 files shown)


flesniak:
CI terminates without an error message after showing:
```
ng_domain_1
2024-07-03T16:19:38Z INF Thank you for trying Cloudflare Tunnel. Doing so, without a Cloudflare account, is a quick way to experiment and try it out. However, be aware that these account-less Tunnels have no uptime guarantee. If you intend to use Tunnels in production you should use a pre-created named tunnel by following: https://developers.cloudflare.com/cloudflare-one/connections/connect-apps
2024-07-03T16:19:38Z INF Requesting new quick Tunnel on trycloudflare.com...
Can not get cf domain.
failed to parse quick Tunnel ID: invalid UUID length: 0
```
Seems like a temporary issue only. Apart from this, ready for review.

neilpang:
1. DNS.yml has never run in your fork -- wire up the secrets and get a passing run per the [DNS-API-Test wiki](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test). Required before review.
2. `$?` is tested after `_debug`, not after `_post`: in `add` the `[ $? -eq 0 ]` is always true, in `rm` the `[ $? -ne 0 ]` is always false, so failures pass silently.
3. `_extract_txt_records`, `_guess_zone`, global `myloc` -- need a `_myloc_` prefix; hooks share one namespace.
4. `${fulldomain#_acme-challenge.}` breaks subdomains and alias mode; probe zones by walking `fulldomain` suffixes with `GET /dns/zone/<candidate>`.
5. Token read via `_readaccountconf_mutable` but saved via `_savedomainconf` -- use `_saveaccountconf_mutable`; only `MYLOC_zone` is per-domain.
6. Add the `dns_myloc_info` block (wiki link in `Docs:`, #5193 in `Issues:`); drop the commented-out block in `add`.

flesniak:
Thanks for your feedback @neilpang!
- DNS.yml runs successfully now
- return codes are now checked properly, good catch!
- prefix added to helper functions
- subdomains are walked up until an existing zone is found instead of guessing
- token is saved as recommended now
- Info block added
- Documentation in the wiki is already updated to match the updated script

neilpang:
1. `dns_myloc_rm` never uses `$txtvalue` -- `_myloc_extract_txt_records` matches on name only, so it deletes every TXT at `_acme-challenge.<domain>`, including a concurrent issuance's record. Match name **and** content.
2. No guard for an empty `MYLOC_token`: requests go out unauthenticated and the user only sees "Get zone failed for all candidates". Add the usual early `_err "You didn't specify MYLOC_token"` / `return 1`.
3. `_myloc_get_zone` still does `${1#_acme-challenge.}` -- redundant now that it walks, and in alias mode it skips the one candidate that can be the zone. Walk from `$1`.
4. Success is decided by `_contains "$response" "error"` / `"content"` over the whole payload: a zone with no records is reported as nonexistent, and a record whose content contains `error` fails the add. Branch on the HTTP status instead: `_code="$(grep "^HTTP" "$HTTP_HEADER" | _tail_n 1 | cut -d " " -f 2 | tr -d "\r\n")"`.
5. `for record in $records` splits on spaces, so a TXT value containing one produces malformed DELETE bodies. Pipe into `while read -r` instead.
6. `Issues: .../5193` points at a closed issue -- reopen it. Minor: `sed -r` is GNU-only (`-E`); `myloc`, `token`, `zone`, `status`, `record` are still unprefixed globals.

flesniak:
Feedback applied to PR, thanks!

1. `dns_myloc_rm` now deletes only a single TXT record with the given `$txtvalue`. The corresponding API request can be fully crafted from the given values, thus the unused `_myloc_extract_txt_records` is removed.
2. Guard for empty token added.
3. I kept the prefix removal using `${1#_acme-challenge.}` on purpose to save one unnecessary API request. I don't know if the DNS API has rate limits, but I'd like to avoid it if possible. I didn't think about alias mode in the first place, but to my understanding `dns_myloc_rm` will be called with `_acme-challenge.acme.example.org` if `acme.example.org` is the alias domain. That would still work fine with removing the prefix. I tested alias mode on a test domain of mine and it worked fine. Is there something I'm missing?
4. HTTP status code evaluation added to all API requests. It still evaluates the response for safety, since the hosting provider claims that the API behavior might change in the future.
5. Code removed as `_myloc_extract_txt_records` is not used anymore.
6. I don't seem to have permission to reopen issue #5193, can you do that or shall I open a new issue?
7. Minor: `sed` call removed as part of  `_myloc_extract_txt_records`. I didn't change the unprefixed variables, probably safer if changed to local variables instead?

neilpang:
1. `dns_myloc_rm`: the "Getting records" GET is dead code -- `_myloc_get_zone` already GET the identical URL and required 200, and `$response` is never read. Dropping it removes exactly the extra API request you wanted to avoid in 3.
2. `dns_myloc_rm`: a failed DELETE logs `_info "Failed to delete record"` and still returns 0, so acme.sh reports the cleanup as successful. Use `_err` and `return 1`.
3. Re. 3: the case you're missing is a delegated challenge zone -- `_acme-challenge.example.com. NS ns1.myloc.de.` hosted as its own zone at myLoc. `${1#_acme-challenge.}` drops the only candidate that can match, so the walk ends in "Get zone failed for all candidates". Starting from `$1` costs one GET per challenge and covers it.
4. Re. 7: `local` is a bashism -- hooks are sourced by whatever shell runs acme.sh (dash, busybox ash, BSD sh), and ShellCheck runs the sh dialect (SC3043). Prefix instead: `myloc`, `token`, `zone`, `status`, `record` -> `_myloc_*`.
5. #5193 is open again.

flesniak:
PR updated:
1. dead code removed
2. `dns_myloc_rm` prints error properly and returns 1 now
3. I checked delegated challenge zones as well and I still can't see how removing a `_acme-challenge.` prefix could lead to the zone not being found, i.e. a record `_acme-challenge.my-own-delegate-zone.com` to be added/removed from this delegate zone still allows the prefix to be removed. The NS records on the original domain won't be touched by acme.sh since it does not have permission to do so anyway? Anyway, to move this further, the prefix removal has been removed.
4. All local variables are prefixed

neilpang:
1. `_myloc_get_zone`: the rename dropped the `$` -- `_debug "Get zone trying _myloc_zone"` and `_debug "Get zone success for _myloc_zone"` now print the literal name. Those are the two lines anyone debugging the walk needs.
2. DNS.yml on a7e271b was cancelled: FreeBSD hung 6h in the vmactions "Installing dependencies" step and hit the timeout, taking the rest of the matrix with it. Infra, not your code -- but the last green run (2ce91dc) predates dropping the `_acme-challenge.` strip, so push the fix from 1 and let it re-run.

Nothing else outstanding; merging once that run is green.

flesniak:
Good catch with the variables in _debug, I missed that skimming through the code. Fix pushed, let's see if CI succeeds.

PR #5215, opened 2024-07-20
[cron] sleep random seconds (<59), if not interactive or forced

oittaa:
https://github.com/acmesh-official/acme.sh/pull/944#issuecomment-707255200

A Let's Encrypt employee said in the comments "we do see peaks at the beginning of minutes and even seconds; the finer-grained time randomization, the better."

This adds a random amount of sleep seconds before beginning the cron job. I considered reading from `/dev/urandom` and so on, but we aren't doing anything security critical here so I thought that just using the process number modulo 59 (the largest prime <= 60) should give decent variability across the systems. The starting hour and minute are already randomized during the installation.


user8446:
Unfortunately this doesn't work on OpenWrt which is ash and not bash.

Also, a sleep of 1-5 or 55-59 would probably defeat the purpose due to being too close to the peaks at the beginning of each second.

So, playing around with it what about a random number between 6 and 54 seconds?
```

random_sec=$((( $RANDOM % 49 ) +5 ))
echo "Sleeping for $random_sec seconds."
sleep $random_sec
```

oittaa:
Is that test broken or what happened there?

```
/home/runner/work/_temp/1c8ca25f-9a95-441c-90d5-270c7ee76975.sh: line 1: docker-compose: command not found
Error: Process completed with exit code 127.
```

https://github.com/acmesh-official/acme.sh/actions/runs/10019783975/job/28809783408?pr=5215#step:4:18

oittaa:
> ```
> random_sec=$(( $RANDOM % 55 +6 ))
> echo Sleeping for $random_sec seconds.
> sleep $random_sec
> ```

And $RANDOM isn't POSIX standard. You can test that with `shellcheck` or something similar.

```sh
#!/bin/sh

echo $RANDOM
```

```
$ shellcheck myscript
 
[Line 3:](javascript:setPosition(3, 6))
echo $RANDOM
     ^-- [SC3028](https://www.shellcheck.net/wiki/SC3028) (warning): In POSIX sh, RANDOM is undefined.
```

Compared to
```sh
#!/bin/sh

echo $$
```

```
$ shellcheck myscript2
No issues detected!
```

user8446:
Thank you for the info.

Since the PID won't change but on reboot, what would be the solution the make it random like $RANDOM?

```
root@OpenWrt:~# echo $$
17372
root@OpenWrt:~# echo $$
17372
```

```
root@OpenWrt:~# echo $RANDOM
23144
root@OpenWrt:~# echo $RANDOM
31538
```

vmmello:
Check how it's done here https://github.com/acmesh-official/acme.sh/blob/dev/acme.sh#L6045

oittaa:
@user8446 Of course you will get the same process within the same shell session. The cron job gets a new process id every time. Google how POSIX process IDs work if you want to know more.

```sh

$ dash -c 'echo $$'
3317086
$ dash -c 'echo $$'
3317116

```

@vmmello I literally referenced that in the first comment. Imagine a hypothetical `random_second=$(_math $_t % 60)` and what would it be?

It would be zero always since the cron job starts when the minute changes.

oittaa:
> Is that test broken or what happened there?
> 
> ```
> /home/runner/work/_temp/1c8ca25f-9a95-441c-90d5-270c7ee76975.sh: line 1: docker-compose: command not found
> Error: Process completed with exit code 127.
> ```
> 
> https://github.com/acmesh-official/acme.sh/actions/runs/10019783975/job/28809783408?pr=5215#step:4:18

Seems like `docker-compose` was removed from the runners and you should use `docker compose` https://github.com/actions/runner-images/issues/9692

oittaa:
The staging tests seem to be broken at the moment, but the all the other tests pass now.

PR #5218, opened 2024-07-22
2 Bugfixes of `dns_dynv6.sh`: 1. domains containing 'id' are now working as intended and 2. hostnames are now cast to lowercase on the fly

IIIPr0t0typ3III:
1. There was a grep for `id` to get the id of the provided zone. This was changed to find `"id":` to only get this field.
2. Since DNS is case insensitive and dynv6 rejects any records containing upper case letters a cast to lowercase was implemented. This fix is necessary to complete the unit tests / github actions required for DNS-API pull requests

### DIFF
--- a/dnsapi/dns_dynv6.sh
+++ b/dnsapi/dns_dynv6.sh
@@ -16,8 +16,8 @@ dynv6_api="https://dynv6.com/api/v2"
 # Please Read this guide first: https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide
 #Usage: dns_dynv6_add  _acme-challenge.www.domain.com  "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
 dns_dynv6_add() {
-  fulldomain=$1
-  txtvalue=$2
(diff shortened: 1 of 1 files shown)


IIIPr0t0typ3III:
The unit tests / github actions requested for a DNS-API run without errors.

dnsapi/dns_dynv6.sh
```diff
@@ -16,8 +16,8 @@ dynv6_api="https://dynv6.com/api/v2"
 # Please Read this guide first: https://github.com/Neilpang/acme.sh/wiki/DNS-API-Dev-Guide
 #Usage: dns_dynv6_add  _acme-challenge.www.domain.com  "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
 dns_dynv6_add() {
-  fulldomain=$1
-  txtvalue=$2
+  fulldomain=$(echo "$1" | tr '[:upper:]' '[:lower:]')
+  txtvalue="$2"
```
Neilpang, line 20: please use `_lower_case()`

IIIPr0t0typ3III: Sorry I overlooked the existence of that function but this is changed now.

Also thanks for the quick response!

PR #5223, opened 2024-07-24
Update teams notification workflow

Linkje:
Update teams notify script to support new notification setup using workflows instead of connectors.
https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/ 
 
Color settings have been dropped since adaptive-card has no support for these values.


Neilpang:
please fix the errors and modify the usage here:

Linkje:
Can I already update that page while this has been unmerged since there are changes?

PR #5228, opened 2024-07-27
Update bark.sh

tomo2403:
[Bark API v2](https://github.com/Finb/bark-server/blob/master/docs/API_V2.md#push) is now fully implemented. BARK_SOUND is now also optional, as intended by the API.


tomo2403:
[Wiki updated](https://github.com/acmesh-official/acme.sh/wiki/notify#16-set-notification-for-ios-bark) ✅

PR #5229, opened 2024-07-27
Update bark.sh

tomo2403:
[Bark API v2](https://github.com/Finb/bark-server/blob/master/docs/API_V2.md#push) is now fully implemented. BARK_SOUND is now also optional, as intended by the API.

Updated workflow for dockerhub to enable own custom images.

### DIFF
--- a/.github/workflows/dockerhub.yml
+++ b/.github/workflows/dockerhub.yml
@@ -51,7 +51,7 @@ jobs:
           echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
       - name: build and push the image
         run: |
-          DOCKER_IMAGE=neilpang/acme.sh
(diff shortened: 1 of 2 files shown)


tomo2403:
[Wiki](https://github.com/acmesh-official/acme.sh/wiki/notify#16-set-notification-for-ios-bark) is already updated

.github/workflows/dockerhub.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
       - name: build and push the image
         run: |
-          DOCKER_IMAGE=neilpang/acme.sh
+          DOCKER_IMAGE=${{ secrets.DOCKER_USERNAME }}/acme.sh
```
Neilpang, line 54: remove this change.

tomo2403: Sorry for that, reverted it.

Issue #4137, opened 2022-06-10
dns_nsd does not update serial number of the zone


AlexeyRudenko:
Dear colleagues,
Thank you for this excellent piece of software!! 
Unfortunately, it is not useful for NSD server, as it does not increment serial number in the zone file. As a result, the updated zone is not tranferred to the secondary server and the whole operation fails.
I checked the file
~/.acme.sh/dnsapi/dns_nsd.sh
It only adds/removes _acme-challenge TXT record to the zone file, and leaves its serial number intact.
Thank you in advance!

sec:
Hi,
I was also checking the script and saw that it lacks serial increment support, but every zonefile can be diffrent, if you ask me, I would create custom reload script and put something like this inside (assuming your serial is only 10 digit text):
```
sed -i '' -E "s/[0-9]{10}/`date +%y%m%d%H%M`/g" /usr/local/etc/nsd/your.zonefile
nsd-control reload
```
and use this script as `Nsd_Command`

AlexeyRudenko:
Thank you for the response! A nice patch! But not all zones contain exactly 10-digit zone number generated as %y%m%d%H%M. And my zone files in particular have verification codes of different services, which may contain 10 consecutive digits. I think it would be more reliable to analyze the SOA record and increase just the serial number. 
Unfortunately I am not an expert in Bash scripting and can not make this change by myself.
Thanks again!

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

PR #5241, opened 2024-08-06
Added zone-serial increment support for dns_nsd challenge

codedump:
As the title says -- inspired by #4137 and my own necessity I wrote a dirty patch to `./dnsapi/dns_nsd.sh` to update the serial number.

Essentially it uses `sed` to parse out the old number. This should work in most circumstances, my own zone files look like this, and this works:
```
...
@                       IN      SOA     ns.example.com. root.example.com. (                                                                                                          
                     2024080615         ; Serial                                                                                                                                       
                           7200         ; Refresh                                                                                                                                      
                            900         ; Retry                                                                                                                                        
                        2419200         ; Expire                                                                                                                                       
                           7200 )       ; Negative Cache TTL                                                                                                                           
;                                                          
...
```

In a nutshell, the parsing algorithm goes like this:
  - look for the `IN SOA` line
  - extract everything until `)`
  - remove comments (i.e. trailing ends from `;` onwards)
  - from the text between `(` and `)` take the 1st entry

This is fairly robust as long as the sysadmin doesn't go out of their way to screw things up. (What they could do to make this fail: insert "IN SOA" in a comment line, split IN and SOA on two different lines, ...generally use less stuff which will, of course, confuse a simple `sed` based parser).

Then another `sed`-call is used to replace the existing serial number by a version which is incremented by 1 (bash's `$[...]` math support). Then `mktemp` is used to create a temporary zonefile with the updated number, which is then `mv`'ed over the original `Nsd_ZoneFile`.

This was tested in a kind-of-an-ad-hoc "dry mode", i.e.:

```
export Nsd_ZoneFile=/tmp/example.com.zonefile
export Nsd_Command="echo hello world"
./acme.sh --server letsencrypt_test --dns dns_nsd --issue --domain example.com"
```

Then I manually verified that the `/tmp/example.com.zonefile` had a properly updated serial number. (The actual zone update / DNS challenge will fail because I'm developing this on a laptop behind a NAT, not on an internet-facing machine with access to a DNS server.)

Hope this helps!

Cheers

### DIFF
--- a/dnsapi/dns_nsd.sh
+++ b/dnsapi/dns_nsd.sh
@@ -9,6 +9,18 @@ Options:
 Issues: github.com/acmesh-official/acme.sh/issues/2245
 '
 
+# args: zonefile
+_local_nsd_get_serial()
+{
+    local _zone_file="$1"
+    cat "$_zone_file" | \
+	sed -n '/IN[ \t]*SOA.*/,/[)]/p' | \
+	sed 's/\([^;]\);\(.*\)/\1/g' | \
+	sed -z 's/\n//g' | \
+	sed 's/\([^(]*\)[(]\([^)]*\)[)]/\2/g' | \
+	sed 's/\([ \t]*\)\([0-9]*\)\(.*\)/\2/g'
+}
+
 # args: fulldomain txtvalue
 dns_nsd_add() {
   fulldomain=$1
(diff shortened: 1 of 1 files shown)


neilpang:
Direction is fine (bump SOA serial so secondaries pick it up via NOTIFY/IXFR), but there are POSIX/portability blockers -- this file is `#!/usr/bin/env sh`, must run under dash/BSD/Solaris:

1. `local` -- not POSIX; drop it, use the `_nsd_` prefix on vars/helpers to avoid collisions.
2. `$[ ... ]` -- bash-only; use `_math "$serial + 1"`.
3. `sed -z` -- GNU-only, breaks on BSD/macOS/Solaris.
4. `rm -rf` -> `rm -f`; use `_mktemp`; drop the useless `cat | sed`.

Two bugs:

5. `sed "s/$zone_serial/.../"` is unanchored -- replaces the first match of that number anywhere (a TTL, an IP octet), not necessarily the SOA serial. Anchor to the SOA.
6. `dns_nsd_rm` never bumps the serial -- asymmetric with `add`, so removals won't propagate. Factor the bump into one helper both call.

Also rebase -- head is your `master`, base moved since 2024.

PR #5245, opened 2024-08-09
RouterOS - make deploy more resilient

nathanejohnson:
Newer releases of RouterOS automatically delete the cert and key files from disk when importing the certs, and that means the existing deploy will fail when it tries to explicitly remove those files.  This still attempts to remove the files but catches the error and moves on instead of bombing like before.

Similarly, if the deploy had failed before but the script was created, subsequent deploys would fail because the script already existed.  This first attempts to remove the script if it exists, and then creates the script.

This has been tested on RouterOS 7.15.3 , but I assume this should be backwards compatible, and it's currently broken on later ROS versions for reasons mentioned above.


cngarrison:
I haven't tested on my routers, but code change looks good. 👍

nathanejohnson:
@cngarrison I don't have anything running 6.x unfortunately, but the two 7.x devices I tried it on worked.

PR #5259, opened 2024-08-21
Update dns_miab.sh

lifeboy:
The MIAB API requires that the txtvlaue to a TXT record includes the "value=" and "ttl=" components as part of the TXT record when adding a new record.


lifeboy:
The change is ready for merging.  Anything else I need to do?

lifeboy:
Should have first create PR to dev.  Apologies.

PR #5262, opened 2024-08-23
Resolve the issue of unable to issue certificates in SAN and CN with the same wildcard domain

ldlb9527:
fix: #5264


ldlb9527:
fix: #5264

Issue #5264, opened 2024-08-23
Unable to issue certificates using the same wildcard domain in both SAN and CN


ldlb9527:
The reproduction process is as follows:
1. Use the following command to issue a certificate
   `./acme1.sh --sign-csr --csr ./domaint.csr --key-file ./private.key --dns dns_dp --home ./  --debug 2`
***
2. When the CN of CSR is c.ldlb.site and the SAN is a.ldlb.site,b.ldlb.site,c.ldlb.site
```shell
v3.0.8
[Fri Aug 23 14:28:23     2024] Running cmd: signcsr
[Fri Aug 23 14:28:23     2024] _csrsubj='c.ldlb.site'
[Fri Aug 23 14:28:23     2024] _csrsubj='c.ldlb.site'
[Fri Aug 23 14:28:23     2024] _dnsAltnames='DNS:a.ldlb.site,DNS:b.ldlb.site,DNS:c.ldlb.site'
[Fri Aug 23 14:28:24     2024] AltNames contains subject
[Fri Aug 23 14:28:24     2024] _excapedAlgnames='DNS:a.ldlb.site,DNS:b.ldlb.site,DNS:c.ldlb.site'
[Fri Aug 23 14:28:24     2024] _escapedSubject='c.ldlb.site'
[Fri Aug 23 14:28:24     2024] _dnsAltnames='DNS:a.ldlb.site,DNS:b.ldlb.site'
[Fri Aug 23 14:28:24     2024] _csrdomainlist='a.ldlb.site,b.ldlb.site'
```
This can issue certificates normally.
***
3. When the CN of CSR is *.c.ldlb.site and the SAN is *.a.ldlb.site,*.b.ldlb.site,*.c.ldlb.site
```shell
v3.0.8
v3.0.8
[Fri Aug 23 13:53:53     2024] Running cmd: signcsr
[Fri Aug 23 13:53:53     2024] _csrsubj='*.c.ldlb.site'
[Fri Aug 23 13:53:53     2024] _csrsubj='*.c.ldlb.site'
[Fri Aug 23 13:53:54     2024] _dnsAltnames='DNS:*.a.ldlb.site,DNS:*.b.ldlb.site,DNS:*.c.ldlb.site'
[Fri Aug 23 13:53:54     2024] AltNames doesn't contain subject
[Fri Aug 23 13:53:54     2024] _csrdomainlist='*.a.ldlb.site,*.b.ldlb.site,*.c.ldlb.site'
```
[...]

PR #5265, opened 2024-08-23
Resolve the issue of unable to issue certificates in SAN and CN with the same wildcard domain

ldlb9527:
fix: #5264

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -464,6 +464,12 @@ _contains() {
   echo "$_str" | grep -- "$_sub" >/dev/null 2>&1
 }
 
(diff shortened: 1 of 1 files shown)


neilpang:
Thanks. This is already fixed in `dev`: `_readSubjectAltNamesFromCSR` now escapes the wildcard `*` (via `tr '*' '#'`) before the match, so a wildcard CN that also appears in the SAN list is handled correctly. #5264 is closed.

(Also note this PR changes `acme.sh`'s file mode from `100755` to `100644`, which would drop the executable bit.)

Closing as superseded.

PR #5224, opened 2024-07-24
Add optional cmd line parameter NSUPDATE_OPT to dns_nsupdate api

gmanic:
To enable TCP communication with nameserver (-v flag of nsupdate) an additional parameter NSUPDATE_OPT was added.
It will be saved/read from config.
Documentation has not been updated, as WIKI is not in forked repo; hence, documentation update makes no sense at this point in time.

This is not a new dns_api.

PR #5285, opened 2024-09-16
URGENT: fix dns_nsupdate when NSUPDATE_OPT is empty

fraenki:
PR #5224 introduced a new parameter `$NSUPDATE_OPT` to `dns_nsupdate`. However, now this DNSAPI is broken, if the new parameter is empty/unset:

```
[Mon Sep 16 15:48:33 CEST 2024] Adding TXT value: XXX for domain: _acme-challenge.example.com
[Mon Sep 16 15:48:33 CEST 2024] adding _acme-challenge.example.com. 60 in txt "XXX"
could not open '': file not found
[Mon Sep 16 15:48:33 CEST 2024] error updating domain
[Mon Sep 16 15:48:33 CEST 2024] Error adding TXT record to domain: _acme-challenge.example.com
[Mon Sep 16 15:48:33 CEST 2024] Please check log file for more details: /var/log/acme.sh/acme.log
```

**This is a major issue, because it will affect almost ALL users of `dns_nsupdate`.**

The root cause is easily demonstrated by the following code snippet:

```
# does not work
$ NSUPDATE_OPT="" nsupdate -k /tmp/foo -d "${NSUPDATE_OPT}"
could not open '': file not found

# works as expected
$ NSUPDATE_OPT="" nsupdate -k /tmp/foo -d $NSUPDATE_OPT
Creating key...
...
```

Note that nsupdate interprets an empty string as a file/input. So when passing an empty parameter to nsupdate, it must not use double quotes.

### DIFF
--- a/dnsapi/dns_nsupdate.sh
+++ b/dnsapi/dns_nsupdate.sh
@@ -39,13 +39,13 @@ dns_nsupdate_add() {
   [ -n "$DEBUG" ] && [ "$DEBUG" -ge "$DEBUG_LEVEL_1" ] && nsdebug="-d"
   [ -n "$DEBUG" ] && [ "$DEBUG" -ge "$DEBUG_LEVEL_2" ] && nsdebug="-D"
   if [ -z "${NSUPDATE_ZONE}" ]; then
-    nsupdate -k "${NSUPDATE_KEY}" $nsdebug "${NSUPDATE_OPT}" <<EOF
+    nsupdate -k "${NSUPDATE_KEY}" $nsdebug $NSUPDATE_OPT <<EOF
 server ${NSUPDATE_SERVER}  ${NSUPDATE_SERVER_PORT}
 update add ${fulldomain}. 60 in txt "${txtvalue}"
 send
 EOF
   else
-    nsupdate -k "${NSUPDATE_KEY}" $nsdebug "${NSUPDATE_OPT}" <<EOF
+    nsupdate -k "${NSUPDATE_KEY}" $nsdebug $NSUPDATE_OPT <<EOF
 server ${NSUPDATE_SERVER}  ${NSUPDATE_SERVER_PORT}
 zone ${NSUPDATE_ZONE}.
 update add ${fulldomain}. 60 in txt "${txtvalue}"
@@ -77,13 +77,13 @@ dns_nsupdate_rm() {
   [ -n "$DEBUG" ] && [ "$DEBUG" -ge "$DEBUG_LEVEL_1" ] && nsdebug="-d"
   [ -n "$DEBUG" ] && [ "$DEBUG" -ge "$DEBUG_LEVEL_2" ] && nsdebug="-D"
   if [ -z "${NSUPDATE_ZONE}" ]; then
-    nsupdate -k "${NSUPDATE_KEY}" $nsdebug "${NSUPDATE_OPT}" <<EOF
+    nsupdate -k "${NSUPDATE_KEY}" $nsdebug $NSUPDATE_OPT <<EOF
 server ${NSUPDATE_SERVER}  ${NSUPDATE_SERVER_PORT}
 update delete ${fulldomain}. txt
 send
 EOF
   else
-    nsupdate -k "${NSUPDATE_KEY}" $nsdebug "${NSUPDATE_OPT}" <<EOF
+    nsupdate -k "${NSUPDATE_KEY}" $nsdebug $NSUPDATE_OPT <<EOF
 server ${NSUPDATE_SERVER}  ${NSUPDATE_SERVER_PORT}
 zone ${NSUPDATE_ZONE}.
 update delete ${fulldomain}. txt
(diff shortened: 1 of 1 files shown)


fraenki:
Yes, the `shellcheck` test is complaining...

```
In dnsapi/dns_nsupdate.sh line 42:
    nsupdate -k "${NSUPDATE_KEY}" $nsdebug $NSUPDATE_OPT <<EOF
                                           ^-----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

But this is desired, it's necessary to fix the bug (and the old `$nsdebug` also does not passt this test).

It would be possible to pass the `shellcheck` test by adding more code. I'll do this is requested by the maintainers.

fraenki:
@Neilpang Please review. If you accept this bugfix, please also issue a new minor release. Thank you!

gmanic:
@fraenki - your fix works fine.

My own full checks worked out fine without double quotes.
Unfortunately, I just added the double quotes due to the `shellcheck` failure. My bad, sorry for that.

Thanks for your fix.

Neilpang:
fix the shellcheck errors

gmanic:
@fraenki 
How about creating a variable with the full set of command line parameters once per function [dns_nsupdate_add, dns_nsupdate_rm] before `if [ -z "${NSUPDATE_ZONE}" ]; then` like this:

`nsupdate_parameters="-k ${NSUPDATE_KEY} ${nsdebug} ${NSUPDATE_OPT}"`

and using this command instead:

`nsupdate "${nsupdate_parameters}" <<EOF`

Should fix the shellcheck error - except the last error regarding usage of $? instead of command...

fraenki:
> How about creating a variable with the full set of command line parameters once per function

I've tried this, but unfortunately it does not work: nsupdate requires double quotes around the filename when using the `-k` option, otherwise it will report an error:

```
could not read key from  "/etc/acme.sh/configs/profile_nsupdate_example/hook.cnf"  .{private,key}: file not found
```

gmanic:
I also tried some alternatives. This one seems to run, taking the correct key file and without `shellcheck` errors:

```shell
  nsupdate_cmd="nsupdate -k ${NSUPDATE_KEY} ${nsdebug} ${NSUPDATE_OPT}"
  if [ -z "${NSUPDATE_ZONE}" ]; then
    ${nsupdate_cmd} <<EOF
``` 
Unfortunately, I hit auth limits on my nameserver to fully test deployment. 🤦‍♂️

For the error check at the end (another `shellcheck` error), I assigned 
`RC=$?`
after the (add and rm) run of nsupdate (once per function, after the full if-block `if [ -z "${NSUPDATE_ZONE}" ]; then`) and extended the error check
`if [ "${RC}" -ne 0 ]; then`

No `shellcheck` errors on shellcheck.net :+1:

gmanic:
In my forked repo, `shellcheck`-test works, but DNS Docker always complains some error I just don't understand...
My [Changes to Master](https://github.com/acmesh-official/acme.sh/compare/master...gmanic:acme.sh:master)

fraenki:
> This one seems to run, taking the correct key file and without shellcheck errors:

Unfortunately, it results in an error, because nsupdate expects the key file to be wrapped in double quotes:

```
could not read key from /tmp/keyfile.{private,key}: file not found
```

gmanic:
Maybe the filename without `.key` extension is the cause of nsupdate trying to add `{.private,.key}`.
My local run (extended with the line "full nsupdate..." - local temporary hack to see the actual commandline) runs without error (domain and txt-record redacted).

```
[Tue Sep 17 12:26:26 PM CEST 2024] adding _acme-challenge.example.com. 60 in txt "LrgB5T1AMk4ah_6psolKcf10kavqtnkfxxxxxxx"
[Tue Sep 17 12:26:26 PM CEST 2024] full nsupdate -k /root/.acme.sh/_acme-challenge.example.com.key -D -v
[Tue Sep 17 12:26:26 PM CEST 2024] Return code of nsupdate: 0
```

fraenki:
> fix the shellcheck errors

@Neilpang Fixed. Please review/merge.

dnsapi/dns_nsupdate.sh
```diff
@@ -39,13 +39,15 @@ dns_nsupdate_add() {
   [ -n "$DEBUG" ] && [ "$DEBUG" -ge "$DEBUG_LEVEL_1" ] && nsdebug="-d"
   [ -n "$DEBUG" ] && [ "$DEBUG" -ge "$DEBUG_LEVEL_2" ] && nsdebug="-D"
   if [ -z "${NSUPDATE_ZONE}" ]; then
-    nsupdate -k "${NSUPDATE_KEY}" $nsdebug "${NSUPDATE_OPT}" <<EOF
+    #shellcheck disable=all
```
Neilpang, line 42: don't disable all,  why not just disable the specific number?

fraenki: @Neilpang Fixed in https://github.com/acmesh-official/acme.sh/commit/9ecd84080b5f5160626de7c886b30831144efbc9.

fraenki:
@Neilpang Would it be possible to get a new version of acme.sh that includes this fix? Thanks!

Neilpang:
@fraenki   here you go:
https://github.com/acmesh-official/acme.sh/releases/tag/3.0.9

PR #5309, opened 2024-09-28
on OpenBSD, add libiconv

jschauma:
acme.sh dnsapi/dns_edgedns.sh invokes iconv(1) which is not provided by OpenBSD in the base system. Adding the libiconv package provides this tool.

This should also help address acmesh-official/acme.sh#4350, albeit indirectly: it looks like that PR cannot be merged because the OpenBSD test fails despite the PR not actually changing anything having to do with iconv. That is, I suspect that the test for OpenBSD failed (or would have failed, had it been run?) prior to that PR being issued.

(This was previously submitted as
https://github.com/acmesh-official/acmetest/pull/28 )


Neilpang:
send to dev branch.

jschauma:
Done.

jschauma:
neil ***@***.***> wrote:
> Merged #5309 into dev.

Thanks!

PR #5334, opened 2024-10-23
added SLEEP_TIME

aathan:
simply adds a variable to configure the SLEEP_TIME between loops

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2205,7 +2205,7 @@ _send_signed_request() {
     _debug2 nonce "$nonce"
     if [ -z "$nonce" ]; then
       _info "Could not get nonce, let's try again."
-      _sleep 2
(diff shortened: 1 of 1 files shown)


Neilpang:
why?

aathan:
Because zerossl was acting strange and I was not able to get a verification done without increasing the delay between polls OR increasing the number of attempts.  While either would increase the total time before failure is declared, polling the service less may help avoid rate limiting imposed by their ingress. It's an easy add, and may help some users.

E.g. $ SLEEP_TIME=60 acme.sh ...

It may also be helpful when debugging acme.sh itself.

Also, why *not*?

jpbion:
Does pull request 6244 (done because GoDaddy had been publishing after quite a delay for me a couple of weeks ago) handle this situation?

PR #6083, opened 2024-11-06
added message thread id for notification to the chat topic

lacosteque:
Hi!
Added the ability to send a notification to the group topic. It was necessary for me, so I will do it. 
If you do not specify the value of the variable, then the notification is sent by default to the chat. 
So the backward compatibility here is 100%

It is only necessary to update the wiki that a new variable $TELEGRAM_MESSAGE_THREAD_ID is added and it is optional.


lacosteque:
ср, 6 нояб. 2024 г. в 15:02, github-actions[bot] ***@***.***>:

Neilpang:
if you need to add a new variable, see how it works for the other variables,  eg: `TELEGRAM_BOT_CHATID`

PR #6094, opened 2024-11-11
StrongSwan deploy hook: swanctl support

sergiustheblack:
acme.sh should support modern StrongSwan's swanctl tool.
The issue is that both legacy ipsec tool and swanctl can co-exist on the machine. So deploy hook maintains both tools independently and sync certificates into each config dir if it exists.

Notable changes:

- ipsec tool selector is moved to search in $PATH which potentially may break wierd installations.
- swanctl doesn't have an analogue to `ipsec --confdir` option, so I have to search for typical conf dir location (tested on Debian 12, Arch Linux, Centos Stream, FreeBSD 14)


Neilpang:
fix the format errors please

sergiustheblack:
> fix the format errors please

fixed, thanks.

PR #6117, opened 2024-11-27
add dns_api for Technitum dns server

qupfer:
This add dns_api support for the Technitium DNS Server (https://technitium.com/dns/)

The Server API match automatically a new record to the "longest matching zone" --> no need to split the domain inside the API script.


qupfer:
Read Dev Guide and github action test was successful

PR #6126, opened 2024-12-03
add environment variable for notify/telegram.sh

SikroYang:
Since the server in mainland China cannot connect to Telegram, add the TELEGRAM_BOT_URLBASE environment variable Please add the proxy address you set. If this variable is not added, it will be the default value (https://api.telegram.org)


SikroYang:
Hi,I have added wiki in [notify](https://github.com/acmesh-official/acme.sh/wiki/notify)

PR #6131, opened 2024-12-05
Fix parameter errors in west.cn

Gucan:
West.cn requires the domain parameter to only specify the domain subject


Neilpang:
测试:

PR #6174, opened 2024-12-28
DNS API: fix structural info for new providers

stokito:
To parse it automatically it needs for an order.


Zennn11zen11:
https://app.binance.com/uni-qr/cpro/Square-Creator-b08e6bfc5a87?l=en&r=818739537&uc=app_square_share_link&us=more

PR #6205, opened 2025-01-18
Add Exponential Backoff for Route53 Throttling in dns_aws.sh

foxdalas:
Hello it is small improvement for Route53 (dns_aws)

**Description:**
- Introduces a simple retry mechanism in aws_rest() with exponential backoff for AWS Route53 throttling errors (Throttling, RequestThrottled, RateLimit, or Rate exceeded).
- No other changes have been made, and the existing logic is preserved.
- Ensures that acme.sh can gracefully handle rate-limited requests to Route53 instead of failing immediately.

**Why:**
- When Route53 is under heavy load, AWS may respond with a throttling error. Without retries, the request fails and the certificate issuance is disrupted.
- This backoff approach retries the request up to five times, doubling the wait each time, allowing enough time for Route53 to recover from throttling.

**How:**
- Added a while loop in `aws_rest()` with a maximum of five attempts (**max_retries=5**).
- On each throttling error, we sleep for exponentially longer intervals (**_sleep_time = base_sleep * 2^(attempt-1)**).
- If throttling persists beyond the maximum attempts, we return an error to maintain backward compatibility.

**Testing:**
- Manually tested by simulating throttling responses. Verified that the script retries and succeeds once AWS is ready to accept requests again.

Thank you for reviewing!

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -323,7 +323,6 @@ aws_rest() {
   _debug2 CredentialScope "$CredentialScope"
 
   StringToSign="$Algorithm\n$RequestDate\n$CredentialScope\n$HashedCanonicalRequest"
-
   _debug2 StringToSign "$StringToSign"
 
   kSecret="AWS4$AWS_SECRET_ACCESS_KEY"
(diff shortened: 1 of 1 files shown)


neilpang:
1. `^` in shell arithmetic is XOR, not exponentiation -- `_math 5 "*" 2 "^" $((attempt-1))` evaluates as `10 XOR (attempt-1)` (11, 8, 9, 14... seconds), so the backoff is neither exponential nor even increasing. POSIX sh has no `**` either; double the value each round instead: `sleep_time=$(_math "$sleep_time" \* 2)`.
2. Once actually exponential, 10 retries from a 5s base sleep up to ~42min on the last round (~85min total) inside a single renewal -- far too long to block in-process. Route53 throttling clears in seconds; cap at ~3-5 retries with a max sleep around 30-60s.
3. The retry path is not a failure yet -- log it with `_info`/`_debug`, not `_err`.
4. Any dns_aws change needs a green DNS.yml run in your fork before merge: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test.

PR #6227, opened 2025-02-10
Fix _check_dns_entries for suggested apex domain substitutions for --domain-alias

zeezooz:
Current checks don't support @.domain.tld or .domain.tld, which are suggested to be used instead of the domain apex with --domain-alias at https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode#7-challenge-alias-or-domain-alias
This fix removes "@." or "." at the beginning of the domain, so the correct domain will be checked.


Neilpang:
docs updated, we should not support this kind of format.

PR #6236, opened 2025-02-17
Deploy certificate to FortiGate firewall using API

NotAnotherHelloWorld:
## This PR adds a new deploy hook for FortiGate firewalls. It:

- Uploads the server and CA certificate to FortiGate via API.
- Sets the uploaded certificate as the active web GUI certificate.
- Removes previously deployed certificate (name stored in acme domain conf).
- Does not remove CA certificates (in case of external dependencies).

## Why is this needed?
FortiGate's native ACME integration:

- Does not support wildcard certificates.
- Does not work with custom management ports (e.g., DNAT web traffic).
- Does not support domain validation.

## Usage
`acme.sh --deploy -d example.com --deploy-hook fortigate --insecure`

> --insecure required in case of invalid pre-existing certificate

## API Requirements
Create a REST API admin with an admin profile assigned the following permissions:

- System → Configuration [Read/Write] (for setting the active GUI certificate)
- VPN → [Read/Write] (for uploading and removing certificates)

Set "Scope" to "global" if using VDOMs. Be aware that the certificate will be visible to all VDOMs.

## Tested On:

- FortiGate 61F v7.4.7 (without VDOMs)
- FortiGate 100F v7.4.6 (with VDOMs)

### DIFF
--- a/deploy/fortigate.sh
+++ b/deploy/fortigate.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env sh
+# Script to deploy a certificate to FortiGate via API and set it as the current web GUI certificate.
+#
+# FortiGate's native ACME integration does not support wildcard certificates or domain validation,
+# and is not supported if you have a custom management web port (eg. DNAT web traffic).
+#
+# REQUIRED:
+#     export FGT_HOST="fortigate_hostname-or-ip"
+#     export FGT_TOKEN="fortigate_api_token"
+#
+# OPTIONAL:
+#     export FGT_PORT="10443"             # Custom HTTPS port (defaults to 443 if not set)
+#
+# Run `acme.sh --deploy -d example.com --deploy-hook fortigate --insecure` to use this script.
+# `--insecure` is required on first run if not already using a valid SSL certificate on firewall.
+
+# Function to parse response
+parse_response() {
+  response="$1"
+  func="$2"
+  status=$(echo "$response" | grep -o '"status":[ ]*"[^"]*"' | sed 's/"status":[ ]*"\([^"]*\)"/\1/')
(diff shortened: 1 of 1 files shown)


NotAnotherHelloWorld:
Sorry about the duplicate PRs–Git and I are a work in progress.

neilpang:
1. Prefix all private functions and working variables with `_fortigate_` -- `parse_response`, `deployer`, `upload_ca_cert`, `set_active_web_cert`, `cleanup_previous_certificate`, `response`, `status`, `payload`, `url`, `var`, etc. Deploy hooks are sourced into the same shell as acme.sh and other hooks; generic names collide.
2. The cert name appends only the date, so a second deploy on the same day (quite common when testing) hits an already-exists name: `FGT_LAST_CERT` equals the new name, the old cert is not deleted, and the import fails. Append a time component, or delete the existing same-name cert before import.
3. Add the usage entry on the [deployhooks wiki](https://github.com/acmesh-official/acme.sh/wiki/deployhooks).

NotAnotherHelloWorld:
Thanks for the review. I have prefixed private helper functions and variables, added time component to cert name and updated the deployhooks wiki.

neilpang:
One remaining point: `grep -o` is not POSIX -- Solaris and some BusyBox greps don't have it. Use the `_egrep_o` helper from acme.sh instead:

```sh
_fortigate_status=$(echo "$_fortigate_response" | _egrep_o '"status":[ ]*"[^"]*"' | cut -d '"' -f 4)
```

Everything else from the previous review is addressed; wiki entry confirmed. Good to merge after this.

NotAnotherHelloWorld:
Thanks, learned something new. Replaced `grep -o` for POSIX.

PR #6241, opened 2025-02-21
Multideploy: Deploy to multiple hooks of the same type

tomo2403:
This hook allows the user to deploy certificates to multiple services at once. It can store configurations for numerous services, even for the same hook.

## Example
You have three Docker containers and a Synology NAS (DSM). However, using the `docker` and `synology_dsm` hooks, you can only deploy to one Docker container with renewals. This problem is solved with Multideploy.

## Sample config file
The file can be named `multideploy.yml` or `multideploy.yaml`. It is stored in the domain folder. `$DOMAIN_DIR` is a variable that allows deploying certificated to a dir named after the certificate's domain to make changes easier. 
```yaml
version: 1.0

configs:
  - name: "default"
    services:
      - "webserver"
      - "webserver2"

services:
  - name: "webserver"
    hook: "docker"
    environment:
      - DEPLOY_DOCKER_CONTAINER_LABEL: "sh.acme.autoload.domain=example.com"
      - DEPLOY_DOCKER_CONTAINER_KEY_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/key.pem"
      - DEPLOY_DOCKER_CONTAINER_CERT_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/cert.pem"
      - DEPLOY_DOCKER_CONTAINER_CA_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/ca.pem"
      - DEPLOY_DOCKER_CONTAINER_FULLCHAIN_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/full.pem"
      - DEPLOY_DOCKER_CONTAINER_RELOAD_CMD: "service nginx force-reload"
  - name: "webserver2"
    hook: "docker"
    environment:
      - DEPLOY_DOCKER_CONTAINER_LABEL: "sh.acme.autoload.domain=example.com"
      - DEPLOY_DOCKER_CONTAINER_KEY_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/key.pem"
      - DEPLOY_DOCKER_CONTAINER_CERT_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/cert.pem"
      - DEPLOY_DOCKER_CONTAINER_CA_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/ca.pem"
      - DEPLOY_DOCKER_CONTAINER_FULLCHAIN_FILE: "/etc/nginx/ssl/$DOMAIN_DIR/full.pem"
      - DEPLOY_DOCKER_CONTAINER_RELOAD_CMD: "service nginx force-reload"
```

## Wiki
Please tell me if you will merge this first before I start writing a wiki entry for this. thx
https://github.com/acmesh-official/acme.sh/wiki/deployhooks#36-deploying-to-multiple-services-with-the-same-hooks

### DIFF
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -29,7 +29,7 @@ multideploy_deploy() {
   fi
   _debug2 "DOMAIN_DIR" "$DOMAIN_DIR"
 
-  _preprocess_deployfile "$DOMAIN_DIR/$MULTIDEPLOY_FILENAME"
+  _preprocess_deployfile "$DOMAIN_DIR/$MULTIDEPLOY_FILENAME" || return 1
 
   MULTIDEPLOY_CONFIG="${MULTIDEPLOY_CONFIG:-$(_getdeployconf MULTIDEPLOY_CONFIG)}"
   if [ -z "$MULTIDEPLOY_CONFIG" ]; then
@@ -47,7 +47,7 @@ multideploy_deploy() {
 
 # deploy_filepath
 _preprocess_deployfile() {
-  deploy_file="$1"
+  _deploy_file="$1"
 
   # Check if yq is installed
   if ! command -v yq >/dev/null 2>&1; then
@@ -56,9 +56,9 @@ _preprocess_deployfile() {
   fi
 
   # Check if deploy file exists and create a default template if not
-  if [ -f "$deploy_file" ]; then
+  if [ -f "$_deploy_file" ]; then
     _debug3 "Deploy file found."
-    _check_deployfile "$deploy_file" "$MULTIDEPLOY_CONFIG"
+    _check_deployfile "$_deploy_file" "$MULTIDEPLOY_CONFIG"
   else
     # TODO: Replace URL with wiki link
     _err "Deploy file not found. Go to https://CHANGE_URL_TO_WIKI to see how to create one."
@@ -66,48 +66,48 @@ _preprocess_deployfile() {
   fi
 }
 
-# deploy_filepath deploy_config
+# deploy_filepath _deploy_config
 _check_deployfile() {
-  deploy_file="$1"
-  deploy_config="$3"
+  _deploy_file="$1"
+  _deploy_config="$3"
 
   # Check version
-  deploy_file_version=$(yq '.version' "$deploy_file")
-  if [ "$MULTIDEPLOY_VERSION" != "$deploy_file_version" ]; then
-    _err "As of $PROJECT_NAME $VER, the deploy file needs version $MULTIDEPLOY_VERSION! Your current deploy file is of version $deploy_file_version."
+  _deploy_file_version=$(yq '.version' "$_deploy_file")
+  if [ "$MULTIDEPLOY_VERSION" != "$_deploy_file_version" ]; then
+    _err "As of $PROJECT_NAME $VER, the deploy file needs version $MULTIDEPLOY_VERSION! Your current deploy file is of version $_deploy_file_version."
     return 1
   fi
 
   # Check if config exists
-  if ! yq e ".configs[] | select(.name == \"$deploy_config\")" "$deploy_file" >/dev/null; then
-    _err "Config '$deploy_config' not found."
+  if ! yq e ".configs[] | select(.name == \"$_deploy_config\")" "$_deploy_file" >/dev/null; then
+    _err "Config '$_deploy_config' not found."
     return 1
   fi
 
   # Extract all services from config
-  services=$(_get_services_list "$deploy_file" "$deploy_config")
+  _services=$(_get_services_list "$_deploy_file" "$_deploy_config")
 
-  if [ -z "$services" ]; then
-    _err "Config '$deploy_config' does not have any services to deploy to."
+  if [ -z "$_services" ]; then
+    _err "Config '$_deploy_config' does not have any services to deploy to."
     return 1
   fi
 
   # Check if extracted services exist in services list
-  for service in $services; do
-    if ! yq e ".services[] | select(.name == \"$service\")" "$deploy_file" >/dev/null; then
-      _err "Service '$service' not found."
+  for _service in $_services; do
+    if ! yq e ".services[] | select(.name == \"$_service\")" "$_deploy_file" >/dev/null; then
+      _err "Service '$_service' not found."
       return 1
     fi
 
     # Check if service has hook
-    if ! yq e ".services[] | select(.name == \"$service\").hook" "$deploy_file" >/dev/null; then
-      _err "Service '$service' does not have a hook."
+    if ! yq e ".services[] | select(.name == \"$_service\").hook" "$_deploy_file" >/dev/null; then
+      _err "Service '$_service' does not have a hook."
       return 1
     fi
 
     # Check if service has environment
-    if ! yq e ".services[] | select(.name == \"$service\").environment" "$deploy_file" >/dev/null; then
-      _err "Service '$service' does not an environment."
+    if ! yq e ".services[] | select(.name == \"$_service\").environment" "$_deploy_file" >/dev/null; then
+      _err "Service '$_service' does not an environment."
       return 1
     fi
   done
@@ -115,26 +115,26 @@ _check_deployfile() {
 
 # deploy_filepath deploy_config
 _get_services_list() {
-  deploy_file="$1"
-  deploy_config="$2"
+  _deploy_file="$1"
+  _deploy_config="$2"
 
-  services=$(yq e ".configs[] | select(.name == \"$deploy_config\").services[]" "$deploy_file")
-  echo "$services"
+  _services=$(yq e ".configs[] | select(.name == \"$_deploy_config\").services[]" "$_deploy_file")
+  echo "$_services"
 }
 
 # deploy_filepath service_names
 _get_full_services_list() {
-  deploy_file="$1"
+  _deploy_file="$1"
   shift
-  service_names="$*"
+  _service_names="$*"
 
-  full_services=""
-  for service in $service_names; do
-    full_service=$(yq e ".services[] | select(.name == \"$service\")" "$deploy_file")
-    full_services="$full_services
-$full_service"
+  _full_services=""
+  for _service in $_service_names; do
+    _full_service=$(yq e ".services[] | select(.name == \"$_service\")" "$_deploy_file")
+    _full_services="$_full_services
+$_full_service"
   done
 
-  echo "$full_services"
+  echo "$_full_services"
 }
 
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env sh
+
+# MULTIDEPLOY_CONFIG="default"
+
+########  Public functions #####################
+
+MULTIDEPLOY_VERSION="1.0"
+MULTIDEPLOY_FILENAME="multideploy.yaml"
+
+# domain keyfile certfile cafile fullchain pfx
+multideploy_deploy() {
+    _cdomain="$1"
+    _ckey="$2"
+    _ccert="$3"
+    _cca="$4"
+    _cfullchain="$5"
+    _cpfx="$6"
+
+    _debug _cdomain "$_cdomain"
+    _debug _ckey "$_ckey"
+    _debug _ccert "$_ccert"
+    _debug _cca "$_cca"
+    _debug _cfullchain "$_cfullchain"
+    _debug _cpfx "$_cpfx"
+
+    DOMAIN_DIR=$_cdomain
+    if echo "$DOMAIN_PATH" | grep -q "$ECC_SUFFIX"; then
+        DOMAIN_DIR="$DOMAIN_DIR"_ecc
+    fi
+    _debug2 "DOMAIN_DIR" "$DOMAIN_DIR"
+
+    _preprocess_deployfile "$DOMAIN_DIR/$MULTIDEPLOY_FILENAME"
+
+    MULTIDEPLOY_CONFIG="${MULTIDEPLOY_CONFIG:-$(_getdeployconf MULTIDEPLOY_CONFIG)}"
+    if [ -z "$MULTIDEPLOY_CONFIG" ]; then
+        MULTIDEPLOY_CONFIG="default"
+        _info "MULTIDEPLOY_CONFIG is not set, so I will use 'default'."
+    else
+        _savedeployconf "MULTIDEPLOY_CONFIG" "$MULTIDEPLOY_CONFIG"
+        _debug2 "MULTIDEPLOY_CONFIG" "$MULTIDEPLOY_CONFIG"
+    fi
+
+    # TODO: Deploy to services
+}
+
+####################  Private functions below #####################
+
+# deploy_filepath
+_preprocess_deployfile() {
+    deploy_file="$1"
+
+    # Check if yq is installed
+    if ! command -v yq >/dev/null 2>&1; then
+        _err "yq is not installed! Please install yq and try again."
+        return 1
+    fi
+
+    # Check if deploy file exists and create a default template if not
+    if [ -f "$deploy_file" ]; then
+        _debug3 "Deploy file found."
+        _check_deployfile "$deploy_file" "$MULTIDEPLOY_CONFIG"
+    else
+        # TODO: Replace URL with wiki link
+        _err "Deploy file not found. Go to https://CHANGE_URL_TO_WIKI to see how to create one."
+        return 1
+    fi
+}
+
+# deploy_filepath deploy_config
+_check_deployfile() {
+    deploy_file="$1"
+    deploy_config="$3"
+
+    # Check version
+    deploy_file_version=$(yq '.version' "$deploy_file")
+    if [ "$MULTIDEPLOY_VERSION" != "$deploy_file_version" ]; then
+        _err "As of $PROJECT_NAME $VER, the deploy file needs version $MULTIDEPLOY_VERSION! Your current deploy file is of version $deploy_file_version."
+        return 1
+    fi
+
+    # Check if config exists
+    if ! yq e ".configs[] | select(.name == \"$deploy_config\")" "$deploy_file" >/dev/null; then
+         _err "Config '$deploy_config' not found."
+        return 1
+    fi
+
+    # Extract all services from config
+    services=$(yq e ".configs[] | select(.name == \"$deploy_config\").services[]" "$deploy_file")
+
+    if [ -z "$services" ]; then
+        _err "Config '$deploy_config' does not have any services to deploy to."
+        return 1
+    fi
+
+    # Check if extracted services exist in services list
+    for service in $services; do
+        if ! yq e ".services[] | select(.name == \"$service\")" "$deploy_file" >/dev/null; then
+            _err "Service '$service' not found."
+            return 1
+        fi
+
+        # Check if service has hook
+        if ! yq e ".services[] | select(.name == \"$service\").hook" "$deploy_file" >/dev/null; then
+            _err "Service '$service' does not have a hook."
+            return 1
+        fi
+
+        # Check if service has environment
+        if ! yq e ".services[] | select(.name == \"$service\").environment" "$deploy_file" >/dev/null; then
+            _err "Service '$service' does not an environment."
+            return 1
+        fi
+    done
+}
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -5,7 +5,8 @@
 ########  Public functions #####################
 
 MULTIDEPLOY_VERSION="1.0"
-MULTIDEPLOY_FILENAME="multideploy.yaml"
+MULTIDEPLOY_FILENAME="multideploy.yml"
+MULTIDEPLOY_FILENAME2="multideploy.yaml"
 
 # domain keyfile certfile cafile fullchain pfx
 multideploy_deploy() {
@@ -56,18 +57,30 @@ multideploy_deploy() {
 
 # deploy_filepath
 _preprocess_deployfile() {
-  _deploy_file="$1"
-
   # Check if yq is installed
   if ! command -v yq >/dev/null 2>&1; then
     _err "yq is not installed! Please install yq and try again."
     return 1
   fi
+  _debug3 "yq is installed."
 
-  # Check if deploy file exists and create a default template if not
-  if [ -f "$_deploy_file" ]; then
-    _debug3 "Deploy file found."
-    _check_deployfile "$_deploy_file" "$MULTIDEPLOY_CONFIG"
+  # Check if deploy file exists
+  for file in "$@"; do
+    _debug3 "Checking file" "$DOMAIN_PATH/$file"
+    if [ -f "$DOMAIN_PATH/$file" ]; then
+      _debug3 "File found"
+      if [ -n "$found_file" ]; then
+        _err "Multiple deploy files found. Please keep only one deploy file."
+        return 1
+      fi
+      found_file="$file"
+    else
+      _debug3 "File not found"
+    fi
+  done
+
+  if [ -n "$found_file" ]; then
+    _check_deployfile "$DOMAIN_PATH/$found_file" "$MULTIDEPLOY_CONFIG"
   else
     # TODO: Replace URL with wiki link
     _err "Deploy file not found. Go to https://CHANGE_URL_TO_WIKI to see how to create one."
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -85,7 +85,7 @@ _check_deployfile() {
   fi
 
   # Extract all services from config
-  services=$(yq e ".configs[] | select(.name == \"$deploy_config\").services[]" "$deploy_file")
+  services=$(_get_services_list "$deploy_file" "$deploy_config")
 
   if [ -z "$services" ]; then
     _err "Config '$deploy_config' does not have any services to deploy to."
@@ -112,3 +112,29 @@ _check_deployfile() {
     fi
   done
 }
+
+# deploy_filepath deploy_config
+_get_services_list() {
+  deploy_file="$1"
+  deploy_config="$2"
+
+  services=$(yq e ".configs[] | select(.name == \"$deploy_config\").services[]" "$deploy_file")
+  echo "$services"
+}
+
+# deploy_filepath service_names
+_get_full_services_list() {
+  deploy_file="$1"
+  shift
+  service_names="$*"
+
+  full_services=""
+  for service in $service_names; do
+    full_service=$(yq e ".services[] | select(.name == \"$service\")" "$deploy_file")
+    full_services="$full_services
+$full_service"
+  done
+
+  echo "$full_services"
+}
+
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -39,10 +39,13 @@ multideploy_deploy() {
     _debug2 "MULTIDEPLOY_CONFIG" "$MULTIDEPLOY_CONFIG"
   fi
 
+  OLDIFS=$IFS
+  file=$(_preprocess_deployfile "$MULTIDEPLOY_FILENAME" "$MULTIDEPLOY_FILENAME2") || return 1
+  _debug3 "File" "$file"
+
   # Deploy to services
-  _services=$(_get_services_list "$DOMAIN_DIR/$MULTIDEPLOY_FILENAME" "$MULTIDEPLOY_CONFIG")
-  _full_services=$(_get_full_services_list "$DOMAIN_DIR/$MULTIDEPLOY_FILENAME" "$_services")
-  _deploy_services "$DOMAIN_DIR/$MULTIDEPLOY_FILENAME" "$_full_services"
+  _services=$(_get_services_list "$file" "$MULTIDEPLOY_CONFIG")
+  _deploy_services "$file" "$_services"
 
   # Save deployhook for renewals
   _debug2 "Setting Le_DeployHook"
@@ -84,6 +87,8 @@ _preprocess_deployfile() {
     _err "Deploy file not found. Go to https://CHANGE_URL_TO_WIKI to see how to create one."
     return 1
   fi
+
+  echo "$DOMAIN_PATH/$found_file"
 }
 
 # deploy_filepath _deploy_config
@@ -155,25 +160,6 @@ _get_services_list() {
   echo "$_services"
 }
 
-# deploy_filepath service_names
-_get_full_services_list() {
-  _deploy_file="$1"
-  shift
-  _service_names="$*"
-
-  _debug3 "Deploy file" "$_deploy_file"
-  _debug3 "Service names" "$_service_names"
-
-  _full_services=""
-  for _service in $_service_names; do
-    _full_service=$(yq e ".services[] | select(.name == \"$_service\")" "$_deploy_file")
-    _full_services="$_full_services
-$_full_service"
-  done
-
-  echo "$_full_services"
-}
-
 # env_list
 _export_envs() {
   _env_list="$1"
@@ -211,6 +197,7 @@ _deploy_services() {
   _debug3 "Services" "$_services"
 
   for _service in $_services; do
+    _debug2 "Service" "$_service"
     _hook=$(yq e ".services[] | select(.name == \"$_service\").hook" "$_deploy_file")
     _envs=$(yq e ".services[] | select(.name == \"$_service\").environment[]" "$_deploy_file")
     _export_envs "$_envs"
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -66,6 +66,7 @@ _preprocess_deployfile() {
   _debug3 "yq is installed."
 
   # Check if deploy file exists
+  IFS=$(printf '\n')
   for file in "$@"; do
     _debug3 "Checking file" "$DOMAIN_PATH/$file"
     if [ -f "$DOMAIN_PATH/$file" ]; then
@@ -79,6 +80,7 @@ _preprocess_deployfile() {
       _debug3 "File not found"
     fi
   done
+  IFS=$OLDIFS
 
   if [ -n "$found_file" ]; then
     _check_deployfile "$DOMAIN_PATH/$found_file" "$MULTIDEPLOY_CONFIG"
@@ -124,6 +126,7 @@ _check_deployfile() {
   fi
   _debug2 "Config has services."
 
+  IFS=$(printf '\n')
   # Check if extracted services exist in services list
   for _service in $_services; do
     _debug2 "Checking service" "$_service"
@@ -145,6 +148,7 @@ _check_deployfile() {
       return 1
     fi
   done
+  IFS=$OLDIFS
 }
 
 # deploy_filepath deploy_config
@@ -166,25 +170,27 @@ _export_envs() {
 
   _secure_debug3 "Exporting envs" "$_env_list"
 
-  for _env in $_env_list; do
-    _key=$(echo "$_env" | cut -d '=' -f1)
-    _value=$(echo "$_env" | cut -d '=' -f2-)
+  IFS=$(printf '\n')
+  echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value' | while IFS='=' read -r _key _value; do
     _savedomainconf "$_key" "$_value"
     _secure_debug3 "Saved $_key" "$_value"
   done
+  IFS=$OLDIFS
 }
 
 _clear_envs() {
   _env_list="$1"
 
   _secure_debug3 "Clearing envs" "$_env_list"
+  env_pairs=$(echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value')
 
-  for _env in $_env_list; do
-    _key=$(echo "$_env" | cut -d '=' -f1)
+  IFS=$(printf '\n')
+  echo "$env_pairs" | while IFS='=' read -r _key _value; do
     _debug3 "Deleting key" "$_key"
     _cleardomainconf "SAVED_$_key"
     unset "$_key"
   done
+  IFS="$OLDIFS"
 }
 
 # deploy_filepath services_array
@@ -196,14 +202,17 @@ _deploy_services() {
   _debug3 "Deploy file" "$_deploy_file"
   _debug3 "Services" "$_services"
 
+  IFS=$(printf '\n')
   for _service in $_services; do
     _debug2 "Service" "$_service"
     _hook=$(yq e ".services[] | select(.name == \"$_service\").hook" "$_deploy_file")
     _envs=$(yq e ".services[] | select(.name == \"$_service\").environment[]" "$_deploy_file")
+
     _export_envs "$_envs"
     _deploy_service "$_service" "$_hook"
     _clear_envs "$_envs"
   done
+  IFS=$OLDIFS
 }
 
 _deploy_service() {
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -151,7 +151,13 @@ _check_deployfile() {
   IFS=$OLDIFS
 }
 
-# deploy_filepath deploy_config
+# Description:
+#   This function retrieves a list of services from the deploy configuration file.
+# Arguments:
+#   $1 - The path to the deploy configuration file.
+#   $2 - The name of the deploy configuration to use.
+# Usage:
+#   _get_services_list "<deploy_file_path>" "<deploy_config_name>"
 _get_services_list() {
   _deploy_file="$1"
   _deploy_config="$2"
@@ -164,7 +170,12 @@ _get_services_list() {
   echo "$_services"
 }
 
-# env_list
+# Description: This function takes a list of environment variables in YAML format,
+#              parses them, and exports each key-value pair as environment variables.
+# Arguments:
+#   $1 - A string containing the list of environment variables in YAML format.
+# Usage:
+#   _export_envs "$env_list"
 _export_envs() {
   _env_list="$1"
 
@@ -178,6 +189,13 @@ _export_envs() {
   IFS=$OLDIFS
 }
 
+# Description:
+#   This function takes a YAML formatted string of environment variables, parses it,
+#   and clears each environment variable. It logs the process of clearing each variable.
+# Arguments:
+#   $1 - A YAML formatted string containing environment variable key-value pairs.
+# Usage:
+#   _clear_envs "<yaml_string>"
 _clear_envs() {
   _env_list="$1"
 
@@ -188,7 +206,7 @@ _clear_envs() {
   echo "$env_pairs" | while IFS='=' read -r _key _value; do
     _debug3 "Deleting key" "$_key"
     _cleardomainconf "SAVED_$_key"
-    unset "$_key"
+    unset -v "$_key"
   done
   IFS="$OLDIFS"
 }
--- a/deploy/multideploy.sh
+++ b/deploy/multideploy.sh
@@ -9,106 +9,106 @@ MULTIDEPLOY_FILENAME="multideploy.yaml"
 
 # domain keyfile certfile cafile fullchain pfx
 multideploy_deploy() {
-    _cdomain="$1"
-    _ckey="$2"
-    _ccert="$3"
-    _cca="$4"
-    _cfullchain="$5"
-    _cpfx="$6"
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+  _cpfx="$6"
 
-    _debug _cdomain "$_cdomain"
-    _debug _ckey "$_ckey"
-    _debug _ccert "$_ccert"
-    _debug _cca "$_cca"
-    _debug _cfullchain "$_cfullchain"
-    _debug _cpfx "$_cpfx"
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+  _debug _cpfx "$_cpfx"
 
-    DOMAIN_DIR=$_cdomain
-    if echo "$DOMAIN_PATH" | grep -q "$ECC_SUFFIX"; then
-        DOMAIN_DIR="$DOMAIN_DIR"_ecc
-    fi
-    _debug2 "DOMAIN_DIR" "$DOMAIN_DIR"
+  DOMAIN_DIR=$_cdomain
+  if echo "$DOMAIN_PATH" | grep -q "$ECC_SUFFIX"; then
+    DOMAIN_DIR="$DOMAIN_DIR"_ecc
+  fi
+  _debug2 "DOMAIN_DIR" "$DOMAIN_DIR"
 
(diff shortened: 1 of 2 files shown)


gilman88:
+1 on this kind of capability at least for wildcard certs. I was recently looking at a situation with multiple mikrotik routers where this would have been helpful.

tomo2403:
@Neilpang done

Neilpang:
let's remove the `configs` part:
```
configs:
  - name: "default"
    services:
      - "webserver"
      - "webserver2"
```

It's not necessary.


In the yaml example, please add some other hooks, not just docker hook. because it should work with any hooks.

Don't use a hardcoded 'multideploy.yml" file, let's make it a env variable, just the others:

```
export  DEPLOY_YAML="/path/to/my/multideploy.yaml"
acme.sh --deploy -d xxxx.com  --deploy-hook  multideploy
```

You can just copy the "$DEPLOY_YAML" file to the domain folder,  it will be easier for the user to use.

tomo2403:
The configurations (`configs`) are intended to simplify the testing/staging process. They allow the user to quickly select or deselect the services the certificate should be deployed to without having to comment out every line. Soon, I also want to implement an overriding functionality similar to Docker (`compose.override.yaml`). Configurations will then be even more necessary to enhance testing and deployment further when multiple certificates are deployed to the same services. This would allow the user to maintain a minimal, non-redundant configuration. That is also the reason for the hardcoded filepath.

Do you agree with this?

Neilpang:
>>The configurations (configs) are intended to simplify the testing/staging process.

no, this is too complicated.

>>  That is also the reason for the hardcoded filepath.

no,  use the env variable to pass value.  it's the same way as others.

tomo2403:
@Neilpang, I removed configs and introduced a variable deploy file name. The wiki is now up to date.

deploy/multideploy.sh
```diff
@@ -0,0 +1,267 @@
...
+# Usage:
+#   _deploy_services "<deploy_file_path>" "<services_list>"
+_deploy_services() {
+  _deploy_file="$1"
+  _debug3 "Deploy file" "$_deploy_file"
+
+  _services=$(yq e '.services[].name' "$_deploy_file")
+  _debug3 "Services" "$_services"
+
+  printf '%s\n' "$_services" | while IFS= read -r _service; do
```
zeocax, line 236:
```suggestion
_service_array=()
printf '%s\n' "$_services" | while IFS= read -r _service; do
  _service_array+=("$_service")
done

for _service in "${_service_array[@]}"; do
```

I think there is an issue caused by the ssh hook. The current ssh hook does not use the -n option, so it consumes stdin. This would cause the "read -r" to end.

Although I believe the default command for the ssh hook should be modified (by adding -n). But considering that the ssh hook saves the commands used, there might be users who have already used the ssh hook before, so it's best to read all the services first, I think.

tomo2403: What issue did you experience with the ssh hook while using multideploy?

lojzik: It's possible, similar problem is with routeros hook. ssh and scp utils in this hooks read stdin and this breaks while statement after first hook with ssh or scp.

zeocax: > What issue did you experience with the ssh hook while using multideploy?

You can test the ssh hook along with other hooks, and you will find that the hooks after the ssh hook will not be processed because read -r has already been closed. As @lojzik said, it seems that this issue is not limited to just the ssh hook.

zeocax (commented):
tested and working fine for me

deploy/multideploy.sh
```diff
@@ -0,0 +1,269 @@
...
+  _cpfx="$6"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+  _debug _cpfx "$_cpfx"
+
+  DOMAIN_DIR=$_cdomain
```
Neilpang, line 52: why  do you need this DOMAIN_DIR ?

deploy/multideploy.sh
```diff
@@ -0,0 +1,269 @@
...
+    return 1
+  fi
+  _debug3 "File" "$file"
+
+  # Deploy to services
+  _deploy_services "$file"
+
+  # Save deployhook for renewals
+  _debug2 "Setting Le_DeployHook"
+  _savedomainconf "Le_DeployHook" "multideploy"
```
Neilpang, line 79: don't save it.   it should be taken care of in the main acme.sh

deploy/multideploy.sh
tomo2403: This can be used in the deploy file to specify the path for a cert if you are going to have multiple certs on the same service. This makes it easier to copy the deploy file to another domain dir.

deploy/multideploy.sh
```diff
@@ -0,0 +1,269 @@
...
+# Usage:
+#   _export_envs "$env_list"
+_export_envs() {
+  _env_list="$1"
+
+  _secure_debug3 "Exporting envs" "$_env_list"
+
+  IFS=$(printf '\n')
+  echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value' | while IFS='=' read -r _key _value; do
+    _value=$(eval echo "$_value")
```
Neilpang, line 193: you don't need to use eval, just use `export  env=value`

deploy/multideploy.sh
```diff
@@ -0,0 +1,269 @@
...
+    _savedomainconf "$_key" "$_value"
```
Neilpang, line 194: don't save it in the domain conf.  everytime, you should just use the values from the multiply.yml file.

deploy/multideploy.sh
tomo2403: If it is not saved, the last hook used by multideploy will be saved and used for renewal

deploy/multideploy.sh
```diff
@@ -0,0 +1,269 @@
...
+}
+
+# Description:
+#   This function takes a YAML formatted string of environment variables, parses it,
+#   and clears each environment variable. It logs the process of clearing each variable.
+# Arguments:
+#   $1 - A YAML formatted string containing environment variable key-value pairs.
+# Usage:
+#   _clear_envs "<yaml_string>"
+_clear_envs() {
```
Neilpang, line 207: no need to clear

deploy/multideploy.sh
```diff
@@ -0,0 +1,269 @@
...
+#   _deploy_service <service_name> <hook>
+_deploy_service() {
+  _name="$1"
+  _hook="$2"
+
+  _debug2 "SERVICE" "$_name"
+  _debug2 "HOOK" "$_hook"
+
+  _info "$(__green "Deploying") to '$_name' using '$_hook'"
+  if echo "$DOMAIN_PATH" | grep -q "$ECC_SUFFIX"; then
```
Neilpang, line 263: why do you need to check ecc suffix here?

deploy/multideploy.sh
Neilpang: which deploy file uses it ?

tomo2403: have a look at the sample config above

deploy/multideploy.sh
Neilpang: use `_deploy()`  instead of `deploy()` function.
`_deploy()` function doesn't save the hook

deploy/multideploy.sh
Neilpang: which one ?

deploy/multideploy.sh
tomo2403: This is needed to make the use of DOMAIN_DIR and all other custom vars work that the user might use

deploy/multideploy.sh
tomo2403: If a hook has optional envs and you have two services that use the same hook, but one service uses the optional envs and the other doesn't, both services will use the optional envs because they are saved by the first service. Clearing them ensures that multideploy only uses the envs specified in the deploy file

deploy/multideploy.sh
tomo2403: was needed for deploy(). can be removed when using _deploy()

deploy/multideploy.sh
tomo2403: My thought was that maybe not every hook checks the exported variables first and maybe takes them directly from the config. Overwriting the config prevents a hook from taking obsolete values.

Gerporgl:
I tested this and I'm looking forward to have this merged.
I needed this to auto renew and publish my wildcard certificate to my synology and nginx servers.
Very useful.

tomo2403:
Hello,

> So it seems like the environment variable isn't being set for some reason, very odd.

1. How many times have you tried this?
2. Could you please share the contents of your `example.com.conf` file?
3. Also, please post the output of your command with debug enabled:
```sh
acme.sh --deploy -d example.com --deploy-hook multideploy --debug 3
```

invario:
> Hello @invario,
> 
> > So it seems like the environment variable isn't being set for some reason, very odd.
> 
> 1. How many times have you tried this?
At least 5 times..  Even completely wiped the multideploy.yml file and copy and pasted from your example and manually typed it in just in case there were something off.

> 2. Could you please share the contents of your `example.com.conf` file?
```
Le_Domain='example.com'
Le_Alt='*.example.com'
Le_Webroot='dns_cf'
Le_PreHook=''
Le_PostHook=''
Le_RenewHook=''
Le_API='https://acme-v02.api.letsencrypt.org/directory'
Le_Keylength='ec-256'
Le_OrderFinalize='https://acme-v02.api.letsencrypt.org/acme/finalize/**BLOCKED**'
CF_Token='**BLOCKED**'
CF_Account_ID=''
CF_Zone_ID='**BLOCKED**'
Le_LinkOrder='https://acme-v02.api.letsencrypt.org/acme/order/2421948647/387838364687'
Le_LinkCert='https://acme-v02.api.letsencrypt.org/acme/cert/065ca48054c135c1fc336ed5367345528824'
Le_CertCreateTime='1748228453'
Le_CertCreateTimeStr='2025-05-26T03:00:53Z'
Le_NextRenewTimeStr='2025-07-24T03:00:53Z'
Le_NextRenewTime='1753326053'
Le_DeployHook='multideploy'

```
This is what the `example.com.conf` looks like prior to running the deploy command.  Aftewards, an additional 3 line are added:
```
DEPLOY_DOCKER_CONTAINER_KEY_FILE='/data/tls/custom/npm-1/privkey.pem'
DEPLOY_DOCKER_CONTAINER_FULLCHAIN_FILE='/data/tls/custom/npm-1/fullchain.pem'
DEPLOY_DOCKER_CONTAINER_RELOAD_CMD='nginx -s reload'
```

> 3. Also, please post the output of your command with debug enabled:
> 
> ```shell
> acme.sh --deploy -d example.com --deploy-hook multideploy --debug 3
> ```

Here is a debug output:
```
[Tue May 27 10:45:33 EDT 2025] Let's find the script directory.
[Tue May 27 10:45:33 EDT 2025] _SCRIPT_='/usr/local/bin/acme.sh'
[Tue May 27 10:45:33 EDT 2025] _script='/root/.acme.sh/acme.sh'
[Tue May 27 10:45:33 EDT 2025] _script_home='/root/.acme.sh'
[Tue May 27 10:45:33 EDT 2025] Using default home: /root/.acme.sh
[Tue May 27 10:45:33 EDT 2025] Using config home: /acme.sh
[Tue May 27 10:45:33 EDT 2025] LE_WORKING_DIR='/root/.acme.sh'
v3.1.2
[Tue May 27 10:45:33 EDT 2025] Running cmd: deploy
[Tue May 27 10:45:33 EDT 2025] Using config home: /acme.sh
[Tue May 27 10:45:33 EDT 2025] default_acme_server
[Tue May 27 10:45:33 EDT 2025] ACME_DIRECTORY='https://acme.zerossl.com/v2/DV90'
[Tue May 27 10:45:33 EDT 2025] _ACME_SERVER_HOST='acme.zerossl.com'
[Tue May 27 10:45:33 EDT 2025] _ACME_SERVER_PATH='v2/DV90'
[Tue May 27 10:45:33 EDT 2025] The domain 'example.com' seems to already have an ECC cert, let's use it.
[Tue May 27 10:45:33 EDT 2025] DOMAIN_PATH='/acme.sh/example.com_ecc'
[Tue May 27 10:45:33 EDT 2025] DOMAIN_CONF='/acme.sh/example.com_ecc/example.com.conf'
[Tue May 27 10:45:33 EDT 2025] _deployApi='/root/.acme.sh/deploy/multideploy.sh'
[Tue May 27 10:45:33 EDT 2025] _cdomain='example.com'
[Tue May 27 10:45:33 EDT 2025] _ckey='/acme.sh/example.com_ecc/example.com.key'
[Tue May 27 10:45:33 EDT 2025] _ccert='/acme.sh/example.com_ecc/example.com.cer'
[Tue May 27 10:45:33 EDT 2025] _cca='/acme.sh/example.com_ecc/ca.cer'
[Tue May 27 10:45:33 EDT 2025] _cfullchain='/acme.sh/cexample.com_ecc/fullchain.cer'
[Tue May 27 10:45:33 EDT 2025] _cpfx='/acme.sh/example.com_ecc/example.com.pfx'
[Tue May 27 10:45:33 EDT 2025] DOMAIN_DIR='example.com_ecc'
[Tue May 27 10:45:33 EDT 2025] MULTIDEPLOY_FILENAME is not set, so I will use 'multideploy.yml'.
[Tue May 27 10:45:33 EDT 2025] Deploy file='/acme.sh/example.com_ecc/multideploy.yml'
[Tue May 27 10:45:33 EDT 2025] Deploy file version is compatible: 1.0
[Tue May 27 10:45:33 EDT 2025] Services='npm'
[Tue May 27 10:45:33 EDT 2025] Config has services.
[Tue May 27 10:45:33 EDT 2025] Checking service='npm'
[Tue May 27 10:45:33 EDT 2025] Service='npm'
[Tue May 27 10:45:33 EDT 2025] SERVICE='npm'
[Tue May 27 10:45:34 EDT 2025] HOOK='docker'
[Tue May 27 10:45:34 EDT 2025] Deploying to 'npm' using 'docker'
[Tue May 27 10:45:34 EDT 2025] _deployApi='/root/.acme.sh/deploy/docker.sh'
[Tue May 27 10:45:34 EDT 2025] _cdomain='example.com'
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_LABEL
[Tue May 27 10:45:34 EDT 2025] The DEPLOY_DOCKER_CONTAINER_LABEL variable is not defined, we use this label to find the container.
[Tue May 27 10:45:34 EDT 2025] See: https://github.com/acmesh-official/acme.sh/wiki/deploy-to-docker-containers
[Tue May 27 10:45:34 EDT 2025] Try use /var/run/docker.sock
[Tue May 27 10:45:34 EDT 2025] _cversion='8.12.1'
[Tue May 27 10:45:34 EDT 2025] _major='8'
[Tue May 27 10:45:34 EDT 2025] _minor='12'
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_KEY_FILE
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_CERT_FILE
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_CA_FILE
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_FULLCHAIN_FILE
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_PFX_FILE
[Tue May 27 10:45:34 EDT 2025] DEPLOY_DOCKER_CONTAINER_RELOAD_CMD
[Tue May 27 10:45:34 EDT 2025] _req='{"label":[""]}'
[Tue May 27 10:45:34 EDT 2025] _req='%7b%22label%22%3a%5b%22%22%5d%7d'
[Tue May 27 10:45:34 EDT 2025] _data
[Tue May 27 10:45:34 EDT 2025] url='http://localhost/containers/json?filters=%7b%22label%22%3a%5b%22%22%5d%7d'
10:45:34.067005 [0-x] == Info: [READ] client_reset, clear readers
10:45:34.067143 [0-0] == Info: [SETUP] added
10:45:34.067222 [0-0] == Info:   Trying /var/run/docker.sock:0...
10:45:34.067332 [0-0] == Info: Connected to localhost (/var/run/docker.sock) port 0
10:45:34.067438 [0-0] == Info: using HTTP/1.x
10:45:34.067505 [0-0] => Send header, 180 bytes (0xb4)
0000: GET /containers/json?filters=%7b%22label%22%3a%5b%22%22%5d%7d HT
0040: TP/1.1
0048: Host: localhost
0059: User-Agent: curl/8.12.1
0072: Accept: */*
007f: Content-Type: application/json
009f: Content-Length: 0
00b2:
10:45:34.067904 [0-0] == Info: Request completely sent off
10:45:34.068005 [0-0] <= Recv header, 17 bytes (0x11)
0000: HTTP/1.1 200 OK
10:45:34.068115 [0-0] == Info: [WRITE] cw_out, wrote 17 header bytes -> 17
10:45:34.068227 [0-0] == Info: [WRITE] download_write header(type=c, blen=17) -> 0
10:45:34.068322 [0-0] == Info: [WRITE] client_write(type=c, len=17) -> 0
10:45:34.068409 [0-0] <= Recv header, 19 bytes (0x13)
0000: Api-Version: 1.47
10:45:34.068520 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=19) -> 0
10:45:34.068617 [0-0] == Info: [WRITE] cw_out, wrote 19 header bytes -> 19
10:45:34.068724 [0-0] == Info: [WRITE] download_write header(type=4, blen=19) -> 0
10:45:34.068828 [0-0] == Info: [WRITE] client_write(type=4, len=19) -> 0
10:45:34.068917 [0-0] <= Recv header, 32 bytes (0x20)
0000: Content-Type: application/json
10:45:34.069036 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=32) -> 0
10:45:34.069134 [0-0] == Info: [WRITE] cw_out, wrote 32 header bytes -> 32
10:45:34.069233 [0-0] == Info: [WRITE] download_write header(type=4, blen=32) -> 0
10:45:34.069347 [0-0] == Info: [WRITE] client_write(type=4, len=32) -> 0
10:45:34.069443 [0-0] <= Recv header, 28 bytes (0x1c)
0000: Docker-Experimental: false
10:45:34.069585 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=28) -> 0
10:45:34.069658 [0-0] == Info: [WRITE] cw_out, wrote 28 header bytes -> 28
10:45:34.069755 [0-0] == Info: [WRITE] download_write header(type=4, blen=28) -> 0
10:45:34.069874 [0-0] == Info: [WRITE] client_write(type=4, len=28) -> 0
10:45:34.069973 [0-0] <= Recv header, 15 bytes (0xf)
0000: Ostype: linux
10:45:34.070058 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=15) -> 0
10:45:34.070160 [0-0] == Info: [WRITE] cw_out, wrote 15 header bytes -> 15
10:45:34.070253 [0-0] == Info: [WRITE] download_write header(type=4, blen=15) -> 0
10:45:34.070352 [0-0] == Info: [WRITE] client_write(type=4, len=15) -> 0
10:45:34.070457 [0-0] <= Recv header, 31 bytes (0x1f)
0000: Server: Docker/27.5.0 (linux)
10:45:34.070567 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=31) -> 0
10:45:34.070684 [0-0] == Info: [WRITE] cw_out, wrote 31 header bytes -> 31
10:45:34.070778 [0-0] == Info: [WRITE] download_write header(type=4, blen=31) -> 0
10:45:34.070873 [0-0] == Info: [WRITE] client_write(type=4, len=31) -> 0
10:45:34.070968 [0-0] <= Recv header, 37 bytes (0x25)
0000: Date: Tue, 27 May 2025 14:45:34 GMT
10:45:34.071078 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=37) -> 0
10:45:34.071182 [0-0] == Info: [WRITE] cw_out, wrote 37 header bytes -> 37
10:45:34.071275 [0-0] == Info: [WRITE] download_write header(type=4, blen=37) -> 0
10:45:34.071369 [0-0] == Info: [WRITE] client_write(type=4, len=37) -> 0
10:45:34.071462 [0-0] <= Recv header, 19 bytes (0x13)
0000: Content-Length: 3
10:45:34.071566 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=19) -> 0
10:45:34.071666 [0-0] == Info: [WRITE] cw_out, wrote 19 header bytes -> 19
10:45:34.071757 [0-0] == Info: [WRITE] download_write header(type=4, blen=19) -> 0
10:45:34.071858 [0-0] == Info: [WRITE] client_write(type=4, len=19) -> 0
10:45:34.071940 [0-0] <= Recv header, 2 bytes (0x2)
0000:
10:45:34.072005 [0-0] == Info: [WRITE] header_collect pushed(type=1, len=2) -> 0
10:45:34.072094 [0-0] == Info: [WRITE] cw_out, wrote 2 header bytes -> 2
10:45:34.072173 [0-0] == Info: [WRITE] download_write header(type=4, blen=2) -> 0
10:45:34.072262 [0-0] == Info: [WRITE] client_write(type=4, len=2) -> 0
10:45:34.072342 [0-0] <= Recv data, 3 bytes (0x3)
0000: [].
10:45:34.072442 [0-0] == Info: [WRITE] cw_out, wrote 3 body bytes -> 3
10:45:34.072517 [0-0] == Info: [WRITE] download_write body(type=1, blen=3) -> 0
10:45:34.072627 [0-0] == Info: [WRITE] client_write(type=1, len=3) -> 0
10:45:34.072709 [0-0] == Info: [WRITE] xfer_write_resp(len=203, eos=0) -> 0
10:45:34.072809 [0-0] == Info: [WRITE] cw-out is notpaused
10:45:34.072887 [0-0] == Info: [WRITE] cw-out done
10:45:34.072942 [0-0] == Info: [READ] client_reset, clear readers
10:45:34.073040 [0-0] == Info: Connection #0 to host localhost left intact
[Tue May 27 10:45:34 EDT 2025] listjson='[]'
[Tue May 27 10:45:34 EDT 2025] Container id:
[Tue May 27 10:45:34 EDT 2025] can not find container id
[Tue May 27 10:45:34 EDT 2025] Error deploying for domain: example.com
[Tue May 27 10:45:34 EDT 2025] Error encountered while deploying.
[Tue May 27 10:45:34 EDT 2025] Setting Le_DeployHook
[Tue May 27 10:45:34 EDT 2025] Success
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,264 @@
...
+#   _export_envs "$env_list"
+_export_envs() {
+  _env_list="$1"
+
+  _secure_debug3 "Exporting envs" "$_env_list"
+
+  IFS=$(printf '\n')
+  echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value' | while IFS='=' read -r _key _value; do
+    _value=$(eval echo "$_value")
+    _savedomainconf "$_key" "$_value"
```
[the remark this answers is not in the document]
invario: Good news. Have tested it on two additional instances for myself.    TrueNAS Core using the `truenas` hook, and TrueNAS Scale also using the `truenas` hook.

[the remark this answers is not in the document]
tomo2403: I have found a solution to this problem. Instead of `_savedeployconf`, I used `_savedomainconf`, which led to this error.

Gerporgl:
> @Gerporgl @invario @egarzadev Could you please pull the latest change and test it again? I can't reproduce this error with my setup.

I tried it and it seems to work fine, I am not get the error I was having anymore.

Thanks!

deploy/multideploy.sh
```diff
@@ -0,0 +1,281 @@
...
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+  _debug _cpfx "$_cpfx"
+
+  DOMAIN_DIR=$_cdomain
+  if echo "$DOMAIN_PATH" | grep -q "$ECC_SUFFIX"; then
+    DOMAIN_DIR="$DOMAIN_DIR"_ecc
+  fi
+  _debug2 "DOMAIN_DIR" "$DOMAIN_DIR"
```
Neilpang, line 56: remove DOMAIN_DIR.
and this 
```
  if echo "$DOMAIN_PATH" | grep -q "$ECC_SUFFIX"; then
    DOMAIN_DIR="$DOMAIN_DIR"_ecc
  fi
  _debug2 "DOMAIN_DIR" "$DOMAIN_DIR"

```

Neilpang: you should never care about th dir of the  domain config.
don't rely on the path and the suffix.

deploy/multideploy.sh
```diff
@@ -0,0 +1,281 @@
...
+_deploy_service() {
+  _name="$1"
+  _hook="$2"
+
+  _debug2 "SERVICE" "$_name"
+  _debug2 "HOOK" "$_hook"
+
+  _info "$(__green "Deploying") to '$_name' using '$_hook'"
+  _deploy "$_cdomain" "$_hook"
+  return $?
```
Neilpang, line 280: remove this `return $?`

tomo2403: Could you please tell me why? This is only used by my hook to determine if the deployment failed to count the hooks that returned with an error. This also solves a problem mentioned here: https://github.com/acmesh-official/acme.sh/pull/6241#issuecomment-2914556098

deploy/multideploy.sh
tomo2403: Removed

deploy/multideploy.sh
Neilpang: you don't have to use `return $?`,    the function will always use the return code of the late command.

jdevera:
This is a very exciting addition! I found this PR after seeing the docs in the wiki and wondering why they did not work 😓 .

I only wanted to leave a minor comment about the configuration format, if you are receptive.

I want to propose a change of the yaml format for environment, from this:

```yaml
    environment:
      - DEPLOY_DOCKER_CONTAINER_LABEL: "sh.acme.autoload.service=example.com"
      - DEPLOY_DOCKER_CONTAINER_KEY_FILE: "/data/tls/custom/npm-1/privkey.pem"
      - DEPLOY_DOCKER_CONTAINER_FULLCHAIN_FILE: "/data/tls/custom/npm-1/fullchain.pem"
      - DEPLOY_DOCKER_CONTAINER_RELOAD_CMD: "nginx -s reload"
```

to this:

```yaml
    environment:
      DEPLOY_DOCKER_CONTAINER_LABEL: "sh.acme.autoload.service=example.com"
      DEPLOY_DOCKER_CONTAINER_KEY_FILE: "/data/tls/custom/npm-1/privkey.pem"
      DEPLOY_DOCKER_CONTAINER_FULLCHAIN_FILE: "/data/tls/custom/npm-1/fullchain.pem"
      DEPLOY_DOCKER_CONTAINER_RELOAD_CMD: "nginx -s reload"
```

The first one is a list of mappings, each of them with only one key, and they are different keys in each, whereas the second is a single mapping where each key is an env var name. Proposing because I found it a bit confusing when reading the wiki.

deploy/multideploy.sh
```diff
@@ -0,0 +1,274 @@
...
+  _debug3 "Services" "$_services"
+
+  _service_list=$(printf '%s\n' "$_services")
+
+  _failedServices=""
+  _failedCount=0
+  for _service in $(printf '%s\n' "$_service_list"); do
+    _debug2 "Service" "$_service"
+    _hook=$(yq e ".services[] | select(.name == \"$_service\").hook" "$_deploy_file")
+    _envs=$(yq e ".services[] | select(.name == \"$_service\").environment[]" "$_deploy_file")
```
kkarger-xqueue: Running into this error:

> Error: !!str has no keys

because environment[] only returns the values but _export_envs expects key-value pairs.

I think it should be
`_envs=$(yq e ".services[] | select(.name == \"$_service\").environment" "$_deploy_file")`

along the suggested changes by @jdevera in https://github.com/acmesh-official/acme.sh/issues/6241#issuecomment-3014287035 it fixes the export issues and enables deploy via the ssh hook in my testing.

tomo2403:
@jdevera Thank you for the suggestion! Your proposed change to the YAML format makes perfect sense and has been implemented.

UsynligAnd:
This addition works perfect in my situation where i need to deploy a single wildcard cert to multiple servers with SSH. 
Great work, cant wait for it to be merged!

deploy/multideploy.sh
```diff
@@ -0,0 +1,274 @@
...
+################################################################################
+# Usage (shown values are the examples):
+# 1. Set optional environment variables
+#   - export MULTIDEPLOY_FILENAME="multideploy.yaml"     - "multideploy.yml" will be automatically used if not set"
+#
+# 2. Run command:
+# acme.sh --deploy --deploy-hook multideploy -d example.com
+################################################################################
+# Dependencies:
+# - yq
```
cheggerdev, line 18: yq does not exist on Synology DSM. Please make yq optional.

tomo2403: yq is essential for reading the YAML file. You can install it with a community package or run acme.sh in Docker.

jdevera:
I'm testing this now with a local copy and saw there is a `$DOMAIN_DIR` variable in the wiki and the PR message above, but I cannot find any reference to it in the code. Is this leftover from a previous version?

tomo2403:
Yes, this is no longer available. https://github.com/acmesh-official/acme.sh/pull/6241#discussion_r2118047565

invario:
> I tried to run this hook on Debian and found plenty of errors. The first one is quotes. I've got an error: `"The deploy hook ""hook_name"" was not found."` The second one is file checking doesn't work. And some more...
> 
> I've fixed them for myself, but I hope that author can fix them, because this hook is very useful. I think so.

That sounds like a configuration error on your end.

UncleVic:
> That sounds like a configuration error on your end. What's your YAML file contain?
I copied it from the wiki

UncleVic:
```yaml
version: "1.0"

services:
  - name: proxmox_NAS
    hook: "proxmoxve"
    environment:
      DEPLOY_PROXMOXVE_SERVER: ""
      DEPLOY_PROXMOXVE_SERVER_PORT: ""
      DEPLOY_PROXMOXVE_NODE_NAME: ""
      DEPLOY_PROXMOXVE_USER: ""
      DEPLOY_PROXMOXVE_USER_REALM: "pam"
      DEPLOY_PROXMOXVE_API_TOKEN_NAME: ""
      DEPLOY_PROXMOXVE_API_TOKEN_KEY: ""

  - name: proxmox_Media
    hook: "proxmoxve"
    environment:
      DEPLOY_PROXMOXVE_SERVER: ""
      DEPLOY_PROXMOXVE_SERVER_PORT: ""
      DEPLOY_PROXMOXVE_NODE_NAME: ""
      DEPLOY_PROXMOXVE_USER: ""
      DEPLOY_PROXMOXVE_USER_REALM: "pam"
      DEPLOY_PROXMOXVE_API_TOKEN_NAME: ""
      DEPLOY_PROXMOXVE_API_TOKEN_KEY: ""
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,274 @@
...
+  _debug3 "Deploy file" "$_deploy_file"
+
+  _services=$(yq e '.services[].name' "$_deploy_file")
+  _debug3 "Services" "$_services"
+
+  _service_list=$(printf '%s\n' "$_services")
+
+  _failedServices=""
+  _failedCount=0
+  for _service in $(printf '%s\n' "$_service_list"); do
```
jdevera: When the service names have spaces, this ends up expanding to each word of the services as different services and gives incorrect results because later it cannot find the service.

deploy/multideploy.sh
```diff
@@ -0,0 +1,274 @@
...
+    return 1
+  fi
+  _debug2 "Config has services."
+
+  IFS=$(printf '\n')
+  # Check if extracted services exist in services list
+  for _service in $_services; do
+    _debug2 "Checking service" "$_service"
+    # Check if service exists
+    if ! yq e ".services[] | select(.name == \"$_service\")" "$_deploy_file" >/dev/null; then
```
jdevera: This check assumes that yq will exit with a non-zero state when it cannot find the selected service, but in my tests with the project's docker image, it exists with a zero every time, so the checks succeed even when the services are invalid.

deploy/multideploy.sh
```diff
@@ -0,0 +1,274 @@
...
+      fi
+      found_file="$file"
+    else
+      _debug3 "File not found"
+    fi
+  done
+  IFS=$OLDIFS
+
+  if [ -n "$found_file" ]; then
+    _check_deployfile "$DOMAIN_PATH/$found_file"
```
jdevera: When this check fails `_check_deployfile` returns 1, but this is not checked here, so execution continues in the presence of errors.

deploy/multideploy.sh
jdevera: This conversation appears as resolved but the change was not applied, was there an agreement somewhere else to do something different?

deploy/multideploy.sh
jdevera: I think you should add the comment above as comment in the code. The comment in the code currently tells what the function does but not why, and that helps reasoning about the code and maintaining it.

jdevera:
@tomo2403 I've found some issues while trying to make this work locally and I've been fixing them. I also applied some of the comments from @Neilpang in this PR that were not yet applied and pushed my changes to this branch: https://github.com/jdevera/acme.sh/tree/multideploy-yaml

You can take a look at the commit messages for details of the changes.

Not sure how to collaborate on this but since the changes are there you can apply them on top of your branch and update this PR, or I can open a new PR with all our changes.

deploy/multideploy.sh
jdevera: I removed this in my branch and the deploy hook gets saved correctly

deploy/multideploy.sh
jdevera: This is fixed in my branch

deploy/multideploy.sh
jdevera: This is fixed in my branch

deploy/multideploy.sh
```diff
@@ -0,0 +1,264 @@
...
+# Usage:
+#   _export_envs "$env_list"
+_export_envs() {
+  _env_list="$1"
+
+  _secure_debug3 "Exporting envs" "$_env_list"
+
+  IFS=$(printf '\n')
+  echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value' | while IFS='=' read -r _key _value; do
+    _value=$(eval echo "$_value")
```
jdevera: I had issues with this `eval` in my testing, where some env vars that contained complex commands got those commands executed at this point (not safe because these commands were meant to run with privileges on different boxes). I fixed it by fixing the quoting of the evalled value, but it got me thinking that this can be a security hole and likely it is not worth having it. `eval` will run **any code** and there is potential for accidents. I recommend starting a first release without this and then exploring options that will do variable expansion without running any other code.

You could provide a set of well-known placeholders, such as the domain name, and replace only those. A very natural way of doing this is using `envsubst` (apk package gettext-envsubst) as an optional dependency and perform variable expansion only if available.

```sh
$ export foo=FOOVALUE
$ export bar=BARVALUE
$ echo 'PRE $foo $bar POST' | envsubst
PRE FOOVALUE BARVALUE POST
$ echo 'PRE $foo $bar POST' | envsubst '$foo'.  # Only the given vars will be expanded.
PRE FOOVALUE $bar POST
```

For now I've left the eval in my branch but I'm 95% convinced it is not worth it.

jdevera: It still brings an issue where the variable interpolation happens locally for strings that perhaps are meant to be interpolated remotely, which can be mitigated by choosing namedspaced names for the placeholders, e.g. `$acmesh_multideploy_domain`

deploy/multideploy.sh
cheggerdev, line 18: Synology does not have yq. SynoCommunity has no package with yq.

Here is a bash function that parses yaml files (taken from [stackoverflow](https://stackoverflow.com/questions/5014632/how-can-i-parse-a-yaml-file-from-a-linux-shell-script)):

```
function parse_yaml {
   local prefix=$2
   local s='[[:space:]]*' w='[a-zA-Z0-9_]*' fs=$(echo @|tr @ '\034')
   sed -ne "s|,$s\]$s\$|]|" \
        -e ":1;s|^\($s\)\($w\)$s:$s\[$s\(.*\)$s,$s\(.*\)$s\]|\1\2: [\3]\n\1  - \4|;t1" \
        -e "s|^\($s\)\($w\)$s:$s\[$s\(.*\)$s\]|\1\2:\n\1  - \3|;p" $1 | \
   sed -ne "s|,$s}$s\$|}|" \
        -e ":1;s|^\($s\)-$s{$s\(.*\)$s,$s\($w\)$s:$s\(.*\)$s}|\1- {\2}\n\1  \3: \4|;t1" \
        -e    "s|^\($s\)-$s{$s\(.*\)$s}|\1-\n\1  \2|;p" | \
   sed -ne "s|^\($s\):|\1|" \
        -e "s|^\($s\)-$s[\"']\(.*\)[\"']$s\$|\1$fs$fs\2|p" \
        -e "s|^\($s\)-$s\(.*\)$s\$|\1$fs$fs\2|p" \
        -e "s|^\($s\)\($w\)$s:$s[\"']\(.*\)[\"']$s\$|\1$fs\2$fs\3|p" \
        -e "s|^\($s\)\($w\)$s:$s\(.*\)$s\$|\1$fs\2$fs\3|p" | \
   awk -F$fs '{
      indent = length($1)/2;
      vname[indent] = $2;
      for (i in vname) {if (i > indent) {delete vname[i]; idx[i]=0}}
      if(length($2)== 0){  vname[indent]= ++idx[indent] };
      if (length($3) > 0) {
         vn=""; for (i=0; i<indent; i++) { vn=(vn)(vname[i])("_")}
         printf("%s%s%s=\"%s\"\n", "'$prefix'",vn, vname[indent], $3);
      }
   }'
}
```

jdevera: The `yq` binary that is required here encapsulates the work of parsing and extracting info from YAML files. As you can imagine from the looks of the code that you pasted, maintaining that could be a nightmare. It will work for some cases today and it will break for a lot other cases. This takes the effort away from the purpose here.

The good news is that even when the `yq` does not build for synology (at least not the ARM7 based one I have, one can build a version for it.

I checked out [the repo](https://github.com/mikefarah/yq) and ran this:

```sh
GOOS=linux GOARCH=arm GOARM=7 go build -o yq_linux_arm7 --ldflags "-X main.GitCommit=$(git rev-parse --short HEAD)$(git status --porcelain | grep -q . && echo '+CHANGES' || true) -X main.GitDescribe=$(git describe --tags --always) -w -s"
```

Which is extracted from the commands the project uses to build for other platforms, and I got a binary that I could copy to my synology NAS and run there.

You nas may have a different processor and may require a different binary, mine is 12 years old, so I don't know what they are providing these days, but you may even find that some of the available binaries already work!

I'd say it's fair to say this hook has this dependency,  especially considering how lightweight it is and how, being written in go, it is very easy to build for any platform and deploy by just copying one file.

jdevera:
@tomo2403 thank you for bringing my changes here! This may be ready for another maintainer review.

invario:
Interestingly enough Alpine installs `yq-go` whether you do `apk add yq-go` or `apk add yq`.

tomo2403:
> Interestingly enough Alpine installs `yq-go` whether you do `apk add yq-go` or `apk add yq`.

@invario, it seems to be the same package: https://github.com/mikefarah/yq?tab=readme-ov-file#alpine-linux



> Also I i think that it would be better to stick to distro provided packages rather than forcing people to use "random" content from github.

@nexces, could you please list the yq versions that work or don't work on your Debian system?

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+#   $1 - The path to the deploy configuration file.
+#   $2 - The name of the deploy configuration to use.
+# Usage:
+#   _check_deployfile "<deploy_file_path>"
+_check_deployfile() {
+  _deploy_file="$1"
+  _debug2 "check: Deploy file" "$_deploy_file"
+
+  # Check version
+  _deploy_file_version=$(yq '.version' "$_deploy_file")
```
invario: Replace `yq` with `yq -r`

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+  _debug2 "check: Config has services."
+  echo "$_services" | while read -r _service; do
+    _debug3 " - $_service"
+  done
+
+  # Check if extracted services exist in services list
+  echo "$_services" | while read -r _service; do
+    _debug2 "check: Checking service: $_service"
+    # Check if service exists
+    _service_config=$(yq e ".services[] | select(.name == \"$_service\")" "$_deploy_file")
```
invario: replace `yq e` with `yq -r`

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+      return 1
+    fi
+
+    _service_hook=$(echo "$_service_config" | yq e ".hook" -)
+    if [ -z "$_service_hook" ] || [ "$_service_hook" = "null" ]; then
+      _err "Service '$_service' does not have a hook."
+      return 1
+    fi
+
+    _service_environment=$(echo "$_service_config" | yq e ".environment" -)
```
invario: replace `yq e` with `yq -r`

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+#   file are used.
+# Arguments:
+#   $1 - A YAML formatted string containing environment variable key-value pairs.
+# Usage:
+#   _clear_envs "<yaml_string>"
+_clear_envs() {
+  _env_list="$1"
+
+  _secure_debug3 "Clearing envs" "$_env_list"
+  env_pairs=$(echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value')
```
invario: replace `yq e -r` with `yq -r`

invario (changes requested):
Hah! What do you know, a few small changes required actually.

For `multideploy.yml`, version must be quoted: `version: "1.0"`

1. Line 128: replace `yq` with `yq -r`
2. Replace all `yq e` with `yq -r`

Tested on both versions of `yq `with success.

deploy/multideploy.sh
invario:
```suggestion
    _service_config=$(yq -r ".services[] | select(.name == \"$_service\")" "$_deploy_file")
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+  echo "$_services" | while read -r _service; do
+    _debug2 "check: Checking service: $_service"
+    # Check if service exists
+    _service_config=$(yq e ".services[] | select(.name == \"$_service\")" "$_deploy_file")
+    if [ -z "$_service_config" ] || [ "$_service_config" = "null" ]; then
+      _err "Service '$_service' not found."
+      return 1
+    fi
+
+    _service_hook=$(echo "$_service_config" | yq e ".hook" -)
```
invario:
```suggestion
    _service_hook=$(echo "$_service_config" | yq -r ".hook" -)
```

deploy/multideploy.sh
invario:
```suggestion
    _service_environment=$(echo "$_service_config" | yq -r ".environment" -)
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+# Arguments:
+#   $1 - A string containing the list of environment variables in YAML format.
+# Usage:
+#   _export_envs "$env_list"
+_export_envs() {
+  _env_list="$1"
+
+  _secure_debug3 "Exporting envs" "$_env_list"
+
+  echo "$_env_list" | yq e -r 'to_entries | .[] | .key + "=" + .value' | while IFS='=' read -r _key _value; do
```
invario:
```suggestion
  echo "$_env_list" | yq -r 'to_entries | .[] | .key + "=" + .value' | while IFS='=' read -r _key _value; do
```

deploy/multideploy.sh
invario:
```suggestion
  env_pairs=$(echo "$_env_list" | yq -r 'to_entries | .[] | .key + "=" + .value')
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+# Usage:
+#   _deploy_services "<deploy_file_path>" "<services_list>"
+_deploy_services() {
+  _deploy_file="$1"
+  _debug3 "Deploy file" "$_deploy_file"
+
+  _tempfile=$(mktemp)
+  trap 'rm -f $_tempfile' EXIT
+
+  yq e '.services[].name' "$_deploy_file" >"$_tempfile"
```
invario:
```suggestion
  yq -r '.services[].name' "$_deploy_file" >"$_tempfile"
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+  _debug3 "Services" "$(cat "$_tempfile")"
+
+  _failedServices=""
+  _failedCount=0
+  while read -r _service <&3; do
+    _debug2 "Service" "$_service"
+    _hook=$(yq e ".services[] | select(.name == \"$_service\").hook" "$_deploy_file")
```
invario:
```suggestion
    _hook=$(yq -r ".services[] | select(.name == \"$_service\").hook" "$_deploy_file")
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+    _envs=$(yq e ".services[] | select(.name == \"$_service\").environment" "$_deploy_file")
```
invario:
```suggestion
    _envs=$(yq -r ".services[] | select(.name == \"$_service\").environment" "$_deploy_file")
```

deploy/multideploy.sh
```diff
@@ -0,0 +1,276 @@
...
+  # Check version
+  _deploy_file_version=$(yq '.version' "$_deploy_file")
+  if [ "$MULTIDEPLOY_VERSION" != "$_deploy_file_version" ]; then
+    _err "As of $PROJECT_NAME $VER, the deploy file needs version $MULTIDEPLOY_VERSION! Your current deploy file is of version $_deploy_file_version."
+    return 1
+  fi
+  _debug2 "check: Deploy file version is compatible: $_deploy_file_version"
+
+  # Extract all services from config
+  _services=$(yq e '.services[].name' "$_deploy_file")
```
invario:
```suggestion
  _services=$(yq -r '.services[].name' "$_deploy_file")
```

deploy/multideploy.sh
invario:
```suggestion
  _deploy_file_version=$(yq -r '.version' "$_deploy_file")
```

jdevera:
> @jdevera, is there any particular reason why you switched from yq to yq-go package in https://github.com/acmesh-official/acme.sh/commit/d2df65dd18b2e07f5edc32c78c2bcddcadb1a877?

I think this was answered, but just in case:

@tomo2403 I found the package to have been renamed to yq-go and at the time got an error trying to install yq.

neilpang:
please create a new issue for bug reporting for this hook.

PR #6244, opened 2025-02-24
DNS Verification Configurable Timeout and Retry Interval

jpbion:
MOTIVE: At least one Domain Name Registrar (DNR) has increased the amount of time between changes you make to your zone file (via API) and their publication to the Internet. This causes DNS verification (via TXT record creation) to time out, as the current timeout is only 20 minutes long, and it retries every 10 seconds. Both the rapid retry rate and the short timeout before failure cause acme.sh to no longer work for verifying domain ownership with such DNRs. Using --dnssleep is also not optimal. It has a downside of not allowing for the DNS propagation to end up working more quickly than the sleep time, and producing the certificate sooner. By allowing a configurable retry and timeout interval, one can both be kind to the DNR (say, by trying it only once every 15-20 minutes), and 'succeeding sooner' on a successful try. Further, --dnssleep disables public DNS checks, which is suboptimal if the issue is simply slow propagation. 

DISCUSSION: An ideal solution would extend the capabilities of the DNR provider scripts to specify what default domain DNS verification timeout and verification retry intervals should be for the DNR in question,so the user does not need to specify them. Then, the user could override them with the two command line options provided here. What I provide in this pull request is NOT this ideal solution, but rather just the two command line options, as getting agreement to extend what DNS API scripts need to provide would take longer, and these options are needed *now*.

DESIGN CHOICE TRADEOFF: 

DOWNSIDE: each user of a slow publisher needs to specify these new parameters on the .acme.sh command line:  if they wish to customize away from the default DNS verification and retry timers, just specify --dns-validate-interval and --dns-validate-timeout and give each of them values. For example, I use an interval of 1200 seconds for interval and 14400 for timeout. This works wonderfully for me. If the per-DNR API option were provided, users would be less likely to need these.

UPSIDE: These command line options are still needed to allow a user to work around slow propagation by their DNR, even if DNR-specific default DNS validation timer values were provided in a DNR's API script, because the DNR will often change propagation behavior without updating acme.sh API scripts. Not having these command line options would require users to modify the acme.sh code.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4200,7 +4200,24 @@ __purge_txt() {
 _check_dns_entries() {
   _success_txt=","
   _end_time="$(_time)"
-  _end_time="$(_math "$_end_time" + 1200)" #let's check no more than 20 minutes.
+
+  # Use default values for DNS timeout (1200 seconds/20 minutes) and DNS retry
+  # interval (10 seconds), unless configured at script invocation to be
+  # something else.
+  _dnstimeout=1200 #default timeout is 20 minutes
+  if [ -n "$Le_DNSValidateTimeout" ]; then
+    # Use configured DNS validation timeout
+    _dnstimeout="$Le_DNSValidateTimeout"
+    _savedomainconf "Le_DNSValidateTimeout" "$Le_DNSValidateTimeout"
+  fi  
+  _dnsinterval=10 #default interval between retries is 10 seconds
+  if [ -n "$Le_DNSValidateInterval" ]; then
+    # Use configured DNS validation retry interval
+    _dnsinterval="$Le_DNSValidateInterval"
+    _savedomainconf "Le_DNSValidateInterval" "$Le_DNSValidateInterval"
+  fi
+
+  _end_time="$(_math "$_end_time" + "$_dnstimeout")" #let's check no longer than this
 
   while [ "$(_time)" -le "$_end_time" ]; do
     _info "You can use '--dnssleep' to disable public dns checks."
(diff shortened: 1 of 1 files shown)


jpbion:
Hi - is it possible for someone to approve the workflows - or at least tell me what steps if any I need to do to get them approved? Thanks!

jpbion:
I’ve committed a change to make shfmt pass - the error it reported was whitespace after a “fi”. I’ve removed the whitespace and committed the one line change. If the workflow could get re-run, that’d be great - thanks!

acme.sh
```diff
@@ -4200,7 +4200,24 @@ __purge_txt() {
 _check_dns_entries() {
   _success_txt=","
   _end_time="$(_time)"
-  _end_time="$(_math "$_end_time" + 1200)" #let's check no more than 20 minutes.
+
```
Neilpang, line 4203: make `_dnstimeout` and `_dnsinterval` as parameters of `_check_dns_entries()` function.

acme.sh
```diff
@@ -4200,7 +4200,24 @@ __purge_txt() {
...
+  # Use default values for DNS timeout (1200 seconds/20 minutes) and DNS retry
+  # interval (10 seconds), unless configured at script invocation to be
+  # something else.
+  _dnstimeout=1200 #default timeout is 20 minutes
+  if [ -n "$Le_DNSValidateTimeout" ]; then
```
Neilpang, line 4208: move this `_savedomainconf` to the `_issue()` function.
The `save` operation should only happen in `_issue()` method

jpbion:
I am closing this pull request for this reason: In thinking about how acme.sh saves domain configuration information for --dnssleep while never reading it, how --dnssleep interplays with the two commands I added, I believe a different approach to solving this problem is needed so I am "erasing the board" of this commit to hopefully submit something that handles all of these items cleanly.

Neilpang:
in the "renew()" function,  you it read everything from the domain conf:
https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L5493

PR #6254, opened 2025-02-28
 dnsapi support for HestiaCP

radumalica:
This adds dnsapi support for HestiaCP control panel.

There are some prerequisites before trying to use this dns api script:

Log in to HestiaCP panel as admin or user
Go to Server -> Configure -> API if logged in as admin otherwise if a normal user, click on your profile on top right, and click Access Keys on the upper menu above "Edit User" field
Choose update-dns-records as Access key role
Copy Host, Access Key, and Secret Key
Login to our HestiaCP server as root, and go to /usr/local/hestia/data/api
The file "update-dns-records" should contain this line in order for this script to work:
ROLE='user'
COMMANDS='v-list-dns-records,v-change-dns-record,v-delete-dns-record,v-add-dns-record'
By default, only v-list-dns-records and v-change-dns-record are enabled.
export these variables in your shell or add them in a file and source them.

 `export HESTIA_HOST="https://panel.domain.com:8083"`
` export HESTIA_ACCESS="your_access_key"`
` export HESTIA_SECRET="your_secret_key"`
` export HESTIA_USER="your_username"`

NOTES:

for wildcard certificates to work, you need to use LetsEncrypt V2 provider, not Alpha ZeroSSL which is default in acme.sh
domains available for requesting SSL certificates will be the ones defined under your HestiaCP username (HESTIA_USER).
Issues: [](https://github.com/acmesh-official/acme.sh/issues/6251)

DNS test is successful but your Github actions randomly fails on docker acmetest in the middle of acme.sh processing stuff.

### DIFF
--- a/.github/workflows/issue.yml
+++ b/.github/workflows/issue.yml
@@ -2,6 +2,8 @@ name: "Update issues"
 on:
   issues:
     types: [opened]
+  issue_comment:
+    types: [created]
   pull_request_target:
     types: [opened]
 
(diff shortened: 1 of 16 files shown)


radumalica:
Added link and content to the documentation.
The test for DNS runs on Docker first, it creates certificates correctly, but the action fails randomly in the middle of processing data.

neilpang:
1. The DNS-API-Test has never run on your fork -- it must pass before review/merge: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test.
2. `_post`'s 5th argument is the Content-Type, not curl options -- every API call here sends `Content-Type: --connect-timeout 10`. Drop that argument.
3. Don't hardcode `_acme-challenge`: the prefix checks, `_sub="_acme-challenge"`, and the literal `"RECORD": "_acme-challenge"` grep in `_find_dns_records` (whose `_name`/`_type` arguments are ignored) all break DNS alias mode, where the record name can be anything. Derive the record name from `fulldomain`.
4. Zone detection strips exactly one label -- `_acme-challenge.www.example.com` yields `www.example.com`, not the zone. Walk the levels probing the API like other hooks' `_get_root`.
5. `rm` deletes every `_acme-challenge` record in the zone and ignores `txtvalue` -- delete only records matching the actual record name and value.
6. Prefix the working variables (`response`, `id`, `value`, `line`, `found_exact`, `removed`, `_sub`, `_data`, `_i`, `_cmd`, ...) with `_hestia_` -- hooks share one shell namespace.
7. The info block is machine-parsed: keep the standard single-line fields (each option on one line with a leading space, `Issues:`, `Docs:` pointing to `dnsapi2#dns_hestiacp`) and move the setup prose to the wiki entry.

radumalica:
opened #7140

PR #6263, opened 2025-03-05
dns_he_ddns: Add empty dns_he_ddns_rm() to remove warnings

markkuleinio:
When the "rm" function is missing, there is stderr output saying "It seems that your API file doesn't define dns_he_ddns_rm", causing some confusion. The function is not doing anything because in HE DDNS API you cannot remove the record with the DDNS key.


markkuleinio:
> Then reply on this message, otherwise, your code will not be reviewed or merged.

Replying as requested.

PR #6296, opened 2025-04-09
dns_1984.hosting.sh: fix logging in by updating session cookie name

paulfertser:
Since at least Mar 29 the 1984.hosting server doesn't return "sessionid" cookie and apparently "cookie1984nammnamm" is used instead for the same purpose.

This results in errors like this
```
One or more cookies are empty: 'csrftoken=JTAJf2Roxl3bwNynzKKIroBqjzgi8d87', ''.
```
and operation fails.

This patch makes DNS TXT method work again with 1984 hosting.

### DIFF
--- a/dnsapi/dns_1984hosting.sh
+++ b/dnsapi/dns_1984hosting.sh
@@ -128,7 +128,7 @@ _1984hosting_login() {
 
   _get "https://1984.hosting/accounts/login/" | grep "csrfmiddlewaretoken"
   csrftoken="$(grep -i '^set-cookie:' "$HTTP_HEADER" | _egrep_o 'csrftoken=[^;]*;' | tr -d ';')"
-  sessionid="$(grep -i '^set-cookie:' "$HTTP_HEADER" | _egrep_o 'sessionid=[^;]*;' | tr -d ';')"
+  sessionid="$(grep -i '^set-cookie:' "$HTTP_HEADER" | _egrep_o 'cookie1984nammnamm=[^;]*;' | tr -d ';')"
 
   if [ -z "$csrftoken" ] || [ -z "$sessionid" ]; then
     _err "One or more cookies are empty: '$csrftoken', '$sessionid'."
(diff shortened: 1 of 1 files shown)


paulfertser:
> **Welcome** First thing: don't send PR to the master branch, please send to the dev branch instead. Please make sure you've read our [DNS API Dev Guide](../wiki/DNS-API-Dev-Guide) and [DNS-API-Test](../wiki/DNS-API-Test). Then reply on this message, otherwise, your code will not be reviewed or merged. Please also make sure to add/update the usage here: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2 We look forward to reviewing your Pull request shortly ✨ 注意: 必须通过了 [DNS-API-Test](../wiki/DNS-API-Test) 才会被 review. 无论是修改, 还是新加的 dns api, 都必须确保通过这个测试.

Hello dear bot and the maintainers. This is a bugfix, not a new DNS API so the guides do not really apply. I've just tested that it fixes the issue with a real life account locally, is that not enough?

paulfertser:
On Wed, Apr 23, 2025 at 11:45:26AM -0700, neil wrote:
> any changes to the hook must pass the test.

Can you please make this requirement explicit in the doc?

Also, do I understand it right that you consider GitHub itself to be
so trusted that one can use production credentials for the DNS API
testing? Or do you recommend everybody to only use a throw-away
account or to probably change password/key after the test is passed?

phedoreanu:
> On Wed, Apr 23, 2025 at 11:45:26AM -0700, neil wrote: any changes to the hook must pass the test.
> Can you please make this requirement explicit in the doc? Also, do I understand it right that you consider GitHub itself to be so trusted that one can use production credentials for the DNS API testing? Or do you recommend everybody to only use a throw-away account or to probably change password/key after the test is passed?

https://checkmarx.com/zero-post/compromised-github-actions-leading-to-credential-leaks/

https://thehackernews.com/2025/03/github-action-compromise-puts-cicd.html

phedoreanu:
tonight-we-test-5c386e

paulfertser:
@Neilpang in my opinion that's not very nice of you to just ignore these security-sensitive questions I raised. Please do educate your contributors about full potential implications of the process you require them to go through.

Neilpang:
@paulfertser   I'm sorry, but I don't get your point.
The password/secert is encrypted by github,  noone else is able to access your secrets.

paulfertser:
> @paulfertser I'm sorry, but I don't get your point. The password/secert is encrypted by github, noone else is able to access your secrets.

Github can, so that's important part of the threat model. Also in comment https://github.com/acmesh-official/acme.sh/pull/6296#issuecomment-2841982142 there are links to research showing how github secrets were exposed to leaks. So it happened before, and it can happen in the future. This needs to be taken into account I think.

PR #6308, opened 2025-04-16
Fix scaleway doc

iaamp:
This updates only the documentation (no code change) of the DNS plugin for Scaleway.

Seemingly there was a change in the API of Scaleway, and while it seems to previously have used an API Token, now it needs the Secret Key of an API Key.

We don't have to change the code here, as it works if just passing the SCW_SECRET_KEY as the existing argument. This is likely ideal, to not break anyones scripted setups.


Neilpang:
Any changes to the hook must be verified by the test.
Even if your chanage is only for doc,  I would recommend you to finish the test.  This is a good way to make sure the hook is still working by now.
Thank

PR #6335, opened 2025-04-26
Add 20i.com DNS and Bearer Token Support for acme.sh

engacs:

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -7577,6 +7577,14 @@ _process() {
         _webroot="$_webroot,$wvalue"
       fi
       ;;
+      --bearer)
+      bearer_value=""
+      if [ "$2" ] && ! _startswith "$2" "-"; then
+        bearer_value="$2"
+        shift
+      fi
+      _bearer="$bearer_value"
+      ;;
     --dnssleep)
       _dnssleep="$2"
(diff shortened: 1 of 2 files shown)


neilpang:
1. Drop the core `--bearer` flag. Provider credentials are passed via provider-specific env vars persisted with `_saveaccountconf_mutable` (your `U20I_Bearer`), never via new acme.sh CLI options -- and a hook must not read the option parser's internal `_bearer`.
2. `local` is not POSIX sh -- hooks run under dash/ash; remove it.
3. Use `_err`/`_info`/`_debug`, not bare `echo`.
4. `_get "https://..." $bearer` does not send the token anywhere -- `_get`'s second argument is the "onlyheader" flag, not an auth token. Export `_H1="Authorization: Bearer ..."` before `_get`/`_post` like other hooks. As written, none of the requests is authenticated.
5. Load/save the token via `_readaccountconf_mutable`/`_saveaccountconf_mutable`, otherwise cron renewal has no credentials.
6. The DNS-API-Test has never run on your fork -- it must pass. A new dnsapi also needs a dedicated tracking issue (`Issues:`) and a dnsapi2 wiki entry (`Docs:`).

PR #6350, opened 2025-05-11
Add WEXBO DNS API

WEXBO:

### DIFF
--- a/dnsapi/dns_wexbo.sh
+++ b/dnsapi/dns_wexbo.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_wexbo_info='wexbo.com
+Site: wexbo.com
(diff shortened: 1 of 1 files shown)


neilpang:
1. The DNS-API-Test has never run on your fork -- configure the secrets and post the link to a green run: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test.
2. `_wexbo_init`'s return code is ignored -- with no token set, `add`/`rm` continue with an empty `Authorization` header. Call it as `_wexbo_init || return 1`.
3. `rm` ignores `txtvalue` and DELETEs by name only: a wildcard cert puts two TXT values on the same name, so this wipes both, and the second `rm` call then fails. Delete by value if the API supports it, and treat a missing record as success (return 0) -- rm is cleanup and must be idempotent.
4. shfmt will fail: tab indentation, trailing semicolons, `dns_wexbo_add(){` spacing, and the missing final newline. Run `shfmt -i 2 -w dnsapi/dns_wexbo.sh`.

PR #6373, opened 2025-05-26
Added dns provider for "Wärner Technologie Services"

LukasWRN:
This PR adds support for the dns-api of "[Wärner Technologie Services](https://www.waerner-techservices.de)"


LukasWRN:
error

PR #6379, opened 2025-05-28
Add 'localcopy' deploy-hook

invario:
PR created to add a `localcopy` deploy-hook that allows users to simply copy the certificates to a target, and then execute a `reloadcmd`.

This was created in anticipation of being used with the `multideploy` deploy-hook in PR #6241.

For example, I have a setup that requires deployment to several docker services in conjunction with two instances of Truenas (Core and Scale).  The TrueNAS deploy-hook is out of date (it works to update the GUI on both Core and Scale) but certain functionality no longer works 100% (it can't even detect Core vs Scale anymore).  There is also an undocumented TrueNAS WS (Websocket) deploy-hook but I believe that only works if running on the same host, not remotely.  TrueNAS users like myself benefit from using https://github.com/danb35/deploy-freenas instead allows for both local and remote deployment.

```
# DEPLOY_LOCALCOPY_CERTIFICATE - /path/to/target/cert.cer
# DEPLOY_LOCALCOPY_CERTKEY - /path/to/target/cert.key
# DEPLOY_LOCALCOPY_FULLCHAIN - /path/to/target/fullchain.cer
# DEPLOY_LOCALCOPY_CA - /path/to/target/ca.cer
# DEPLOY_LOCALCOPY_RELOADCMD - "echo 'this is my cmd'"
```

In my own use case, my `DEPLOY_LOCALCOPY_RELOADCMD` is `/root/deploy-freenas/deploy_freenas.py -c /root/deploy-freenas/deploy_config_truenas2`

Wiki entry added:
https://github.com/acmesh-official/acme.sh/wiki/deployhooks#39-deploy-to-local-directory-via-copy

### DIFF
--- a/deploy/localcopy.sh
+++ b/deploy/localcopy.sh
@@ -21,82 +21,85 @@
 
 #domain keyfile certfile cafile fullchain
 localcopy_deploy() {
-  _cdomain="$1"
-  _ckey="$2"
-  _ccert="$3"
-  _cca="$4"
-  _cfullchain="$5"
+_cdomain="$1"
+_ckey="$2"
+_ccert="$3"
+_cca="$4"
+_cfullchain="$5"
 
-  _debug _cdomain "$_cdomain"
-  _debug _ckey "$_ckey"
-  _debug _ccert "$_ccert"
-  _debug _cca "$_cca"
-  _debug _cfullchain "$_cfullchain"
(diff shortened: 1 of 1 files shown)


invario:
Closed in order to redo the PR properly using a dev fork on my repo.

PR #6386, opened 2025-06-03
dns_ovh specific permission for record removal

amk1969:
The update adds a specific API permissions required to remove the record after domain validation.

### DIFF
--- a/dnsapi/dns_ovh.sh
+++ b/dnsapi/dns_ovh.sh
@@ -224,7 +224,7 @@ _ovh_authentication() {
   _H3=""
   _H4=""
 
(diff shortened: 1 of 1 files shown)


neilpang:
The DNS-API-Test was never run — the DNS workflow on your `ovhupd` branch only hit the `Fail` job (no test secrets configured).

Please follow https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test to run it with your OVH credentials, using a consumer key created with the new access rules, and post the link to the passing run.

amk1969:
> The DNS-API-Test was never run — the DNS workflow on your `ovhupd` branch only hit the `Fail` job (no test secrets configured).

As mentioned in the previous comment, the update is not related to domain validation. It only adds a missing permission into initial, one time setup of the API account. Anyways, for the sake of established rules, I have created a temporary api key and rebased the branch, so the test was happy to complete.

PR #6391, opened 2025-06-09
New dnsapi support for OpenProvider.eu using REST API

Lambiek12:
This add support for [OpenProvider.eu / .com](https://www.openprovider.com/) using their newer REST API.

With this newer version there is no full DNS zone update needed anymore (including existing A, AAAA, CAA, CNAME, MX, NS, PTR, SOA, SRV, and TXT) for a single validation request.

Only the single TXT record is added and removed for the validation request.

_Note: 
- added as new dnsapi option as the original SOAP/XML API is still online and working with [dns_openprovider.sh](https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_openprovider.sh)
- created new fork due to action workflow issues on original PR https://github.com/acmesh-official/acme.sh/pull/6123

### DIFF
--- a/dnsapi/dns_openprovider_rest.sh
+++ b/dnsapi/dns_openprovider_rest.sh
@@ -113,7 +113,7 @@ _openprovider_rest() {
   export _H1="Content-Type: application/json"
   export _H2="Accept: application/json"
   export _H3="Authorization: Bearer $openproviderauthtoken"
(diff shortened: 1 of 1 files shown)


RenevdBroek:
great, missed the reference to the new pull request.

PR #6424, opened 2025-06-30
truenas_ws: Add ability to deploy to remote TrueNAS server via WebSockets

rglidden:
Add support for deploying to remote TrueNAS instances via WebSockets by adding support for the `DEPLOY_TRUENAS_HOSTNAME` and `DEPLOY_TRUENAS_PROTOCOL` options that were already mentioned in the script but not implemented. Defaults to `ws://localhost` to match the existing behaviour if the options are not supplied.

Increased UI restart delay to 15 seconds to give the UI a little more time to restart.

Tested on TrueNAS Scale 24.10.2.2 and 25.04.1

### DIFF
--- a/deploy/truenas_ws.sh
+++ b/deploy/truenas_ws.sh
@@ -39,13 +39,13 @@ _ws_call() {
   _debug "_ws_call arg2" "$2"
   _debug "_ws_call arg3" "$3"
   if [ $# -eq 3 ]; then
-    _ws_response=$(midclt -K "$DEPLOY_TRUENAS_APIKEY" call "$1" "$2" "$3")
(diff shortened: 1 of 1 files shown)


rglidden:
> you need to save and read. see how the other variables are saved and read.

Sorry, I missed that. Updated.

deploy/truenas_ws.sh
```diff
@@ -304,7 +322,7 @@ truenas_ws_deploy() {
   _info "Restarting WebUI..."
   _ws_response=$(_ws_call "system.general.ui_restart")
   _info "Waiting for UI restart..."
-  sleep 6
+  sleep 15
```
neilpang: use function `_sleep`

rglidden: Done.

PR #6436, opened 2025-07-07
Add 'localcopy' deploy-hook 

invario:
Deploy-hook to very simply copy files to set directories and then execute whatever reloadcmd the admin needs afterwards. This can be useful for configurations where the "multideploy" hook (in development) is used or when an admin wants ACME.SH to renew certs but needs to manually configure deployment via an external script (e.g. The deploy-freenas script for TrueNAS Core/Scale

https://github.com/danb35/deploy-freenas/

Note: replaces my earlier PR #6379 which I closed.

### DIFF
--- a/deploy/localcopy.sh
+++ b/deploy/localcopy.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env sh
+
+# Deploy-hook to very simply copy files to set directories and then
+# execute whatever reloadcmd the admin needs afterwards. This can be
+# useful for configurations where the "multideploy" hook (in development)
+# is used or when an admin wants ACME.SH to renew certs but needs to
+# manually configure deployment via an external script
+# (e.g. The deploy-freenas script for TrueNAS Core/Scale
+# https://github.com/danb35/deploy-freenas/ )
+#
+#
+# Environment variables to be utilized are as follows:
+#
+# DEPLOY_LOCALCOPY_CERTIFICATE - /path/to/target/cert.cer
+# DEPLOY_LOCALCOPY_CERTKEY - /path/to/target/cert.key
+# DEPLOY_LOCALCOPY_FULLCHAIN - /path/to/target/fullchain.cer
+# DEPLOY_LOCALCOPY_CA - /path/to/target/ca.cer
+# DEPLOY_LOCALCOPY_RELOADCMD - "echo 'this is my cmd'"
+
+########  Public functions #####################
+
+#domain keyfile certfile cafile fullchain
+localcopy_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
+  _debug _cfullchain "$_cfullchain"
+
+  _getdeployconf DEPLOY_LOCALCOPY_CERTIFICATE
+  _getdeployconf DEPLOY_LOCALCOPY_CERTKEY
+  _getdeployconf DEPLOY_LOCALCOPY_FULLCHAIN
+  _getdeployconf DEPLOY_LOCALCOPY_CA
+  _getdeployconf DEPLOY_LOCALCOPY_RELOADCMD
+
+  if [ "$DEPLOY_LOCALCOPY_CERTIFICATE" ]; then
+    _info "Copying certificate"
+    _debug "Copying $_ccert to $DEPLOY_LOCALCOPY_CERTIFICATE"
+    if ! eval "cp $_ccert $DEPLOY_LOCALCOPY_CERTIFICATE"; then
+      _err "Failed to copy certificate, aborting."
+      return 1
+    fi
+    _savedeployconf DEPLOY_LOCALCOPY_CERTIFICATE "$DEPLOY_LOCALCOPY_CERTIFICATE"
+  fi
+
+  if [ "$DEPLOY_LOCALCOPY_CERTKEY" ]; then
+    _info "Copying certificate key"
+    _debug "Copying $_ckey to $DEPLOY_LOCALCOPY_CERTKEY"
+    if ! eval "cp $_ckey $DEPLOY_LOCALCOPY_CERTKEY"; then
+      _err "Failed to copy certificate key, aborting."
+      return 1
+    fi
+    _savedeployconf DEPLOY_LOCALCOPY_CERTKEY "$DEPLOY_LOCALCOPY_CERTKEY"
+  fi
+
+  if [ "$DEPLOY_LOCALCOPY_FULLCHAIN" ]; then
+    _info "Copying fullchain"
+    _debug "Copying $_cfullchain to $DEPLOY_LOCALCOPY_FULLCHAIN"
+    if ! eval "cp $_cfullchain $DEPLOY_LOCALCOPY_FULLCHAIN"; then
+      _err "Failed to copy fullchain, aborting."
+      return 1
+    fi
+    _savedeployconf DEPLOY_LOCALCOPY_FULLCHAIN "$DEPLOY_LOCALCOPY_FULLCHAIN"
+  fi
+
+  if [ "$DEPLOY_LOCALCOPY_CA" ]; then
+    _info "Copying CA"
+    _debug "Copying $_cca to $DEPLOY_LOCALCOPY_CA"
+    if ! eval "cp $_cca $DEPLOY_LOCALCOPY_CA"; then
+      _err "Failed to copy CA, aborting."
+      return 1
+    fi
+    _savedeployconf DEPLOY_LOCALCOPY_CA "$DEPLOY_LOCALCOPY_CA"
+  fi
+
+  _reload=$DEPLOY_LOCALCOPY_RELOADCMD
+  _debug "Running reloadcmd $_reload"
+
+  if [ -z "$_reload" ]; then
+    _info "Reloadcmd not provided, skipping."
+  else
+    _info "Reloading"
+    if eval "$_reload"; then
+      _info "Reload successful."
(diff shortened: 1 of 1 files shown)


rbicker:
Hi,

We have a use case where we need to deploy a combined PEM file (containing both the full certificate chain and the private key) to multiple web services. While acme.sh itself supports specifying the same path for both `--key-file` and `--fullchain-file` (or `--cert-file`), the current localcopy deploy hook does not account for this scenario—it performs separate copy operations, resulting in one file overwriting the other.

Would it be possible to enhance the hook by adding logic similar to the snippet below (around line 41), to detect when the cert and key targets are the same and generate a proper combined PEM?

```
  _combined_target=""
  _combined_srccert=""

  if [ "$DEPLOY_LOCALCOPY_CERTKEY" ] && \
     { [ "$DEPLOY_LOCALCOPY_CERTKEY" = "$DEPLOY_LOCALCOPY_FULLCHAIN" ] || \
       [ "$DEPLOY_LOCALCOPY_CERTKEY" = "$DEPLOY_LOCALCOPY_CERTIFICATE" ]; }; then

    _combined_target="$DEPLOY_LOCALCOPY_CERTKEY"

    if [ "$DEPLOY_LOCALCOPY_CERTKEY" = "$DEPLOY_LOCALCOPY_FULLCHAIN" ]; then
      _combined_srccert="$_cfullchain"
      DEPLOY_LOCALCOPY_FULLCHAIN=""
    else
      _combined_srccert="$_ccert"
      DEPLOY_LOCALCOPY_CERTIFICATE=""
    fi

    _info "Creating combined PEM at $_combined_target"
    _tmpfile="$(mktemp)"
    if ! cat "$_combined_srccert" "$_ckey" > "$_tmpfile"; then
      _err "Failed to build combined PEM file"
      return 1
    fi
    if ! mv "$_tmpfile" "$_combined_target"; then
      _err "Failed to move combined PEM into place"
      return 1
    fi

    DEPLOY_LOCALCOPY_CERTKEY=""
    _savedeployconf DEPLOY_LOCALCOPY_CERTKEY "$_combined_target"
  fi
```
This change would make the deploy hook more robust in real-world deployment scenarios (e.g. HAProxy), without affecting current behavior for users who specify separate files.

Thanks!

invario:
> PEM


```
    if ! cat "$_combined_srccert" "$_ckey" > "$_tmpfile"; then
      _err "Failed to build combined PEM file"
      return 1
    fi
    if ! mv "$_tmpfile" "$_combined_target"; then
      _err "Failed to move combined PEM into place"
      return 1
    fi
```

Thanks for the suggestion! Is there any reason a temp file has to be used as opposed to just concatenating directly to the target?

invario:
> Hi Please excuse my late response, unfortunately I have missed your reply. I opted for using a temporary file mainly as a safety measure, to avoid partially written or corrupted files in case something goes wrong during the concatenation (e.g. disk full, permissions, interruption, etc.).

I used your code and made some changes and incorporated it.  Let me know if it works for you, thanks!

rbicker:
Thanks for the quick reply!
Good point about the existing file. It should be easy to handle by checking if the target file already exists before copying attributes, for example:

```bash
if [ -e "$_combined_target" ]; then
  cp --attributes-only "$_combined_target" "$_tmpfile" || _err "Failed to copy file attributes"
fi
```


That way, ownership and permissions are preserved when a file already exists (e.g., during renewal), without needing to rely on reloadcmd. Also, other deploy hooks like nginx or apache already handle permissions internally, so keeping that behavior consistent here would avoid confusion for users who expect similar handling.

And got it about the PR process — I’ll keep an eye out for @Neilpang's review. Thanks again for your work on this hook; it’s been really helpful in my setup!

invario:
> Thanks for the quick reply! Good point about the existing file. It should be easy to handle by checking if the target file already exists before copying attributes, for example:
> 
> ```shell
> if [ -e "$_combined_target" ]; then
>   cp --attributes-only "$_combined_target" "$_tmpfile" || _err "Failed to copy file attributes"
> fi
> ```
> 
> That way, ownership and permissions are preserved when a file already exists (e.g., during renewal), without needing to rely on reloadcmd. Also, other deploy hooks like nginx or apache already handle permissions internally, so keeping that behavior consistent here would avoid confusion for users who expect similar handling.
> 
> And got it about the PR process — I’ll keep an eye out for @Neilpang's review. Thanks again for your work on this hook; it’s been really helpful in my setup!

Sorry for the delay in getting back to you.  So as it turns out, `cp` should preserve existing permissions when overwriting the file, but I'm using `mv` and that's why the permissions are being lost. Any reason you see why I can't just use `cp` to overwrite the existing permissions, and then subsequently `rm` the temp file?

meineerde:
Generally, I'd prefer you just use a plain `cat "..." > "$_combined_target"` (or the equivalent `echo`) and just report any errors happening there. That way, existing permissions and ownerships of the target file  would be fully preserved in a way which may not be possible otherwise. This also doesn't require to write into the target file's directory but just to the file itself, allowing minimal permissions to be set on the target.

If this is not desired, many operating systems (at least Linux and MacOS) also offer the [`install` tool](https://linux.die.net/man/1/install) to copy or link files while preserving or explicitly setting permissions, owner, ... This may be more robust than a `cp` or `mv`.

In any case, you should make sure that any file containing private key material is set to file mode `600` to avoid the key material to be world-readable. acme.sh itself uses the following approach in the `_installcert` function which I think should work fine (apart from the TOCTOU race-conditions which are hard to avoid in plain shell without using the `install` command):

```
if [ -f "$target" ]; then
  cat "$source >"$target" || return 1
else
  touch "$target" || return 1
  chmod 600 "$target"
  cat "$source" >"$target" || return 1
fi
```

deploy/localcopy.sh
```diff
@@ -0,0 +1,197 @@
...
+        chmod 600 "$DEPLOY_LOCALCOPY_CERTIFICATE"
+      ); then
+        _err "Failed to copy certificate, aborting."
+        return 1
+      fi
+    fi
+    if ! cat "$_ccert" >"$DEPLOY_LOCALCOPY_CERTIFICATE"; then
+      _err "Failed to copy certificate, aborting."
+      return 1
+    fi
```
hoser21, line 87: When `$DEPLOY_LOCALCOPY_CERTIFICATE != $DEPLOY_LOCALCOPY_CERTKEY`, I think the certificate path needs to be saved to the config. Otherwise, the certificate won't be copied on the deploy hook.

```suggestion
    fi
    _savedeployconf DEPLOY_LOCALCOPY_CERTIFICATE "$DEPLOY_LOCALCOPY_CERTIFICATE"
```

invario: Absolutely right, thank you!

meineerde:
> Pushed update. Also removed use of temp file.

For files which do not contain key material, the permissions are likely too restrictive now as you now create all files with mode `600`.

As mentioned a few times, the `_installcert` function in acme.sh does this differently. See https://github.com/acmesh-official/acme.sh/blob/1bd2922bc37cddba97765af2ae12ad5441c91a74/acme.sh#L6039-L6088

For the following files in the localcopy deploy hook, acme.sh would just uses `cp` instead of `touch` / `chmod` / `cat`:

* `$DEPLOY_LOCALCOPY_CERTIFICATE`
* `$DEPLOY_LOCALCOPY_FULLCHAIN`
* `$DEPLOY_LOCALCOPY_CA`

In addition to that, your `touch` / `chmod` approach for files containing key material (`$_combined_target`, `$DEPLOY_LOCALCOPY_CERTKEY`, `$DEPLOY_LOCALCOPY_PFX`) could possibly succeed unexpectedly:

```sh
if ! (
  touch "$DEPLOY_LOCALCOPY_CERTKEY"
  chmod 600 "$DEPLOY_LOCALCOPY_CERTKEY"
); then
  #...
fi
```

With your approach, I believe, the result of the `touch` command is ignored, even if it fails. This may not be extremely bad here, because if `touch` fails, the `chmod` command is also likely to fail.

However, it may still be a good idea to convert this to the acme.sh logic to ensure that the `touch` succeeds and to ignore any chmod errors (in case we are not allowed to set the file mode). If we then can not write to the file with `cat` later, an error is still thrown.

That would make sure that the behavior of acme.sh itself is the same as this hook and to avoid some edge-cases.

invario:
> For files which do not contain key material, the permissions are likely too restrictive now as you now create all files with mode `600`.

> 
> As mentioned a few times, the `_installcert` function in acme.sh does this differently. See
> 
> https://github.com/acmesh-official/acme.sh/blob/1bd2922bc37cddba97765af2ae12ad5441c91a74/acme.sh#L6039-L6088
> 
> For the following files in the localcopy deploy hook, acme.sh would just uses `cp` instead of `touch` / `chmod` / `cat`:
> 
> * `$DEPLOY_LOCALCOPY_CERTIFICATE`
> * `$DEPLOY_LOCALCOPY_FULLCHAIN`
> * `$DEPLOY_LOCALCOPY_CA`
> 
> In addition to that, your `touch` / `chmod` approach for files containing key material (`$_combined_target`, `$DEPLOY_LOCALCOPY_CERTKEY`, `$DEPLOY_LOCALCOPY_PFX`) could possibly succeed unexpectedly:
> 
> ```shell
> if ! (
>   touch "$DEPLOY_LOCALCOPY_CERTKEY"
>   chmod 600 "$DEPLOY_LOCALCOPY_CERTKEY"
> ); then
>   #...
> fi
> ```
> 
> With your approach, I believe, the result of the `touch` command is ignored, even if it fails. This may not be extremely bad here, because if `touch` fails, the `chmod` command is also likely to fail.
> 
> However, it may still be a good idea to convert this to the acme.sh logic to ensure that the `touch` succeeds and to ignore any chmod errors (in case we are not allowed to set the file mode). If we then can not write to the file with `cat` later, an error is still thrown.
> 
> That would make sure that the behavior of acme.sh itself is the same as this hook and to avoid some edge-cases.

I looked at the code, it uses `cp` only for the backup operation, but `cat` for the rest, so I followed that and continued to use `cat`. I think this hook should basically be the same now in terms of the logic?

I appreciate the feedback, thanks!

invario:
Thanks for the merge!

PR #6442, opened 2025-07-10
Support certificate profile selection

### LINKED ISSUE #6193: Feature Request: Support ACME Profiles
Could acme.sh consider supporting ACME profiles?
Currently the profile is only available on Let's Encrypt's staging endpoint and still a draft on IETF, but it's nice if acme.sh would be able to support this feature when/if the draft would made it to published. 
https://letsencrypt.org/2025/01/09/acme-profiles/

JensSpanier:
In preparation for the [upcoming support of IP address certificates at Letsencrypt](https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate/), I have added the option to [select certificate profiles](https://letsencrypt.org/2025/01/09/acme-profiles/).

So if you want to issue an IP address certificate, this is the command for Letsencrypt (staging):
`acme.sh --issue -d 203.0.113.195 -w /home/username/public_html --server letsencrypt_test --certificate-profile shortlived --days 6`

Closes #6193

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4416,6 +4416,7 @@ issue() {
   _preferred_chain="${15}"
   _valid_from="${16}"
   _valid_to="${17}"
+  _certificate_profile="${18}"
 
   if [ -z "$_ACME_IS_RENEW" ]; then
     _initpath "$_main_domain" "$_key_length"
@@ -4490,6 +4491,11 @@ issue() {
     _savedomainconf "Le_Preferred_Chain" "$_preferred_chain" "base64"
   else
     _cleardomainconf "Le_Preferred_Chain"
+  fi
+  if [ "$_certificate_profile" ]; then
+    _savedomainconf "Le_Certificate_Profile" "$_certificate_profile"
+  else
+    _cleardomainconf "Le_Certificate_Profile"
   fi
 
   Le_API="$ACME_DIRECTORY"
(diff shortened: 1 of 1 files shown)


trieloff:
thank you, this looks interesting. I hope the maintainers will get around to a review.

acme.sh
```diff
@@ -7689,6 +7702,10 @@ _process() {
       _valid_to="$2"
       shift
       ;;
+    --certificate-profile)
```
neilpang: please give it another short name, eg:
```
--certificate-profile | --profile
```

JensSpanier: Done. However, I am a little concerned that it might appear as if `--profile` and `--no-profile` belong together.

neilpang: yes,  please change it to `--cert-profile`

Neilpang:
please also add the usage here: https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L6962

Neilpang:
please add a new page in the wiki:
https://github.com/acmesh-official/acme.sh/wiki

acme.sh
ghen2: Will the stored config file option stay `Le_Certificate_Profile`? (for people already using this patch)

neilpang: yes, the name Le_Certificate_Profile doesn't matter.  whatever you like.

acme.sh
```diff
@@ -6996,6 +7006,8 @@ Parameters:
                                       If no match, the default offered chain will be used. (default: empty)
                                       See: $_PREFERRED_CHAIN_WIKI
 
+  --certificate-profile <profile>   If the CA offers profiles, select the desired profile
```
neilpang: add the short name here too: `--cert-profile`

acme.sh
JensSpanier: Changed it to `--certificate-profile | --cert-profile`

acme.sh
JensSpanier: Done. Also added the link to [wiki](https://github.com/acmesh-official/acme.sh/wiki/Profile-selection)

PR #6453, opened 2025-07-17
dns_cyon: invalid regexes

micha4w:
This is a very simple PR

The dns_cyon script did not work for me, turns out the regexes used are invalid.

(no error message were shown because _egrep_o pipes errors to /dev/null)


neilpang:
Already fixed in dev by 282b0485 -- the identical change. Thanks anyway; closing as superseded.

PR #6473, opened 2025-08-12
Add sleep before each REST call to Constellix to prevent rate limit

woutd:
Constellix is getting more strict for many requests per second. This PR adds a sleep function before each REST request like many other DNS API's already do.


Neilpang:
https://github.com/woutd/acme.sh/actions/runs/16904824086

PR #6485, opened 2025-08-20
Fix ipv6 cert cannot be found

Fangliding:
the old code only check the folder with "." `for di in "${CERT_HOME}"/*.*/; do`
if we have a ipv6 ip cert like this `2001:db8::_ecc`
because it doesn't contain "." it cannot be found and renewed

### DIFF
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -232,7 +232,7 @@ jobs:
         envs: 'TEST_DNS TestingDomain TEST_DNS_NO_WILDCARD TEST_DNS_NO_SUBDOMAIN TEST_DNS_SLEEP CASE TEST_LOCAL DEBUG http_proxy https_proxy TokenName1 TokenName2 TokenName3 TokenName4 TokenName5 ${{ secrets.TokenName1}} ${{ secrets.TokenName2}} ${{ secrets.TokenName3}} ${{ secrets.TokenName4}} ${{ secrets.TokenName5}}'
         prepare: pkg install -y socat curl
         usesh: true
-        copyback: false
+        sync: nfs
         run: |
           if [ "${{ secrets.TokenName1}}" ] ; then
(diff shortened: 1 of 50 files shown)


Fangliding:
But this will cause the log to display some incorrect folders (although they are safely skipped)
```
[Wed Aug 20 05:53:38 AM BST 2025] Renewing: 'ca'
[Wed Aug 20 05:53:38 AM BST 2025] 'ca' is not an issued domain, skipping.
[Wed Aug 20 05:53:38 AM BST 2025] Skipped ca
[Wed Aug 20 05:53:38 AM BST 2025] Renewing: 'deploy'
[Wed Aug 20 05:53:38 AM BST 2025] 'deploy' is not an issued domain, skipping.
[Wed Aug 20 05:53:38 AM BST 2025] Skipped deploy
[Wed Aug 20 05:53:38 AM BST 2025] Renewing: 'dnsapi'
[Wed Aug 20 05:53:38 AM BST 2025] 'dnsapi' is not an issued domain, skipping.
[Wed Aug 20 05:53:38 AM BST 2025] Skipped dnsapi
[Wed Aug 20 05:53:38 AM BST 2025] Renewing: 'notify'
[Wed Aug 20 05:53:38 AM BST 2025] 'notify' is not an issued domain, skipping.
[Wed Aug 20 05:53:38 AM BST 2025] Skipped notify
```
shall we skip this 4 folders?

Fangliding:
> Just a thought but as a naive approach could you just glob over `"${CERT_HOME}"/*[.:]*/; do` in order to get directories containing . or : since IPv6 addresses likely have a : in them?

Is all shell can support regex?

coderjoe:
This is not a regular expression. It is a directory pattern match using globbing.
You can find out more in the bash manual under the topic of pattern matching.

It is pretty widely supported among shells and at very least works in bash, dash, and sh which are listed as this project's supported shells.

Issue #4519, opened 2023-02-24
Report bugs to Nanelo DNS API


AnTheMaker:
Please report bugs you come across when using the [Nanelo](https://nanelo.com) DNS Integration here.

AnTheMaker:
Good catch, @QuadPiece - thank you! I've fixed it and created a PR which adds support for team API keys!

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

PR #6504, opened 2025-09-07
[Important DNS Provider Update] Support Nanelo DNS Team- & Workspace-specific API keys

AnTheMaker:
Nanelo Team- and Workspace-specific API keys require the "domain" parameter to be set containing the DNS zone name (unlike the Domain-specific API keys). So I've added a function to detect the root DNS zone and set the required parameter as described here: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide#3-detect-which-part-is-your-root-zone

### DIFF
--- a/dnsapi/dns_nanelo.sh
+++ b/dnsapi/dns_nanelo.sh
@@ -27,8 +27,16 @@ dns_nanelo_add() {
   fi
   _saveaccountconf_mutable NANELO_TOKEN "$NANELO_TOKEN"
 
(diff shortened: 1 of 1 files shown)


AnTheMaker:
This addresses https://github.com/acmesh-official/acme.sh/issues/4519#issuecomment-3258720182. The fix seems to work fine, I've tested it locally and all CI checks have passed as well.

Neilpang:

PR #6528, opened 2025-09-24
dns_infoblox docs: fix typo in error message and add option to docs

fadecore:
## Problem
The error message for the infoblox server variable has a typo. If you use the variable name from the error you are not using the correct one with capital S -> Infoblox_**S**erver .  
There is a top area that defines available options, but the **Infoblox_View** is missing as potential Option that can be set.

## Solution
Change the error message and add the Infoblox_View option in the header of the script.


fadecore:
not opened correctly

PR #6547, opened 2025-10-02
Fix typo (var "excapedAlgnames" -> "escapedAltnames")

joellimberg:


neilpang:
The base is stale -- this block was restructured on dev and the variable now appears in 4 places, so the PR no longer applies cleanly. Renamed all occurrences in https://github.com/acmesh-official/acme.sh/commit/2058a77d, credited to you. Closing.

PR #6556, opened 2025-10-09
Add deploy hook for KeyHelp

JensSpanier:
Although the [wiki](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#33-deploy-your-cert-to-keyhelp) states that KeyHelp is already supported, there is no deploy hook for it yet (see #5061).

This hook uses [KeyHelp's API](https://app.swaggerhub.com/apis-docs/keyhelp/api/2.12) and supports deployment to multiple KeyHelp instances.
First, an attempt is made to update an existing certificate based on the name. If this fails because the certificate cannot be found, a new certificate is created.

I will update the wiki as soon as the PR is merged.

### DIFF
--- a/deploy/keyhelp.sh
+++ b/deploy/keyhelp.sh
@@ -50,7 +50,7 @@ keyhelp_deploy() {
 
   for _host in $_hosts; do
     _key="$(_getfield "$_keys" "$_i" " ")"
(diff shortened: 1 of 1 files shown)


Neilpang:
yes,  use a new script name.

Issue #6554, opened 2025-10-07
Migrate to the new Hetzner DNS API


jooola:
Dear maintainers,

This project holds a DNS provider for the old Hetzner DNS API.

A new [Hetzner DNS API](https://docs.hetzner.cloud/reference/cloud#dns) has been [released](https://docs.hetzner.cloud/changelog#2025-10-07-dns-beta).

We recommend any consumer of the old [Hetzner DNS API](https://dns.hetzner.com/api-docs) to migrate to the new one.

The new Hetzner DNS API is currently in **beta**, which will likely end (reaching GA) on 10 November 2025. After the beta ended, it will no longer be possible to create new zones in the old DNS API (the old API will still be functional until May 2026). See the [Hetzner DNS beta FAQ](https://docs.hetzner.com/networking/dns/faq/beta) for more details.

Links:

- Hetzner DNS documentation: https://docs.hetzner.com/networking/dns
- Hetzner DNS API reference: https://docs.hetzner.cloud/reference/cloud#dns
- User migration guide to the new DNS system: https://docs.hetzner.com/networking/dns/migration-to-hetzner-console/process

If you have any additional questions, please let me know.

PR #6563, opened 2025-10-13
Add Hetzner Cloud DNS provider (dns_hetznercloud)

PKlapp:
This PR adds a new DNS provider `dns_hetznercloud` for the **Hetzner Cloud DNS API**, addressing [#6554](https://github.com/acmesh-official/acme.sh/issues/6554).
It is separate from the existing `dns_hetzner` provider (legacy/console API) and does not modify it.

**Why**
Hetzner operates two distinct DNS APIs (Console/legacy vs. Cloud). This provider targets the Cloud DNS API to avoid confusion and enable users who manage DNS via the Cloud API.

**Implementation**

* Pure shell using only acme.sh helper functions (no external dependencies).
* Default TTL for new TXT RRsets: **120s** (overridable).
* Configurable API base URL (defaults to `https://api.hetzner.cloud/v1`).

**CI**

* DNS-API workflow passes on a forked run.

**Compatibility**

* No changes to `dns_hetzner`.
* New provider: `dns_hetznercloud`.

---

#### Hetzner **Cloud** DNS — `dns_hetznercloud`

Provider for the Hetzner **Cloud** DNS API (not the legacy/console API).

**Environment variables**

* `HETZNER_TOKEN` *(required)* — Personal Access Token for the Cloud DNS API (**read & write**).
* `HETZNER_TTL` *(optional)* — TTL in seconds for TXT RRsets (default **120**).
* `HETZNER_API` *(optional)* — Override API base URL (default `https://api.hetzner.cloud/v1`).

**Example**

```bash
export HETZNER_TOKEN="your_token_here"
# optional:
# export HETZNER_TTL=120
# export HETZNER_API="https://api.hetzner.cloud/v1"

acme.sh --issue --dns dns_hetznercloud -d test.example.com
```

### DIFF
--- a/dnsapi/dns_hetznercloud.sh
+++ b/dnsapi/dns_hetznercloud.sh
@@ -8,11 +8,13 @@ Options:
 Optional:
  HETZNER_TTL Custom TTL for new TXT rrsets (default 120)
  HETZNER_API Override API endpoint (default https://api.hetzner.cloud/v1)
+ HETZNER_MAX_ATTEMPTS Number of 1s polls to wait for async actions (default 120)
 Issues: github.com/acmesh-official/acme.sh/issues
 '
 
 HETZNERCLOUD_API_DEFAULT="https://api.hetzner.cloud/v1"
 HETZNERCLOUD_TTL_DEFAULT=120
+HETZNER_MAX_ATTEMPTS_DEFAULT=120
 
 ########  Public functions #####################
 
@@ -57,6 +59,9 @@ dns_hetznercloud_add() {
 
   case "${_hetznercloud_last_http_code}" in
   200 | 201 | 202 | 204)
+    if ! _hetznercloud_handle_action_response "TXT record add"; then
+      return 1
+    fi
     _info "Hetzner Cloud TXT record added."
     return 0
     ;;
@@ -116,6 +121,9 @@ dns_hetznercloud_rm() {
     fi
     case "${_hetznercloud_last_http_code}" in
     200 | 201 | 202 | 204)
+      if ! _hetznercloud_handle_action_response "TXT record remove"; then
+        return 1
+      fi
       _info "Hetzner Cloud TXT record removed."
       return 0
       ;;
@@ -170,6 +178,17 @@ _hetznercloud_init() {
     return 1
   fi
   _saveaccountconf_mutable HETZNER_TTL "${HETZNER_TTL}"
+
+  HETZNER_MAX_ATTEMPTS="${HETZNER_MAX_ATTEMPTS:-$(_readaccountconf_mutable HETZNER_MAX_ATTEMPTS)}"
+  if [ -z "${HETZNER_MAX_ATTEMPTS}" ]; then
+    HETZNER_MAX_ATTEMPTS="${HETZNER_MAX_ATTEMPTS_DEFAULT}"
+  fi
+  attempts_check=$(printf "%s" "${HETZNER_MAX_ATTEMPTS}" | tr -d '0-9')
+  if [ -n "${attempts_check}" ]; then
+    _err "HETZNER_MAX_ATTEMPTS must be an integer value."
+    return 1
+  fi
+  _saveaccountconf_mutable HETZNER_MAX_ATTEMPTS "${HETZNER_MAX_ATTEMPTS}"
 
   return 0
 }
@@ -290,9 +309,15 @@ _hetznercloud_parse_zone_fields() {
   if [ -z "${zone_id}" ] || [ -z "${zone_name}" ]; then
     return 1
   fi
+  zone_name_trimmed=$(printf "%s" "${zone_name}" | sed 's/\.$//')
+  if zone_name_ascii=$(_idn "${zone_name_trimmed}"); then
+    zone_name="${zone_name_ascii}"
+  else
+    zone_name="${zone_name_trimmed}"
+  fi
   _hetznercloud_zone_id="${zone_id}"
   _hetznercloud_zone_name="${zone_name}"
-  _hetznercloud_zone_name_lc=$(printf "%s" "${zone_name}" | sed 's/\.$//' | _lower_case)
+  _hetznercloud_zone_name_lc=$(printf "%s" "${zone_name}" | _lower_case)
   return 0
 }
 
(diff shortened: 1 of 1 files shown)


mpkossen:
> Thanks for the suggestion. I’ve added a clear note to the wiki to avoid confusion with the legacy provider.
> 
> It now explicitly states that `dns_hetznercloud` uses the Hetzner Cloud DNS API, requires a new token from the Hetzner Cloud, and that users must set `HETZNER_TOKEN` (not `HETZNER_Token`).
> 
> Do you think I should rename the env var to something even clearer like `HETZNERCLOUD_TOKEN` to further reduce mix-ups?

I have no opinion on that to be honest.

You're a lifesaver, by the way. Your PR saved me a ton of time and headache.

PKlapp:
@Laufwerk
I’m not a maintainer of this project or the OPNsense plugin, so I can’t give a timeline.
I don’t use OPNsense, so I can only share how it works on pfSense+ as a concrete example. If OPNsense is structured in a similar way, this should point you in the right direction.
This manual addition is a temporary workaround until `dns_hetznercloud` is included in the official OPNsense acme plugin. If it works on OPNsense, please drop a short note here so others can follow more easily.

1. Copy `dns_hetznercloud.sh` to
   `/usr/local/pkg/acme/dnsapi/`

2. Add this method to `/usr/local/pkg/acme/acme.inc`:

```php
$acme_domain_validation_method['dns_hetznercloud'] = [
  'name' => 'DNS-Hetzner Cloud',
  'fields' => [
    'HETZNER_TOKEN' => [
      'name' => 'hetzner_token',
      'description' => 'Hetzner Cloud API Token',
      'columnheader' => 'API Token',
      'type' => 'password',
    ],
  ]
];
```

After that, DNS-Hetzner Cloud should be selectable in the GUI.

---

@mpkossen Thanks a lot for the kind words. Happy it saved you time.

dnsapi/dns_hetznercloud.sh
```diff
@@ -0,0 +1,431 @@
...
+  if ! _hetznercloud_api GET "/zones?name=${encoded_candidate}"; then
+    return 1
+  fi
+  if [ "${_hetznercloud_last_http_code}" != "200" ]; then
+    if [ "${_hetznercloud_last_http_code}" = "404" ]; then
+      return 1
+    fi
+    _hetznercloud_log_http_error "Hetzner Cloud zone search failed" "${_hetznercloud_last_http_code}"
+    return 1
+  fi
```
jooola, line 288: I am not sure this is needed, calling `GET /zones/<zone-name>` should already cover getting the zone by its name.

Also  `GET /zones?name=<zone-name>` does an exact match, same as the `GET /zones/<zone-name>` endpoint, so I don't think there is an added benefit.

dnsapi/dns_hetznercloud.sh
```diff
@@ -0,0 +1,431 @@
...
+
+  if ! _hetznercloud_api POST "${_hetznercloud_rrset_action_add}" "${add_payload}"; then
+    return 1
+  fi
+
+  case "${_hetznercloud_last_http_code}" in
+  200 | 201 | 202 | 204)
+    _info "Hetzner Cloud TXT record added."
+    return 0
+    ;;
```
jooola, line 67: `add_records` and `remove_records` are async tasks (actions) triggered by the POST call above.

The returned actions should be waited upon, to make sure the changes are fully applied.

You may use the status field in the action object to check whether it is running, it failed or succeeded.

I recommend to poll the `GET /actions/<action-id>` endpoint to get an update of the running action.

dnsapi/dns_hetznercloud.sh
```diff
@@ -0,0 +1,431 @@
...
+      return 1
+    fi
+    if ! _hetznercloud_api POST "${_hetznercloud_rrset_action_remove}" "${remove_payload}"; then
+      return 1
+    fi
+    case "${_hetznercloud_last_http_code}" in
+    200 | 201 | 202 | 204)
+      _info "Hetzner Cloud TXT record removed."
+      return 0
+      ;;
```
jooola, line 129: `add_records` and `remove_records` are async tasks (actions) triggered by the POST call above.

The returned actions should be waited upon, to make sure the changes are fully applied.

You may use the status field in the action object to check whether it is running, it failed or succeeded.

I recommend to poll the `GET /actions/<action-id>` endpoint to get an update of the running action.

dnsapi/dns_hetznercloud.sh
```diff
@@ -0,0 +1,431 @@
...
+  fi
+
+  if ! _hetznercloud_prepare_zone "${fulldomain}"; then
+    _err "Unable to determine Hetzner Cloud zone for ${fulldomain}"
+    return 1
+  fi
+
+  if ! _hetznercloud_get_rrset; then
+    return 1
+  fi
```
jooola, line 101: I think you should be able to remove this API request, as if the rrset does not exists, the remove_records will also fail with a 404.

dnsapi/dns_hetznercloud.sh
```diff
@@ -0,0 +1,431 @@
...
```
jooola, line 38: I think you should be able to remove this API request, as if the rrset does not exists, the add_records request will automatically create a new one.

PKlapp:
Thanks for the suggestions, I’ve implemented both points. IDNs are converted to punycode for zone and record names before calling the API, and the provider now waits for the returned actions to complete by polling `/actions/<id>` with sensible limits and proper error reporting.

dnsapi/dns_hetznercloud.sh
PKlapp: I kept the extra call as a defensive fallback. First we try `GET /zones/<name>`, and if that returns 404 or isn’t supported we fall back to `GET /zones?name=<name>`.

dnsapi/dns_hetznercloud.sh
PKlapp: Implemented: After `add_records`, the script now waits for the returned actions to complete by polling `/actions/<id>` with sensible limits and proper error reporting.

dnsapi/dns_hetznercloud.sh
PKlapp: Implemented: After `remove_records`, the script now waits for the returned actions to complete by polling `/actions/<id>` with sensible limits and proper error reporting.

dnsapi/dns_hetznercloud.sh
PKlapp: The request is kept to prevent unnecessary failed actions and the ensuing action polling when there is effectively nothing to change.

dnsapi/dns_hetznercloud.sh
PKlapp: The request is kept to prevent unnecessary failed actions and the ensuing action polling when there is effectively nothing to change.

Laufwerk:
> > Unfortunately, there doesn't seem to be an acme.inc file, and the ACME configurations are spread across the entire system. I haven't been able to make the adjustments. I'm hoping for a quick solution from the plugin maintainer.
> 
> @Laufwerk I am interested to know if you did find any solution in the meantime? Waiting for the plugin fix to.

@archham 
Hey, unfortunately, I haven't found a solution in the meantime. The only advice I have is that we shouldn't migrate to the new DNS zone unless we have to.
The plugin owners will certainly take action before the mandatory migration.

dnsapi/dns_hetznercloud.sh
egandro: That fallback makes sense. I think I have seen this also in the cert-manager webhook code that is e.g. used in kubernetes

jooola: > That fallback makes sense. I think I have seen this also in the cert-manager webhook code that is e.g. used in kubernetes

Which webhook are you referring to?

egandro:
> @archham Hey, unfortunately, I haven't found a solution in the meantime. The only advice I have is that we shouldn't migrate to the new DNS zone unless we have to. The plugin owners will certainly take action before the mandatory migration.

We are at the day after :( There is no more old Console support.

Proxmox is also broken: <https://forum.proxmox.com/threads/acme-plugin-hetzner-dns-console-is-moving-to-the-hetzner-console.174363/>

Minzilein2003:
What exactly is preventing the PR from being approved? What is missing/not yet right?

Minzilein2003:
> so if old dns api is gone, why is there no new script for opnsense?

Once the PR is approved, the PR can be approved by OPNsense Integration.
[https://github.com/opnsense/plugins/pull/5020](https://github.com/opnsense/plugins/pull/5020)

egandro:
Thank you very much!

cpzengel:
#bashclub thanks a lot

hebbet:
no you need to generate a new one in console.hetzner.com in the project with your dns settings for the domain.

Security > API Tokens

https://console.hetzner.com/projects/PROJECTID/security/tokens

Issue #6576, opened 2025-10-22
[Question]How many domains can a zeroSSL 90 day wildcard certificate support at most


ningkexin:
I received the following error directly when applying for 16 domain names
```bash
The retryafter=86400 value is too large (> 600), will not retry anymore.
```

ignoramous:
Looks related to fixes made for #5293 as the `retryAfter` header set by ZeroSSL REST API goes up to `86400` seconds (equals 1 day), which prompts acme.sh (v3.1.0 onward) to exit.

```diff
2025-10-29T04:06:46.9009483Z [Wed Oct 29 04:06:46 UTC 2025] Processing. The CA is processing your order, please wait. (1/30)
2025-10-29T04:06:46.9035516Z [Wed Oct 29 04:06:46 UTC 2025] Sleep 2 seconds before verifying again
2025-10-29T04:06:48.9082896Z [Wed Oct 29 04:06:48 UTC 2025] Checking
2025-10-29T04:06:48.9110363Z [Wed Oct 29 04:06:48 UTC 2025] =======Sending Signed Request=======
2025-10-29T04:06:48.9137583Z [Wed Oct 29 04:06:48 UTC 2025] url='https://acme.zerossl.com/v2/DV90/authz/<deadb33f>'
2025-10-29T04:06:48.9163942Z [Wed Oct 29 04:06:48 UTC 2025] payload
2025-10-29T04:06:48.9189304Z [Wed Oct 29 04:06:48 UTC 2025] Use cached jwk for file: /home/runner/.acme.sh/ca/acme.zerossl.com/v2/DV90/account.key
2025-10-29T04:06:48.9288646Z [Wed Oct 29 04:06:48 UTC 2025] Use _CACHED_NONCE='<deadb33f>'
2025-10-29T04:06:48.9314967Z [Wed Oct 29 04:06:48 UTC 2025] nonce='<deadb33f>'
2025-10-29T04:06:48.9871701Z [Wed Oct 29 04:06:48 UTC 2025] POST
2025-10-29T04:06:48.9895779Z [Wed Oct 29 04:06:48 UTC 2025] _post_url='https://acme.zerossl.com/v2/DV90/authz/<deadb33f>'
2025-10-29T04:06:48.9923939Z [Wed Oct 29 04:06:48 UTC 2025] body='{"protected": "<deadb33f>", "payload": "", "signature": "<defadb33f>"}'
[...]

Issue #5293, opened 2024-09-18
Retry-After headers are ignored during challenge validation


miken32:
More of a feature request than a bug. While checking the status of a processing authorization, `Retry-After` headers that the server sends are ignored. Instead a [fixed 2 second retry interval](https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L5101) is used. There is some code in `_send_signed_request()` to deal with these headers but it's only triggered by [an error response](https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L2254).

The RFC doesn't explicitly require clients to obey the header:
>[7.5.1](https://datatracker.ietf.org/doc/html/rfc8555/#section-7.5.1).  Responding to Challenges
>...
>    To check on the status of an authorization, the client sends a POST-
>    as-GET request to the authorization URL, and the server responds with
>    the current authorization object.  In responding to poll requests
>    while the validation is still in progress, the server MUST return a
>    200 (OK) response and MAY include a Retry-After header field to
>    suggest a polling interval to the client.
> 

But it might be nice. Implementing this would also necessitate a change in the number of max attempts from 30 to something dynamic.

I'd submit a PR but, even as someone pretty familiar with shell scripting, I find this code pretty bewildering.

Debug log
-----------------
Note the response headers contain a retry-after header, but the retry interval is set to 2 seconds.
```
[Wed 18 Sep 2024 17:11:13 MDT] acme.sh:_send_signed_request:2241        responseHeaders='HTTP/2 200 
server: nginx/1.26.2
content-type: application/json
x-powered-by: PHP/8.3.11
retry-after: 30
cache-control: no-cache, private
date: Wed, 18 Sep 2024 23:11:13 GMT
replay-nonce: wpUoY-KwoM_nD4Xdeex2hTF9drgUiw_nEH-3_v-LurI
link: <https://localhost/acme>;rel="index"
'
[Wed 18 Sep 2024 17:11:13 MDT] acme.sh:_send_signed_request:2244        code='200'
[...]

PR #6587, opened 2025-10-29
Let max retries exit the CA processing step instead of retryAfter

ignoramous:
For background see:
- #6576

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2305,7 +2305,7 @@ _send_signed_request() {
           _sleep $_sleep_overload_retry_sec
           continue
         else
-          _info "The retryafter=$_retryafter value is too large (> 600), will not retry anymore."
+          _err "The retryafter=$_retryafter value is too large (> 600), will not retry anymore."
         fi
       fi
       if _contains "$_body" "JWS has invalid anti-replay nonce" || _contains "$_body" "JWS has an invalid anti-replay nonce"; then
@@ -2323,7 +2323,7 @@ _send_signed_request() {
     fi
     return 0
   done
-  _info "Giving up sending to CA server after $MAX_REQUEST_RETRY_TIMES retries."
+  _err "Giving up sending to CA server after $MAX_REQUEST_RETRY_TIMES retries."
   return 1
 
 }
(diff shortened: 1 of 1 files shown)


ignoramous:
> if the retryafter is too large, it usually means the ca is refusing your requests ... Maybe 600 is not that large to give up. how about make it 3600?

What I've found with ZeroSSL is, when there are 8+ domains in a single cert request, it takes a while (I've seen it take 10mins). But their `retrtAfter` is set to `84600` (1 day), which is over-the-top, as the response is usually available within 10mins.

neilpang:
By design, `Retry-After` > 600s means the CA is refusing -- the client must stop, not poll against it. Fast-failing is cheap: the next cron run re-orders and the CA reuses pending authzs. This change would block the serial cron run up to ~29x600s per domain, polling 6x faster than asked, then still fail with Timeout. Not merging; ZeroSSL's `86400` (#6576) needs a different solution.

The log-level fixes are welcome as a separate PR, minus one: the `_err` in `_send_signed_request`'s retryafter branch falls through to `return 0` -- keep `_info` or return 1.

ignoramous:
> This change would block the serial cron run up to ~29x600s per domain, polling 6x faster than asked, then still fail with Timeout

That because the original PR that introduced the bug never implemented the other part:

"But it might be nice. Implementing this would also necessitate a change in the number of max attempts from 30 to something dynamic." https://github.com/acmesh-official/acme.sh/issues/5293#issue-2534935761

> ZeroSSL's 86400 (#6576) needs a different solution.

What solution would that be? We are using 3.0.9 since a year or so due to this and would like to move to newer versions and not maintain a separate patch.

> By design, Retry-After > 600s means the CA is refusing -- the client must stop, not poll against it

Is this in the specs or some arbitrary design choice?

> Fast-failing is cheap: the next cron run re-orders and the CA reuses pending authzs

Not sure what this means, but in our runs, a retry has never picked up previous "pending authzs". We've got a new cert while the previous cert had also been published but our acme.sh process quit due to a high `retry-after` & never later picked up the already issued certs. Clearly, we are doing something wrong. But the assumption that "pending authzs" are picked up (as some guarantee) don't quite hold in the environment we operate in, at least.

PR #6589, opened 2025-10-30
Add priority, tags and title to ntfy notification

de-es:
Make the ntfy.sh notifications easier to distinguish at a first glance.


de-es:
No changes to the wiki necessary, nothing changed here.

PR #6602, opened 2025-11-05
DNS-Challenge: fix DESec.io -> Fix return values in add txt record function

rICTx-T1D:
Changed return values for error handling in txt record addition. DESec.io create txt-record but PUT was indicated as failed

Test pass OK


Neilpang:
why return 0 when there is add error ?

PR #6641, opened 2025-11-26
Add support for TEST_DNS_TXT_LEN secret in DNS API workflow (for acmetest PR)

kubax:
**Summary**

This PR updates the DNS workflow to support the new `TEST_DNS_TXT_LEN` secret introduced in acmetest PR.  
The related changes in the testing framework are implemented in my PR to acmetest:

**acmetest PR:** https://github.com/acmesh-official/acmetest/pull/34

The updated acmetest adds the ability to configure the length of TXT values used during DNS API testing — a requirement for testing systems such as *acme-dns*, which expect TXT records with an exact length. To ensure acme.sh can pass this configuration to all test environments, the workflow needs to be updated accordingly.

---

### 1. Adds support for the `TEST_DNS_TXT_LEN` secret
- Introduces the new secret into all DNS test jobs  
- Ensures that every environment (Linux/Docker, macOS, Windows/Cygwin, BSD variants, Solaris/OmniOS) passes the value through to `letest.sh` or the acmetest Docker runtime  
- Keeps the value optional — if the secret is not defined, the default behavior remains unchanged

### 2. Updates Docker-based testing
- Adds `TEST_DNS_TXT_LEN` to `docker.env` so that the variable is available inside the acmetest container  
- Ensures consistent behavior with the native runner environments

### 3. Updates VM-based testing (FreeBSD, OpenBSD, NetBSD, DragonFlyBSD, Solaris, OmniOS)
- Adds `TEST_DNS_TXT_LEN` to the `envs:` variable list so that the VM receives the configuration

---

## Why This Change Is Needed

The updated acmetest now allows generating TXT values of a configurable size.  
This is required because some DNS ACME challenge implementations — specifically **acme-dns** — need TXT records of a fixed length to function correctly.

Without updating the acme.sh workflow:

- acmetest cannot receive the new configuration  
- TXT length cannot be adjusted during CI  
- DNS API tests relying on acme-dns will not work correctly  
- the CI behavior between Docker and native platforms would become inconsistent

With this PR, the workflow fully supports the new functionality.

---

## Backward Compatibility

- If `TEST_DNS_TXT_LEN` is not defined, all workflows behave exactly as before.  
- The default TXT length remains unchanged internally (handled by acmetest).  
- No changes are required for existing DNS API providers or test setups.

---

## Summary

This PR ensures that acme.sh’s DNS testing workflow is compatible with the upcoming enhancements in acmetest by exposing the new `TEST_DNS_TXT_LEN` secret to all supported test environments.  
It enables full and consistent testing of DNS providers that require fixed-length TXT values while preserving existing behavior for all others.

### DIFF
--- a/.github/workflows/DNS.yml
+++ b/.github/workflows/DNS.yml
@@ -450,7 +450,7 @@ jobs:
       run: cd .. && git clone --depth=1 https://github.com/acmesh-official/acmetest.git  && cp -r acme.sh acmetest/
     - uses: vmactions/solaris-vm@v1
       with:
-        envs: 'TEST_DNS TestingDomain TEST_DNS_NO_WILDCARD TEST_DNS_NO_SUBDOMAIN TEST_DNS_SLEEP CASE TEST_LOCAL DEBUG http_proxy https_proxy TEST_DNS_TXT_LEN TokenName1 TokenName2 TokenName3 TokenName4 TokenName5 ${{ secrets.TokenName1}} ${{ secrets.TokenName2}} ${{ secrets.TokenName3}} ${{ secrets.TokenName4}} ${{ secrets.TokenName5}}'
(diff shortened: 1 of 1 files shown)


Neilpang:
no, this is not necessary,  just make the tested txt length to be exact 43.

Issue #6443, opened 2025-07-12
invalid domain


github2sean:
mydns uesed dnspod and write id and key in .zshrc

execute 
acme.sh --issue --server letsencrypt --dns dns_dp -d '*.demo.com' --key-file '/usr/local/nginx/certs/*.demo.com.key' --fullchain-file '/usr/local/nginx/certs/*.demo.com.crt'

something wrong:
invalid domain
[Sat Jul 12 16:17:25 CST 2025] Error adding TXT record to domain: _acme-challenge.demo.com
[Sat Jul 12 16:17:25 CST 2025] Please add '--debug' or '--log' to see more information.
[Sat Jul 12 16:17:25 CST 2025] See: https://github.com/acmesh-official/acme.sh/wiki/How-to-debug-acme.sh

kshartman:
I did that and it still failing (using dns_aws).  Also dns_aws is calling undefined function _error.  I think it should be _err

kshartman:
> I did that and it still failing (using dns_aws). Also dns_aws is calling undefined function _error. I think it should be _err

My problem was user error (wrong aws key) but dns_aws still is calling _error instead of _err

PR #6644, opened 2025-11-28
fix "dns_aws.sh: line 164: _error: command not found" #6443

SunMar:
Fixing a small typo that causes an error to appear when something is wrong during AWS DNS verification.

```
~/.acme.sh/dnsapi/dns_aws.sh: line 164: _error: command not found
```


SunMar:
Done, tests pass in Actions

PR #6659, opened 2025-12-04
Added Sotoon dns api

erfantkerfan:
Added Sotoon dns api.
CI is ok: https://github.com/erfantkerfan/acme.sh/actions/runs/19917644325
I'm adding this api as a current employee.

### DIFF
--- a/dnsapi/dns_sotoon.sh
+++ b/dnsapi/dns_sotoon.sh
@@ -1,0 +1,319 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_sotoon_info='Sotoon.ir
+Site: Sotoon.ir
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_sotoon
+Options:
+ Sotoon_Token API Token
+ Sotoon_WorkspaceUUID Workspace UUID
+ Sotoon_WorkspaceName Workspace Name
+Issues: github.com/acmesh-official/acme.sh/issues/6656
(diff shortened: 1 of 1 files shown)


erfantkerfan:
Hi there, Any update on this PR?

rfakit:
This PR is important for our production use of acme.sh. Could you please review it when possible or let us know if there are any blockers?

Thank you.

dnsapi/dns_sotoon.sh
```diff
@@ -0,0 +1,300 @@
...
+    _err_sotoon "Error removing record"
+    return 1
+  fi
+}
+
+####################  Private functions below ##################################
+
+_get_root() {
+  domain=$1
+  i=2
```
neilpang: i starts from `1`

erfantkerfan: Starting i=2 is intentional, it skips checking the full subdomain which will never be a domain zone api. This saves one API call per execution. Starting from i=1 would work but wastes a single API request. I checked and other people have used this too with other providers in the code base.

erfantkerfan: changed, Done

erfantkerfan:
I rebased, cleanup the commits and reran the ci jobs. everything ran okay but zerossl jobs fail, since zerossl refuses to generate certificate for .ir domain because of sanctions placed on Iran domain. its done on my side. could you check if you can merge it now?
@Neilpang

PR #6687, opened 2025-12-18
Ssh noscp permissions

kapouer:
When using ssh hook, copying without using "scp", the resulting key file has default permissions, which are, on many servers, world-readable.
This simply fixes that.

Users of DEPLOY_SSH_USE_SCP, will be smart enough to use "scp -p".


kapouer:
Indeed, will fix

dga-nagra:
I had split the change into 2 lines for readability in the meantime.
Sorry, I wasn't expecting a feedback so fast !

The result is the same, I have submitted a PR on the author's repository

kapouer:
@dga-nagra very cool, merged.

PR #6691, opened 2025-12-18
fix: improve Servercow DNS API

DerLinkman:
This pull request refactors and improves the `dns_servercow.sh` script, focusing on better handling of TXT records in the ServerCow DNS API integration, clearer error messages, and minor code cleanup. The changes enhance reliability when adding or removing DNS TXT records and improve user feedback.

### TXT record management improvements
* Enhanced logic in `dns_servercow_add` to correctly handle existing TXT records: it now detects if the content is an array and prepends the new value, or extracts the single value and builds an array if needed. This ensures multiple TXT values are supported for a single subdomain.
* Added explicit debug statements for key operations, such as getting TXT records, adding new records, and deleting records, making troubleshooting easier.

### Error handling and messaging
* Standardized and clarified error messages for missing API credentials and record operation failures, improving user guidance and consistency. 

### Code cleanup and documentation
* Removed outdated usage comments and added clearer function usage documentation at the top of public functions. 
* Minor whitespace and formatting adjustments for improved readability.

It also ran through the DNS Test but failed due to a VM issue i guess (as all previous Systems passed) see: https://github.com/DerLinkman/acme.sh/actions/runs/20341035479

### DIFF
--- a/dnsapi/dns_servercow.sh
+++ b/dnsapi/dns_servercow.sh
@@ -12,26 +12,25 @@ Author: Jens Hartlep
 
 SERVERCOW_API="https://api.servercow.de/dns/v1/domains"
 
-# Usage dns_servercow_add _acme-challenge.www.domain.com "abcdefghijklmnopqrstuvwxyz"
+########  Public functions #####################
+
+#Usage: add  _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
 dns_servercow_add() {
   fulldomain=$1
   txtvalue=$2
 
-  _info "Using servercow"
-  _debug fulldomain "$fulldomain"
-  _debug txtvalue "$txtvalue"
-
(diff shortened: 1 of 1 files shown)


neilpang:
Rebase onto current dev and re-run DNS.yml in your fork -- the January run's only red job was Omnios and its logs have expired, so flake vs real can't be told.

1. shfmt will fail: the file's trailing newline was dropped, and several added blank lines carry trailing whitespace.
2. Wildcard + base domain share one TXT record here; the first `rm` deletes the whole record by name, so the second `rm` for the same name hits "no such record" and returns 1. Treat a missing record as success in `rm`.

neilpang:
The failed run looks self-inflicted rather than a code problem: you pushed twice within a minute, so two DNS runs (17:06:14 and 17:06:41) validated against the same `_acme-challenge` name concurrently and each saw the other's TXT values -- exactly what the error says (`Incorrect TXT record ... (and 1 more)`).

Please: (1) delete any leftover `_acme-challenge` TXT records for the test domain in the Servercow panel, (2) trigger ONE DNS.yml run on the current head and let it finish alone. If a clean solo run still fails with values that match what the log just added, then we look at the array-extraction change.

Issue #6124, opened 2024-12-01
acme.sh image requires root access when using Docker


sebschlicht:
While the acme.sh README explicitly states that no root/sudo access is required, the Docker image apparently can only be run as root in Docker.

The image does not respect PUID/PGID environment variables.
More importantly, the `acme.sh` binaries become inaccessible when using other means to go rootless (e.g. Docker's `user` directive).

This comes with some additional security threats (e.g. container escapes would grant root access to the host) and all acquired certificates are owned by root.

Steps to reproduce
------------------
1. `docker run -u "1000:1000" --rm neilpang/acme.sh`

Debug log
-----------------
Not applicable, acme.sh can not be called. CLI output:

```
/usr/local/bin/--help: line 2: /root/.acme.sh/acme.sh: Permission denied
```

dzamsari:
I am also facing the same issue. The container immediately exited if I use --user flag and when I run `docker logs acme.sh`, this is the output: `crond: can't open or create /var/run/crond.pid: Permission denied`

Neilpang:
fixed,  try it again.

invario:
Submitted PR #6709 which uses supercronic, open to feedback, suggestions, etc.

invario:
@github-cli #6709 was merged into `master` and is in the `latest` Docker image. I strongly suggest reading the additional information here regarding non-root deployment to see if the changes impact you.
https://github.com/acmesh-official/acme.sh/wiki/Run-acme.sh-in-docker#3-run-acmesh-as-a-docker-daemon

github-cli:
I have /acme.sh/crontab which is the same crontab I had before which worked fine but now it does not execute the crontab entries I have, neither the @reboot nor the others, the following example entry is not executed or rather does not show up in the docker logs but did when I used normal crontab as root, executing the commands manually as non-root user works fine...
here is one example entry that does not work:
```
* * * * * echo "=== test ===" >> /proc/1/fd/1 2>&1
```
EDIT: of course all files that are mounted (/acme.sh) belong to the acme user and this is the docker logs output:
```
acme.sh  | Running Supercronic using crontab at /acme.sh/crontab
acme.sh  | time="2026-03-14T16:31:23+01:00" level=info msg="reaping dead processes"
acme.sh  | time="2026-03-14T16:31:23+01:00" level=info msg="read crontab: /acme.sh/crontab"

```

github-cli:
[...]

PR #6709, opened 2025-12-23
Docker with non-root using supercronic

invario:
Replaces cronie with supercronic to allow non-root users to have cronjobs. 'crontab' is generated in LE_CONFIG_HOME which is used by supercronic.

Note that `acme.sh --installcronjob` and `--uninstallcronjob` when run as a non-root user will fail but neither of should be used in `daemon` mode anyway.

**Additional considerations:**

1. All users: if you are using the `docker` deploy-hook and therefore mounting `/var/run/docker.sock`, you must ensure your non-root user has permission to read/write to `/var/run/docker.sock`. I was able to do this by runner the container with `user: 1000:999` (my host's GID for docker is 999). If using Docker Compose, use `GROUP_ADD` to accomplish the same thing.
2. Existing users: if you are using the `ssh` deploy-hook, take note of where your `.ssh` keys were stored. **LE_CONFIG_HOME (/acme.sh) is used as home for the image, so keys will be stored in `/acme.sh/.ssh` regardless of the user.** Previously, you may have stored them in `/root/.ssh`.  Depending on the user's configuration, this is potentially a breaking change for `ssh`.
~~3.The image runs, by default, as non-root `USER '1000:1000'` is set in the `Dockerfile`.~~
4. All users:  if `crontab` is missing, `entry.sh` will generate one in LE_CONFIG_HOME (/acme.sh). Since this is a VOLUME, subsequent changes to `crontab` made by the user will persist.
5. All users: User must specify UID 1000 when running the container. The image still defaults to root.
6. Existing users: User needs to ensure /acme.sh has proper read/write permissions for the non-root user.

Resolves:
#6124

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,8 +34,7 @@ RUN addgroup -g 1000 acme && adduser -h $LE_CONFIG_HOME -s /bin/sh -G acme -D -H
 
 RUN cd /install_acme.sh && ([ -f /install_acme.sh/acme.sh ] && /install_acme.sh/acme.sh --install || curl https://get.acme.sh | sh) && rm -rf /install_acme.sh/
 
-RUN ln -s $LE_WORKING_DIR/acme.sh /usr/local/bin/acme.sh \
-  && crontab -l | grep acme.sh | sed 's#> /dev/null##' > $LE_CONFIG_HOME/crontab
+RUN ln -s $LE_WORKING_DIR/acme.sh /usr/local/bin/acme.sh
 
 RUN for verb in help \
   version \
@@ -74,6 +73,13 @@ RUN for verb in help \
 
 RUN printf "%b" '#!'"/usr/bin/env sh\n \
 if [ \"\$1\" = \"daemon\" ];  then \n \
+  if [ ! -f \"\$LE_CONFIG_HOME/crontab\" ]; then \n \
+     echo \"\$LE_CONFIG_HOME/crontab not found, generating one\" \n \
+     time=\$(date -u \"+%s\") \n \
+     random_minute=\$((\$time % 60)) \n \
+     random_hour=\$((\$time / 60 % 24)) \n \
+     echo \"\$random_minute \$random_hour * * * \\\"\$LE_WORKING_DIR\\\"/acme.sh --cron --home \\\"\$LE_WORKING_DIR\\\" --config-home \\\"\$LE_CONFIG_HOME\\\"\" > \"\$LE_CONFIG_HOME\"/crontab \n \
+  fi \n \
   echo \"Running Supercronic using crontab at \$LE_CONFIG_HOME/crontab\" \n \
   exec -- /usr/bin/supercronic \"\$LE_CONFIG_HOME/crontab\" \n \
 else \n \
@@ -81,8 +87,6 @@ else \n \
 fi\n" >/entry.sh && chmod +x /entry.sh && chmod -R o+rwx $LE_WORKING_DIR && chmod -R o+rwx $LE_CONFIG_HOME
 
 VOLUME /acme.sh
-
-USER 1000:1000
 
 ENTRYPOINT ["/entry.sh"]
 CMD ["--help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN apk --no-cache add -f \
   libidn \
   jq \
   yq-go \
-  cronie
+  supercronic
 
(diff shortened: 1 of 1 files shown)


invario:
On second thought, creating a new user/group named acme is unnecessary, also removing the `USER '1000:1000'` directive from `Dockerfile`.  Users can specify whatever user desired.

invario:
Rebased.

mill1000:
I tested this and it appears to be working. Have to wait until the next schedule cron to ensure it runs.

To switch from a root setup to non-root I also had to
1. `chown` the acme.sh/ directory on my host (see docker compose below)
2. Use `group_add` to ensure the docker deploy still works

My compose
```yaml
services:
  acmesh:
    image: acmesh:noroot # Local image I built to test
    user: 1000:1000 # Set user/group
    group_add:
      - "998" # Add docker group for docker deploy
    command: daemon
    container_name: acme.sh
    restart: unless-stopped
    volumes:
      - ./acme.sh:/acme.sh
      - /var/run/docker.sock:/var/run/docker.sock
```

invario:
Updated Dockerfile to `chown -R acme:acme /acme.sh` and also set an `ENV HOME=/acme.sh`.  

Existing users need to ensure `/acme.sh` has proper read/write permissions for the non-root user.

github-cli:
maybe it would be better to copy the /etc/crontab/root to /acme.sh/crontab if one exists but the other does not instead of just creating the default crontab with a random time.
that way custom crontabs would still have their original content

also note that supercronic seems to not know about the @reboot command like cron does (I used this to start a renewal attempt at container creation, that way I can just recreate the container to let it run instead of waiting a day until the cron ran)
and it is also very verbose in the docker logs... a bit too verbose for me as it clutters up all the nice logs I created with crontab there

do you know if supercronic supports some way to start a command at creation?
great that this can run as non-root now though :)

invario:
> maybe it would be better to copy the /etc/crontab/root to /acme.sh/crontab if one exists but the other does not instead of just creating the default crontab with a random time. that way custom crontabs would still have their original content
> 
We can't do this because of permissions. The crontab file for a user is owned by that user.

> also note that supercronic seems to not know about the @reboot command like cron does (I used this to start a renewal attempt at container creation, that way I can just recreate the container to let it run instead of waiting a day until the cron ran) and it is also very verbose in the docker logs... a bit too verbose for me as it clutters up all the nice logs I created with crontab there

> do you know if supercronic supports some way to start a command at creation? great that this can run as non-root now though :)

I was not familiar with the `@reboot` directive but let me look into a workaround for you. But off the top of my head, your best bet may be to launch your container using a custom entrypoint script to run your command, and then launch the normal entrypoint.

Regarding logging, I need to look into it.

github-cli:
I found that supercronic does not support anything like the reboot expression:
https://github.com/aptible/supercronic/tree/main/cronexpr

so yes, I guess I have to start with a different entry script and then continue to run supercronic in some less verbose mode
I made supercronic silent by running it with -quiet and -passthrough-logs
it does log any command output to the container logs, of course you can circumvent it by not putting the output of the commands into stdout
Before Supercronic I did it the other way around, I specifically wrote everything into the container logs
I will adjust my crontab to account for this (much easier this way than manually putting everything into the container log)

PS: I only wrote this so others who might have the same issues know what has changed, I hope it does not come across as a complaint! Thank you very much for this contribution that makes it possible to run acme.sh as non-root! :-)

invario:
Not at all. I don't mind working to make things better. I appreciate the input!

github-cli:
except of not making supercronic so verbose, I wouldn't know what to make better :)
I ended up copying the entry.sh script and just replacing the supercronic start commands with this:

```
### BEGIN Customized stuff in entry.sh
#   echo "Running Supercronic using crontab at $LE_CONFIG_HOME/crontab"
#   exec -- /usr/bin/supercronic "$LE_CONFIG_HOME/crontab"
   echo "=== $(date +'%d.%m.%Y %H:%M:%S') Container startup with Supercronic using crontab at $LE_CONFIG_HOME/crontab ==="
   if [ -f "$LE_CONFIG_HOME/cronstart" ]; then
      /bin/sh "$LE_CONFIG_HOME/cronstart"
   fi
   exec -- /usr/bin/supercronic -passthrough-logs -quiet "$LE_CONFIG_HOME/crontab"
### END of customizations inside this script
```
this starts /acme.sh/cronstart on every container start (if the file exists) and then runs supercronic in a more silent mode, it still prints supercronic error messages but silences those info messages

PR #6722, opened 2025-12-29
Add script/hook for Proxmox Mail Gateway deployment via API

nexces:
Added script for Proxmox Mail Gateway deployment via API.

Fully based on @jdevera work with PVE/PBS scripts.

Requires separate calls for SMTP and proxy certificates.

### DIFF
--- a/deploy/proxmoxmg.sh
+++ b/deploy/proxmoxmg.sh
@@ -1,0 +1,145 @@
+#!/usr/bin/bash
+
+# Deploy certificates to a proxmox mail gateway using the API.
+#
+# Environment variables that can be set are:
+# `DEPLOY_PROXMOXMG_SERVER`: The hostname of the proxmox backup server. Defaults to
+#                            _cdomain.
+# `DEPLOY_PROXMOXMG_SERVER_PORT`: The port number the management interface is on.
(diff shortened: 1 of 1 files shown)


neilpang:
1. `#!/usr/bin/bash` -> `#!/usr/bin/env sh` -- hooks must be POSIX sh.
2. `_debug2 DEPLOY_PROXMOXMG_PASSWORD ...` logs the password in plaintext at `--debug 2` -- use `_secure_debug2`. Same for the ticket / auth cookie header.
3. `\s` in the `_egrep_o` patterns is a GNU extension -- BSD/busybox `grep -E` won't match it; use `[[:space:]]*`.
4. `_post` returns 0 on HTTP 4xx/5xx, so `_retval` catches neither a failed login nor a rejected upload -- with a wrong password this deploys nothing and still reports success. Error out when the extracted ticket is empty, and validate the upload response body instead of `$?`.
5. Prefix `response`, `_retval`, `_json_payload` with `_proxmoxmg_` -- hooks share one shell namespace.
6. Add the usage entry on the [deployhooks wiki](https://github.com/acmesh-official/acme.sh/wiki/deployhooks).

PR #6724, opened 2025-12-30
Base64 encode pfx password

### LINKED ISSUE #6721: Escaping of singe quote's in stored pfx password
When the password used with '--toPkcs' contains single quotes (that are escaped correctly on the command line) the command succeeds but the password written to 'Le_PFXPassword' in the conf file does not escape those single quotes.

This leeds to later renewals failing with 'syntax error: unterminated quoted string'

h-haaks:
Pfx passwords may contain characters like `'` and `;` 
When saved to config without encoding and then sourced in renew the password is not parsed correctly.

Base64 encoding the password avoids this.

fixes #6721


h-haaks:
@neilpang I would really like to address some security issues related to this bug. How do I do that?

PR #6725, opened 2025-12-30
Make sure the grep command  works on Linux as well as BSD variants to login to a Synology NAS with 2FA enabled

arjanvanderveen:
I had troubles getting the login to the Synology NAS working on OpenBSD. It turned out there are differences between
Linux grep and BSD variants of grep. This commit works in all operating systems (did not test Windows btw).

### DIFF
--- a/deploy/synology_dsm.sh
+++ b/deploy/synology_dsm.sh
@@ -234,7 +234,7 @@ synology_dsm_deploy() {
     fi
   fi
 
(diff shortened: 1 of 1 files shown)


neilpang:
Correct fix -- `[0-9]*` can match the empty string and BSD grep's `-o` handles empty matches differently from GNU. dev has the same pipeline in three places though, so all three are fixed in https://github.com/acmesh-official/acme.sh/commit/660a5e32, credited to you. Closing.

PR #6744, opened 2026-01-05
Restrict access to domain conf

h-haaks:
When toPkcs() is used the keystore password is saved to domain config.
Because of this the file should only be accessible by owner.


neilpang:
don't change the file permissions everytime.  only change it to 600 when you first create it.

h-haaks:
> don't change the file permissions everytime. only change it to 600 when you first create it.

Hmm, isn't that exactly what I do by putting it in `if [ ! -f "$DOMAIN_CONF" ]; then`?

PR #6785, opened 2026-02-12
Allow more than one DNS server for HA environments

alexandergott-afk:
This pull request allows you to specify more than one DNS server in this format: IP,IP,IP,IP

It is still possible to specify only one IP address without having to change anything compared to the current configuration.


alexandergott-afk:
I need a different editor.

PR #6795, opened 2026-02-18
Implement error tracking in SSH deployment script

ogmueller:
Add error handling to track deployment issues across multiple servers. Send notification on deployment error.

Currently if you deploy certificates to multiple servers using SSH, you will always get a success message. If one or all server fail due to any kind of connection errors, you will not see it in the end result (though you do get some error output). But the main problem is, that there is no notification nor a return code indicating an error.

This pull request improves error handling in the `ssh_deploy` function of the deployment script. Now, if deployment to any SSH server fails, the script will remember the failure, send a notification if appropriate, and return a non-zero exit code after attempting all servers.

### DIFF
--- a/deploy/ssh.sh
+++ b/deploy/ssh.sh
@@ -170,10 +170,19 @@ ssh_deploy() {
     _info "Required commands batched and sent in single call to remote host"
   fi
 
+  _returnCode=0
   _deploy_ssh_servers="$DEPLOY_SSH_SERVER"
   for DEPLOY_SSH_SERVER in $_deploy_ssh_servers; do
     _ssh_deploy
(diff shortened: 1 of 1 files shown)


neilpang:
The return-code tracking is the right fix. But deploy hooks must not call `_send_notify` — it's core-internal, nothing under `deploy/` calls it, and `$_set_level` is only set inside `renew()`, so in a standalone `--deploy` run your `[ $_set_level -ge $NOTIFY_LEVEL_ERROR ]` is a test syntax error (and you pass `$RENEW_SKIP` as the status for an *error*). Drop the notify block and just return non-zero — the caller handles notification. Also rebase onto current `dev` so CI runs.

Issue #6809, opened 2026-02-26
Report bugs to Calrissia DNS API


wardhus:
Report bugs or issues with the Calrissia DNS API here.
Thanks for your feedback!

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

Issue #6182, opened 2025-01-02
Unclear documentation


loxal:
It's unclear what the difference between `--renew` and `--issue` is. It would be great if `--renew` would be a `--safe-renew`, meaning an idempotent operation that generates a new cert or renews a certs if one already exists. Making INIT a one-off operation, introduces a lot of complexity on the callee side. 

What is the difference between ` --install-cert` and `--issue -d example.com -w /var/www/html`? Does the second command perform the same actions as ` --install-cert`?

loxal:
My comment was not related to keeping the acme.sh script up-to-date.

Neilpang:
>> It would be great if --renew would be a --safe-renew, meaning an idempotent operation that generates a new cert or renews a certs

That's what it is now.


read the main readme file, you will know everything you want:
https://github.com/acmesh-official/acme.sh

PR #6811, opened 2026-02-26
Add Calrissia.be API

wardhus:
Added to the wiki:
image

Created a issue for support: https://github.com/acmesh-official/acme.sh/issues/6809

Workflow passed: https://github.com/wardhus/acme.sh/actions/runs/22413282212

### DIFF
--- a/dnsapi/dns_calrissia.sh
+++ b/dnsapi/dns_calrissia.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_calrissia_info='Calrissia.be DNS API
+Site: calrissia.be
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_calrissia
+Options:
+ CALRISSIA_TOKEN Personal access token
+Issues: github.com/acmesh-official/acme.sh/pull/6802
+Author: Ward Hus
+'
+
+CALRISSIA_API="https://my.calrissia.com/api"
+
+dns_calrissia_add() {
+  fulldomain="$1"
+  txtvalue="$2"
+
+  _calrissia_load_token || return 1
+
+  if ! _calrissia_get_root "$fulldomain"; then
+    _err "Unable to find domain in Calrissia account for: $fulldomain"
+    return 1
+  fi
+
+  _debug "domain='$_domain' id='$_domain_id' sub='$_sub_domain'"
+  _info "Adding TXT record for $fulldomain"
+
+  _body="{\"name\":\"$_sub_domain\",\"type\":\"TXT\",\"content\":\"$txtvalue\",\"ttl\":120,\"prio\":0}"
+  _response="$(_calrissia_request POST "/domain/$_domain_id/record" "$_body")"
+
+  if ! _contains "$_response" '"id"'; then
+    _err "Failed to create TXT record: $_response"
+    return 1
+  fi
+
+  _record_id="$(printf "%s" "$_response" | _egrep_o '"id" *: *[0-9]+' | _egrep_o '[0-9]+')"
+  _key="$(_calrissia_key "$fulldomain" "$txtvalue")"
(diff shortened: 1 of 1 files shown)


dnsapi/dns_calrissia.sh
```diff
@@ -0,0 +1,124 @@
...
+    return 1
+  fi
+  _saveaccountconf_mutable CALRISSIA_TOKEN "$CALRISSIA_TOKEN"
+}
+
+# Sets _domain, _domain_id, _sub_domain for a given FQDN.
+_calrissia_get_root() {
+  _fqdn="$1"
+
+  i=2
```
neilpang: i starts from `1`

dnsapi/dns_calrissia.sh
```diff
@@ -0,0 +1,124 @@
...
+    _candidate="$(printf "%s" "$_fqdn" | cut -d . -f "$i"-)"
+    [ -z "$_candidate" ] && return 1
+
+    _debug "Trying root domain: $_candidate"
+    _response="$(_calrissia_request GET "/domain?full_domain_name=$_candidate")"
+    _debug2 "Response: $_response"
+
+    _domain_id="$(printf "%s" "$_response" |
+      _egrep_o '"id":[[:space:]]*[0-9]+' |
+      head -1 |
```
neilpang: don't use ` head -1 `,  use `_head_n 1`

wardhus:
I think we're ready to merge.

wardhus:
Dear @neilpang,

Can this be merged?

neilpang:
**1. Don't persist the record id locally — look it up from the API every time.** `add` saves `CALRISSIA_RECORD_ID_*`/`CALRISSIA_DOMAIN_ID_*` to domainconf and `rm` only reads them back, erroring if absent. Drop that state: in `rm`, list the domain's records, match on name + TXT value, delete by the id you get back, and treat "not found" as success. That survives state loss (other machine, cleared conf).

**2.** Info block: `Issues:` points to PR #6802 — it should be your tracking issue #6809. And `Docs:` points to the old `dnsapi` page — should be `dnsapi2#dns_calrissia`.

neilpang:
Your fork's DNS run now fails in `_calrissia_get_root`: the lookup for your own TestingDomain returns no `"id"`, so it walks down to `be` and gives up. Fix that and get the run green — required before merge.

wardhus:
Thanks for your feedback @neilpang ! The token was expired, i updated it and the action should go green now.

neilpang:
Code looks good. Two wiki fixes needed:

1. Wiki says `CALRISSIA_API_KEY`; the script reads `CALRISSIA_TOKEN`.
2. Wiki's "Report bugs" link points to #6182 (a different provider); should be #6809.

Two API questions:

1. Is `GET /domain?full_domain_name=...` an exact match? `_calrissia_get_root` takes the first `"id"` without verifying the returned domain, so a partial match on `_acme-challenge.example.com` would leave `_sub_domain` empty.
2. At the zone apex (e.g. challenge alias), add sends `"name":""` -- does the API accept an empty name, or does it expect `@`?

wardhus:
Thanks @neilpang! Answers below — both tested against the live API:

**1. Yes, it's an exact match.** `GET /domain?full_domain_name=...` only returns a domain when the query equals the zone name exactly. Verified: `full_domain_name=wardhus.be` returns the zone, while longer or partial queries (`_acme-challenge.wardhus.be`, `sub.wardhus.be`, `ardhus.be`, `wardhu`) all return `{"domains":[]}`. A longer FQDN never matches, so `_calrissia_get_root` can only stop at the exact registered zone and `_sub_domain` can't be left empty by a partial match.

**2. The empty name is accepted.** `POST /domain/{id}/record` with `"name":""` returns HTTP 200 and stores it as the apex record — verified with a live add + delete. (`"@"` happens to be accepted too, but the empty name is the intended form for the apex.)

The two wiki items are also fixed: the section now uses `CALRISSIA_TOKEN` and the bug-report link points to #6809.

Issue #6831, opened 2026-03-07
Report bugs for Hostinger.com API


microwavenby:
Please add bugs or issues with the Hostinger.com DNSAPI here

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

PR #6843, opened 2026-03-10
Microwavenby dns hostinger

microwavenby:
- [x] Test run as required by https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test: https://github.com/microwavenby/acme.sh/actions/runs/22861239290
- [x] Wiki documentation added to https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_hostinger
- [x] Issue created for tracking bugs or problems: https://github.com/acmesh-official/acme.sh/issues/6831 
- [x] Info block added to `dnsapi/dns_hostinger.sh`
- [x] Pull request against `dev` branch

### 201. Use Hostinger API:

Hostinger is a VPS and Domain registrar with DNS functionality.

You will need an API token; you can read how to create and use an API token in the [Hostinger API authentication documentation](https://developers.hostinger.com/#description/authentication)

The API documentation for DNS zone management is located at [https://developers.hostinger.com](https://developers.hostinger.com/#tag/dns-zone)

There is only one necessary credential, the API token stored in HOSTINGER_Token:

```sh
export HOSTINGER_Token=EXAMPLE_TRi7VCHHwXZhf9BVEUzzdYcPVLjJECTWX34mhxWQ
```

To issue a cert:

```sh
./acme.sh --issue --dns dns_hostinger -d example.nl -d *.example.nl
```

Report any bugs or issues to the open github issue [here](https://github.com/acmesh-official/acme.sh/issues/6831)

### DIFF
--- a/dnsapi/dns_hostinger.sh
+++ b/dnsapi/dns_hostinger.sh
@@ -1,0 +1,165 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_hostinger_info='Hostinger
+Site: Hostinger.com
+Domains: hostinger.nl
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_hostinger
+Options:
+ HOSTINGER_Token API Key
+Issues: https://github.com/acmesh-official/acme.sh/issues/6831
+Author: Sasha Reid <github@sasha.hackl.es>
+'
+
+HOSTINGER_Api="https://developers.hostinger.com/api/dns/v1/zones"
+
+########  Public functions #####################
+
+#Usage: add  _acme-challenge.www.domain.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_hostinger_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  HOSTINGER_Token="${HOSTINGER_Token:-$(_readaccountconf_mutable HOSTINGER_Token)}"
+
+  if [ -z "$HOSTINGER_Token" ]; then
+    HOSTINGER_Token=""
+    _err "You didn't specify a Hostinger API Key yet."
+    _err "Please read the documentation for the Hostinger API authentication at https://developers.hostinger.com/#description/authentication"
+    return 1
+  fi
+  _saveaccountconf_mutable HOSTINGER_Token "$HOSTINGER_Token"
+
+
+  _debug "First detect the root zone"
+  if ! _get_root "$fulldomain"; then
+    _err "invalid domain"
+    return 1
+  fi
+  _debug _sub_domain "$_sub_domain"
+  _debug _domain "$_domain"
+
+  _debug "Getting existing records"
+  _hostinger_rest GET "${_domain}"
+
+  if ! echo "$response"; then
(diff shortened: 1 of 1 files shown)


leandroscardua:
Hi @microwavenby ,

Are you working on this request?
if you have block or something else, Let me know may i can help you

neilpang:
**1. `dns_hostinger_rm` never loads the token** — it goes straight to `_get_root` without reading `HOSTINGER_Token` from accountconf, so any flow where `rm` runs without `add` in the same process gets a 401. Load it like `add` does.
**2. `rm` deletes by name+type, not value** — the DELETE filter has no content match, so it wipes every TXT at the name, including the second wildcard challenge value. If the API can't filter by content, read the records and PUT back the remaining values instead.
**3.** `if ! echo "$response"; then` is always false (`echo` always succeeds) — dead check, twice.
**4.** `Docs:` points to the old `dnsapi` page — should be `dnsapi2#dns_hostinger`.

leandroscardua:
HI @microwavenby ,

Are you planning to continue to work with this change?
If not, let me know
I'm really interesting to make it work as you already have it almost ready to use

microwavenby:
> HI @microwavenby ,
> 
> Are you planning to continue to work with this change? If not, let me know I'm really interesting to make it work as you already have it almost ready to use

Hey there - sorry for the long wait on the reply. 
I'll address Neil's comments tomorrow morning and update the PR.

neilpang:
Latest DNS run failure is only the DragonFlyBSD VM (curl missing in the image, pre-check fails before acme.sh even installs) -- not your code; all platforms that actually hit the Hostinger API passed.

**1. `rm` parsing relies on `.*?`, which POSIX ERE does not have.** `_egrep_o` (grep -E) treats it as greedy `.*` on GNU, and the sed/BRE fallback used when egrep is missing matches nothing. Reproduced with GNU grep: with any record set after the `_acme-challenge` TXT set in the GET response (A/CNAME/SPF -- any real zone), `\{\"name\":\"$_sub_domain\",\"records\":\[.*?\].*?TXT.*?\}` captures to the end of the JSON, then `\[.*?\]` spans from the first `[` to the last `]`:

```
REMAINING=[[{"content":"\"VALUE_TWO\"",...}],"ttl":120,"type":"TXT"},{"name":"@","records":[{"content":"1.2.3.4",...}]]
```

So `[ "$remaining_records" != "[]" ]` never reaches the DELETE branch (the last TXT value is never cleaned up), and the rebuilt `remaining_json` (`"content":"\\".*?\\""` is greedy too) can swallow `\"..\"` contents from other names -- e.g. an SPF/DKIM TXT -- and PUT them under `_acme-challenge`. CI only passes because the test zone has no other records and rm failures don't fail the issue flow. Rewrite without lazy quantifiers -- e.g. `_normalizeJson`, split records with `sed 's/},{/}\n{/g'`, then per-line exact match on the content value.

**2.** rm's PUT branch says "Added, OK" / "Add txt record error" -- copy/paste from add.

**3.** `_get_root`: a 401/429/5xx body contains neither "Domain name is not valid" nor `[]`, so the very first probe (the full challenge domain) is accepted as the root zone. Test for a positive zone-response signature instead.

PR #6862, opened 2026-03-16
[dnsapi][dns_yc] The value of the 'name' field in the API FQDN

alexey-gavrilov-flant:
The actual value of the field must be an FQDN.
This is not explicitly stated in the API documentation, but becomes apparent when you try to use it. Additionally, entries are displayed as FQDNs in the web interface.

### DIFF
--- a/dnsapi/dns_yc.sh
+++ b/dnsapi/dns_yc.sh
@@ -85,13 +85,13 @@ dns_yc_add() {
   _debug _domain "$_domain"
 
   _debug "Getting txt records"
-  if ! _yc_rest GET "zones/${_domain_id}:getRecordSet?type=TXT&name=$_sub_domain"; then
+  if ! _yc_rest GET "zones/${_domain_id}:getRecordSet?type=TXT&name=$_sub_domain$_domain"; then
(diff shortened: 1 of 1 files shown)


alexey-gavrilov-flant:
I followed the instructions and created the secrets, but CI still wouldn't start.
I removed access to the Yandex API so I wouldn't forget.

neilpang:
https://github.com/alexey-gavrilov-flant/acme.sh/actions/runs/23150025873

neilpang:
**1. `dns_yc_rm` still uses the bare `$_sub_domain`** — you only changed `add`, so cleanup queries/deletes the wrong name. Both must use the FQDN.
**2. The concatenation `$_sub_domain$_domain` is only correct on one code path.** The `YC_Zone_ID` path builds `_sub_domain` with `cut -c` and keeps a trailing dot; the folder-walk path (`cut -d . -f`) doesn't, so you get `_acme-challenge.wwwexample.com` — missing the dot. Normalize the two paths or join with an explicit `.`.
**3.** The DNS-API-Test still hasn't run green on your fork — required before merge.

PR #6865, opened 2026-03-17
Fix off-by-one error preventing the final poll to succeed

brevilo:
When the final poll (`_link_cert_retry` at 29) returns, the status is never checked again. So even a `valid` status goes unnoticed. It's a pre-test loop after all.

Solution: check the status once more after the loop.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -5324,6 +5324,11 @@ $_authorizations_map"
     _link_cert_retry="$(_math $_link_cert_retry + 1)"
   done
 
+  # cover case where the final poll returned 'valid'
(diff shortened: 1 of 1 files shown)


neilpang:
how about increasing `_MAX_CERT_RETRY=31`  or even `_MAX_CERT_RETRY=40`?
does that work ?

brevilo:
No, and why should it? It's always the *final* poll that's not checked, no matter which absolute iteration that might be. IOW, it's an off-by-one error, independent of `_MAX_CERT_RETRY`.

Issue #6763, opened 2026-01-24
Add an option to override retryafter in acme.sh


whirvis:
Sometimes when using `acme.sh`, the HTTP headers will return a `Retry-After` value that is far too high. This causes the program to exit early. This happened to me when trying to use it with the NameCheap API.
```
[Sat Jan 24 03:52:54 UTC 2026] Using CA: https://acme.zerossl.com/v2/DV90
[Sat Jan 24 03:52:54 UTC 2026] Multi domain='DNS:whirvis.io,DNS:*.whirvis.io'
[Sat Jan 24 03:52:56 UTC 2026] Getting webroot for domain='whirvis.io'
[Sat Jan 24 03:52:56 UTC 2026] Getting webroot for domain='*.whirvis.io'
[Sat Jan 24 03:52:56 UTC 2026] Adding TXT value: k5rYN-4ck1aure7tgJImRd4A_JlA78KIjZaNc1LDvuo for domain: _acme-challenge.whirvis.io
[Sat Jan 24 03:52:58 UTC 2026] The TXT record has been successfully added.
[Sat Jan 24 03:52:58 UTC 2026] Adding TXT value: JtQo3AiB7Y-N_QEP6S5yNS0ch8KmT_L3x8heQIVUuEM for domain: _acme-challenge.whirvis.io
[Sat Jan 24 03:53:00 UTC 2026] The TXT record has been successfully added.
[Sat Jan 24 03:53:00 UTC 2026] Let's check each DNS record now. Sleeping for 20 seconds first.
[Sat Jan 24 03:53:21 UTC 2026] You can use '--dnssleep' to disable public dns checks.
[Sat Jan 24 03:53:21 UTC 2026] See: https://github.com/acmesh-official/acme.sh/wiki/dnscheck
[Sat Jan 24 03:53:21 UTC 2026] Checking whirvis.io for _acme-challenge.whirvis.io
[Sat Jan 24 03:53:21 UTC 2026] Not valid yet, let's wait for 10 seconds then check the next one.
[Sat Jan 24 03:53:33 UTC 2026] Checking whirvis.io for _acme-challenge.whirvis.io
[Sat Jan 24 03:53:33 UTC 2026] Success for domain whirvis.io '_acme-challenge.whirvis.io'.
[Sat Jan 24 03:53:33 UTC 2026] Let's wait for 10 seconds and check again.
[...]

Issue #5233, opened 2024-08-01
Order status is 'processing', let's sleep and retry.


eleaner:
Steps to reproduce
------------------
I have no idea how to reproduce it
I am running 
` "/root/.acme.sh"/acme.sh --cron --home "/root/.acme.sh" --log --debug 2`
everything seems to work, success after success and then it gets stuck on 'processing' status

Debug log
-----------------
providing part of the log, this gets repeated multiple times
```
[Thu Aug  1 20:36:36 UTC 2024] code='200'
[Thu Aug  1 20:36:36 UTC 2024] original='{"status":"processing","expires":"2024-10-30T20:33:39Z","identifiers":[{"type":"dns","value":"kisiel.top"},{"type":"dns","value":"*.kisiel.top"}],"authorizations":["https://acme.zerossl.com/v2/DV90/authz/gtkFJ-tX3AHxmFy2GCjxhA","https://acme.zerossl.com/v2/DV90/authz/C0UEvOQhlGwqdd3xiJ7eaw"],"finalize":"https://acme.zerossl.com/v2/DV90/order/72pXXARu5kK5CNXPjrvTDg/finalize"}'
[Thu Aug  1 20:36:36 UTC 2024] response='{"status":"processing","expires":"2024-10-30T20:33:39Z","identifiers":[{"type":"dns","value":"kisiel.top"},{"type":"dns","value":"*.kisiel.top"}],"authorizations":["https://acme.zerossl.com/v2/DV90/authz/gtkFJ-tX3AHxmFy2GCjxhA","https://acme.zerossl.com/v2/DV90/authz/C0UEvOQhlGwqdd3xiJ7eaw"],"finalize":"https://acme.zerossl.com/v2/DV90/order/72pXXARu5kK5CNXPjrvTDg/finalize"}'
[Thu Aug  1 20:36:36 UTC 2024] Order status is 'processing', let's sleep and retry.
[Thu Aug  1 20:36:36 UTC 2024] _retryafter='15'
[Thu Aug  1 20:36:36 UTC 2024] Sleeping for 15 seconds then retrying
[Thu Aug  1 20:36:52 UTC 2024] Polling order status: https://acme.zerossl.com/v2/DV90/order/72pXXARu5kK5CNXPjrvTDg
[Thu Aug  1 20:36:52 UTC 2024] =======Sending Signed Request=======
[...]

PR #6869, opened 2026-03-21
Add configurable sleep for slow CA order processing

denisfoulon:
This PR keeps the change minimal and shell-compatible.

It adds configurable waiting for slow ACME order processing:
- LE_PROCESSING_MIN_SLEEP
- LE_MIN_RETRY_SLEEP
- LE_MAX_RETRY_AFTER

This helps with CAs returning Retry-After: 0 or taking longer to finalize certificates with many SANs.

Tested successfully with a slow CA on a certificate containing many SANs.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -2299,11 +2299,11 @@ _send_signed_request() {
 
       _retryafter=$(echo "$responseHeaders" | grep -i "^Retry-After *: *[0-9]\+ *" | cut -d : -f 2 | tr -d ' ' | tr -d '\r')
       if [ "$code" = '503' ]; then
-        _sleep_overload_retry_sec=$(( $_retryafter > ${LE_MIN_RETRY_SLEEP:-5} ? $_retryafter : ${LE_MIN_RETRY_SLEEP:-5} ))
+        _sleep_overload_retry_sec=$(( ${_retryafter:-0} > ${LE_MIN_RETRY_SLEEP:-5} ? ${_retryafter:-0} : ${LE_MIN_RETRY_SLEEP:-5} ))
         if [ -z "$_sleep_overload_retry_sec" ]; then
           _sleep_overload_retry_sec=5
         fi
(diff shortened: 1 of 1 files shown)


denisfoulon:
Related to #5233, #6103, #6763.

This PR improves handling of slow order processing and retry timing for ACME servers returning Retry-After: 0 or taking longer to finalize certificates with many SANs.

This follow-up PR removes the unrelated shebang change from the previous attempt and targets dev only. The diff is now limited to the retry/sleep handling for slow CA order processing.

denisfoulon:
Hi everyone,

I've updated the PR to address the points raised during the review:

    Robustness: Added ${_retryafter:-0} to prevent arithmetic syntax errors if the header is missing.

    Consistency: Replaced hardcoded 600 thresholds with the ${LE_MAX_RETRY_AFTER} variable.

    Best Practices: Switched from sleep to the internal _sleep function.

    Coverage: Updated the logic to ensure the configured minimum sleep is respected even if the CA returns Retry-After: 0.

denisfoulon:
I've applied the formatting fixes (arithmetic spacing and indentation) to comply with shfmt. Tests should be greener now! Thanks for your help on this.

denisfoulon:
Friendly follow-up on this PR.

I completely understand that maintainers are busy and reviews take time. If there are any concerns, requested changes, or improvements needed, I'd be happy to update the implementation.

Thank you for your time and for maintaining acme.sh.

neilpang:
acme.sh already honors the `Retry-After` header for slow order processing, per RFC 8555 — a compliant CA is expected to return a sensible `Retry-After` and the client waits accordingly. Adding client-side sleep knobs to work around a CA that sends `Retry-After: 0` (or none) isn't something we want to carry; that behavior should be fixed on the CA side. Thanks, but we'll pass on this one.

denisfoulon:
Thanks for the review and the explanation. I understand the decision and appreciate the feedback. We'll keep this change in our local fork for our own use. Thanks!

PR #6885, opened 2026-03-30
Add Abion.com DNS API

stelar7:


neilpang:
**1. `add` overwrites existing TXT records at the same name.** The PATCH in `dns_abion_add` sets `records[$_sub_domain][TXT]` to a single-element array, and Abion replaces the whole TXT list for that name. So wildcard certs (base + `*` place two challenge values on the same `_acme-challenge` name) break — the second `add` wipes the first. `add` needs to read the existing TXT records, append the new value, and PATCH the union, the same read-modify-write `rm` already does.

**2. Run the DNS-API-Test on your fork — it has to pass before this can be merged.** There are no Actions runs on your fork at all right now.

**3. shfmt is failing** — run `shfmt -i 2 -w dnsapi/dns_abion.sh`.

Minor: `rm` rebuilds the remaining records with a hardcoded `"ttl":60`, which resets the TTL of any other TXT records at that name; and the `Options:` block is missing a `Docs:` line.

PR #6901, opened 2026-04-12
Add certificate upload via CPanel JSON API

josephabbey:
This adds support for uploading certificates through the same API used be `dns_cpanel.sh`, which means that they can use the same credentials, this is different than the UAPI version. You will notice that a lot of the functions are the same (I reused them), but I do not see a standard way in this repo of sharing function between DNS and deploy hooks.

I know the API is technically deprecated, but the UAPI authentication method doesn't work for my provider.

This has been running in production on my home instance for several months, working well.

I am happy to write some docs for this on the wiki once it is approved.

Thanks for your work!!

..

### DIFF
--- a/deploy/deploy_cpanel.sh
+++ b/deploy/deploy_cpanel.sh
@@ -95,30 +95,6 @@ _myget() {
 
 ########  Private functions #####################
 
-# Internal utility to process YML from UAPI - looks at main_domain, sub_domains, addon domains and parked domains
-#[response]
-__cpanel_parse_response() {
-  if [ $# -gt 0 ]; then resp="$*"; else resp="$(cat)"; fi
-
-  echo "$resp" |
-    sed -En \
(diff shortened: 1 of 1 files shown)


neilpang:
**1. The private key is sent in a GET query string** — it ends up in webserver/proxy access logs. Use `_post` with the parameters in the request body.
**2. Use UAPI, not API 2.** cPanel's docs: "The cPanel API 2 system is deprecated. We strongly recommend that you use UAPI instead." Switch to the remote UAPI call (`execute/SSL/install_ssl`) — required before merge.
**3.** Prefix the helpers/globals (`_myget`, `_checkcredentials`, `_result`) with `_cpanel_` — generic names collide when sourced.
**4.** Persist config via `_getdeployconf`/`_savedeployconf`, not accountconf.
**5.** CI hasn't run on this branch — rebase onto current `dev`.

Issue #6919, opened 2026-04-20
The script, executing "crontab -l", writes an error to the crontab file


alexeyp0708:
For example, if the crontab file has not been created for the user, then “crontab -l” will display the message "no crontab for acme" with execution status "1" . The installcronjob function of the acme.sh script will write this text to the crontab file.
https://github.com/acmesh-official/acme.sh/blob/563415d21fea772b24c1d4703d5245ccb52ce232/acme.sh#L6277

To fix

`$_CRONTAB -l 2>/dev/null | { ...`

neilpang:
Fixed in dev by #6920 (`2>/dev/null` added to the `crontab -l` calls, as you suggested). `acme.sh --upgrade -b dev` to verify; it will be in the next release.

PR #6920, opened 2026-04-20
Fix write error $_CRONTAB -l in crontab file

alexeyp0708:
https://github.com/acmesh-official/acme.sh/issues/6919

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -6274,13 +6274,13 @@ installcronjob() {
     return 1
   fi
   _info "Installing cron job"
-  if ! $_CRONTAB -l | grep "$PROJECT_ENTRY --cron"; then
+  if ! $_CRONTAB -l 2>/dev/null | grep "$PROJECT_ENTRY --cron"; then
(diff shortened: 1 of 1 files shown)


acme.sh
```diff
@@ -6233,7 +6227,7 @@ installcronjob() {
   _initpath
   _CRONTAB="crontab"
   if [ -f "$LE_WORKING_DIR/$PROJECT_ENTRY" ]; then
-    lesh="\"$LE_WORKING_DIR\"/$PROJECT_ENTRY"
+    lesh="\"$LE_WORKING_DIR/$PROJECT_ENTRY\""
```
neilpang: revert this line.

alexeyp0708: Removed the fixes!

But this correction resulted in good readability of the path. For example, the path `"name\full name"\project` looks strange! Although the correct look would be `"name\full name\project"`
In any case, both options work when passing arguments. But in displaying script execution information, the path will be displayed as `"name\full name"\project`
I think you can fix this yourself

PR #6944, opened 2026-05-05
[dnsapi] add ServerCore DNS API hook

kot1grun:
Adds `dnsapi/dns_servercore.sh` for DNS-01 challenges via [ServerCore](https://servercore.com/),
## Usage
```sh
export SCore_Login_ID=<login_id> # login identifier
export SCore_Project_Name="<project_name>" # project name
export SCore_Login_Name="<username>" # username
export SCore_Pswd="<user_password>" # password for created user
```
Login identifier can be found in the upper right corner of the control panel.
After that you can issue the certificate with acme.sh, for example:
```sh
./acme.sh --issue --dns dns_servercore -d example.net -d *.example.net
```
## Testing

Tested against production ServerCore API with Let's Encrypt staging. Issue and cleanup both work.

### DIFF
--- a/dnsapi/dns_servercore.sh
+++ b/dnsapi/dns_servercore.sh
@@ -1,0 +1,345 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_servercore_info='Servercore.com
+Site: Servercore.com
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_serevercore
+Options:
+   SCore_Login_ID Account ID
+   SCore_Project_Name Project name
+   SCore_Login_Name Service user name
+   SCore_Pswd Service user password
+   SCore_Expire Token lifetime. In minutes (0-1440). Default "1400"
+'
+
+SCore_Api="https://api.servercore.com/domains"
+auth_uri="https://cloud.api.servercore.com/identity/v3/auth/tokens"
+_score_sep='#'
+
+########  Public functions #####################
+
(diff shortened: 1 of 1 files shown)


neilpang:
1. The DNS-API-Test has never run on your fork branch — configure the secrets (`TokenName1` etc., see the [DNS-API-Test wiki](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test)) and get it green; required before merge.
2. Info block: `Docs:` points to the old `dnsapi` page and has a typo (`dns_serevercore`) — should be `dnsapi2#dns_servercore`; and add an `Issues:` field pointing to #6945.
3. Prefix the globals/helpers: `auth_uri` and `_get_auth_token` are unprefixed and can collide — use `_score_`/`SCore_` naming like the rest of the file.

neilpang:
Four more for the same pass:

1. `_debug2 "_H1:" "$_H1"` logs the Keystone token at `--debug 2` — the exact level users paste into issues; use `_secure_debug2` or drop it.
2. `rm`: "Txt record not found" / "can not find _record_seg" must `return 0`, not 1 — the record may already be gone, a failing rm breaks cleanup.
3. Line 155: `info ...` → `_info` (as-is it is "command not found").
4. Line 146: `\n` in a sed replacement and `:a;N;s/\n/,/;ta` are GNU-only — breaks BSD/macOS sed; the MacOS job of the DNS test would catch this. Needs a POSIX rewrite.

PR #6952, opened 2026-05-07
ARI - Update renewal condition in issue()

sim0n-v:
The renewal condition in `renew()` supports ARI but the one in `issue()` does not.
This PR enable renewal when ARI is updated by the CA (e.g. in case of mass revocation event, see [this post from Let's Encrypt](https://community.letsencrypt.org/t/lets-encrypt-2026-mass-revocation-simulation/245960)).

Condition copied from line 5856.


sim0n-v:
Looks like it is better to update the `Le_NextRenewTimeStr` and `Le_NextRenewTime` based on ARI updates.

PR #6969, opened 2026-05-17
Fix dns_namesilo_rm failing to remove TXT record

### LINKED ISSUE #6907: dns_namesilo_rm() fails to remove DNS challenge records because NameSilo dnsListRecords API does not return the full domain
The following line in dns_namesilo_rm() function in dnsapi/dns_namesilo.sh:

`_record_id=$(echo "$response" | _egrep_o "<record_id>([^<]*)</record_id><type>TXT</type><host>$fulldomain</host>" | _egrep_o "<record_id>([^<]*)</record_id>" | sed -r "s/<record_id>([^<]*)<\/record_id>/\1/" | tail -n 1)`

 fails to retrieve the record id. This results in an error like:
```
Removing txt: <redacted> for domain: _acme-challenge.mydomainredacted.com
Empty record id, it seems no such record.
```
and a failure to remove the DNS challenge records. This makes the script not repeatable, because the DNS queries end up getting the old records, then failing.

The reason for this failure is because the NameSilo `dnsListRecords` API does not return the full domain name in the `<host>` section of the XML. Perhaps this is only true for parked domains, I do not know. Here's a sample result of `dnsListRecords` for a parked domain, with my values redacted:

``` XML
<?xml version="1.0" encoding="UTF-8"?>
<namesilo>
<request>
<operation>dnsListRecords</operation>
<ip>my-ip-redacted</ip>
</request>
<reply>
<code>300</code>
<detail>success</detail>
<resource_record><record_id>record-id-1-redacted</record_id><type>CNAME</type><host>www</host><value>parking.namesilo.com</value><ttl>3603</ttl><distance>0</distance></resource_record>
<resource_record><record_id>record-id-2-redacted</record_id><type>A</type><host>@</host><value>ip-redacted</value><ttl>3603</ttl><distance>0</distance></resource_record>
<resource_record><record_id>record-id-3-redacted</record_id><type>A</type><host>@</host><value>ip-redacted</value><ttl>3603</ttl><distance>0</distance></resource_record>
<resource_record><record_id>record-id-4-redacted</record_id><type>A</type><host>@</host><value>ip-redacted</value><ttl>3603</ttl><distance>0</distance></resource_record>
<resource_record><record_id>record-id-5-redacted</record_id><type>TXT</type><host>_acme-challenge</host><value>value-1-redacted</value><ttl>7200</ttl><distance>0</distance></resource_record>
<resource_record><record_id>record-id-6-redacted</record_id><type>TXT</type><host>_acme-challenge</host><value>value-2-redacted</value><ttl>7200</ttl><distance>0</distance></resource_record>
</reply>
</namesilo>
```

You can see above that the `<host>` tag does not contain the full domain name. The script needs to look for the subdomain, like `_acme-challenge` or the custom challenge alias, not `$fulldomain` in the `<host>` tag.

regisvidal-bitmapz:
Fixes #6907

NameSilo dnsListRecords API returns abbreviated hostnames rather than FQDNs, so matching against $fulldomain never succeeded and challenge records were never deleted.

Match against $txtvalue instead, which uniquely identifies the record

### DIFF
--- a/dnsapi/dns_namesilo.sh
+++ b/dnsapi/dns_namesilo.sh
@@ -65,7 +65,7 @@ dns_namesilo_rm() {
   if _namesilo_rest GET "dnsListRecords?version=1&type=xml&key=$Namesilo_Key&domain=$_domain"; then
     retcode=$(printf "%s\n" "$response" | _egrep_o "<code>300")
     if [ "$retcode" ]; then
-      _record_id=$(echo "$response" | _egrep_o "<record_id>([^<]*)</record_id><type>TXT</type><host>[^<]*</host><value>$txtvalue</value>" | _egrep_o "<record_id>([^<]*)</record_id>" | sed -r "s/<record_id>([^<]*)<\/record_id>/\1/" | tail -n 1)
+      _record_id=$(echo "$response" | _egrep_o "<record_id>([^<]*)</record_id><type>TXT</type><host>$_sub_domain</host><value>$txtvalue</value>" | _egrep_o "<record_id>([^<]*)</record_id>" | sed -r "s/<record_id>([^<]*)<\/record_id>/\1/" | tail -n 1)
       _debug _record_id "$_record_id"
       if [ "$_record_id" ]; then
         _info "Successfully retrieved the record id for ACME challenge."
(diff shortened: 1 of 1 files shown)


bilogic:
image

seems not fixed, I'm using a sub.sub domain, i.e. https://sub1.sub2.apex.com

Issue #6965, opened 2026-05-14
ARI suggestedWindow has started, but cert not renewed


hebbet:
following the revoke of YE and YR of lets encrypt all serts issued by that got a new ari window for renewal. 

https://community.letsencrypt.org/t/2026-05-08-gen-y-cross-certified-subordinate-cas-missing-serverauth-eku/247105/2

when i run

acme.sh --renew -d DOMAIN.NAME
[Thu May 14 01:02:21 PM CEST 2026] Renewing: 'DOMAIN.NAME'
[Thu May 14 01:02:21 PM CEST 2026] Renewing using Le_API=https://acme-v02.api.letsencrypt.org/directory
[Thu May 14 01:02:23 PM CEST 2026] ARI suggestedWindow has started (2026-05-14T10:02:23Z), proceeding with renewal.
[Thu May 14 01:02:23 PM CEST 2026] Domains not changed.
[Thu May 14 01:02:23 PM CEST 2026] Skipping. Next renewal time is: 2026-06-30T11:57:00Z
[Thu May 14 01:02:23 PM CEST 2026] Add '--force' to force renewal.

server time:
2026-05-14T13:05:26.55Z, so over three hours after start of ARI suggestedWindow it is not renewed.

it works of course when i run it with --force.

hebbet:
> acme.sh --renew -d DOMAIN.NAME --debug 2
[Thu May 14 01:07:51 PM CEST 2026] Let's find the script directory.
[Thu May 14 01:07:51 PM CEST 2026] _SCRIPT_='/root/.acme.sh/acme.sh'
[Thu May 14 01:07:51 PM CEST 2026] _script='/root/.acme.sh/acme.sh'
[Thu May 14 01:07:51 PM CEST 2026] _script_home='/root/.acme.sh'
[Thu May 14 01:07:51 PM CEST 2026] Using config home: /root/.acme.sh
[Thu May 14 01:07:51 PM CEST 2026] LE_WORKING_DIR='/root/.acme.sh'
https://github.com/acmesh-official/acme.sh
v3.1.4
[Thu May 14 01:07:51 PM CEST 2026] Running cmd: renew
[Thu May 14 01:07:51 PM CEST 2026] _renewServer
[Thu May 14 01:07:51 PM CEST 2026] Using config home: /root/.acme.sh
[Thu May 14 01:07:51 PM CEST 2026] default_acme_server='https://acme-v02.api.letsencrypt.org/directory'
[...]

PR #6953, opened 2026-05-07
ARI - Add support for Mass Revocation

sim0n-v:
Ref: 
* [Let'sEncrypt](https://community.letsencrypt.org/t/lets-encrypt-2026-mass-revocation-simulation/245960)
* [CABF](https://cabforum.org/working-groups/server/baseline-requirements/requirements/#4911-reasons-for-revoking-a-subscriber-certificate)

ARI should support imminent renewal (renew in 24 hours, before revocation by the CA).

First option was to add support for ARI on `issue()` (see https://github.com/acmesh-official/acme.sh/pull/6952).
I think it is better to simply update the `$Le_NextRenewTime` and `$Le_NextRenewTimeStr` when ARI "start" is in the past.

PR #6971, opened 2026-05-17
Add a configurable cron interval

e-nomem:
This PR is an implementation of my idea from https://github.com/acmesh-official/acme.sh/issues/6965#issuecomment-4466230338 and pulls both #6939 and #6953 as they currently are and adds a `--cron-interval` CLI flag. This should let the cron job figure out if the next cron run will happen before the ARI renewal window ends and trigger an early renewal as necessary. It ensures that we will never overshoot the 30 minute renewal window for the shortlived profile while also not renewing immediately after the 7 day renewal window opens for the classic profile.

Since this PR effectively contains 3 separate PRs in it, here's a link to just my changes without the other two: https://github.com/acmesh-official/acme.sh/compare/3b1f3c88cd3bb50e2cf57ecd8bdd0757b6304f88...e-nomem:acme.sh:configurable-cron-interval

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -6549,7 +6549,7 @@ installcronjob() {
     fi
     $_CRONTAB -l 2>/dev/null | {
       cat
-      echo "$random_minute ${random_hour},$(_math $random_hour + 6),$(_math $random_hour + 12),$(_math $random_hour + 18) * * * $lesh --cron --home \"$LE_WORKING_DIR\" $_c_entry> /dev/null"
+      echo "$random_minute $random_hour,$(_math $random_hour + 6),$(_math $random_hour + 12),$(_math $random_hour + 18) * * * $lesh --cron --home \"$LE_WORKING_DIR\" $_c_entry> /dev/null"
     } | $_CRONTAB_STDIN
   fi
   if [ "$?" != "0" ]; then
--- a/acme.sh
+++ b/acme.sh
@@ -182,6 +182,8 @@ _DNSCHECK_WIKI="https://github.com/acmesh-official/acme.sh/wiki/dnscheck"
 
 _PROFILESELECTION_WIKI="https://github.com/acmesh-official/acme.sh/wiki/Profile-selection"
 
+_ARI_WIKI="https://github.com/acmesh-official/acme.sh/wiki/ARI"
+
 _DNS_MANUAL_ERR="The dns manual mode can not renew automatically, you must issue it again manually. You'd better use the other modes instead."
 
 _DNS_MANUAL_WARN="It seems that you are using dns manual mode. please take care: $_DNS_MANUAL_ERR"
@@ -7662,6 +7664,10 @@ Parameters:
 
   --password <password>             Add a password to exported pfx file. Use with --to-pkcs12.
 
+  --cron-interval <interval hours>  Sets the cron interval when installing the cron job or used by the renew commands to detect if
+                                      early cert renewal is required when using ARI. Only valid for '--install', '--install-cronjob',
+                                      '--renew', '--renew-all', and '--cron'.
+                                      See: $_ARI_WIKI
 
 "
 }
--- a/acme.sh
+++ b/acme.sh
@@ -6477,7 +6477,7 @@ _install_win_taskscheduler() {
   _info "Please input your Windows password for: $(__green "$_myname")"
   _password="$(__read_password)"
   #SCHTASKS.exe '/create' '/SC' 'DAILY' '/TN' "$_WINDOWS_SCHEDULER_NAME" '/F' '/ST' "00:$_randomminute" '/RU' "$_myname" '/RP' "$_password" '/TR' "$_winbash -l -c '$_lesh --cron --home \"$LE_WORKING_DIR\" $_centry'" >/dev/null
-  echo SCHTASKS.exe '/create' '/SC' 'HOURLY' '/MO' '6' '/TN' "$_WINDOWS_SCHEDULER_NAME" '/F' '/ST' "0$_randomhour:$_randomminute" '/RU' "$_myname" '/RP' "$_password" '/TR' "\"$_winbash -l -c '$_lesh --cron --home \"$LE_WORKING_DIR\" $_centry'\"" | cmd.exe >/dev/null
+  echo SCHTASKS.exe '/create' '/SC' 'HOURLY' '/MO' '6' '/TN' "$_WINDOWS_SCHEDULER_NAME" '/F' '/ST' "$(printf '%02d:%02d' "$_randomhour" "$_randomminute")" '/RU' "$_myname" '/RP' "$_password" '/TR' "\"$_winbash -l -c '$_lesh --cron --home \"$LE_WORKING_DIR\" $_centry'\"" | cmd.exe >/dev/null
   echo
 
 }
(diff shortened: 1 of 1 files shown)


e-nomem:
@sim0n-v this contains both your PRs

acme.sh
```diff
@@ -5840,14 +5856,39 @@ renew() {
...
+          Le_NextRenewTime=$(_math "$_ari_start_t" + "$_ari_offset")
+          Le_NextRenewTimeStr=$(_time2str "$Le_NextRenewTime")
+          _info "ARI suggestedWindow: $(__green "$(_time2str "$_ari_start_t")") to $(__green "$(_time2str "$_ari_end_t")")"
+          _info "Next renewal time picked from ARI window: $(__green "$Le_NextRenewTimeStr")"
+          _savedomainconf Le_NextRenewTime "$Le_NextRenewTime"
+          _savedomainconf Le_NextRenewTimeStr "$Le_NextRenewTimeStr"
+        fi
+        _next_cron_run_t="$(_time "$_cron_interval")"
+        if [ "$Le_NextRenewTime" ] && ([ "$(_time)" -ge "$Le_NextRenewTime" ] || [ "$_next_cron_run_t" -ge "$_ari_end_t" ]); then
           _ari_should_renew="1"
```
sim0n-v, line 5883: `_ari_should_renew` can be removed as per https://github.com/acmesh-official/acme.sh/pull/6953#discussion_r3252473643
(same on line 5848 and 5897)

acme.sh
```diff
@@ -5840,14 +5856,39 @@ renew() {
...
+          if [ "$_next_cron_run_t" -ge "$_ari_end_t" ]; then
```
sim0n-v: As per [LE's ARI Imprementation Guide Step 5](https://letsencrypt.org/2024/04/25/guide-to-integrating-ari-into-existing-acme-clients#step-5-selecting-a-specific-renewal-time) or [RFC9773 Section 4.2](https://www.rfc-editor.org/rfc/rfc9773.html#name-renewalinfo-objects), I think you should replace `_ari_end_t` by `Le_NextRenewTime`.
> Otherwise, if the selected time is before the next time that the client would wake up normally, attempt renewal immediately. 

This enable renewal if the next cron is after the uniformly chosen time inside ARI suggestedWindow.

Edit: this also targets the next line (5883) and the one above (5880 with the same condition).

acme.sh
```diff
@@ -6513,12 +6580,17 @@ installcronjob() {
...
+  fi
+  _c_entry="--home \"$LE_WORKING_DIR\" --cron-interval \"$_cron_interval\" "
   if [ "$_c_home" ]; then
-    _c_entry="--config-home \"$_c_home\" "
+    _c_entry="${_c_entry}--config-home \"$_c_home\" "
   fi
   _t=$(_time)
   random_minute=$(_math $_t % 60)
-  random_hour=$(_math $_t / 60 % 24)
+  random_hour=$(_math $_t / 60 % 6)
```
sim0n-v: I think you have to make `% ${_cron_interval}`: `% 6` gets an integer between 0 and 5 (included).
Here, we want to get an hour between 0 and `_cron_interval - 1`.

If `_cron_interval` is 24 hours, it will be the old behaviour (e.g. `random_hour=$(_math $_t / 60 % 24)`).

acme.sh
```diff
@@ -5948,12 +5992,19 @@ renewAll() {
     fi
     d=$(basename "$di")
     _debug d "$d"
+    _d_ari="$d.ari"
+    _debug d_ari "$d_ari"
```
sim0n-v: `d_ari` is not set, `_d_ari` is.
(this is my mistake from https://github.com/acmesh-official/acme.sh/pull/6953)

acme.sh
```diff
@@ -6549,13 +6621,13 @@ installcronjob() {
     fi
     $_CRONTAB -l 2>/dev/null | {
       cat
-      echo "$random_minute $random_hour * * * $lesh --cron --home \"$LE_WORKING_DIR\" $_c_entry> /dev/null"
+      echo "$random_minute $random_hour/$_cron_interval * * * $lesh --cron $_c_entry> /dev/null"
```
sim0n-v: I tried `$random_hour/6` in https://github.com/acmesh-official/acme.sh/pull/6939/changes/647c7338908c35944a27b80d40b0dab3b453d0a4 but didn't work. Don't really know why...

That's why I came up with a list of hours instead (`$random_hour,$random_hour+$_cron_interval*1,$random_hour+$_cron_interval*2` etc. until `$random_hour+$_cron_interval*(24 / $_cron_interval)`).

acme.sh
sim0n-v: Updated in https://github.com/acmesh-official/acme.sh/pull/6953

acme.sh
sim0n-v: Or it should be kept to enable renewal when selected time is before the next scheduled run.

e-nomem: Yeah, I added it back because it's in use again after #6966 was merged.

acme.sh
e-nomem: I think it's functionally the same thing but I can update it so it's clearer what the condition is. `[ "$(_time "$_cron_offset")" -ge "$Le_NextRenewTime" ]`

acme.sh
e-nomem: Good catch

acme.sh
e-nomem: I think it's because `$random_hour/$_cron_interval` is technically non-standard syntax. The left side is supposed to be a range expression so we probably want `${random_hour}-23/$_cron_interval`

acme.sh
```diff
@@ -7589,6 +7665,10 @@ Parameters:
 
   --password <password>             Add a password to exported pfx file. Use with --to-pkcs12.
 
+  --cron-interval <interval hours>  Sets the cron interval when installing the cron job or used by the renew commands to detect if
+                                      early cert renewal is required when using ARI. Only valid for '--install', '--install-cronjob',
+                                      '--renew', '--renew-all', and '--cron'.
+                                      See: $_ARI_WIKI
```
e-nomem, line 7671: Added this with the assumption that the Wiki will be updated to clarify the interaction between this flag and the ARI feature if this PR is merged.

acme.sh
```diff
@@ -5931,6 +5974,9 @@ renewAll() {
   _server="$2"
   _debug "_server" "$_server"
 
+  _cron_interval="$3"
+  debug "_cron_interval" "$_cron_interval"
```
sim0n-v, line 5978: Typo: should be `_debug`

acme.sh
```diff
@@ -5782,6 +5798,8 @@ renew() {
 
   _isEcc="$2"
   _renewServer="$3"
+  _cron_interval="$4"
```
sim0n-v, line 5801: I don't think that upgrading `acme.sh` will update the cronjob as well. This value might be not set.
ARI is enabled by default so maybe add something like:
```
if [ -z "$_cron_interval" ]; then
  _cron_interval=24
fi
```

acme.sh
```diff
@@ -1850,8 +1860,14 @@ _calcjwk() {
   __CACHED_JWK_KEY_FILE="$keyfile"
 }
 
+# [offset_hours]
 _time() {
-  date -u "+%s"
+  _now_unix="$(date -u "+%s")"
+  if [ "$1" ]; then
+    _offset_sec="$(_math "$1" * 3600)"
+    _now_unix="$(_math "$_now_unix" + "$_offset_sec")"
```
sim0n-v, line 1868: I get some errors with these (running bash and sh): I think `*` and `+` should be escaped.

The first line (with `*`) prints the current dir.
The second line (with `+` ) cannot understand the symbol.

Fixed with escaped char.

neilpang:
This bundles #6939 and #6953 plus the new flag — a 3-in-1 can't be reviewed or merged as is. Let's settle those two PRs first, then rebase this down to just the `--cron-interval` change. Also CI hasn't run on this branch.

PR #6981, opened 2026-05-22
Update author and API version in dns_timeweb.sh

augin:
the previous script uses an outdated API and is not functional. I offer my working script with API v2


neilpang:
This replaces a convention-compliant implementation with a script that breaks several hard rules, so it can't be accepted in this form:

1. It **redefines `_debug()`**, overriding acme.sh's core function for the whole run.
2. Raw `curl` — must use `_get`/`_post`. `echo >&2` — must use `_err`/`_info`.
3. Record id persisted to `/tmp/timeweb_*.id` — don't store state; look the record up from the API in `rm` by name + value.
4. Generic helper names (`_load_token`, `_add_txt_record`) — prefix with `_timeweb_`.
5. `TW_Token` must be loaded/saved via `_readaccountconf_mutable`/`_saveaccountconf_mutable`; comments must be English; shfmt is failing.
6. The DNS-API-Test must pass in your fork.

If v1 is really dead (please link an announcement), the right move is to update the existing implementation's endpoints to v2, keeping its structure — not replace the file.

Issue #6964, opened 2026-05-13
Check for `replaces` error not matching for Actalis CA


JensSpanier:
Actalis returns this error if they couldn't find the cert to replace:

```json
{"status": 400, "type": "urn:ietf:params:acme:error:malformed", "detail": "Cannot find any certificate to replace for given AKI and Serial Number"}
```

So this check doesn't catch this error:

https://github.com/acmesh-official/acme.sh/blob/7c91ee1fc0a22a9b2a3fbc4ddc57f55fd8da9d71/acme.sh#L4898

sim0n-v:
Hey, I think the error you get (without more context) is because of a missconstructed CertID as per https://github.com/acmesh-official/acme.sh/issues/6942.
Can you ensure that ARI is fetched correctly and/or try the latest version? If ARI fails, then the `replace` should fail as well.

But I agree that the current code will add `replace` even if the CA changes (e.g. switching from a directory to another) and will cause an error.

JensSpanier:
Thanks for the clarification.

I tested the request against `https://acme-api.actalis.com/acme/renewal-info/XXX` and it worked correctly, so the generated ARI itself seems to be valid.

My guess is that Actalis may have changed or fixed their implementation in the meantime. Another possibility could be that the `replaces` field is only supposed to be sent once the specified renewal window has actually been reached. However, I just performed such a renewal before the indicated renewal window, and it worked successfully.

I’ll close this issue for now since it is working now (in this case probably due to changes on Actalis’ side). However, the CA switch should also work better now with the recent changes. Thanks!

PR #6983, opened 2026-05-24
ARI - Add support for switching ACME Server during renewal

sim0n-v:
Fix for https://github.com/acmesh-official/acme.sh/issues/6964

Following the initial [ACME RFC Section 9.7.4](https://datatracker.ietf.org/doc/html/rfc8555#section-9.7.4):
* Error Type starts with: `urn:ietf:params:acme:error:`
* Error Type ends in particular with ([ref](https://www.iana.org/assignments/acme/acme.xhtml#acme-error-types)):
  * `alreadyReplaced`
  * `malformed`

This PR add support of exact Error Types `urn:ietf:params:acme:error:alreadyReplaced` and `urn:ietf:params:acme:error:malformed`.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -4895,7 +4895,7 @@ issue() {
     # (Let's Encrypt) may also reject with a malformed error if the prior cert
     # was issued by a different issuer / different CA. Retry without "replaces"
     # whenever the failure mentions ARI or the replaces field.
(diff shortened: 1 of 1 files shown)


neilpang:
just add `urn:ietf:params:acme:error:malformed` don't change the existing values

sim0n-v:
Done.

I still think that it should be the same format:
* `urn:ietf:params:acme:error:malformed` vs `malformed`
* `urn:ietf:params:acme:error:alreadyReplaced` vs `alreadyReplaced`
For now we have both.

PR #6991, opened 2026-05-31
Add dns_gcloud_rest for Google Cloud DNS

cbarrick:
Add dns_gcloud_rest for Google Cloud DNS

This script is an alternative to the existing dns_gcloud script but
aimed at embeded environments. The existing dns_gcloud script depends
on the gcloud CLI, which is over 60 MB on ARM and over 85 MB on x86_64.
In many embeded environments, this dependency is too large to install.

This new script, dns_gcloud_rest, interfaces with Google Cloud DNS
using the REST API. It implements the service account auth flow and
requires the user to provision a service account with permission to
manage DNS zones.

The Google Cloud REST API uses JSON and this script uses regex to
extract data from the response bodies. I have done my best to implement
robust patterns, but regex is fundamentally not able to parse JSON. The
response handling is therefore sensitive to non-semantic details of the
response bodies, like newline placement. That said, some responses from
GCP are minified and some are pretty; this script handles responses in
both formats adequately.

This script is ported from a similar script I contributed the the
[OpenWRT DDNS scripts][1]. OpenWRT provides tools for JSON parsing and
formatting, which makes that version of the script more robust and
succinct than this verion.

I have tested the script in several environments to ensure portability:
- Bash 3.2.57 with macOS coreutils on macOS
- Bash 3.2.57 with GNU coreutils on macOS
- busybox 1.37.0 on OpenWRT

[1]: https://github.com/cbarrick/openwrt_packages/blob/ddns-gcp/net/ddns-scripts/files/usr/lib/ddns/update_gcp_v1.sh

### DIFF
--- a/dnsapi/dns_gcloud_rest.sh
+++ b/dnsapi/dns_gcloud_rest.sh
@@ -34,15 +34,16 @@ dns_gcloud_rest_add() {
   _save_config || return 1
 
   access_token=$(_get_access_token) || return 1
+  zone=$(_pick_zone "${access_token}" "${domain}") || return 1
 
   _debug "Adding TXT record for ${domain}: ${record}"
 
   # If a matching ResourceRecordSet exists, the record is inserted into it.
   # Otherwise a new ResourceRecordSet is created.
-  if _has_rrset "${access_token}" "${domain}" TXT; then
-    _patch_add_rrset "${access_token}" "${domain}" TXT "${record}"
+  if _has_rrset "${access_token}" "${zone}" "${domain}" TXT; then
+    _patch_add_rrset "${access_token}" "${zone}" "${domain}" TXT "${record}"
   else
-    _create_rrset "${access_token}" "${domain}" TXT "${record}"
+    _create_rrset "${access_token}" "${zone}" "${domain}" TXT "${record}"
   fi
 }
 
@@ -55,16 +56,17 @@ dns_gcloud_rest_rm() {
   _save_config || return 1
 
   access_token=$(_get_access_token) || return 1
+  zone=$(_pick_zone "${access_token}" "${domain}") || return 1
 
   _debug "Removing TXT record for ${domain}: ${record}"
 
   # If the ResourceRecordSet has more than one record, the record is removed
   # from it. Otherwise, if the ResourceRecordSet contains only this record,
   # the entire ResourceRecordSet deleted.
-  if _has_many_records "${access_token}" "${domain}" TXT; then
-    _patch_remove_rrset "${access_token}" "${domain}" TXT "${record}"
-  elif _has_record "${access_token}" "${domain}" TXT "${record}"; then
-    _delete_rrset "${access_token}" "${domain}" TXT
+  if _has_many_records "${access_token}" "${zone}" "${domain}" TXT; then
+    _patch_remove_rrset "${access_token}" "${zone}" "${domain}" TXT "${record}"
+  elif _has_record "${access_token}" "${zone}" "${domain}" TXT "${record}"; then
+    _delete_rrset "${access_token}" "${zone}" "${domain}" TXT
   fi
 }
 
@@ -477,14 +479,14 @@ _pick_zone() {
   # Find the zone whose name is the longest suffix of the given domain.
   _debug "Determining the correct zone for DNS record."
   suffix_len=0
-  for z in $zones; do
+  for z in ${zones}; do
     dns_suffix=$(_get_dns_suffix_for_zone "${access_token}" "${z}")
     _debug dns_suffix "${dns_suffix}"
     case "${domain}." in # Include trailing dot in the pattern to match.
-    *$dns_suffix)
-      if [ "${#z}" -gt "${suffix_len}" ]; then
+    *.${dns_suffix})
+      if [ "${#dns_suffix}" -gt "${suffix_len}" ]; then
         zone="${z}"
-        suffix_len="${#z}"
+        suffix_len="${#dns_suffix}"
       fi
       ;;
     esac
@@ -501,10 +503,9 @@ _pick_zone() {
 # Fetch a ResourceRecordSet from the Cloud DNS API.
 _get_rrset() {
   access_token="$1"
-  domain="$2"
-  record_type="$3"
-
-  zone=$(_pick_zone "${access_token}" "${domain}") || return 1
+  zone="$2"
+  domain="$3"
+  record_type="$4"
 
   _debug "Fetching RRSet: ${domain} ${record_type}."
   _H1="Authorization: Bearer ${access_token}"
@@ -522,10 +523,9 @@ _get_rrset() {
 # Like _get_rrset but quieter.
 _has_rrset() {
   access_token="$1"
-  domain="$2"
-  record_type="$3"
-
-  zone=$(_pick_zone "${access_token}" "${domain}") || return 1
+  zone="$2"
+  domain="$3"
+  record_type="$4"
 
   _debug "Fetching RRSet: ${domain} ${record_type}."
   _H1="Authorization: Bearer ${access_token}"
@@ -539,10 +539,11 @@ _has_rrset() {
 # Determine if a ResourceRecordSet has multiple records.
 _has_many_records() {
   access_token="$1"
-  domain="$2"
-  record_type="$3"
+  zone="$2"
+  domain="$3"
+  record_type="$4"
 
-  rrset=$(_get_rrset "${access_token}" "${domain}" "${record_type}") || return 1
+  rrset=$(_get_rrset "${access_token}" "${zone}" "${domain}" "${record_type}") || return 1
   count=$(echo "${rrset}" | _extract_records | wc -l)
   if [ "${count}" -lt 2 ]; then
     return 1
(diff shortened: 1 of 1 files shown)


cbarrick:
The DNS API test for MidnightBSD is broken. Not because of this commit, but because the CI fails to install curl in the VM. I filed a separate PR to remove MidnightBSD from the test matrix: acmesh-official#6997.

cbarrick:
DNS Test passed: https://github.com/cbarrick/acme.sh/actions/runs/26727746327

Excluding MidnightBSD, as discussed above.

cbarrick:
The DNS API Dev Guide says to add documentation to the wiki.

Since this is the second implementation for Google Cloud DNS, I am proposing that the existing Google Cloud DNS docs be updated to include documentation for both scripts as follows:

> ```
> <a name="dns_gcloud" />
> <a name="dns_gcloud_rest" />
> ```
> 
> ## 49. Use Google Cloud DNS API to automatically issue cert
> 
> There are two scripts to choose from when working with Google Cloud DNS:
> 
> - `dns_gcloud` uses the `gcloud` CLI to interact with GCP. This is the most
>   flexible option, but the `gcloud` CLI is a heavy dependency and is not always
>   available in embeded environments.
> 
> - `dns_gcloud_rest` uses the Google Cloud REST API to interact with GCP. This
>   is much more lightweight and is suitable for embeded environments.
> 
> Both scripts support [DNS alias mode][].
> 
> [DNS alias mode]: https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode
> 
> ### Using `dns_gcloud`
> 
> First you need to authenticate to gcloud.
> 
> ```
> gcloud init
> ```
> 
> **The `dns_gcloud` script uses the active gcloud configuration and
> credentials.** There is no logic inside `dns_gcloud` to override the project
> and other settings. If needed, create additional [gcloud configurations][]. You
> can change the configuration being used without *activating* it; simply set the
> `CLOUDSDK_ACTIVE_CONFIG_NAME` environment variable.
> 
> To issue a certificate you can:
> 
> ```sh
> export CLOUDSDK_ACTIVE_CONFIG_NAME=default  # see the note above
> ./acme.sh --issue --dns dns_gcloud -d example.com -d *.example.com
> ```
> 
> [gcloud configurations]: https://cloud.google.com/sdk/gcloud/reference/topic/configurations
> 
> ### Using `dns_gcloud_rest`
> 
> First create a new [service account][] in the GCP project that owns your DNS
> zone and provision a service account key. This will be used to authenticate with
> GCP.
> 
> Then grant the DNS Administrator role to the service account in [Cloud IAM][].
> 
> To use the script, you must set the following environment variables:
> 
> - `GCP_PROJECT`: The project that owns the DNS zone.
> - `GCP_SERVICE_ACCOUNT`: The service account name. The service account must
>   have permission to update DNS records. Example:
>   `my-dns-service-account@my-dns-project.iam.gserviceaccount.com`
> - `GCP_SERVICE_ACCOUNT_KEY`: The service account key. You can paste the key
>   directly into this variable or set this variable to a path to the key file.
>   This script supports JSON keys or the raw private key as a PEM file. P12 keys
>   are not supported. File names must end with `*.json` or `*.pem`.
> 
> To issue a certificate:
> 
> ```sh
> # See the note above about setting these variables.
> export GCP_PROJECT=my-project
> export GCP_SERVICE_ACCOUNT=my-dns-service-account@my-dns-project.iam.gserviceaccount.com
> export GCP_SERVICE_ACCOUNT_KEY=/path/to/my/service/account/key.json
> 
> ./acme.sh --issue --dns dns_gcloud_rest -d example.com -d *.example.com
> ```
> 
> [service account]: https://console.cloud.google.com/iam-admin/serviceaccounts
> [Cloud IAM]: https://console.cloud.google.com/iam-admin/iam

cbarrick:
I think this is ready for review! CC @neilpang 

TL;DR - I use acme.sh on my router. The existing `dns_gcloud` script has a 60 MB dependency that I can't install with my limited space, so I wrote an alternative without that dependency.

cbarrick:
Since acmesh-official#6997 was closed, I rebased this PR to not be built on top of that PR.

cbarrick:
Err, I just realized that the command I use to unescape literal `\n` with a newline is incorrect.

Commit was using this this:

```
sed "s/\\n/\\${NL}/g"
```

But is should be this:

```
sed "s/\\\n/\\${NL}/g"
```

Where `${NL}` is a variable holding a literal newline.

The extra backslash is needed because this is a double quoted string.

This wasn't caught in the DNS Test because the secret I set for the service account key doesn't have a `\n` in it.

And I didn't notice this when I was doing manual testing earlier because I originally used a different command (`sed 's/\\n/\n/g'`) that I found out doesn't work on OpenBSD.

I fixed this on line 162 and force pushed to update this PR.

```diff
diff --git a/dnsapi/dns_gcloud_rest.sh b/dnsapi/dns_gcloud_rest.sh
index 42236185..45e51b48 100644
--- a/dnsapi/dns_gcloud_rest.sh
+++ b/dnsapi/dns_gcloud_rest.sh
@@ -159,7 +159,7 @@ _get_json_str() {
     sed 's/"\(.*\)"/\1/' |
     _json_decode |
     sed 's/\\"/"/g' |
-    sed "s/\\n/\\${NL}/g"
+    sed "s/\\\n/\\${NL}/g"
 }
```

neilpang:
Thanks for the contribution, this is well structured and the self-contained JWT/service-account auth flow looks correct. A few things from a review:

**Blocking - `_pick_zone` picks the wrong zone when multiple zones match**

When finding the longest-suffix zone, the comparison uses the length of the managed-zone *resource name* (`${#z}`) instead of the matched *DNS suffix* length (`${#dns_suffix}`):

```sh
*$dns_suffix)
  if [ "${#z}" -gt "${suffix_len}" ]; then   # should be ${#dns_suffix}
    zone="${z}"
    suffix_len="${#z}"                        # should be ${#dns_suffix}
  fi
  ;;
```

The managed-zone name is arbitrary and unrelated to suffix depth, so when a domain matches both e.g. `example.com.` and `sub.example.com.` (delegation / DNS-alias setups), the wrong zone can be selected - which is exactly the case this logic exists to handle. For reference, the existing `dns_gcloud` selects by `dnsName` label depth. Please compare `${#dns_suffix}` instead. (Also worth resetting `zone=""` at the top of `_pick_zone`, since it's a global.)

**Minor**

- `_delete_rrset` passes the global `${request}` as the DELETE body, but it's never set in that function (it carries a stale/empty value from a previous call). DELETE ignores the body so it's harmless, but please pass `""` explicitly.
- `_pick_zone` is called 2-3 times per add/rm, and each call does one `_list_zones` request plus one request per zone to resolve its `dnsName` (an N+1 pattern). Functionally fine, but caching the resolved zone would cut a lot of API calls.
- The JWT `iat`/`exp` are emitted as JSON strings (`"iat":"..."`). RFC 7519 specifies these as numeric `NumericDate`; Google accepts the string form (your test passed), but numbers would be more correct.
- `_has_record` / `_remove_rrdata` use substring matching (`grep -F` / `grep -Fv`). ACME tokens are fixed-length base64url so a collision is very unlikely, and this matches the existing `dns_gcloud` style, so it's fine - just flagging it.

**Security note (not a bug)**

`GCP_SERVICE_ACCOUNT_KEY` is persisted to `account.conf` in plaintext via `_saveaccountconf_mutable`. That's consistent with how acme.sh stores other secrets, but a service-account private key is high-value - it'd be good for the wiki docs to recommend the file-path form and note that `account.conf` permissions matter.

Once `_pick_zone` is fixed (and the `_delete_rrset` body tidied up), this is good to merge from my side. Thanks again!

cbarrick:
That's a great catch with the zone logic using the wrong field. Thanks for that! I did test with multiple zones, but I guess the resource names I used happened to have the same relative lengths as the DNS suffixes. An early draft of this script had the `_list_zones` function returning DNS suffixes rather than resource names, but I clearly forgot to update this loop when I changed that.

Regarding `GCP_SERVICE_ACCOUNT_KEY`, the use I expect is for users to provide a path to a key file that is appropriately secured, but in some circumstances passing inline is acceptable or even the only available option (like in the unit tests). And I agree that users should be given enough information to make an informed decision about which style to use. Ideally, the service account key should be of no greater value than secrets used in other DNS scripts because the service account can be limited to a very fine grained scope. The proposed wiki instructions say to only give it DNS permissions, not e.g. permission to start VMs or make other networking changes. I will update the proposed wiki instructions to be more clear about the persistence and security considerations with service accounts.

I can also correct all of the minor concerns, except I may not change the `_has_record` check.

Thanks for the great review, and I'll let you know when it's ready for another look.

cbarrick:
Proposed addition to the wiki text:

> #### Security considerations
>
> Like all DNS API scripts that handle secrets, the account credentials are perissted to `~/.acme.sh/account.conf` to be available at renewal time. This applies to the `GCP_SERVICE_ACCOUNT_KEY`. If you are passing a service account key inline, you must ensure that read access to `account.conf` is only granted to the user that runs acme.sh (this is the default). If you are passing a path to a service account key file, you must ensure that access to the key file is similarly restricted.
>
> The best practice when creating a Google Cloud service account key is to use a service account that is reserved exclusively for acme.sh and to only grant that account permission to modify DNS. Avoid granting broad roles like Owner or Editor.

cbarrick:
@neilpang Please take another look!

cbarrick:
(Force pushed to correct a typo in a commit message)

neilpang:
**1. The DNS-API-Test on your fork is actually failing** — CheckToken passes but the Docker job dies at "Run acmetest" in ~34s, on both recent runs. It must run green before merge.

**2. Prefix your private helpers.** `_load_config`, `_save_config`, `_print`, `_get_json_str`, `_is_error`, `_pick_zone`, `_get_access_token`, etc. are all generic names. dnsapi scripts are sourced into the same shell as acme.sh and other hooks, so these will collide. Rename them `_gcloud_rest_*` (see `_hostup_*`, `_ab_*` in other plugins).

**3. A new dnsapi needs a dedicated "Report bugs to ..." tracking issue** (like #6700 for Hostup), referenced in an `Issues:` field in the info block — currently absent.

**4. Add the wiki usage entry** on [dnsapi2](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2), and point `Docs:` there (`dnsapi2#dns_gcloud_rest`) instead of the old `dnsapi` page.

PR #7014, opened 2026-06-07
Update dns_hostup.sh to v2 API

hostup:
Update to v2 api support; developer.hostup.se otherwise functionally the same.

It passes all action tests:
https://github.com/hostup/acme.sh/actions/runs/27082315990

### DIFF
--- a/dnsapi/dns_hostup.sh
+++ b/dnsapi/dns_hostup.sh
@@ -55,7 +55,6 @@ dns_hostup_add() {
 
   record_name_fqdn="$(_hostup_fqdn "$fulldomain")"
   if _hostup_find_record "$HOSTUP_ZONE_ID" "$record_name_fqdn" "$hostup_add_txtvalue"; then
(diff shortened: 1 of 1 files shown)


hostup:


Let me know if anything else need to be done from my end!

neilpang:
Thanks for the v2 update. A few suggestions before merge:

1. **Dead code in `dns_hostup_rm`:** `record_id_cached` is fetched and `_debug`-logged but never used — the old "delete by cached id" fast path is gone, so it's now vestigial. Either drop the lookup or restore it as a fast path.
2. **Possible duplicate `Content-Type` header:** `_hostup_rest` sets both `export _H2="Content-Type: application/json"` and passes `application/json` as the 5th arg to `_post`, so two `Content-Type` headers may be emitted. Keeping just one (dropping `_H2`) is cleaner.
3. **Minor:** legacy `HOSTUP_RECORD_<zone>_<domain>` (value-less) conf entries only get cleared on an `rm` of that exact domain, otherwise they linger as stale account-conf cruft. Harmless, just noting it.

None of these block merge — with at least #1 and #2 tidied up I'm happy to take it. Thanks!

hostup:
New test is running for last changes https://github.com/hostup/acme.sh/actions/runs/28341709993

PR #7023, opened 2026-06-15
Fix StrongSwan hook

sergiustheblack:
Split CAs into separate files since StrongSwan does not support multiple certs in CA bundle

https://hgot07.hatenablog.com/entry/2026/02/15/230930

### DIFF
--- a/deploy/strongswan.sh
+++ b/deploy/strongswan.sh
@@ -62,6 +62,12 @@ strongswan_deploy() {
 
 ####################  Private functions below ##################################
 
+__split_ca() {
+  _certfile="${1}"
+  _targetdir="${2}"
+  awk -v dir="${_targetdir}" '/-----BEGIN CERTIFICATE-----/{c++} c > 1 {print > (dir "/ca-" (c-1) ".cer")}' "${_certfile}"
+}
+
 __deploy_cert() {
   _cdomain="${1}"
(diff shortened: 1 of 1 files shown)


neilpang:
The awk split is correct, but two issues before merge:

**1. Orphaned CA files never cleaned up.** `__split_ca` only writes `ca-N.cer`, never clears the dir first. So: (a) the old single bundle from previous acme.sh versions (`$(basename "$_cca")`) stays and StrongSwan keeps loading it — the very bug this fixes; (b) if a renewal has fewer intermediates, a stale `ca-2.cer` lingers. Clear old `ca-*.cer` (and the legacy bundle) before splitting.

**2. `stroke` mode still writes the full bundle.** The `cat "$_cfullchain" > .../$_dir_ca/...` line drops the whole multi-cert fullchain into the CA dir — exactly what StrongSwan can't handle. Should stroke get the same split?

neilpang:
One issue: stale files. The previously deployed `$(basename _cca)` (e.g. `ca.cer`) is left behind in the CA dir after upgrading to this version, and if a renewal produces a shorter chain, old `ca-N.cer` files linger — swanctl keeps loading and trusting them. Please remove the hook's previously written `ca-*.cer` / old CA file before splitting.

PR #7027, opened 2026-06-16
fix(dns_desec): fix rate limit and compatibility issues

astehlik:
When acme.sh is executed for multiple domains you might run into errors when using deSEC because they have a rate limit of 2 write requests per second, see https://desec.readthedocs.io/en/latest/rate-limits.html#api-request-throttling

To prevent this, a sleep of 1 second is added after each write request.

Additionally two issues are fixed that were detected in failing GitHub workflows:

* The DNS name is converted to lowercase before talking to the API to prevent validation errors. The deSEC API only accepts lowercase DNS names
* The Regex used to extract old TXT values is now POSIX compatible to fix tests on OpenBSD

### DIFF
--- a/dnsapi/dns_desec.sh
+++ b/dnsapi/dns_desec.sh
@@ -39,6 +39,7 @@ dns_desec_add() {
     _err "invalid domain"
     return 1
   fi
+  _sub_domain=$(echo "$_sub_domain" | _lower_case)
   _debug _sub_domain "$_sub_domain"
   _debug _domain "$_domain"
 
@@ -100,7 +101,7 @@ dns_desec_rm() {
     _err "invalid domain"
     return 1
   fi
-
+  _sub_domain=$(echo "$_sub_domain" | _lower_case)
   _debug _sub_domain "$_sub_domain"
   _debug _domain "$_domain"
 
@@ -150,6 +151,8 @@ _desec_rest() {
   if [ "$m" != "GET" ]; then
     _secure_debug2 data "$data"
     response="$(_post "$data" "$ep" "" "$m")"
+    _info "Sleeping 1s to respect deSEC write rate limit"
+    sleep 1
   else
     response="$(_get "$ep")"
   fi
--- a/dnsapi/dns_desec.sh
+++ b/dnsapi/dns_desec.sh
@@ -49,7 +49,7 @@ dns_desec_add() {
   _desec_rest GET "$REST_API/$_domain/rrsets/$_sub_domain/TXT/"
 
   if [ "$_code" = "200" ]; then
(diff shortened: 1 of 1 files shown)


astehlik:

I had to make two additional fixes but now the Pipeline ran successfully:

* The subdomain name is now transformed to lowercase, because the deSEC API only allows lowercase chars and `acmetestXyzRandomName` which is used in the tests contains uppercase chars
* Also a non POSIX compliant regex is fixed to make the tests run on OpenBSD

https://github.com/astehlik/acme.sh/actions/runs/27911106572/

neilpang:
The sleep fixes the rate limit fine, but this PR actually bundles three changes under a "rate limit" title:

1. `sleep 1` after each write — works. One nit: please use `_sleep 1` instead of bare `sleep 1` (it's the acme.sh convention and shows a countdown in interactive mode).
2. Lowercasing `_sub_domain` — reasonable (deSEC's rrset path is case-sensitive), but unrelated to rate limiting and not mentioned in the description.
3. `\S*` → `[^ ]*` in the records regex — this is actually a good portability fix, since `\S` isn't POSIX and breaks the `_egrep_o` sed fallback and BSD/busybox grep. Also unrelated to the title.

#2 and #3 are both worth keeping — could you just update the title/description to cover all three? No need to split. With `sleep 1` → `_sleep 1`, this is good to merge. Thanks!

astehlik:
@neilpang Thank you for the feedback :slightly_smiling_face: 

I've adjusted your points (added fixes to description, replaced `sleep` with `_sleep`).

Sadly, the new build currently fails, because GhostBSD seems to have mirror issues:

https://github.com/astehlik/acme.sh/actions/runs/28331391983/

marrobHD:
Hello, the right fix would be to implement bulk requests to dnssec: https://desec.readthedocs.io/en/latest/dns/rrsets.html#bulk-operations

@astehlik would you like to implement that?

neilpang:
The hook already uses the bulk endpoint: `PUT /api/v1/domains/<domain>/rrsets/` with a JSON array body. It sends one element per call because acme.sh calls `dns_desec_add <name> <value>` once per challenge, each in its own subshell that re-sources the hook file (`acme.sh:5446-5463`) -- the hook never sees the other names and keeps no state between calls. There is nothing to batch at the hook level.

Real batching needs a new core entry point that hands the hook all (name, value) pairs at once, i.e. a change to the dnsapi contract shared by ~150 hooks. That's a separate FR, not a dns_desec change.

Note that `_sleep 1` only covers the 2/s bucket -- the per-domain write scope is also 15/min 100/h 300/day, so many SANs under one deSEC domain can still 429. The hook-level fix for that is honoring `Retry-After` on 429, not bulk.

PR #7033, opened 2026-06-19
Adding custom Port definitions for truenas

SirNomad49:
### Summary
This PR introduces support for configuring a custom port when connecting to the TrueNAS websocket API. Previously, the script assumed the default port based on the protocol (`ws`/`wss`). With this change, deployments can target environments where TrueNAS is exposed on non-standard ports.

---

### Changes
- ✅ Added new optional environment variable:
  - `DEPLOY_TRUENAS_PORT`
- ✅ Extended URI construction to include port when provided:
  - `ws(s)://<host>:<port>/websocket`
- ✅ Added validation to ensure the port is numeric
- ✅ Updated environment checks and debug output
- ✅ Persisted port via `_savedeployconf`
- ✅ Improved error messaging to include port configuration
- ✅ Extended return codes:
  - `8: Invalid port`

---

#### Default (no port)
```sh
export DEPLOY_TRUENAS_PROTOCOL="wss"
export DEPLOY_TRUENAS_HOSTNAME="truenas.example.local"
```
Result: 
`wss://truenas.example.local/websocket`
#### with custom port
```sh
export DEPLOY_TRUENAS_PROTOCOL="wss"
export DEPLOY_TRUENAS_HOSTNAME="truenas.example.local"
export DEPLOY_TRUENAS_PORT="8443"
```
Result: 
`wss://truenas.example.local:8443/websocket`


neilpang:
Thanks for the PR, but as written the port feature is a no-op. The new `if/else` builds `_ws_uri` correctly with the port, but the old line right after it is still there:

```sh
_ws_uri="$DEPLOY_TRUENAS_PROTOCOL://$DEPLOY_TRUENAS_HOSTNAME/websocket"
```

That unconditionally overwrites `_ws_uri` with the port-less URI, so `DEPLOY_TRUENAS_PORT` never actually takes effect. Please delete that leftover line.

Also shfmt is failing — the new `# Check port` / `if` block sits at column 0 instead of the function's 2-space indent, plus a blank line with trailing whitespace. `shfmt -i 2 -w deploy/truenas_ws.sh` fixes it.

After that, please re-test with a port set and confirm the URI actually comes out as `wss://host:8443/websocket`. Thanks!

PR #7048, opened 2026-06-25
fix(dns_infomaniak): correctly detect API errors

tsueri:
The add/rm success check never rejected anything: for any non-empty API
response it always reported "Record added"/"Record deleted" and returned
0, so the _err branch was dead code. A valid key looked fine only because
the API call genuinely created the record; an invalid key returning
{"result":"error"} produced the same "Record added" output even though
nothing was created.


tsueri:
I read the DNS API Dev Guide. All tests pass.

neilpang:
Nice catch on the `dns_infomaniak` bug — the analysis is spot on. `_contains` only sets an exit code and prints nothing to stdout, so `[ ! "$(... | _contains ...)" ]` was always true for any non-empty response, making the `_err` branch dead code. The fix (checking `_contains`'s exit status directly) is correct, and keeping the `"result":"success"` marker matches the Infomaniak v2 API response format. 👍

Could you drop the `README.md` changes from this PR, though? The ZeroSSL banner rework is unrelated to the DNS API fix, and it also touches the sponsor link (the tracking param and the banner images), which I'd rather keep as a separate, deliberate change — especially since the banner was just updated on `master` (`New Banner`, the current HEAD), so this hunk would likely conflict with or revert that.

If you'd still like to propose the dark/light-mode banner, please open a separate PR for it and we can discuss it there. With the README hunk removed, the `dns_infomaniak` fix is good to merge.

Thanks!

tsueri:
Thanks for the review! The README banner change was unrelated and slipped in from the sync merge. I've removed it from this PR. Should be good to go.

tsueri:
Btw: I created a issue [here](https://github.com/acmesh-official/acmetest/issues/35) after I found this bug. I propose a negative test path to find this class of failures. Have a look! What do you think?

PR #7051, opened 2026-06-27
feat: introduce --no-sudo-check flag

mrl5:
## use case
As an admin I want to have a cron job that opens port `80` only for certificate renewal without a need of running acme.sh as root but still in standalone mode.

cert init:
```console
./acme.sh --issue  -d example.com  --standalone
```
cron:
```console
sudo -u acmesh /home/acmesh/acme.sh/acme.sh --cron 
```

## issue
currently to bypass sudo constraint I can use `--force`, but together with `--cron` flag I end up with certificate renewal on each call
```console
sudo -u acmesh /home/acmesh/acme.sh/acme.sh --cron --force
```

## proposed solution
introduce `--no-sudo-check` flag so that cron job will renew certificates only when they are about to expire and I can still invoke it with sudo
```diff
- sudo -u acmesh /home/acmesh/acme.sh/acme.sh --cron
+ sudo -u acmesh /home/acmesh/acme.sh/acme.sh --cron --no-sudo-check
```

## additional context
I have this daily cron job and the core of it includes this 3 lines, where `setcap` allows non root users to bind into low ports (like 80)
```console
sudo setcap CAP_NET_BIND_SERVICE=+ep /usr/bin/socat1
sudo -u acmesh /home/acmesh/acme.sh/acme.sh --cron --no-sudo-check
sudo setcap -r /usr/bin/socat1
```
the goal that I want to achieve is:
1. open port 80 only when cert needs to be renewed
2. don't run acme.sh (and other child processes) as root
3. use standalone mode that relies on socat

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -8056,6 +8056,9 @@ _process() {
     -f | --force)
       FORCE="1"
       ;;
+    --no-sudo-check)
+      NO_SUDO_CHECK="1"
+      ;;
     --treat-skip-as-success | --treatskipassuccess)
       _TREAT_SKIP_AS_SUCCESS="1"
       ;;
@@ -8488,11 +8491,14 @@ _process() {
 
   if [ "${_CMD}" != "install" ]; then
     if [ "$__INTERACTIVE" ] && ! _checkSudo; then
-      if [ -z "$FORCE" ]; then
+      if [ -z "$FORCE" ] && [ -z "$NO_SUDO_CHECK" ]; then
         #Use "echo" here, instead of _info. it's too early
         echo "It seems that you are using sudo, please read this page first:"
         echo "$_SUDO_WIKI"
         return 1
+      fi
+      if ! [ -z "$NO_SUDO_CHECK" ]; then
+        echo "Running as $(whoami)"
       fi
     fi
     __initHome


neilpang:
Thanks for the PR and the detailed write-up of the use case 👍

Before merging I want to make sure the flag is actually needed for the cron scenario you describe, because I think the sudo check may already not fire there.

The sudo check only runs when the shell is interactive:

```sh
if [ "$__INTERACTIVE" ] && ! _checkSudo; then
```

and `__INTERACTIVE` is only set when stdout is a tty:

```sh
__INTERACTIVE=""
if [ -t 1 ]; then
  __INTERACTIVE="1"
fi
```

In a real crontab job stdout is not a tty, so `__INTERACTIVE` is empty and the whole sudo-check block is skipped already. That means `sudo -u acmesh ./acme.sh --cron` (without `--force` and without this new flag) should run fine and only renew when certs are about to expire.

Could you confirm whether the issue actually reproduces from a real cron entry (non-tty), as opposed to running the command by hand in a terminal? If it only shows up interactively, the flag is mostly a convenience for manual testing rather than a fix for the cron path — which is fine, but it changes how we'd document it.

Thanks again!

mrl5:
ow.. you might be actually right. I'll close the PR after double checking your hint. Thanks!

mrl5:
yup, you were right @neilpang -- thanks for detailed explanation!

PR #7054, opened 2026-06-29
Remove deprecated Linode API v3 DNS plugin

ChihweiLHBird:
The `dns_linode.sh` plugin relied on the retired Linode API v3.
Users should use `dns_linode_v4.sh` instead, which is already in `dnsapi`.


neilpang:
update the wiki first:

https://github.com/acmesh-official/acme.sh/wiki/dnsapi

ChihweiLHBird:
Hi @neilpang, I don't see any Linode APIv3 usage in the wiki. The only Linode content left is APIv4, which is the latest operative version of the API.

neilpang:
Confirmed - the wiki is already v4-only (dnsapi section 14, dns_linode_v4 / LINODE_V4_API_KEY); there is no v3 content left anywhere in it.

v3 is dead upstream too: `https://api.linode.com/?api_key=...&api_action=domain.list` now 301s to `https://api.linode.com/v4/`, which 404s. The hook cannot work for anyone.

Merging.

PR #7056, opened 2026-06-29
dns_inwx: fix IDN zone detection without python dependency

### LINKED ISSUE #7038: dns_inwx.sh: zone detection fails for IDN/punycode domains (INWX returns zone names in Unicode)
## Bug Description

`dns_inwx.sh`'s `_get_root()` function fails to detect the zone for IDN domains
specified in punycode ACE form (e.g. `xn--bro-6qa.de`). The DNS-01
challenge TXT record ends up being added to the `de` TLD zone instead of the
correct zone, which obviously fails.

## Reproduction

Issue a certificate that includes a punycode SAN for an IDN domain managed in INWX:

```bash
acme.sh --issue --dns dns_inwx --keylength ec-256 \
  -d www.buro.de \
  -d buro.de \
  -d www.xn--bro-6qa.de \
  -d xn--bro-6qa.de

Error Output (--debug)

[...] First detect the root zone
[...] get root
[...] h='www.xn--bro-6qa.de'
[...] h='xn--bro-6qa.de'
[...] h='de'
[...] _sub_domain='_acme-challenge.www.xn--bro-6qa'
[...] _domain='de'
[...] Adding record
[...] body='... <string>de</string> ... <string>_acme-challenge.www.xn--bro-6qa</string> ...'
[...] Error
[...] Error adding TXT record to domain: _acme-challenge.www.xn--bro-6qa.de

The non-IDN SANs (www.buro.de, buro.de) succeed because those zone names
appear literally in INWX's nameserver.list response.

Root Cause

_get_root() calls nameserver.list once and then walks candidate zone names
from left to right, calling _contains "$response" "$h" for each. INWX's API
returns zone names in Unicode form (e.g. büro.de) even when the domain
was registered as an IDN. So when acme.sh searches for the punycode form
xn--bro-6qa.de, _contains returns false, the loop falls through to de,
and the record is added to the wrong zone.

The same pattern likely affects any DNS provider that returns zone names in
Unicode in its list API.

Fix

Add a Unicode fallback immediately after the _contains check in _get_root().
When the punycode form is not found, convert the label to Unicode via python3
and check again. If the Unicode form matches, use it as _domain (INWX's API
also expects Unicode in nameserver.createRecord).

--- a/dnsapi/dns_inwx.sh
+++ b/dnsapi/dns_inwx.sh
@@ _get_root()
     if _contains "$response" "$h"; then
       _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-"$p")
       _domain="$h"
       return 0
     fi
+    # IDN fallback: INWX returns zone names in Unicode; convert punycode label and retry
+    _h_idn=$(python3 -c "import sys; d=sys.argv[1]; print('.'.join(p.encode('ascii').decode('idna') if p.startswith('xn--') else p for p in d.split('.')))" "$h" 2>/dev/null)
+    if [ -n "$_h_idn" ] && [ "$_h_idn" != "$h" ] && _contains "$response" "$_h_idn"; then
+      _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-"$p")
+      _domain="$_h_idn"
+      return 0
+    fi
     p=$i
     i=$(_math "$i" + 1)

With this patch, _domain is set to büro.de (Unicode) and
nameserver.createRecord succeeds because that is exactly what INWX expects.

Environment

- acme.sh version: 3.1.4
- DNS provider: INWX (api.domrobot.com/xmlrpc/)
- Debian 13 (Trixie), python3 available
- IDN domain with punycode SAN; zone returned as Unicode by INWX nameserver.list

Notes

- python3 is used for punycode→Unicode conversion because it is available
everywhere; if idn/idn2 is preferred for portability, those can be
substituted with a fallback.
- The fallback is a no-op for non-IDN domains (the xn-- prefix check
short-circuits immediately).
- Cert renewal also fails without this patch (not only initial issuance).

bluenenschloss:
Fixes #7038

## Problem

INWX returns zone names in Unicode form (e.g. `lünenschloß.de`) even when the domain was registered as an IDN. When acme.sh passes the SAN in punycode (`xn--lnenschlo-o1a42a.de`), `_contains "$response" "$h"` never matches and `_get_root` falls through to the TLD, placing the ACME DNS challenge record in the wrong zone.

## Fix

Replace the previous `python3`-based ACE→Unicode conversion with `_idn()`, which already exists in acme.sh for this purpose:

- Guard with `_contains "$h" "xn--"` to short-circuit non-IDN domains
- Extract `<string>` values from the `nameserver.list` XML response
- Encode each via `_idn()` and compare to `$h` (ACE form)
- On match, store the original Unicode zone name as `_domain` for `createRecord`
- If `idn` is not installed, `_idn()` returns empty → comparison fails → falls through silently (no regression)

This removes the `python3` hard-dependency and uses the same IDN convention as the rest of acme.sh.

## Testing

Tested on Debian 13 with `xn--lnenschlo-o1a42a.de` (lünenschloß.de) — certificate issuance and renewal both succeed.

### DIFF
--- a/dnsapi/dns_inwx.sh
+++ b/dnsapi/dns_inwx.sh
@@ -315,7 +315,7 @@ _get_root() {
     # IDN fallback: INWX returns Unicode zone names; when $h is ACE/punycode,
     # encode each zone name via _idn() and compare — no python dependency.
     if _contains "$h" "xn--"; then
-      _zone_unicode=$(printf "%s" "$response" | _egrep_o '<string>[^<]*</string>' |
+      _zone_unicode=$(printf "%s" "$response" | _egrep_o '<string>[^<]*' |
         sed 's/<[^>]*>//g' | while IFS= read -r _z; do
         if [ "$(_idn "$_z")" = "$h" ]; then
           printf "%s" "$_z"
--- a/dnsapi/dns_inwx.sh
+++ b/dnsapi/dns_inwx.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/bash
 # shellcheck disable=SC2034
 dns_inwx_info='INWX.de
 Site: INWX.de
@@ -311,6 +311,22 @@ _get_root() {
       _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-"$p")
       _domain="$h"
       return 0
+    fi
+    # IDN fallback: INWX returns Unicode zone names; when $h is ACE/punycode,
+    # encode each zone name via _idn() and compare — no python dependency.
+    if _contains "$h" "xn--"; then
+      _zone_unicode=$(printf "%s" "$response" | grep -o '<string>[^<]*</string>' | \
+        sed 's/<[^>]*>//g' | while IFS= read -r _z; do
+          if [ "$(_idn "$_z")" = "$h" ]; then
+            printf "%s" "$_z"
+            break
+          fi
+        done)
+      if [ -n "$_zone_unicode" ]; then
+        _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-"$p")
+        _domain="$_zone_unicode"
+        return 0
+      fi
     fi
     p=$i
     i=$(_math "$i" + 1)
--- a/dnsapi/dns_inwx.sh
+++ b/dnsapi/dns_inwx.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env sh
 # shellcheck disable=SC2034
 dns_inwx_info='INWX.de
 Site: INWX.de
@@ -315,13 +315,13 @@ _get_root() {
     # IDN fallback: INWX returns Unicode zone names; when $h is ACE/punycode,
     # encode each zone name via _idn() and compare — no python dependency.
     if _contains "$h" "xn--"; then
-      _zone_unicode=$(printf "%s" "$response" | grep -o '<string>[^<]*</string>' | \
+      _zone_unicode=$(printf "%s" "$response" | _egrep_o '<string>[^<]*</string>' |
         sed 's/<[^>]*>//g' | while IFS= read -r _z; do
-          if [ "$(_idn "$_z")" = "$h" ]; then
-            printf "%s" "$_z"
-            break
-          fi
-        done)
+        if [ "$(_idn "$_z")" = "$h" ]; then
+          printf "%s" "$_z"
+          break
+        fi
+      done)
       if [ -n "$_zone_unicode" ]; then
         _sub_domain=$(printf "%s" "$domain" | cut -d . -f 1-"$p")
         _domain="$_zone_unicode"


bluenenschloss:
DNS-API-Test passed on the fork: https://github.com/bluenenschloss/acme.sh/actions/runs/28369726203

The Docker (Linux), macOS, Windows, and FreeBSD jobs all succeeded. The GhostBSD job failed due to a QEMU VM infrastructure timeout (orphan processes terminated at cleanup), unrelated to the `dns_inwx.sh` change.

bluenenschloss:
**Note on IDNA2003 vs IDNA2008 limitation**

This fix uses `_idn()` as suggested, which internally calls the `idn` command (GNU Libidn, IDNA2003). For most IDN domains this works correctly.

However, domains containing the German sharp-s `ß` expose an IDNA2003/IDNA2008 incompatibility: IDNA2003 maps `ß` → `ss` before encoding, while IDNA2008 preserves it. For example:

- `lünenschloß.de` via `idn` (IDNA2003) → `xn--lnenschloss-thb.de`
- `lünenschloß.de` via `idn2` (IDNA2008) → `xn--lnenschlo-o1a42a.de`

If INWX stores such a zone in Unicode form and the ACME client passes the IDNA2008 ACE name, the `_idn()`-based comparison will not match. This is a pre-existing limitation of `_idn()` in acme.sh itself (it only calls `idn`, not `idn2`) and is not specific to this fix — the same mismatch would affect `txtdomain=$(_idn "$txtdomain")` and other callers. Noting it here for visibility.

neilpang:
`_idn()` approach is right, logic looks correct. Two CI fixes:

**1. Revert the shebang** (`#!/usr/bin/bash` → `#!/usr/bin/env sh`). It's wrong anyway (hardcoded path, dnsapi is POSIX sh), and it's the *only* reason ShellCheck fails — the SC2001 errors are on lines 175-178, which you didn't touch; they only appear because shellcheck now lints as bash. Reverting makes ShellCheck green.

**2. shfmt:** drop the `\` after the pipe and reindent — `shfmt -i 2 -w dnsapi/dns_inwx.sh`.

Optional: `_egrep_o` instead of raw `grep -o` (more portable).

With those two, good to merge. Thanks!

bluenenschloss:
Done — pushed [7391bfe](https://github.com/bluenenschloss/acme.sh/commit/7391bfe4):

1. Shebang reverted to `#!/usr/bin/env sh`
2. `grep -o` → `_egrep_o`
3. `shfmt -i 2 -w dnsapi/dns_inwx.sh`

Thanks for the review!

neilpang:
The DNS-API-Test on your fork is currently failing — it needs to pass before this can be merged. Please get that run green.

One optional improvement: `_egrep_o '<string>[^<]*</string>'` contains a `/` (in `</string>`), which collides with the `s/…/…/` delimiter in `_egrep_o`'s sed fallback, so it breaks on systems without `egrep -o`. Dropping the closing tag — `<string>[^<]*` — works in both backends (the following `sed 's/<[^>]*>//g'` still strips the tag).

bluenenschloss:
Fixed in [10b47cb](https://github.com/bluenenschloss/acme.sh/commit/10b47cb03fda99f24e3af595aa4892d2f7b0af7c) — dropped `</string>` from the `_egrep_o` pattern so the `/` in the closing tag no longer collides with the sed fallback delimiter.

Regarding the CI: only GhostBSD is failing, and it is a `vmactions/ghostbsd-vm@v1` VM infrastructure failure (`orphan processes terminated at cleanup`), not related to the code change. Same failure mode as the previous run. Docker, macOS, Windows, and FreeBSD all pass.

Issue #7057, opened 2026-07-01
Report bugs to Glesys DNS API


tonikarppi:
Leave comments here to report bugs with the Glesys DNS API. Please read the [acme.sh wiki entry](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_glesys) for the provider before reporting a bug.

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

PR #7059, opened 2026-07-01
Add Glesys dnsapi provider

tonikarppi:
This PR adds a DNS API provider for Glesys.

Followed instructions on DNS API Dev Guide. CI secrets are configured as documented.

For clarity: I'm not affiliated with Glesys. This was implemented using their [public API](https://github.com/GleSYS/api-docs/wiki/API-Documentation).

- [Wiki entry](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_glesys).
- [Bug report issue](https://github.com/acmesh-official/acme.sh/issues/7057).

### DIFF
--- a/dnsapi/dns_glesys.sh
+++ b/dnsapi/dns_glesys.sh
@@ -1,0 +1,240 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_glesys_info='Glesys
+Site: Glesys.se
+Docs: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_glesys
+Options:
+ GLESYS_API_KEY Generated API key.
+ GLESYS_PROJECT_ID Project ID for the API key (e.g. cl12345).
+ GLESYS_API API endpoint. Default "https://api.glesys.com/domain".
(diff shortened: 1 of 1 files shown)


neilpang:
Two changes before merge:

**1. Don't persist the record id locally — look it up from the API every time.** Right now `add` saves the record id to domainconf (`GLESYS_RECORD_ID_<sha256>`) and `rm` reads it back to delete. Please drop that state entirely: in `rm`, call GleSYS `/listrecords` for the domain, match on host + `data` (the txt value), and delete by the id you get back. That survives state loss (other machine, cleared conf, records created elsewhere). Also treat "no matching record found" as success (`return 0`), not an error.

**2. Wrong error message in `add`.** The `response_code != 200` branch in `dns_glesys_add` says "...when attempting to **remove** TXT record" — should be "add".

tonikarppi:
Thanks for the feedback, @neilpang. I've updated the branch with the requested changes. 

Also updated the wiki docs with the change in permissions needed from this changes. Let me know if you have additional suggestions, otherwise I can squash the commits for merge.

Issue #7079, opened 2026-07-04
Add support for ACME Account Key Rollover


sim0n-v:
Hi,

The rotation of the acme account key is not implemented (yet).
The `ACME_KEY_CHANGE` URL is fetched from directory but never used.

Added in my todo:
* Ability to renew the acme account key
* Ability to change the acme account key type (e.g. rsa to ecc or ecc to rsa)

Ref: [RFC8555 §7.3.5](https://www.rfc-editor.org/info/rfc8555/#section-7.3.5)

PR #7080, opened 2026-07-04
Add support for Account Key Rollover

sim0n-v:
Response to https://github.com/acmesh-official/acme.sh/issues/7079, implements the ACME Account `key-change` (a.k.a Account Key Rollover).
This makes it possible to rotate the ACME account key as described in [RFC8555 §7.3.5](https://datatracker.ietf.org/doc/html/rfc8555/#section-7.3.5).

Tested with the following directories:
* LE Staging ✅
* LE (prod) ✅
```
acme.sh --server <directory> --register-account
acme.sh --server <directory> --update-account-key -ak 3072
acme.sh --server <directory> --update-account-key -ak P-384
```
Not tested with directories requiring EAB, but it should work as well.

What is supported:
* Rekey with the same algorithm/keylength
* Rekey with another algorithm/keylength (RSA<--->ECC and from a keysize to another)

_Nb: Not specifying `-ak` (or `--accountkeylength`) will use the default (for now: P-256)._

Just in case, the old key is still saved in `${ACCOUNT_KEY_PATH}.old`.

I can add a wiki page if needed.


neilpang:
1. `if [ -z "_accUri" ]` -- missing `$`, the guard never fires. With no `ACCOUNT_URL` in ca.conf the request goes out with `"account": ""` and an empty `kid`. Compare `updateaccount()`: `[ -z "$_accUri" ]`.

2. Inner protected header is built from `JWK_HEADERPLACE_PART1` (`{"nonce": "`), so the inner JWS contains `"nonce": ""`. RFC 8555 sec 7.3.5: the inner JWS MUST omit `nonce`. LE/Pebble pass only because go-jose parses `"nonce":""` as empty and Boulder only rejects a non-empty inner nonce; strict CAs reject as malformed. Build `{"alg":...,"url":...,"jwk":...}` explicitly.

3. 409 branch calls `_error` -- undefined (helper is `_err`), prints `command not found`. Both 409/else branches use `exit 1`; siblings use `return 1`.

4. `CA_KEY_HASH` not refreshed after the swap -- next `--issue`/`--renew` mismatches and re-runs `_regAccount` (breaks EAB CAs with single-use credentials). Add `_savecaconf CA_KEY_HASH "$(__calcAccountKeyHash)"` and recalculate `ACCOUNT_THUMBPRINT` like `updateaccount()`.

5. `[ -z "$ACME_KEY_CHANGE" ]` returns 1 silently -- add an `_err`.

6. Swap is unsafe: a second rollover overwrites the only `.old` backup; if the second `mv` fails, `ACCOUNT_KEY_PATH` is gone while the server already switched keys. Use `mv -f`, swap in a reverse-safe order. `_calcjwk`/`_sign` failure paths leave an orphan `.new` key -- `rm -f` there too.

7. Add a `showhelp` entry for `--update-account-key`.

8. `length`, `url`, `body` are generic unprefixed globals -- use `_`-prefixed names.

sim0n-v:
Hey!
Thanks for the review.

I dropped the `.old` key file since it is linked to the `ca.conf` and the `CA_KEY_HASH`.
The goal is to use the `.new` key file (temporarily), check everything then overwrite the key.

neilpang:
1. `ACCOUNT_THUMBPRINT` / `CA_KEY_HASH` are still derived from the OLD key: `_send_signed_request ... "$ACCOUNT_KEY_PATH"` re-runs `_calcjwk` on the old key (so `$jwk` is old when `__calc_account_thumbprint` runs), and `__calcAccountKeyHash` reads `ACCOUNT_KEY_PATH` before the `mv`. The saved hash then mismatches on the next run and `_regAccount` re-fires -- the EAB problem from round 1. Fix: `mv` right after the 200 check, then `__cache_jwk_path=""` and `_calcjwk "$ACCOUNT_KEY_PATH"` (the jwk cache is keyed by path and would return the old jwk otherwise), then recompute thumbprint and hash.
2. Round-1 point 5 still open: `[ -z "$ACME_KEY_CHANGE" ]` returns 1 silently. Also the `_err "Server does not expose keyChange url."` sits on the `_calcjwk` failure branch -- the two messages are swapped.
3. Drop the commented-out `# mv -f ... .old` line.

Rest of round 1 is addressed; dropping `.old` is fine.

sim0n-v:
I get it (I think): the `$jwk` is not updated if `$__CACHED_JWK_KEY_FILE = $keyfile`. Since the file is overwritten, the path remains the same.

So I re-run `_calcjwk "$ACCOUNT_KEY_PATH_NEW"` before updating everything.

neilpang:
Last one: in the 200 branch, a `_calcjwk "$ACCOUNT_KEY_PATH_NEW"` failure `rm -f`s the new key -- but the server has already switched to it at that point, so deleting it locks the account out. After a 200 the key file must survive unconditionally: `mv -f` first, then refresh the jwk with `__CACHED_JWK_KEY_FILE="" ; _calcjwk "$ACCOUNT_KEY_PATH"` (the cache still holds the old jwk for that path) and recompute the thumbprint/hash. If the recalc somehow fails, warn -- don't delete.

Rest of rounds 1-2 is all addressed.

sim0n-v:
Nice one!

Fixed: the key is overwritten when the server replies with 200. The old key is not useful anymore.

neilpang:
Two more things before merge:

1. Add a wiki page documenting this feature -- what `--update-account-key` does (RFC 8555 sec 7.3.5 key rollover), when to use it, and usage examples. Link it here once created.
2. Please manually verify the rollover actually works against other CAs besides Let's Encrypt -- e.g. ZeroSSL and Google Trust Services (both EAB-based) -- and report the results here.

sim0n-v:
Interesting! Tested with:
* zerossl: ❌ `Account key rollover error: {"type":"urn:ietf:params:acme:error:serverInternal","status":500,"detail":"Not yet implemented"}`
* ssl.com: ❌ `Account key rollover error: {"message":"Missing Authentication Token"}`

Seems like this feature is not wildly supported (yet).

I think it is only a matter of time: the [Chrome Root Program Policy](https://googlechrome.github.io/chromerootprogram/) specifies that (§1.3.3.1.1)
> If ACME is supported: 
Each endpoint MUST support the following capabilities, as specified in [RFC 8555](https://www.rfc-editor.org/rfc/rfc8555): **keyChange**, newAccount, newNonce, newOrder, and revokeCert.

I will create a wiki mentioning that this feature is not yet supported by all CAs.

Can the CI integrate these tests? I had to create SSL.com and ZeroSSL account for testing purposes.

neilpang:
https://github.com/acmesh-official/acmetest/commit/af6802e354bf9d66a030f1391af49ef7ed99dde8

PR #7102, opened 2026-07-07
Add DNS API support for ps.kz

PyBorov:
### Add DNS API support for ps.kz

This adds `dns_pskz.sh`, a DNS-01 plugin for [ps.kz](https://ps.kz), a
Kazakhstan-based hosting provider and domain registrar.

#### Background

ps.kz doesn't publish API documentation for their DNS console. The
GraphQL schema used by this plugin was recovered via introspection
against their DNS console endpoint, and the correct auth header
(`X-User-Token`) was confirmed against their (separately documented)
Cloud API docs, since the DNS endpoint itself has none.

Two things worth flagging for review, since they shaped the
implementation:

- **API tokens are bound to a single ps.kz account at creation time.**
  The token format is `secret.accountId.userId`, but the `accountId`
  segment is cosmetic — changing it doesn't change which account the
  token actually resolves to (confirmed empirically). If a user's
  zones are split across multiple accounts, they need a separate
  token/issuance per account. This is documented in the README entry
  and in a comment at the top of the script.
- **Zone names are returned with a trailing dot** (e.g. `example.kz.`),
  matching standard DNS FQDN notation. The plugin's zone-resolution
  logic (`_pskz_find_zone`) accounts for this.

#### How it works

`_pskz_find_zone` resolves the correct zone for a given domain by
querying `dns.zones(searchName: ...)` and walking up the label
hierarchy until it finds an exact match the token has access to (e.g.
for `_acme-challenge.wiki.example.kz` it tries
`_acme-challenge.wiki.example.kz`, `wiki.example.kz`, `example.kz`,
`kz`). This means subdomains resolve correctly without the user having
to specify the parent zone explicitly.

#### Checklist

- [x] Implements `dns_pskz_add()` / `dns_pskz_rm()` per the [DNS API
      Dev Guide](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide).
- [x] Uses only standard acme.sh helpers (`_post`, `_err`, `_info`,
      `_debug`, `_debug2`, `_contains`, `_readaccountconf_mutable`,
      `_saveaccountconf_mutable`) — no external dependencies (no `jq`,
      etc).
- [x] `shellcheck -s sh dns_pskz.sh` passes with zero warnings.
- [x] Manually tested end-to-end (issue + renew + cleanup) against a
      real ps.kz account and zone — *(to be checked before submitting;
      the logic has been unit-tested against mocked `_post` responses,
      see below, but not yet run against the live API through acme.sh
      itself)*.
- [x] `_pskz_find_zone` and `_pskz_extract_record_id` were additionally
      unit-tested against mocked `_post` responses covering: record
      creation, record cleanup, and the cleanup no-op path when the
      expected TXT record is already gone.
- [x] Added a README entry with usage instructions (see diff).

A companion Python plugin for Certbot
([`certbot-dns-pskz`](https://github.com/PyBorov/certbot-dns-pskz)) and
a variant for Proxmox VE's built-in ACME client exist in the same
repo, implementing the same zone-resolution logic, for anyone who
wants to cross-check the approach.

Happy to adjust naming, structure, or add anything else the
maintainers want before merge.

Wiki:
## Use ps.kz DNS API

ps.kz is a Kazakhstan-based hosting and domain registrar. This plugin
uses their DNS GraphQL API.

First you need to log in to your ps.kz console and generate an API
token. The token is bound to a single ps.kz account at creation time —
if your zones are split across multiple accounts, you'll need one
token (and one issuance) per account.

```sh
export PSKZ_Token="xxxxxxxxxxxxxxxx.accountid.userid"
```

To issue a cert:

```sh
acme.sh --issue --dns dns_pskz -d example.kz -d '*.example.kz'
```

The `PSKZ_Token` will be saved in `~/.acme.sh/account.conf` and reused
automatically on renewal.


neilpang:
1. No DNS.yml run in your fork — the DNS-API-Test must pass there before review. Your own PR body says the plugin has never run against the live API through acme.sh; mock tests don't count.
2. shfmt fails, run `shfmt -w dnsapi/dns_pskz.sh`.
3. Use the structured `dns_pskz_info='...'` header (Site/Docs/Options/Issues/Author — see any recent dnsapi file), not free-form comments.
4. New dnsapi needs a tracking issue (in `Issues:`) and a dnsapi2 wiki entry (in `Docs:`).
5. `sed 's/},{/}\n{/g'`: BSD sed treats `\n` as literal `n`, split fails and `head -n 1` can return the wrong record id. Use `tr` or a literal newline.

PR #7107, opened 2026-07-08
fix: dnsexit api rejects a TTL of zero, changed to 1 (minute)

fuzzkat:
`dns_dnsexit.sh` submits DNS record additions with "ttl":0 in the JSON payload. DNSExit's current API (https://api.dnsexit.com/dns/) rejects this with: `{"code":3,"message":"Missing Required Parameters - ttl must be >= 1 (ttl is in minutes; the value is multiplied by 60 internally)"}`

The script does not inspect the API response body, so it proceeds as if the record was added. Let's Encrypt validation then fails because the _acme-challenge TXT record never publishes.

**Expected:** TXT challenge record is published; certificate is issued.

**Actual:** Task log shows "Add TXT record" and "Sleeping 150 seconds..." but the record never appears on DNSExit's authoritative nameservers. LE returns status: invalid for the challenge.

**Verification of API behaviour:**
Same payload the script sends, reproduced with curl:
```
curl -sS -X POST https://api.dnsexit.com/dns/ \
  -H 'Content-Type: application/json' -H "apikey: <redacted>" \
  -d '{"domain":"example.com","add":{"type":"TXT","name":"_acme-challenge.host","content":"test","ttl":0,"overwrite":false}}'
```
→ `{"code":3,"message":"Missing Required Parameters - ttl must be >= 1 ..."}`

Same payload with `"ttl":1`:
→ `{"code":0,"details":["Successfully added TXT ..."],"message":"Success"}`

Fix: change `"ttl":0` to `"ttl":1` in the add payload (TTL is in minutes per DNSExit's API).

### DIFF
--- a/dnsapi/dns_dnsexit.sh
+++ b/dnsapi/dns_dnsexit.sh
@@ -25,7 +25,7 @@ dns_dnsexit_add() {
     return 1
   fi
 
-  _dnsexit_zone_op add ',"ttl":0,"overwrite":false'
+  _dnsexit_zone_op add ',"ttl":1,"overwrite":false'
 }
 
 #Usage: fulldomain txtvalue


neilpang:
Please run the [DNS-API-Test](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test) in your own fork with your `DNSEXIT_API_KEY` set as a repository secret, and post a link to the passing run here. This is required for any dnsapi change before review/merge.

neilpang:
The NetBSD job died at VM boot (qemu terminated during setup) -- an infra flake, not your change. Everything that actually ran passed, but the rest of the matrix was skipped when the chain aborted. Hit "Re-run failed jobs" on that workflow run so the remaining platforms execute, and post the green link -- then this is ready to merge.

PR #7110, opened 2026-07-09
Docker: update crontab used, disable uninstall/install-cronjob cmds

invario:
- Docker: Update `entry.sh` to generate a `crontab` that runs more frequently keeping in line with PR #6939 
- Disable the `--install-cronjob` and `--uninstall-cronjob` commands when it detects it is running in a Docker container since the `crontab` command is not applicable anymore in Docker with `supercronic`


invario:
Hmm, this is unfortunate as it's causing a failure due to the way the tests are built. I'm not sure of the best way to proceed. I can just remove the changes I made to `--install-cronjob` and `--uninstall-cronjob` I suppose which would allow the tests to pass then.

invario:
I am closing this PR and will replace it with a new one that only updates the cron expression used in `entry.sh`.  

Replaced by PR #7111

Issue #7121, opened 2026-07-13
Report bugs to Route53 IAM Roles Anywhere (dns_aws X.509 auth)


kisst:
This is the support/tracking issue for authenticating the **Route53 DNS API (`dns_aws`) via AWS IAM Roles Anywhere** (X.509 client-certificate auth), as an alternative to long-lived `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`.

Report bugs or questions about the Roles Anywhere auth path here.

### What it does

acme.sh signs a `CreateSession` request with a client certificate's private key, receives **temporary** AWS credentials, and uses them with the existing Route53 SigV4 path. No long-lived secret and no dedicated IAM user are stored on the host — the certificate assumes an IAM role directly.

### Configuration

```sh
export AWS_RA_TRUST_ANCHOR_ARN="arn:aws:rolesanywhere:<region>:<acct>:trust-anchor/<id>"
export AWS_RA_PROFILE_ARN="arn:aws:rolesanywhere:<region>:<acct>:profile/<id>"
export AWS_RA_ROLE_ARN="arn:aws:iam::<acct>:role/<role>"
# optional: AWS_RA_CERT (default ~/.aws/rolesanywhere/certificate.pem),
#           AWS_RA_KEY  (default ~/.aws/rolesanywhere/private-key.pem),
#           AWS_RA_REGION (else parsed from the trust-anchor ARN),
#           AWS_RA_DURATION (default 3600, range 900-43200)
acme.sh --issue --dns dns_aws -d example.com -d '*.example.com'
```

Supersedes the original request in #4303. Implementation PR: to follow.

When reporting a problem, please include the `--debug 2` log (redact the signature / any credentials).

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

Issue #4303, opened 2022-09-09
Switch Route53 authentication from access key to certificate


kisst:
In 2022 July AWS [announced](https://aws.amazon.com/about-aws/whats-new/2022/07/aws-identity-access-management-iam-roles-anywhere-workloads-outside-aws/) the option to authenticate external to AWS entities based on X.509 certificates.
Switching away from long lived access keys, should be the trivial way to go, specially for a project which deals with certs already.
More details and a quick demo [here](https://aws.amazon.com/blogs/security/extend-aws-iam-roles-to-workloads-outside-of-aws-with-iam-roles-anywhere/).

neilpang:
Closing as stale: no activity for over 18 months. If this still applies on the current release, please open a new issue with the `--debug 2` log.

PR #7122, opened 2026-07-13
dns_aws: add IAM Roles Anywhere (X.509) authentication

kisst:
## Summary

Adds **AWS IAM Roles Anywhere** (X.509 client-certificate) authentication to the Route53 DNS API (`dns_aws`), as an alternative to long-lived `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`.

Implements the long-standing request in #4303. Tracking/bug issue: #7121.

## Why

A static access key is a bearer secret with no expiry and **requires a dedicated IAM user** (per AWS, "access keys are long-term credentials for an IAM user or the AWS account root user"; roles can't hold permanent keys). Roles Anywhere lets the host prove identity with a certificate signed by a registered trust anchor, assume an IAM **role**, and receive **temporary** credentials — no standing secret and no IAM user on the host. A natural fit for a tool that already manages certificates.

## What changed

- New credential source `_use_roles_anywhere()` in `dnsapi/dns_aws.sh`, alongside the existing container/instance-role sources. It produces the same `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN` and sets `_using_role=true`, so the temporary credentials and the private key are never persisted.
- The `CreateSession` request reuses the file's existing SigV4 scaffolding. Deltas vs. `aws_rest`: `AWS4-X509-{RSA,ECDSA}-SHA256` (auto-selected from the client key type), the `X-Amz-X509` header, StringToSign signed with the cert private key (`openssl dgst -sign` → native PKCS#1 / ASN.1-DER), and the cert **serial in decimal** as the `Credential`.
- The decimal serial needs a bignum hex→dec conversion (serials are up to 160 bits); done in **pure shell** — no `bc`/`awk`/`jq`/`python` — to stay within the project's portability rules.
- `dns_aws_info` metadata extended with the `AWS_RA_*` options.
- Docs: `docs/dnsapi/dns_aws-roles-anywhere.md` (wiki-ready) covering why, obtaining the cert (ACM PCA **and** self-managed CA), storage under `~/.aws/rolesanywhere/`, and rotation.

## Configuration

```sh
export AWS_RA_TRUST_ANCHOR_ARN="arn:aws:rolesanywhere:<region>:<acct>:trust-anchor/<id>"
export AWS_RA_PROFILE_ARN="arn:aws:rolesanywhere:<region>:<acct>:profile/<id>"
export AWS_RA_ROLE_ARN="arn:aws:iam::<acct>:role/<role>"
# optional: AWS_RA_CERT, AWS_RA_KEY, AWS_RA_REGION, AWS_RA_DURATION
acme.sh --issue --dns dns_aws -d example.com -d '*.example.com'
```

## Testing

- Verified end-to-end for **RSA and EC P-256** client certs: correct canonical request, StringToSign, decimal-serial `Credential`, and a valid PKCS#1 / DER signature (EC signature confirmed as ASN.1 `SEQUENCE` of two integers, which is what Roles Anywhere requires).
- Bignum hex→dec checked against known vectors including two real 160-bit serials.
- `shellcheck -e SC2181 -e SC2089` clean; `shfmt -i 2` clean; `sh -n` clean.
- `dns_aws_rm` independently re-runs all credential setup (add/rm run in isolated subshells), consistent with the Dev Guide.

## Points open for maintainer preference

1. **Precedence** — when the `AWS_RA_*` ARNs are set, Roles Anywhere is currently used in preference to explicit `AWS_ACCESS_KEY_ID`/`SECRET` (a deliberate opt-in shouldn't lose to a stale key in `account.conf`). Trivial to reorder to run *after* explicit keys if preferred.
2. **Signing location** — the asymmetric signing is kept local to `dns_aws.sh` rather than extending core `_sign()` (core's `_sign()` re-encodes ECDSA to raw r‖s, which RA rejects), to minimise blast radius. Happy to move it into core if you'd rather.

## Known limitations (documented)

- Signs with SHA-256 (covers RSA and NIST P-256). P-384/P-521 and ML-DSA not yet handled.
- Passphrase-protected keys and `X-Amz-X509-Chain` not yet supported; PKCS#11/TPM users can point `AWS_RA_*` at the official `aws_signing_helper`.
- Under `--debug 2`, a failed request dump includes the `Authorization` header (the request signature) — this is the same pre-existing behaviour as the existing SigV4 path (`aws_rest`), not new to this change.

### DIFF
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -351,14 +351,17 @@ _use_roles_anywhere() {
   _ra_cert="${AWS_RA_CERT:-$HOME/.aws/rolesanywhere/certificate.pem}"
   _ra_key="${AWS_RA_KEY:-$HOME/.aws/rolesanywhere/private-key.pem}"
   if _startswith "$AWS_RA_CERT" "-----BEGIN"; then
-    _ra_cert_tmp="$(_mktemp)"
+    if ! _ra_cert_tmp="$(_aws_ra_write_temp "$AWS_RA_CERT")"; then
+      return 1
+    fi
     _ra_cert="$_ra_cert_tmp"
-    printf "%s" "$AWS_RA_CERT" >"$_ra_cert"
   fi
   if _startswith "$AWS_RA_KEY" "-----BEGIN"; then
-    _ra_key_tmp="$(_mktemp)"
+    if ! _ra_key_tmp="$(_aws_ra_write_temp "$AWS_RA_KEY")"; then
+      [ -n "$_ra_cert_tmp" ] && rm -f "$_ra_cert_tmp"
+      return 1
+    fi
     _ra_key="$_ra_key_tmp"
-    printf "%s" "$AWS_RA_KEY" >"$_ra_key"
   fi
 
   _use_roles_anywhere_impl
@@ -367,6 +370,30 @@ _use_roles_anywhere() {
   [ -n "$_ra_cert_tmp" ] && rm -f "$_ra_cert_tmp"
   [ -n "$_ra_key_tmp" ] && rm -f "$_ra_key_tmp"
   return "$_ra_ret"
+}
+
+# contents
+# Writes PEM contents to a private temp file and echoes its path. The file is created
+# under an owner-only umask and its permissions tightened to 600 before the private key
+# is written, so the key is never exposed to other users -- this also hardens the
+# predictable-name _mktemp fallback, which returns a path without creating the file.
+_aws_ra_write_temp() {
+  _ra_tmp="$(_mktemp)"
+  if [ -z "$_ra_tmp" ]; then
+    _err "Roles Anywhere: unable to create a secure temporary file."
+    return 1
+  fi
+  (
+    umask 077
+    : >"$_ra_tmp" || exit 1
+    chmod 600 "$_ra_tmp" 2>/dev/null
+    printf "%s" "$1" >"$_ra_tmp"
+  ) || {
+    _err "Roles Anywhere: unable to write the temporary certificate/key file."
+    rm -f "$_ra_tmp"
+    return 1
+  }
+  printf "%s" "$_ra_tmp"
 }
 
 # Implementation of the Roles Anywhere exchange. Reads $_ra_cert / $_ra_key (resolved
--- a/dnsapi/dns_aws.sh
+++ b/dnsapi/dns_aws.sh
@@ -6,6 +6,13 @@ Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_aws
 Options:
  AWS_ACCESS_KEY_ID API Key ID
  AWS_SECRET_ACCESS_KEY API Secret
+ AWS_RA_TRUST_ANCHOR_ARN IAM Roles Anywhere trust anchor ARN. Optional, enables X.509 cert auth.
+ AWS_RA_PROFILE_ARN IAM Roles Anywhere profile ARN. Optional.
+ AWS_RA_ROLE_ARN IAM role ARN to assume via Roles Anywhere. Optional.
+ AWS_RA_CERT Path to the Roles Anywhere client certificate PEM. Optional, default "~/.aws/rolesanywhere/certificate.pem".
+ AWS_RA_KEY Path to the Roles Anywhere client private key PEM. Optional, default "~/.aws/rolesanywhere/private-key.pem".
+ AWS_RA_REGION Roles Anywhere region. Optional, default parsed from the trust anchor ARN.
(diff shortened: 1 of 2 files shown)


kisst:
Acknowledged — going through the checklist:

- **Base branch:** this PR targets `dev` (not `master`). ✅
- **DNS API Dev Guide:** read and followed — `dns_aws_add`/`dns_aws_rm` re-run all credential setup independently (isolated subshells), config via `_readaccountconf_mutable`/`_saveaccountconf_mutable`, HTTP via `_post`, headers via `_H1..`, no `awk`/`jq`, `_debug2`/`_secure_debug2` logging, `_err` + `return 1` on failure. `shellcheck -e SC2181 -e SC2089`, `shfmt -i 2`, and `sh -n` are clean.
- **DNS-API-Test:** ready to run. This is an auth-method change to the existing `dns_aws` plugin, so `TEST_DNS=dns_aws` with a Route53 `TestingDomain` exercises it. Because Roles Anywhere needs the client cert **and** key, and the test harness injects secrets as env vars, I added support for `AWS_RA_CERT`/`AWS_RA_KEY` to hold the **PEM contents** directly (not only a file path) — so the test can be driven entirely from secrets:
  - `AWS_RA_TRUST_ANCHOR_ARN`, `AWS_RA_PROFILE_ARN`, `AWS_RA_ROLE_ARN`
  - `AWS_RA_CERT` = certificate PEM contents, `AWS_RA_KEY` = private key PEM contents

  The certificate must be a real end-entity cert from a CA registered as the trust anchor (it can't be self-issued by acme.sh — that would be circular). I've verified the full CreateSession signing end-to-end for both RSA and EC P-256 keys locally.
- **Wiki usage:** documented. `docs/dnsapi/dns_aws-roles-anywhere.md` in this PR is wiki-ready (anchor + subsection heading) for the Route53 entry on the dnsapi page; happy to have it merged onto the wiki. The in-code `dns_aws_info` block is updated with all `AWS_RA_*` options.

Tracking/bug issue: #7121. Original request: #4303.

Let me know your preference on the two open points in the PR description (credential precedence, and keeping the asymmetric signing local vs. in core `_sign()`).

neilpang:
1. Run the [DNS-API-Test](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test) in your fork with your `AWS_RA_*` secrets and post the passing run link — your fork has no DNS.yml runs yet. Required before merge.

2. No direct openssl calls in a dnsapi hook — use core helpers, extending core where needed (this settles your open point 2):
   - key-type detection → existing `_isRSA` / `_isEcc`; serial → existing `_getSerial`.
   - signing → extend core `_sign()` with an optional third arg `outform`; `"der"` returns the native openssl output (RSA PKCS#1 / ECDSA ASN.1 DER) and honors the passed hash alg (the JWS path hardcodes `-sha$__ECC_KEY_LEN`):
     ```sh
     if [ "$outform" = "der" ]; then
       $_sign_openssl -"$alg" | _base64
       return
     fi
     ```
     Hook side: `_sign "$_ra_key" "sha256" der | _dbase64 | _hex_dump | tr -d ' '`
   - cert → base64 DER: add a core helper next to `_getSerial`:
     ```sh
     _getCertDerB64() {
       _cert="$1"
       ${ACME_OPENSSL_BIN:-openssl} x509 -in "$_cert" -outform DER | _base64
     }
     ```
   - `_sign()` is core, so add letest.sh tests in a companion [acmetest](https://github.com/acmesh-official/acmetest) PR: sign fixed data with RSA and EC keys in `der` mode, `openssl dgst -verify` both, assert the EC output starts with `30` (ASN.1 SEQUENCE).

3. Nothing ever saves `AWS_RA_*` (`_using_role=true` skips the save block), so cron renewal has no config and silently falls back to instance-role/static keys. Contradicts your doc. Save the 3 ARNs plus `AWS_RA_CERT`/`AWS_RA_KEY` only when they are file paths — never persist inline PEM contents.

4. Drop `docs/dnsapi/dns_aws-roles-anywhere.md` — usage docs live on the wiki. I'll add your content there when this merges.

5. An encrypted key makes openssl prompt on /dev/tty and an interactive `--issue` hangs. Grep the key for `ENCRYPTED` and `_err` early.

6. 1-2 of 3 ARNs set → RA silently skipped, generic "haven't specified api key" error. Warn on partial config.

7. `echo "$_ra_response"` — dash/busybox `echo` mangles backslashes; use `printf "%s\n"`.

8. Validate `AWS_RA_DURATION` numeric 900-43200 before interpolating into JSON.

Open point 1: keep RA precedence as is. The protocol implementation checks out — verified against AWS docs, the official credential helper, and 160-bit serial test vectors.

PR #7124, opened 2026-07-14
fix(dns_oci): read ~/.oci/config before cached account.conf values

Djelibeybi:
The OCI DNS plugin reads tenancy, user, region and the API signing key in the wrong precedence order: it consults acme.sh's cached `account.conf` values *before* `~/.oci/config`. Because those values are cached at issuance, a value cached then permanently shadows the config file, so editing `~/.oci/config` afterwards has no effect and renewals keep using stale credentials.

This PR makes `~/.oci/config` authoritative over the cache. I will update the wiki page to document this change after the PR is merged.

### DIFF
--- a/dnsapi/dns_oci.sh
+++ b/dnsapi/dns_oci.sh
@@ -115,12 +115,15 @@ _oci_config() {
     _clearaccountconf_mutable OCI_CLI_PROFILE
   fi
 
-  OCI_CLI_TENANCY="${OCI_CLI_TENANCY:-$(_readaccountconf_mutable OCI_CLI_TENANCY)}"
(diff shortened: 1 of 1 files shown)


Djelibeybi:
> Then reply on this message, otherwise, your code will not be reviewed or merged. 

Tests passed: https://github.com/Djelibeybi/acme.sh/actions/runs/29322405090/job/87057031553

neilpang:
please update wiki.

Djelibeybi:
> please update wiki.

Done: https://github.com/acmesh-official/acme.sh/wiki/How-to-use-Oracle-Cloud-Infrastructure-DNS/03a35a96ff6a120b37af3859db5f163362dec057

PR #7127, opened 2026-07-14
feat(dns_oci): support OCI Resource Principal authentication

Djelibeybi:
Allow the OCI DNS API to authenticate with an OCI Resource Principal so acme.sh can solve dns-01 challenges from inside OCI non-instance compute (Container Instances, Functions, OKE Workload Identity, Data Science, etc) without a user API key. When Resource Principal material is not present, the existing user-principal path (~/.oci/config or OCI_CLI_* environment variables) is used unchanged.

Will link to the test results when the job has finished.

### DIFF
--- a/dnsapi/dns_oci.sh
+++ b/dnsapi/dns_oci.sh
@@ -1,8 +1,11 @@
 #!/usr/bin/env sh
 # shellcheck disable=SC2034
 dns_oci_info='Oracle Cloud Infrastructure (OCI)
- If OCI CLI configuration file ~/.oci/config has a DEFAULT profile then it will be used.
-Site: Cloud.Oracle.com
+ Resource principal auth is selected automatically when OCI_RESOURCE_PRINCIPAL_VERSION=2.2.
+ API-key auth from OCI CLI config or OCI_CLI_* values is used otherwise.
+ delegated subzones are supported by selecting the most-specific accessible zone.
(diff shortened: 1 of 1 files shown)


neilpang:
Wait for the DNS test to go green -- the previous run on this branch failed, so I'll hold off until you link a passing one.

Two things meanwhile:

1. `_oci_openssl_sign_with_passphrase` shells out to `openssl dgst -sign -passin file:...` directly. No dnsapi does that -- the two that once did (`dns_transip`, `dns_yc`) left it commented out. Core `_sign` is the only signing path hooks should use; if it needs a passphrase, extend `_sign` with an optional passphrase argument and call that.

2. Please keep the `Author:` line in the info block -- 106 hooks carry it, and moving it into a copyright comment drops it from the plugin metadata.

Also, this bundles a fair amount beyond RP auth: a rewrite of `_oci_config` (just reworked in #7124), JSON escaping, the authz-error heuristics, path normalization, and the stray `))` fix. Splitting the cleanup out from the feature would make both easier to review and to bisect later.

Djelibeybi:
Yeah, I did this first and noticed the other bug later but wanted to get that committed before this larger feature. The failing test is due to test flakiness, not the plugin (a rerun will clear it).

Djelibeybi:
Closing this out, will open two new PRs.

PR #4654, opened 2023-06-05
Deploy opengear

Olen:
Support for Opengear Operations Manager.

OpenGear official documentation 
https://opengear.zendesk.com/hc/en-us/articles/8141687682203-Installing-a-HTTPS-certificate-from-the-CLI

PR #7136, opened 2026-07-17
Add deploy hook for Opengear Operations Manager

Olen:
Deploy certificates to **Opengear Operations Manager** and **Lighthouse** appliances over SSH, using `ogconfig-cli`.

Opengear official documentation: https://opengear.zendesk.com/hc/en-us/articles/8141687682203-Installing-a-HTTPS-certificate-from-the-CLI

This supersedes the previously closed #4654 and reworks the script to follow the [code of conduct](https://github.com/acmesh-official/acme.sh/wiki/Code-of-conduct):

- `#!/usr/bin/env sh` shebang instead of `bash`
- `printf '%b'` instead of the non-portable `echo -e`
- `_base64` (openssl) instead of the GNU-only `base64 -w0`
- Config is loaded with `_getdeployconf` (env overrides saved) and only written back with `_savedeployconf` **after** a successful deploy, so a failed run never persists config and cron renewals keep working
- Falls back to the Lighthouse `ogconfig-cli` path when the standard one is absent

Passes `shellcheck -e SC2181 -e SC2089` and `shfmt -i 2`.

### Usage

```sh
export OPENGEAR_USER="root"    # required
export OPENGEAR_HOST="om1234"  # optional, defaults to the domain name
acme.sh --deploy -d example.com --deploy-hook opengear
```

SSH key auth and password-less `sudo` must be set up on the appliance first. A usage entry for the deployhooks wiki page is included above and can be added on merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/deploy/opengear.sh
+++ b/deploy/opengear.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env sh
+
+# Script to deploy a certificate to an Opengear Operations Manager (and
+# Lighthouse) appliance over SSH.
+#
+# See the Opengear documentation:
+# https://opengear.zendesk.com/hc/en-us/articles/8141687682203-Installing-a-HTTPS-certificate-from-the-CLI
+#
+# Returns 0 when successful, otherwise an error.
+#
+# SSH must be able to log in to the remote host without a password (exchange
+# SSH keys first) and the user must have password-less sudo access. Validate
+# that you can log in to USER@HOST from the host running acme.sh before using
+# this script.
+#
+# The following environment variables are used on the first run and then
+# saved (encrypted) to the deploy configuration:
+#
+#   export OPENGEAR_USER="root"    # required
+#   export OPENGEAR_HOST="om1234"  # optional, defaults to the domain name
+
+Le_Deploy_ssh_cmd="ssh"
+
+#domain keyfile certfile cafile fullchain
+opengear_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  _debug _cdomain "$_cdomain"
+  _debug _ckey "$_ckey"
+  _debug _ccert "$_ccert"
+  _debug _cca "$_cca"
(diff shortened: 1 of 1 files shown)


neilpang:
1. The remote command is broken as written: the single quotes in `printf '%b\n'` collide with the `sh -c "'$_cmd'"` wrapper in `_ssh_remote_cmd`. After ssh arg-joining and the remote shell's re-parsing, the format string degrades to `%bn`, so ogconfig-cli receives the last line as `pushn` with no trailing newline -- the certificate is set in the config tree but never pushed. Reproduced locally with a fake-ssh harness:
   ```
   set services.https.certificate =CERTB64AAAA
   set services.https.private_key =KEYB64BBBB
   pushn
   ```
   #4654's `echo -e` did not have this problem; it was introduced by this rewrite, so this version cannot have been tested on a real appliance. Simplest robust fix: drop the nested quoting entirely and feed the config over stdin, selecting the binary on the remote side:
   ```sh
   printf '%s\n' "set services.https.certificate =$_og_cert" \
     "set services.https.private_key =$_og_key" "push" \
     | ssh "$OPENGEAR_USER@$OPENGEAR_HOST" \
       "OG=/usr/bin/ogconfig-cli; [ -x \$OG ] || OG=/usr/unsupported/bin/ogconfig-cli; sudo \$OG"
   ```
2. `sudo printf ... | /usr/bin/ogconfig-cli` privileges the wrong side of the pipe: ogconfig-cli runs unprivileged. If it needs root (why else require password-less sudo), sudo must wrap ogconfig-cli. Carried over from #4654.
3. Top-level `Le_Deploy_ssh_cmd="ssh"` executes at source time and collides with deploy/ssh.sh's legacy config variable of the same name (ssh.sh migrates it to DEPLOY_SSH_CMD). Use an `_opengear_`-prefixed name; same for `_ssh_remote_cmd`.
4. Please confirm a successful deploy on a real Operations Manager / Lighthouse after the fix -- deploy hooks have no CI coverage.

PR #7140, opened 2026-07-19
feat: add dnsapi for HestiaCP

radumalica:
## Summary

Adds a DNS API hook for [HestiaCP](https://hestiacp.com) (`dns_hestiacp`), using the HestiaCP JSON API (`v-list-dns-records`, `v-add-dns-record`, `v-delete-dns-record`).

Supersedes the earlier draft reviewed in this repo; all review comments have been addressed:

- `_post` is called with `application/json` as the Content-Type argument (the stray `--connect-timeout 10` argument is gone)
- the record name is derived from `fulldomain` relative to the detected zone, so DNS alias mode works; nothing is hardcoded to `_acme-challenge`
- the root zone is detected by walking the domain labels and probing `v-list-dns-records`, handling `_acme-challenge.www.example.co.uk`-style names
- `rm` deletes only TXT records matching both the challenge record name and `txtvalue`
- all working variables are prefixed with `_hestia_`
- the info block uses the standard single-line machine-parsed fields with `Docs:` pointing to `dnsapi2#dns_hestiacp`

Wildcard certs are supported: add appends a second TXT record with the same name (HestiaCP permits duplicate record names), and records are added/removed with `restart=yes` so bind serves them immediately.

Report bugs to: https://github.com/acmesh-official/acme.sh/issues/6251

## Test plan

- [ ] DNS-API-Test workflow run on the fork per https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test (secrets: `TEST_DNS=dns_hestiacp`, `TestingDomain`, `HESTIA_HOST`/`HESTIA_ACCESS`/`HESTIA_SECRET`/`HESTIA_USER` via TokenName1-4, `TEST_DNS_SLEEP=120`)
- [ ] dnsapi2 wiki entry added with `<a name="dns_hestiacp"/>` anchor and API key setup instructions

### DIFF
--- a/dnsapi/dns_hestiacp.sh
+++ b/dnsapi/dns_hestiacp.sh
@@ -11,7 +11,6 @@ Options:
 Issues: github.com/acmesh-official/acme.sh/issues/6251
 Author: Radu Malica <radu.malica@gmail.com>
 '
-# Copyright (c) 2026 Radu Malica
 
 ########  Public functions #####################
 
--- a/dnsapi/dns_hestiacp.sh
+++ b/dnsapi/dns_hestiacp.sh
@@ -182,7 +182,7 @@ _hestia_find_records() {
   _hestia_ftype=$2
 
   echo "$_hestia_response" | tr -d '\n' | sed 's/},/}\
-/g' | grep -- "\"RECORD\": \"$_hestia_fname\"" | grep -- "\"TYPE\": \"$_hestia_ftype\"" | while read -r _hestia_line; do
+/g' | grep -F -- "\"RECORD\": \"$_hestia_fname\"" | grep -F -- "\"TYPE\": \"$_hestia_ftype\"" | while read -r _hestia_line; do
     _hestia_id=$(echo "$_hestia_line" | _egrep_o '"ID": "[^"]*' | cut -d '"' -f 4)
     _hestia_value=$(echo "$_hestia_line" | _egrep_o '"VALUE": "[^"]*' | cut -d '"' -f 4)
     if [ -n "$_hestia_id" ]; then
--- a/dnsapi/dns_hestiacp.sh
+++ b/dnsapi/dns_hestiacp.sh
@@ -1,0 +1,192 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_hestiacp_info='HestiaCP Server API
+Site: hestiacp.com
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_hestiacp
+Options:
+ HESTIA_HOST Panel URL. E.g. "https://panel.example.com:8083"
+ HESTIA_ACCESS API access key
+ HESTIA_SECRET API secret key
+ HESTIA_USER Username owning the DNS zones. Default "admin". Optional.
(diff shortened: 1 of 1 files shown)


neilpang:
1. DNS-API-Test on your fork has not passed yet ([latest run](https://github.com/radumalica/acme.sh/actions/runs/29694273814)). Two separate problems in the log:
   - `Error: ID=40 already exists` on add: a leftover TXT record from an earlier failed run. Clean up all `_acme-challenge` TXT records in the test zone before re-running.
   - Wildcard validation failed with 3 TXT values served, and your secondary NS 193.37.138.2 returns SERVFAIL for the name, which breaks Let's Encrypt secondary validation. Both nameservers must serve the test zone consistently.

   A green DNS run on the fork is required for merge.
2. Nit: in `_hestia_find_records`, use `grep -F` for the RECORD/TYPE matches -- the record name contains dots, which are regex wildcards.

radumalica:
Fixed second issue and also tests are green now. It was a problem on my end with a test domain that had 2 different name servers and AXFR between them was broken.

neilpang:
Not green yet -- the [latest run](https://github.com/radumalica/acme.sh/actions/runs/29738767702) failed at NetBSD: after the TXT records were added, both nameservers answered SERVFAIL for `_acme-challenge.<domain>` for the whole 5-minute propagation check (`EDE(22): No Reachable Authority at delegation`), and the jobs after NetBSD (DragonFlyBSD through Haiku) never ran. Same symptom as the previous run, minutes after the earlier platforms passed the same flow, so the nameserver setup is still intermittently broken. Please re-run until the whole workflow is green.

radumalica:
everything is fixed from my side, but in the process i've found a bug in hestiaCP documented here: https://github.com/hestiacp/hestiacp/issues/5533 . Actually, repeated dns requests done from the CI/CD pipeline triggers serial updating up to more than 99 digits, HestiaCP doesn't reset the serial back to 00 but instead it goes to 3 digits (100,101,etc) which breaks the zone and causing the DNS servers to report SERVFAIL.

neilpang:
Good find. The [run for the current head](https://github.com/radumalica/acme.sh/actions/runs/29870374486) is still red though (Solaris, same 5-minute SERVFAIL; it ran before your fix). One caveat before re-running: a full workflow run is 15 jobs (Docker alone tests ~10 images), each doing several add/remove cycles with `restart=yes`, so one run alone generates 100+ zone updates within a single UTC day -- enough to trip the serial bug again near the tail of the chain. That matches both previous full runs breaking at NetBSD/Solaris. Patch your panel for hestiacp/hestiacp#5533 first, then trigger a full run on the current head; merge needs the whole workflow green on it.

radumalica:
Yeah, probably i will patch my installed version on my hosting where the tests are running against, and run the pipeline again with a test commit of some sort (comment or similar). Then will check to see if it's working so i don't block this PR until HestiaCP fixes their software. 

Will do that today or tomorrow tops whenever i have some time, and report here.

radumalica:
re-ran last failed [job ](https://github.com/radumalica/acme.sh/actions/runs/29870374486 )and all green now, after i applied my suggested fix on my own Hestia server:

PR #7145, opened 2026-07-22
Feat: Shelly deploy hook for firmware 2.0.0+

Qhilm:
## feat: add Shelly Gen3+ deploy hook

   Adds a deploy hook for [Shelly smart home devices](https://www.shelly.com) (Gen3+, tested on Gen4 with
 firmware 2.0.0) that installs Let's Encrypt HTTPS server certificates via the device's JSON-RPC API over HTTP.

According to [Shelly release notes](https://shelly-api-docs.shelly.cloud/gen2/changelog/), Firmware 2.0.0 minimum is required.

   ### Usage

   ```sh
   export SHELLY_HOST=192.168.1.100
   export SHELLY_PASSWORD=mysecret    # only if auth is enabled on the device
   acme.sh --deploy -d shelly.example.com --deploy-hook shelly
   ```

   ### Configuration

   | Variable | Required | Default | Description |
   |----------|----------|---------|-------------|
   | `SHELLY_HOST` | ✅ | — | IP or hostname of the Shelly device |
   | `SHELLY_PASSWORD` | ❌ | — | Admin password (omit if device auth is disabled) |
   | `SHELLY_USER` | ❌ | `admin` | Username for digest auth |
   | `SHELLY_REBOOT` | ❌ | `1` | Set to `0` to skip auto-reboot |

   ### How it works

   1. **Auth handshake** — probes the device with a protected RPC method (`Shelly.GetStatus`). If the device
 returns HTTP 401, it extracts the `realm`, `nonce`, and `qop` from the `WWW-Authenticate` header and computes an
 RFC 7616 SHA-256 HTTP Digest Authorization header. If the device responds without challenge, it proceeds without
 auth.

   2. **Upload** — clears any existing certificate and key, then uploads the fullchain and private key via
 `Shelly.PutHTTPServerCert` and `Shelly.PutHTTPServerKey`. PEM data is JSON-escaped using acme.sh's built-in
 `_json_encode`.

   3. **Reboot** — calls `Shelly.Reboot` to apply the new certificate. Can be disabled with `SHELLY_REBOOT=0`.

   ### Authentication details

   Firmware 2.0.0+ uses **standard RFC 7616 HTTP Digest** — the `Authorization` header, not the legacy JSON-RPC
 body `auth` object. The nonce counter (`nc`) is incremented for each request as required by the spec.

   ### Testing

   - **Self-test**: `deploy/test_shelly.sh` validates JSON encoding, RFC 7616 hash computation, Authorization
 header format, and the full deploy flow with mocked HTTP.
   - **Real device**: Tested against Shelly 2PM Gen4 (S4SW-002P16EU, firmware 2.0.0) with digest auth enabled —
 all 4 RPC calls (clear cert, upload cert, clear key, upload key) returned HTTP 200 with valid responses.

   ### Files

   | File | Lines | Purpose |
   |------|-------|---------|
   | `deploy/shelly.sh` | ~265 | Deploy hook |
   | `deploy/test_shelly.sh` | ~280 | Self-test suite |


neilpang:
1. `_secure_debug "Shelly ha1: $_shelly_ha1"` — `_secure_debug` takes label and value as two arguments and only masks the second. With a single argument the whole string, including the secret, is printed at `--debug 1`. ha1 is password-equivalent for this realm. Use `_secure_debug "Shelly ha1" "$_shelly_ha1"`; same for the Authorization header and nonce lines.
2. `$RANDOM` is bash-only. Hooks are sourced and run in whatever shell runs acme.sh (dash, busybox ash, BSD sh), where `$RANDOM` is empty, making the cnonce a constant string of zeros. `[ "${ACME_OPENSSL_BIN:-openssl}" ]` is also always true. Since `_digest` already requires openssl, drop the fallback branch entirely.
3. The raw curl re-request is unnecessary: `_ACME_CURL` already includes `--dump-header $HTTP_HEADER`, so after the first probe `_post` the 401 `WWW-Authenticate` header is already in `$HTTP_HEADER` for both curl and wget. Parse `$HTTP_HEADER` and drop the direct curl invocation (hooks must not call curl directly). You can also detect auth from the 401 status line there instead of inferring it from an empty body.
4. The reboot bypasses `_shelly_rpc` and reuses the previous request's `_H1`, whose `nc` was already consumed; a server enforcing RFC 7616 replay protection rejects it and `|| true` hides the failure, so with auth enabled the cert is never applied. Rebuild the header before the reboot request, or send it via `_shelly_rpc` and tolerate the dropped connection.
5. `export HTTPS_INSECURE=1` has no effect here (every URL is `http://`) but stays exported and disables certificate verification for all later requests in the same run (other deploy/notify hooks). Remove it. Also clear `_H1` before returning so the Shelly Authorization header is not sent to unrelated endpoints.
6. Remove `deploy/test_shelly.sh`: anything under deploy/ resolves as a hook name (`--deploy-hook test_shelly`), and the repo does not carry per-hook test scripts (it is also bash + python3). Keep it in your fork.
7. Both files are missing the trailing newline; shfmt CI will fail.
8. Prefix the helper variables with `_shelly_` (`_body`, `_ret`, `_method`, `_params`, `_ha2`, `_digest_response`, `_cert_data`, `_key_data`, `_headers_file`): hooks share one shell with core, and `_post` itself uses `_ret`.
9. Please add a usage entry for the hook to the deployhooks wiki page and link it from the PR.

Qhilm:
1. _secure_debug — split into label + value: _secure_debug "Shelly ha1" "$_shelly_ha1", same for nonce and the
    Authorization header.

 2. $RANDOM fallback — removed. _digest already requires openssl, so the $RANDOM branch was both bash-only dead
    code and a constant-zero cnonce in dash/ash. OpenSSL rand -hex unconditionally now.

 3. Raw curl re-request — dropped entirely. _post already dumps headers into $HTTP_HEADER for both curl and wget,
    so the handshake now parses $HTTP_HEADER directly. Auth detection also switched from "empty body" heuristic
    to checking the HTTP 401 status line via a new _shelly_has_auth_challenge helper.

 4. Reboot nc replay — reboot now goes through _shelly_rpc instead of raw _post, so the auth header gets rebuilt
    with the correct nonce counter before the call. The || true is replaced with || _debug "..." to tolerate the
    dropped connection without silently swallowing other errors.

 5. HTTPS_INSECURE leak — removed the export (every URL is http://, so it was a no-op on the Shelly side but
    disabled cert verification for all subsequent hooks). Added export _H1="" before return 0 so the Shelly
    Authorization header doesn't leak to unrelated endpoints.

 6. deploy/test_shelly.sh — deleted from the branch. Kept in the fork.

 7. Trailing newline — fixed.

 8. Unprefixed helper variables — _body, _ret, _method, _params, _ha2, _digest_response, _cert_data, _key_data,
    _headers_file all renamed to _shelly_* variants. _ret in particular now uses _shelly_ret to avoid colliding
    with _post.

 9. Wiki entry — added to the deployhooks page.

Qhilm:
don't merge yet, found an issue, working on it.

Qhilm:
Added fixes for:

### Bug 1: Non-JSON responses passed as success

 _shelly_rpc considered a response successful as long as it was non-empty and didn't contain the string "error". The Shelly's
 rate limiter returns plain-text Too Many Requests, which sailed through both checks. Now validates the response starts with {
  and contains "id" before treating it as a valid JSON-RPC response.

 ### Bug 2: Clear-before-upload breaks HTTPS

 The original flow cleared each file before uploading:

 ```
   clear cert → upload cert → clear key → upload key → reboot
 ```

 But the Shelly docs state: "When you remove the certificate or key, the device automatically removes all three files (cert,
 key, CA bundle)." So the key-clear step wipes the cert we just uploaded, leaving only the key at boot time. The device's
 mbedtls_pk_check_pair then silently skips the HTTPS listener — no error, just no port 443.

 Fix: removed the clear steps entirely. Uploading overwrites in place, so clearing is unnecessary and actively harmful. New
 flow:

 ```
   upload cert → upload key → reboot → HTTPS up on port 443 ✅
 ```

 The restart_required field in the upload responses correctly transitions from false (cert-only, no matching key yet) to true
 (pair complete), and the device serves the certificate after reboot. Tested end-to-end with both self-signed and LE
 certificates.

neilpang:
CI is red on the latest push:
1. shfmt: redirections must have no space — `_json_encode <"$_cfullchain"` / `_json_encode <"$_ckey"`. Run `shfmt -w -i 2 deploy/shelly.sh`.
2. ShellCheck SC2090 on the `export _H1="Authorization: $_shelly_auth_header"` line. It is a false positive here (`_post` expands `-H "$_H1"` quoted), so add `# shellcheck disable=SC2090` above that line; SC2089 is already excluded globally.
3. Nit: use the core wrapper `_tail_n 1` instead of `tail -1` in `_shelly_has_auth_challenge` — core itself parses the status line as `grep "^HTTP" "$HTTP_HEADER" | _tail_n 1 | cut -d " " -f 2`.

neilpang:
SC2090 now fires on the `else` branch (`export _H1=""`) — shellcheck tracks `_H1` across both paths. Move the single `# shellcheck disable=SC2090` to just above `if [ -n "$_shelly_auth_header" ]; then`: a directive placed before the `if` covers the whole block. Verified locally that `shellcheck -e SC2181 -e SC2089` passes with that change.

PR #7150, opened 2026-07-23
dns_yc: fix TXT record removal failing with "Unknown key file format"

gasRU76:
## Problem

When issuing/renewing a certificate with `--dns dns_yc` (Yandex Cloud DNS), the TXT record cleanup step fails:

```
Unknown key file format.
invalid domain
Error removing txt for domain: _acme-challenge.example.com
```

The TXT record is never removed.

## Root cause

`dns_yc_add()` rebuilds `YC_SA_Key_File` from `YC_SA_Key_File_PEM_b64` / `YC_SA_Key_File_Path` before doing anything else:

https://github.com/acmesh-official/acme.sh/blob/dev/dnsapi/dns_yc.sh#L25-L35

`dns_yc_rm()` never does this — it goes straight to `_get_root` → `_yc_rest` → `_yc_login` without `$YC_SA_Key_File` ever being set.

Per the [DNS API Dev Guide](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide):

> Rm() ... repeat all the preparation stages of the add() function, since they are executed in different sub-shells and variables are not passed.

That's exactly what bites here: when `_yc_login()` needs to sign a fresh JWT during the removal phase (the IAM token obtained during `add()` isn't available), it calls `openssl dgst -sign` with an empty/unset key path, which fails with `Unknown key file format`. The resulting auth failure then causes the next authenticated call inside `_get_root` to fail, which surfaces as the misleading `invalid domain` message, and `dns_yc_rm` aborts.

This is unrelated to #6862 (which was about FQDN handling in the `name` field and has already been resolved independently — `dev` already appends the trailing dot identically in both `add` and `rm`).

## Fix

Add the same key-restoration block used in `dns_yc_add()` to `dns_yc_rm()`, before it's needed by `_get_root`/`_yc_login`. No `_savedomainconf` calls are needed here since `add()` already persists these values.

## Testing

Verified against a real Yandex Cloud service account, folder and DNS zone using `--staging`:

- **Before the fix**: `--issue --dns dns_yc` succeeds at adding the TXT record, but the removal step reproducibly fails with `Unknown key file format` / `invalid domain`, matching the reported bug.
- **After the fix**: full `--issue --dns dns_yc --staging` run succeeds end-to-end — TXT record is added, a fresh login happens during removal (confirming add/rm don't share the IAM token across the subshell boundary), and the TXT record is actually deleted (`Delete, OK`), followed by `Cert success.`
- `shellcheck dnsapi/dns_yc.sh` shows no new warnings (the sole existing SC2181 warning predates this change).

### DIFF
--- a/dnsapi/dns_yc.sh
+++ b/dnsapi/dns_yc.sh
@@ -22,7 +22,12 @@ dns_yc_add() {
   fulldomain="$(echo "$1". | _lower_case)" # Add dot at end of domain name
   txtvalue=$2
 
+  # YC_SA_Key_File_PEM_b64/Path are always persisted to the domain conf below,
+  # so they must be recovered from there first (account conf is only a
+  # fallback for the YC_Folder_ID case, see the SA_ID/SA_Key_ID save below).
+  YC_SA_Key_File_PEM_b64="${YC_SA_Key_File_PEM_b64:-$(_readdomainconf YC_SA_Key_File_PEM_b64)}"
   YC_SA_Key_File_PEM_b64="${YC_SA_Key_File_PEM_b64:-$(_readaccountconf_mutable YC_SA_Key_File_PEM_b64)}"
+  YC_SA_Key_File_Path="${YC_SA_Key_File_Path:-$(_readdomainconf YC_SA_Key_File_Path)}"
   YC_SA_Key_File_Path="${YC_SA_Key_File_Path:-$(_readaccountconf_mutable YC_SA_Key_File_Path)}"
 
   if [ "$YC_SA_Key_File_PEM_b64" ]; then
@@ -34,9 +39,13 @@ dns_yc_add() {
     _savedomainconf YC_SA_Key_File_Path "$YC_SA_Key_File_Path"
   fi
 
+  YC_Zone_ID="${YC_Zone_ID:-$(_readdomainconf YC_Zone_ID)}"
   YC_Zone_ID="${YC_Zone_ID:-$(_readaccountconf_mutable YC_Zone_ID)}"
+  YC_Folder_ID="${YC_Folder_ID:-$(_readdomainconf YC_Folder_ID)}"
   YC_Folder_ID="${YC_Folder_ID:-$(_readaccountconf_mutable YC_Folder_ID)}"
+  YC_SA_ID="${YC_SA_ID:-$(_readdomainconf YC_SA_ID)}"
   YC_SA_ID="${YC_SA_ID:-$(_readaccountconf_mutable YC_SA_ID)}"
+  YC_SA_Key_ID="${YC_SA_Key_ID:-$(_readdomainconf YC_SA_Key_ID)}"
   YC_SA_Key_ID="${YC_SA_Key_ID:-$(_readaccountconf_mutable YC_SA_Key_ID)}"
 
   if [ "$YC_SA_ID" ] && [ "$YC_SA_Key_ID" ] && [ "$YC_SA_Key_File" ]; then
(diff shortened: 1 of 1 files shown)


neilpang:
Please run the [DNS-API-Test](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test) in your own fork with your Yandex Cloud secrets and post a link to the passing run. Your fork currently has no Actions runs. A manual test log does not substitute.

neilpang:
The Omnios job hung and timed out, so Tribblix/OpenIndiana/Haiku/Hurd were skipped. Please re-run the DNS workflow in your fork until all jobs pass, then post the link to the fully green run.

gasRU76:
@neilpang I already restarted this job four times, but it keeps getting stuck for an hour every time. Do you have any suggestions on how to get it moving?

neilpang:
Root cause of the Omnios hang found: OmniOS native `egrep -o` infinite-loops on empty-matchable patterns, and dns_yc's `_egrep_o "[^:]*$"` (used in `_yc_login` token extraction) triggers it -- unrelated to your change. Fixed in dev (830782fd). Please rebase your branch onto dev and re-run the DNS workflow; it should now pass all jobs.

gasRU76:
@neilpang 
DNS-API-Test passed on the fork, all jobs green including Omnios: https://github.com/gasRU76/acme.sh/actions/runs/30146620272

neilpang:
`grep -Fxv` breaks on Solaris: /usr/bin/grep supports neither `-F` nor `-x` (verified on 11.4: "illegal option -- F/x"), so `_remaining_txtvalue` is always empty there and the preserve logic silently falls back to whole-rrset deletion, with a grep usage error on stderr on every rm. `tr '\n' ',' | sed 's/,$//'` has a second issue: its input has no trailing newline and Solaris sed drops an unterminated last line. The DNS test stays green either way because the fallback also returns `done: true`, so CI cannot catch this.

Portable replacement without grep/sed:

```sh
_remaining_txtvalue=""
for _v in $(echo "$exists_txtvalue" | tr -d '[] ' | tr ',' ' '); do
  if [ "$_v" != "\"$txtvalue\"" ]; then
    _remaining_txtvalue="$_remaining_txtvalue$_v,"
  fi
done
_remaining_txtvalue="${_remaining_txtvalue%,}"
```

gasRU76:
@neilpang Thanks for your code — I think everything works now.

Re-ran the DNS workflow on the fork, all jobs pass green (Solaris included): https://github.com/gasRU76/acme.sh/actions/runs/30152552065

neilpang:
`updateRecordSets` has no `merges` field -- only `deletions` and `additions` ([REST](https://yandex.cloud/en/docs/dns/api-ref/DnsZone/updateRecordSets), [gRPC](https://yandex.cloud/en/docs/dns/api-ref/grpc/DnsZone/updateRecordSets)). The preserve branch therefore never removes anything and the TXT record stays in the zone. That is a regression: current dev deletes the whole rrset, which does clean both values up.

CI can't catch it -- `_clearupdns` runs `dns_yc_rm` in a subshell and ignores its exit code (acme.sh:3730), so even a rejected request leaves the run green.

`merges` is valid on `upsertRecordSets`, and that method's `deletions` does exactly what you want ("Delete only specified records from corresponding record sets"), which removes the need for the getRecordSet read-modify-write entirely:

```sh
_yc_rest POST "zones/$_domain_id:upsertRecordSets" "{\"deletions\": [ { \"name\":\"$_sub_domain\",\"type\":\"TXT\",\"ttl\":\"120\",\"data\":[\"$txtvalue\"]}]}"
```

Otherwise keep a single `updateRecordSets` call with `deletions` (old rrset) + `additions` (remaining).

Verify against a real zone with `-d example.com -d *.example.com` and confirm the records are actually gone afterwards -- the DNS test does not prove removal.

Separately: the `_readdomainconf` commit is mostly redundant -- `renew()` sources `$DOMAIN_CONF` (acme.sh:6127), so those vars are already set on renewal.

gasRU76 (commented):
Thanks again for catching this.
I fixed the code as you suggested and tested it against a real domain using:

```sh
./acme.sh --issue --staging --dns dns_yc -d 'test2.vlprojects.tech' -d '*.test2.vlprojects.tech'
```

The TXT DNS record was successfully removed after the certificate was issued.

CI re-run in progress on the fork: https://github.com/gasRU76/acme.sh/actions/runs/30196947585

neilpang:
Fix looks right, and the live test plus the fully green run cover it -- for `YC_SA_Key_File_PEM_b64` mode.

`YC_SA_Key_File_Path` mode is still broken, by a pre-existing bug right next to your change: `_yc_login` unconditionally deletes `$YC_SA_Key_File` (line 278). In Path mode that variable is the user's own key file (lines 38 / 141), so the first `--issue` permanently deletes it. Verified against your branch with stubbed helpers:

```
before login: -rw-r--r-- 1 ... user-sa-key.json
after  login: No such file or directory
```

Consequences: the key is gone, `dns_yc_rm` then hits exactly the "Unknown key file format" this PR is named after, and every later renewal fails the same way. CI never sees it because the test supplies the key as `YC_SA_Key_File_PEM_b64`.

Only the decoded temp file should be deleted. In both `dns_yc_add()` and `dns_yc_rm()`:

```sh
  if [ "$YC_SA_Key_File_PEM_b64" ]; then
    echo "$YC_SA_Key_File_PEM_b64" | _dbase64 >private.key
    YC_SA_Key_File="private.key"
    _yc_key_is_temp=1
  else
    YC_SA_Key_File="$YC_SA_Key_File_Path"
    _yc_key_is_temp=""
  fi
```

and in `_yc_login`:

```sh
  if [ "$_yc_key_is_temp" ]; then
    rm -f "$YC_SA_Key_File"
  fi
```

Pre-existing, not caused by your change -- but it is the same failure this PR is titled after, so better fixed here than left behind.

gasRU76:
I fixed the code as you suggested too and tested it against a real domain using YC_SA_Key_File_Path=./cert.crt

`./acme.sh --issue --staging --dns dns_yc -d 'test3.vlprojects.tech' -d '*.test3.vlprojects.tech'`

The certificate was issued without any problems, and the DNS record was removed without any errors.

gasRU76:
While comparing this file against other RSA-signing dnsapi providers (dns_oci.sh, dns_transip.sh), found and fixed one more issue in the same family as the domain-vs-account conf mismatch:

The failure branch in `dns_yc_add()` (when `YC_SA_ID`/`YC_SA_Key_ID`/`YC_SA_Key_File` aren't all present) only called `_clearaccountconf`. But `YC_Zone_ID`, `YC_Folder_ID`, and the key material are persisted via `_savedomainconf` (domain conf), and `YC_SA_ID`/`YC_SA_Key_ID` may have been saved via `_saveaccountconf_mutable` in Folder_ID mode (which stores under a `SAVED_` prefix, only cleared by `_clearaccountconf_mutable`). So on invalid config, stale values were left behind in whichever store the plain `_clearaccountconf` call didn't actually touch.

Fixed to clear both domain conf and the mutable account conf variant for each key. Verified by seeding both `domain.conf` and `account.conf` with leftover values, triggering this branch, and confirming both files end up empty afterward.

PR #7160, opened 2026-07-24
ARI: pick renewal offset with openssl CSPRNG

frank499:
The ARI suggestedWindow offset was computed as `now % window`. Use openssl CSPRNG to distribute evenly across the window, and that same-batch certs no longer cluster.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -1915,6 +1915,18 @@ _time() {
   date -u "+%s"
 }
 
+# _rand_below <max>
+_rand_below() {
+  _rb_max="$1"
+  _rb_rand="$(${ACME_OPENSSL_BIN:-openssl} rand -hex 4 2>/dev/null)"
+  if [ "$_rb_rand" ]; then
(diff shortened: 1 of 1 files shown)


neilpang:
RFC 9773 sec 4.2: "Select a uniform random time within the suggested window". `now % window` isn't random -- `$(_time)` is constant within a run, so certs sharing a window all get the same offset. Good change; two fixes first:

**1.** Extract a named helper (e.g. `_rand_below <max>`) used by both sites, and add `letest.sh` tests in acmetest -- core function changes need them, and an inline expression can't be tested.

**2.** Fall back to the old expression when `openssl rand` returns empty. `$((0x & 0x7FFFFFFF))` is a syntax error (dash: `expecting ')'`), so `_ari_offset` and then `Le_NextRenewTime` end up empty; `renew()` saves that empty value, the guard stops firing, and the cert renews on every cron run.

Hex in `$(( ))` is portable (verified on dash/bash), but wait for the full matrix.

PR #7164, opened 2026-07-26
deploy/keyhelp_api: allow custom cert name, skip key on PUT, URL-encode name

JensSpanier:
## Summary
- Add `DEPLOY_KEYHELP_NAME` to optionally override the certificate name sent to KeyHelp (falls back to `$_cdomain` when unset).
- Stop sending `private_key` in the PUT (update) request body. KeyHelp's update endpoint doesn't need the key, and for certificates issued with `--signcsr` there is no private key available at all (acme.sh never sees it), so requiring it here would break that workflow. The key is now only read and included when falling back to the POST (create) request for a new entry.
- URL-encode the certificate name when building the PUT request URL, since names may contain characters that are not valid unencoded in a URL path segment.

### DIFF
--- a/deploy/keyhelp_api.sh
+++ b/deploy/keyhelp_api.sh
@@ -30,9 +30,9 @@ keyhelp_api_deploy() {
   fi
 
   if [ -z "$DEPLOY_KEYHELP_NAME" ]; then
-    _name="$_cdomain"
+    _keyhelp_name="$_cdomain"
   else
-    _name="$DEPLOY_KEYHELP_NAME"
+    _keyhelp_name="$DEPLOY_KEYHELP_NAME"
   fi
 
   # Save current values
@@ -43,8 +43,8 @@ keyhelp_api_deploy() {
   _request_cert="$(tr '\n' ':' <"$_ccert" | sed 's/:/\\n/g')"
   _request_ca="$(tr '\n' ':' <"$_cca" | sed 's/:/\\n/g')"
 
-  _put_body="{
-    \"name\": \"$_name\",
+  _keyhelp_put_body="{
+    \"name\": \"$_keyhelp_name\",
     \"components\": {
       \"certificate\": \"$_request_cert\",
       \"ca_certificate\": \"$_request_ca\"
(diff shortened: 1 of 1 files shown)


neilpang:
1. Dropping `private_key` from the PUT breaks key rotation. acme.sh regenerates the domain key on renewal when `--always-force-new-domain-key` is set, and also when `--force` is combined with a changed `--keylength`. KeyHelp then keeps the old key and receives the new cert -- mismatch, TLS breaks. This doesn't have to be either/or: include `private_key` when the key file is present and non-empty, omit it only when it isn't (the `--signcsr` case). One body builder covers both cases and drops the duplicated JSON.

2. `[ ! -f "$_ckey" ]` should be `[ ! -s "$_ckey" ]`. acme.sh can leave a zero-byte key file -- `createDomainKey` itself tests with `-s`, and `Le_Next_Domain_Key` is emptied via `echo "" >` rather than removed. With `-f` alone the POST sends `"private_key": ""`.

3. Does the KeyHelp v2 update endpoint actually accept `components` without `private_key`? Please link the API docs or paste a run log. If it's required server-side, every renewal fails.

4. `_savedeployconf DEPLOY_KEYHELP_NAME` is wrapped in `[ -n ... ]` while the two lines above it are unconditional. It also makes the value unclearable once saved, since `_getdeployconf` falls back to the saved value whenever the env var is empty. Save it unconditionally.

5. Prefix the new variables with the provider name (`_keyhelp_name` etc.). No actual collision today -- deploy hooks are sourced in a subshell -- just consistency with the rest of the tree.

6. `DEPLOY_KEYHELP_NAME` needs an entry on the deployhooks wiki page.

`_url_encode` usage and the 404 -> POST fallback are correct.

JensSpanier:
Thanks for your review @neilpang!

1. KeyHelp doesn't allow to change the private key once the certificate is created. It's neither possible via UI nor API. Sending `private_key` on PUT gets ignored. https://app.swaggerhub.com/apis-docs/keyhelp/api/2.14
2. fixed
3. Yes, see 1.
4. fixed
5. Done. New variables are now `_keyhelp_name`, `_keyhelp_put_body`, `_keyhelp_put_url`, `_keyhelp_post_body`, `_keyhelp_post_url`, `_keyhelp_name_encoded`
6. Done, see #7165

neilpang:
- `DEPLOY_KEYHELP_NAME` is interpolated raw into the JSON body; a `"` or `\` in the name breaks it.
- `_keyhelp_name_encoded` is recomputed on every loop iteration; hoist it above the `for`.
- Follow-up, not this PR: with `--always-force-new-domain-key`, or `--force` plus a changed `--keylength`, acme.sh rotates the domain key and KeyHelp keeps the old one against the new cert. PUT can't fix that -- it needs delete + recreate, or at least an `_info` warning.

JensSpanier:
Thanks for the review @neilpang! The first two points are addressed:

- `DEPLOY_KEYHELP_NAME` is now JSON-escaped (`\` and `"`) before being interpolated into the PUT/POST bodies, via a dedicated `_keyhelp_name_json_escaped` variable.
- `_keyhelp_name_encoded` (renamed `_keyhelp_name_url_encoded`) is now computed once above the `for` loop instead of on every iteration.

On the domain-key-rotation point: agreed this is a real gap, but I'd like to punt it to a follow-up rather than fold it into this PR. It's not a simple delete+recreate — KeyHelp entries can be referenced by multiple existing domains/vhosts pointing at the old certificate, and swapping the entry out would require updating all of those references too, not just the certificate itself. I'll think through the right approach separately.

PR #7166, opened 2026-07-29
Refactor dns_freemyip.sh for enhanced compatibility

ExtremeFiretop:
Updated dns_freemyip.sh for better readability and compatibility with ASUSWRT-Merlin. 
The Merlin repository’s base BusyBox configuration currently disables the BusyBox rev applets, It's possible other embedded systems like routers do not have tools such tools.

```
Admin@GT-BE98_Pro:/tmp/home/root#
Admin@GT-BE98_Pro:/tmp/home/root# which rev 2>/dev/null || echo "rev is missing"
rev is missing
Admin@GT-BE98_Pro:/tmp/home/root# which sed 2>/dev/null || echo "sed is missing"
/bin/sed
Admin@GT-BE98_Pro:/tmp/home/root# which seq 2>/dev/null || echo "seq is missing"
/usr/bin/seq
Admin@GT-BE98_Pro:/tmp/home/root# which tee 2>/dev/null || echo "tee is missing"
/usr/bin/tee
Admin@GT-BE98_Pro:/tmp/home/root#
```

We also don't use a path for:  **/dev/fd/2** and we instead use: **/proc/self/fd/2**
While neither pathname is guaranteed by POSIX it seems that /dev/fd exists across more Unix-like operating systems, while /proc/self/fd is specifically associated with Linux:
```
Admin@GT-BE98_Pro:/tmp/home/root# ls -ld /dev/fd /proc/self/fd 2>/dev/null
dr-x------    2 Admin    root             0 Jul 29 13:59 /proc/self/fd
Admin@GT-BE98_Pro:/tmp/home/root#
```

Instead I replaced the logic with something a little more portable.

### DIFF
--- a/dnsapi/dns_freemyip.sh
+++ b/dnsapi/dns_freemyip.sh
@@ -6,8 +6,7 @@ Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_freemyip
 Options:
  FREEMYIP_Token API Token
 Issues: github.com/acmesh-official/acme.sh/issues/6247
-Author: Recolic Keghart <root@recolic.net>, @Giova96
-Modified by: ExtremeFiretop for ASUSWRT-Merlin compatibility
+Author: Recolic Keghart <root@recolic.net>, @Giova96, ExtremeFiretop
 '
 
 FREEMYIP_DNS_API="https://freemyip.com/update?"
--- a/dnsapi/dns_freemyip.sh
+++ b/dnsapi/dns_freemyip.sh
@@ -12,6 +12,7 @@ Author: Recolic Keghart <root@recolic.net>, @Giova96, ExtremeFiretop
 FREEMYIP_DNS_API="https://freemyip.com/update?"
 
 ################ Public functions ################
+
 #Usage: dns_freemyip_add    fulldomain    txtvalue
 dns_freemyip_add() {
   fulldomain="$1"
@@ -26,8 +27,10 @@ dns_freemyip_add() {
     _err "Please specify your token and try again."
     return 1
   fi
+  
   #save the credentials to the account conf file.
   _saveaccountconf_mutable FREEMYIP_Token "$FREEMYIP_Token"
+  
   if _is_root_domain_published "$fulldomain"; then
     _err "freemyip API don't allow you to set multiple TXT record for the same subdomain!"
     _err "You must apply certificate for only one domain at a time!"
@@ -36,6 +39,7 @@ dns_freemyip_add() {
     _debug "If you are testing this workflow in github pipeline or acmetest, please set TEST_DNS_NO_SUBDOMAIN=1 and TEST_DNS_NO_WILDCARD=1"
     return 1
   fi
+  
   # txtvalue must be url-encoded. But it's not necessary for acme txt value.
   _freemyip_get_until_ok "${FREEMYIP_DNS_API}token=$FREEMYIP_Token&domain=$fulldomain&txt=$txtvalue" 2>&1
   return $?
@@ -55,6 +59,7 @@ dns_freemyip_rm() {
     _err "Please specify your token and try again."
     return 1
   fi
+  
   #save the credentials to the account conf file.
   _saveaccountconf_mutable FREEMYIP_Token "$FREEMYIP_Token"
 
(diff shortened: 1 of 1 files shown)


neilpang:
Please run `DNS.yml` in your fork with the real FreeMyIP credentials and confirm that it passes. The current PR run has `hasToken=false`, so all functional DNS test jobs were skipped. ShellCheck and shfmt alone do not test the changed add/remove flow.

ExtremeFiretop:
> Please run `DNS.yml` in your fork with the real FreeMyIP credentials and confirm that it passes. The current PR run has `hasToken=false`, so all functional DNS test jobs were skipped. ShellCheck and shfmt alone do not test the changed add/remove flow.

I ran DNS.yml in my fork using real FreeMyIP credentials.

The initial Windows failure was caused by repeated curl timeouts to the FreeMyIP API. Windows passed when I reran the failed jobs, so that appears to have been a transient runner or network issue.

The rerun then reached Tribblix and exposed another portability issue: the two retry loops currently depend on seq. On Tribblix, neither loop executed. The log went directly from the verification message to the final failure in the same second, with no DNS lookup retries or FreeMyIP API attempts.

Tribblix, why do you do this to me? 😅

The seq dependency predates this PR, I didn't introduce the seq dependency. So I'm assuming this workflow must be relatively new bar to pass?

Anyways, the requested compatibility workflow exposed it. Back to the drawing board: I’ll work to try and replace both loops with POSIX shell counters and rerun the full workflow.

I’m putting the PR back into draft until the fix is ready and the full DNS.yml run passes.

ExtremeFiretop:
@neilpang 

All good to go now. Thanks.

neilpang:
Two changes before merge:

**1. Token leak at the default log level.**

```sh
_err "Failed to request freemyip API: $_fmi_url . Server does not say 'OK'"
```

`$_fmi_url` contains `token=<FREEMYIP_Token>`. `_err` writes unconditionally to stderr, `$LOG_FILE` and syslog -- no `--debug` needed -- and this line fires exactly when a user is about to paste the output into an issue. The URL adds nothing over the `_info` line already printed, so drop it:

```sh
_err "Failed to request freemyip API. Server does not say 'OK'"
```

The `_debug` line above it carries the token too, but core `_get` already does `_debug url "$url"`, so masking it in the hook would change nothing -- leave it.

**2. Use the core wrapper instead of raw `grep`.**

```sh
if _contains "$_fmi_response" "OK"; then
  return 0
fi
```

Right now the response is printed twice: once by `printf ... >&2`, once by `grep` writing its match to stdout.

Optional: rename `i` to `_fmi_i` (everything else in the file is `_fmi_*`, and hooks are sourced into one shared shell), and revert the unrelated blank-line deletions so the diff stays on the portability fix.

Rest verified locally: the `sed` `_get_root` is output-identical to `rev | cut -d . -f 1-3 | rev` for 1-6 labels and is plain POSIX BRE.

ExtremeFiretop:
> Two changes before merge:
> 
> **1. Token leak at the default log level.**
> 
> ```shell
> _err "Failed to request freemyip API: $_fmi_url . Server does not say 'OK'"
> ```
> 
> `$_fmi_url` contains `token=<FREEMYIP_Token>`. `_err` writes unconditionally to stderr, `$LOG_FILE` and syslog -- no `--debug` needed -- and this line fires exactly when a user is about to paste the output into an issue. The URL adds nothing over the `_info` line already printed, so drop it:
> 
> ```shell
> _err "Failed to request freemyip API. Server does not say 'OK'"
> ```
> 
> The `_debug` line above it carries the token too, but core `_get` already does `_debug url "$url"`, so masking it in the hook would change nothing -- leave it.
> 

This seems like unrelated issues though that predate this PR again.
Usually, I create a single PR per issue, anyways that's how my Github org handles them when possible.

Did you want me to just do this all in this PR now? 
Or create separate PRs for the existing security issues?

ExtremeFiretop:
@neilpang 

I did the refactoring live in Github, so sorry for the noisey commit history.
Feel free to squash all that down.

Hope that addresses every one of your concerns. My initial concern was only resolving the rev dependency.
I played along though ;) I don't mind the challenge!

neilpang:
Code looks right now -- all three points addressed, and `seq` / `rev` / `/dev/fd` are all gone. `dash -n`, `sh -n` and ShellCheck (sh dialect) are clean here.

DNS.yml has not gone green on `9776ab7` though: both runs failed, and the earlier passing run on `96220fa` is deleted, so there is no green run left on record.

Not the diff. The test domain has a leftover TXT record from an earlier aborted run:

```
"Answer":[{"name":"_acme-challenge.***","type":16,"TTL":62,"data":"\"PiM9Ycr...R3JwI\""}]
'***' already has a TXT record published!
freemyip API don't allow you to set multiple TXT record for the same subdomain!
```

That value is not the one the run was trying to add. `_is_root_domain_published` hits it on the first `_ns_lookup` and `dns_freemyip_add` aborts by design -- same path as before this PR, and ubuntu has `seq`, so the old loop behaved identically.

Clear the record, then re-run DNS.yml once:

```
curl "https://freemyip.com/update?token=<TOKEN>&domain=<yourdomain>.freemyip.com&txt="
```

Don't push while a run is in flight -- a cancelled run dies between add and remove and leaves the record behind, which is most likely how this one got stuck.

ExtremeFiretop:
> ```
> curl "https://freemyip.com/update?token=<TOKEN>&domain=<yourdomain>.freemyip.com&txt="
> ```
> 
> Don't push while a run is in flight -- a cancelled run dies between add and remove and leaves the record behind, which is most likely how this one got stuck.

Ahhh! I was trying to determine why it got stuck.
I could tell by the logs it was due to a txt record but I couldn't tell how the stale record happened.

Thanks for explaining. Give me an hour to re-run the DNS.yml and we should be good :)
Thanks for your review, was a pleasure @neilpang

ExtremeFiretop:
@neilpang 

All good now, you have all greens.
Thanks again for your time. I'll take these lessions learned if/when we ever cross paths again!

The ball is in your court.
Joel S.

Issue #1261, opened 2018-02-13
Update dns api to support v2 wildcard cert


Neilpang:
To support v2 wildcard cert,  we need to add 2 txt record for the same domain. 
for example:

```sh
_acme-challenge.example.com   TXT   "this is txt value 1"
_acme-challenge.example.com   TXT   "this is txt value 2"
```

In many dns api hooks,  in the `dns_xx_add()` function, they try to `UPDATE` the existing txt record, instead of `ADD` a new record.
This was a good practice for ACME v1,  but it's not good in ACME v2.

In ACME v2,  we just need to add new txt record all the time in the `dns_xx_add()` function,
And in the `the `dns_xx_rm()` function`, we must delete the txt record based on the according to the specified txt value.

See my changes:

https://github.com/Neilpang/acme.sh/commit/ea25492c2823a4971b3d2eb28fa0dcd2b5105db9#diff-51fe23dd1a90a481487dbca5b9c3ae24

https://github.com/Neilpang/acme.sh/commit/72f54ca6c13c33943aafbacaee03423907f5e738#diff-d48ca70b90232acffb2b5b9d1ec2938a

https://github.com/Neilpang/acme.sh/commit/584fb2904b6b63cf13d71ba20fa184ade62730a0#diff-f272833bc0ccf326ea343539e829f1d3

Neilpang:
@wpk-   Please update the rm function of  dns_ad  hook.

Neilpang:
@baiyangliu   please support the  rm function for  dns_ali hook.

Neilpang:
@martgras   please  make sure  dns_azure hook works in this way.

Neilpang:
@boyanpeychev   please update  dns_cloudns hook

boyanpeychev:
@Neilpang please accept this pull request and I will proceed with this implementation:
https://github.com/Neilpang/acme.sh/pull/1094

Neilpang:
@boyanpeychev   accepted,  please go ahead.

Neilpang:
@pho3nixf1re   please update dns_dnsimple hook

Neilpang:
@dkerr64   please update dns_freedns.sh

Neilpang:
@fcrozat   please update dns_gandi_livedns.sh

Neilpang:
@justmwa    please update dns_me.sh

Neilpang:
@justmwa   please also update dns_nsone.sh

[...]

Issue #7071, opened 2026-07-02
Report bugs to WEDOS (dns_wedos) DNS API


jforman96:
Please report any issues with the WEDOS WAPI DNS API (dnsapi/dns_wedos.sh) here.

Related PR: #7072

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

Issue #7064, opened 2026-07-02
Report bugs to Volcano Engine DNS API


gonghao:
This is the place to report bugs in the Volcano Engine DNS API.

If you experience a bug, please report it in this issue.

[Related Official Docs - TrafficRoute DNS 套件/云解析 DNS](https://www.volcengine.com/docs/6758/2222459?lang=zh)

### Timeline
- [Released] 2027.07.03 PR Merged, and master branch synced
- [NOT Released] 2027.07.03 PR created #7069 
- [NOT Released] 2027.07.02 Wiki edited #7066 #7076
- [NOT Released] 2027.07.02 Init Bug Reporter

neilpang:
Closing this tracking issue for now to keep the issue list clean. It remains the place to report problems with this provider -- if you hit a bug, comment here and the issue will be reopened.

PR #7170, opened 2026-08-02
feat: added nexdns dnsapi

morozov-alexey:
Adds `dnsapi/dns_nexdns.sh`, a DNS-01 hook for [NexDNS](https://nexdns.tech), an authoritative DNS hosting service with a REST API. It is the only file this PR touches.

`dns_nexdns_add` finds the zone that owns the challenge name and creates the TXT record in it. `dns_nexdns_rm` lists the TXT records at that name, picks the one carrying exactly this challenge value, and deletes it by id.

### Options

| Variable | Required | Description |
|---|---|---|
| `NEXDNS_Token` | yes | API token, created at <https://nexdns.tech/settings/api-keys> on a plan that includes API access |
| `NEXDNS_Api` | no | API base url, default `https://api.nexdns.tech/v1` |

### Usage

```sh
export NEXDNS_Token="your-api-token"
acme.sh --issue --dns dns_nexdns -d example.com -d '*.example.com'
```

The token is saved with `_saveaccountconf_mutable` on the first issue, so renewals need nothing in the environment. `NEXDNS_Api` is saved only when it differs from the default and cleared otherwise, so a user who never set it does not end up with a pinned copy of the default in `account.conf`.

acme.sh's own propagation check covers the wait after the record is added. For anyone who prefers a fixed wait instead, 60 seconds is the figure our documentation gives.

### Conventions

- POSIX `sh` with `#!/usr/bin/env sh`; `return` everywhere, never `exit`.
- Every request goes through `_get` / `_post` with `_H1`..`_H3`. No direct `curl`, and no assumption about which transport acme.sh picked.
- No `jq`, `awk`, `grep -A`, `grep -Po`, `rev`, `xargs` or `iconv`. Parsing is `_normalizeJson`, `tr`, `grep`, `_egrep_o`, `_head_n` and `cut`.
- `_get_root` starts its counter at `1`, so DNS alias mode works.
- `dns_nexdns_rm` removes the record **by TXT value** (#1261) rather than rewriting the record set. That matters here: a certificate for `example.com` plus `*.example.com` puts two different challenge values at one name, and both have to survive until both authorizations are done.
- Credentials are read with `_readaccountconf_mutable` and written with `_saveaccountconf_mutable`, in the add path only.
- Verbose output is `_debug2`. The token appears only in the `_H1` header, never in a log line.
- TTL is hardcoded at 120 rather than introducing a variable for it.
- The `dns_nexdns_info` metadata block is present and points at `dnsapi2`.

### Zone lookup and error handling

Zone discovery walks the label list and asks the API for each candidate. The search parameter matches on a substring, so a page can contain zones that merely contain the candidate; the hook takes the id of the entry whose name is exactly the candidate, not the first row.

A rejected request comes back as `{"error":{"code":..,"message":..}}`, and the hook surfaces that message rather than a generic failure. A delete answers 204 with an empty body, so for that one the status line from `$HTTP_HEADER` is what it checks.

A `429` is waited out and the request retried, up to four attempts, in the shape `dns_hetznercloud.sh` and `dns_bunny.sh` already use. One certificate costs a handful of requests, but a renewal sweep over many of them meets the account's per-minute budget, and that run is unattended. `Retry-After` is treated as a floor rather than as the answer: a limiter that reports the time one token needs at an average release rate can name a second when nothing frees for a minute, so the wait grows across attempts. A header that is not a plain count of seconds - RFC 9110 also allows a date - falls through to that backoff rather than being parsed, since comparing a date numerically would abort the hook outright.

### Testing

Linters, as the workflows run them:

```
$ shellcheck -e SC2181 -e SC2089 dnsapi/dns_nexdns.sh
(no output)

$ shfmt -l -w -i 2 . ; git diff --exit-code
(no file listed, no diff)
```

`sh -n` and `dash -n` are clean as well.

Both public functions were then driven end to end against a stub of the API that logs every request it receives, once over curl and once with `ACME_USE_WGET=1`, so the request shapes are executed rather than eyeballed:

| Scenario | Result |
|---|---|
| `dns_nexdns_add` | walks `_acme-challenge.www.example.com` down to `example.com`, POSTs `{"name":"_acme-challenge.www","type":"TXT","content":"...","ttl":120}`, returns 0 |
| zone list containing a decoy | given both `notexample.com` and `example.com` for the search `example.com`, selects `example.com` |
| `dns_nexdns_rm` | given two challenge records at one name, deletes the id whose value matches and leaves the other alone, returns 0 |
| `dns_nexdns_rm` with the value absent | reports that the record is already gone and returns 0, with no delete issued |
| rejected token | prints the message the API returned and returns 1 |
| body reformatted with whitespace | still parses, via `_normalizeJson` |
| `429` with `Retry-After: 1` | waits the backoff rather than the second it was offered, retries, returns 0 with the body of the second response |

The same four scenarios were then run against the real API on a live zone (2026-07-31), which is where the stub cannot speak for itself: the added record was confirmed to resolve from `1.1.1.1` and `8.8.8.8` and from the authoritative nameserver, and the selective removal was checked with two challenge records present at the same name - the matching value went, the other stayed. The zone was left as it was found.

DNS-API-Test: not yet run on this branch, since it needs the fork secrets. I will run it and link the run here before asking for review.

### Documentation

I will add the usage section to the [dnsapi2](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2) wiki page. `README.md` needs no change, since it links to the wiki list rather than enumerating providers.

### DIFF
--- a/dnsapi/dns_nexdns.sh
+++ b/dnsapi/dns_nexdns.sh
@@ -1,0 +1,239 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_nexdns_info='NexDNS
+Site: nexdns.tech
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_nexdns
+Options:
+ NEXDNS_Token API token. Can be created at https://nexdns.tech/settings/api-keys
+ NEXDNS_Api API base url. Default "https://api.nexdns.tech/v1". Optional.
+Issues: github.com/acmesh-official/acme.sh/issues
+Author: NexDNS <https://github.com/nexdns>
+'
+
+NEXDNS_Api_Default="https://api.nexdns.tech/v1"
+
+########  Public functions #####################
+
+#Usage: dns_nexdns_add   _acme-challenge.www.example.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_nexdns_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  if ! _nexdns_init; then
+    return 1
+  fi
+
+  _saveaccountconf_mutable NEXDNS_Token "$NEXDNS_Token"
+  if [ "$NEXDNS_Api" != "$NEXDNS_Api_Default" ]; then
+    _saveaccountconf_mutable NEXDNS_Api "$NEXDNS_Api"
+  else
+    _clearaccountconf_mutable NEXDNS_Api
+  fi
+
+  _debug "First detect the root zone"
+  if ! _get_root "$fulldomain"; then
+    _err "Cannot find the zone of $fulldomain in this NexDNS account."
+    return 1
+  fi
+  _debug _sub_domain "$_sub_domain"
+  _debug _domain "$_domain"
+  _debug _domain_id "$_domain_id"
+
+  _info "Adding the TXT record for $fulldomain"
+  if ! _nexdns_rest POST "zones/$_domain_id/records" "{\"name\":\"$_sub_domain\",\"type\":\"TXT\",\"content\":\"$txtvalue\",\"ttl\":120}"; then
+    return 1
+  fi
+
+  _info "The TXT record has been added."
+  return 0
+}
+
+#Usage: dns_nexdns_rm   _acme-challenge.www.example.com   "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_nexdns_rm() {
+  fulldomain=$1
+  txtvalue=$2
+
+  if ! _nexdns_init; then
+    return 1
+  fi
+
+  _debug "First detect the root zone"
+  if ! _get_root "$fulldomain"; then
+    _err "Cannot find the zone of $fulldomain in this NexDNS account."
+    return 1
+  fi
+  _debug _sub_domain "$_sub_domain"
+  _debug _domain "$_domain"
+  _debug _domain_id "$_domain_id"
+
+  _info "Removing the TXT record for $fulldomain"
+  if ! _nexdns_rest GET "zones/$_domain_id/records?type=TXT&name=$_sub_domain"; then
+    return 1
+  fi
+
+  #All the challenge records share one name and one type, so the value is the
+  #only thing that tells them apart. A certificate covering example.com and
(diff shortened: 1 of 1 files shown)


morozov-alexey:
DNS-API-Test passed on my fork, all 17 platforms: https://github.com/morozov-alexey/acme.sh/actions/runs/30764952969

Each one issued a staging certificate for a domain plus its wildcard, then deactivated and revoked it. The two challenges share one `_acme-challenge` label, so that run also covers `dns_nexdns_rm` deleting the record whose value matches and leaving the other in place.

I will add the usage section to the [dnsapi2](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2) wiki page once this is merged.

neilpang:
1. The dnsapi2 wiki page has no nexdns section, so the Docs: anchor in
   the info block does not resolve. It has to land before the merge,
   not after.

2. Issues: points at the generic issue list. Open a tracking issue for
   the provider and reference that number -- 122 of the 193 hooks in
   the tree carry a specific one.

3. Line 204: the 429 wait has no ceiling. Retry-After: 3600 sleeps for
   an hour, and three of those can stack in one hook call. A wait that
   large is a refusal rather than a schedule -- cap it, and return 1
   past the cap so the next cron run picks it up.

4. Line 217: the error check fires on {"error":null,...} too, turning a
   success into a failure with an empty message. Match on "error":{ or
   test the status code first.

1 and 2 block the merge; 3 and 4 do not.

morozov-alexey:
All four done. Wiki section is in, #7179 is the tracking issue.

neilpang:
1. #7179 is titled "dns_nexdns: tracking issue". The convention in the
   tree is "Report bugs to <Provider> DNS API" -- see #7071, #7064,
   #7057, #6835. Rename it to "Report bugs to NexDNS DNS API
   (dns_nexdns)".

2. DNS-API-Test has not passed on 0a0758e. The run on that commit got
   "Invalid API key." back from the first zones request, so the token
   in your fork has lapsed since the 08-02 run. Refresh it and
   dispatch the workflow again -- the code changed after the green
   run, so that result no longer covers what is on the branch.

morozov-alexey:
Both done. Green on 0a0758e: https://github.com/morozov-alexey/acme.sh/actions/runs/31102779716

PR #7171, opened 2026-08-03
Add deploy hooks for Caddy (remote), EdgeRouter, HAOS, ReadyNAS, UISP, and UniFi OS Server for Mac

Pablo1:
Six new deploy hooks, each targeting a device/service with no existing coverage in `deploy/`:

- `caddy_remote.sh` — Caddy on a remote host via SSH (systemd service, sudo tee + systemctl reload)
- `edgerouter.sh` — Ubiquiti EdgeRouter (EdgeOS), SSH + sudo tee to `/config/ssl/`, lighttpd restart
- `haos.sh` — Home Assistant OS, SCP to `/ssl/`, `ha core restart`
- `readynas.sh` — Netgear ReadyNAS OS 6, SSH pipe into frontview's apache2 cert file, apache2 restart
- `uisp.sh` — UISP (nico640/docker-unms Podman image), copies to usercert dir + container restart
- `unifios_mac.sh` — UniFi OS Server for Mac, via its local REST API (login, list, delete stale cert, upload, activate) since no existing hook covers this app's Postgres-backed cert storage

All six pass `shellcheck -e SC2181 -e SC2089` and `shfmt -i 2` (verified locally against the same flags used in this repo's CI).

### DIFF
--- a/deploy/caddy_remote.sh
+++ b/deploy/caddy_remote.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env sh
+# Deploy hook for a remote Caddy instance via SSH.
+#
+# Deploys cert files to a remote host via SSH + sudo tee, fixes ownership
+# to caddy:caddy mode 640, then reloads Caddy via systemctl. Designed for
+# hosts where Caddy runs as a systemd service under the `caddy` user.
+#
+# Settings:
+#   DEPLOY_CADDY_REMOTE_USER - SSH username (required)
+#   DEPLOY_CADDY_REMOTE_HOST - SSH hostname or IP (required)
+#   DEPLOY_CADDY_REMOTE_CERT_DIR - remote cert directory (default: "/etc/caddy/certs")
+#
+# Example:
+#   export DEPLOY_CADDY_REMOTE_USER="acmeuser"
+#   export DEPLOY_CADDY_REMOTE_HOST="192.168.1.30"
+#   acme.sh --deploy -d example.com --deploy-hook caddy_remote
+
+caddy_remote_deploy() {
+  _cdomain="$1"
+  _ckey="$2"
+  _ccert="$3"
+  _cca="$4"
+  _cfullchain="$5"
+
+  _getdeployconf DEPLOY_CADDY_REMOTE_USER
+  if [ -z "$DEPLOY_CADDY_REMOTE_USER" ]; then
+    _err "DEPLOY_CADDY_REMOTE_USER is required."
+    return 1
+  fi
+
+  _getdeployconf DEPLOY_CADDY_REMOTE_HOST
+  if [ -z "$DEPLOY_CADDY_REMOTE_HOST" ]; then
+    _err "DEPLOY_CADDY_REMOTE_HOST is required."
+    return 1
+  fi
+
+  _getdeployconf DEPLOY_CADDY_REMOTE_CERT_DIR
+  if [ -z "$DEPLOY_CADDY_REMOTE_CERT_DIR" ]; then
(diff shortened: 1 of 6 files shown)


neilpang:
shellcheck does not pass. With this repo's exact CI command,
`shellcheck -e SC2181 -e SC2089 **/*.sh`, dev is exit 0 and these six
files make it exit 1 -- three SC2002 hits at caddy_remote.sh:49,
caddy_remote.sh:56, edgerouter.sh:60. (Actions' bash has globstar off,
so `**/*.sh` expands one level and deploy/ is in scope.) shfmt -i 2 is
clean. The Shellcheck job has not actually run here -- it is held at
the first-time-contributor approval gate, so the green check is only
the pull_request_target "Update issues" job.

Please split this: one PR per hook, each with evidence it ran against
real hardware. Six unrelated devices in one commit cannot be reviewed
or reverted independently.

caddy_remote, edgerouter, haos and readynas look like configuration of
the existing deploy/ssh.sh rather than new hooks. Each reduces to "copy
N files to remote paths, run one remote command" = DEPLOY_SSH_KEYFILE /
CERTFILE / CAFILE / FULLCHAIN plus DEPLOY_SSH_REMOTE_CMD (plus
DEPLOY_SSH_USE_SCP=yes for haos). ssh.sh additionally gives backups,
multiple servers, port parsing and a configurable ssh/scp command.
What can ssh.sh not do here?

unifios_mac.sh:

  - Five raw curl calls. Hooks must use _get / _post, so that proxy,
    CA bundle, timeouts, the wget fallback and --debug tracing work.
    -k also hardcodes insecure TLS instead of honoring HTTPS_INSECURE.
    For cookie sessions, see deploy/peplink.sh and deploy/ruckus.sh:
    read Set-Cookie from $HTTP_HEADER, send it back as
    _H1="Cookie: ...".

  - Hard dependency on python3 for JSON. No deploy hook does this;
    core has _json_encode, _dbase64, _egrep_o, _url_encode.

  - The password and the session JWT are passed as python3 -c argv,
    so any local user can read them from ps.

  - The old certificate is deleted before the new one is uploaded --
    and that is the active cert. If the upload or activation then
    fails, the server is left on its self-signed cert. Upload,
    activate, then delete.

  - It uploads $_ccert, the leaf only; deploy/unifi.sh uses
    $_cfullchain. Clients without a cached intermediate fail the
    handshake.

The four SSH hooks:

  - StrictHostKeyChecking=accept-new silently trusts an unknown host
    key on the very run that carries the private key; ssh.sh
    deliberately requires the key to be exchanged beforehand.
    accept-new also needs OpenSSH 7.6+ and is a hard error on older
    clients and on Dropbear.

  - Only caddy_remote sets BatchMode=yes; the other three can block on
    a password prompt.

  - _ssh_base, _ssh_opts, _ssh_target, _scp_opts are unprefixed
    globals in edgerouter, haos and readynas. Hooks are sourced into
    one shell (acme.sh:6682), so private variables need the provider
    prefix; caddy_remote's `_cr_*` is the right pattern.

  - caddy_remote's remote "sudo tee $DIR/..." path is unquoted.

uisp.sh: no _exists podman guard, podman hardcoded, "podman stop -t
120" stalls the deploy for up to two minutes, and the usercert
directory is never created.

All six: use _tail_n 1 rather than tail -n1; add the usual _debug of
the incoming arguments; call _savedeployconf right after reading the
settings, not only after every step succeeded -- otherwise a first run
that fails at the last step discards the credentials the user
exported.

None of the six has an entry on the deployhooks wiki page, and there is
no linked issue or user request for any of them.

Pablo1:
Closing to split into per-hook PRs per review feedback. Will resubmit individually once each hook has been re-tested against real hardware.

Issue #7181, opened 2026-08-07
Report bugs to uisp deploy hook


Pablo1:
This is the place to report bugs in the `uisp` deploy hook.

If you experience a bug with `--deploy-hook uisp`, please report it in this issue. (Please include the `--debug` output to assist in diagnosing any problems.)

Thanks!

PR #7183, opened 2026-08-07
Add UISP (nico640/docker-unms) deploy hook

Pablo1:
## Summary

Adds a deploy hook for UISP (Ubiquiti's ISP management platform), commonly self-hosted via the community [nico640/docker-unms](https://hub.docker.com/r/nico640/docker-unms) container image -- there's no official first-party container, so this is the de facto way to self-host UISP outside Ubiquiti's cloud. No existing deploy hook covers this.

## How it works

UISP reads its TLS certificate from a `usercert` directory (`custom.crt`/`custom.key`) mounted into the container, and needs a container restart to pick up a new cert -- there's no live-reload API. The hook copies the fullchain/key into that directory and restarts the container.

- `DEPLOY_UISP_CONTAINER_CMD` selects the container engine -- defaults to `podman`, but can be set to `docker` (or anything else podman/docker-CLI-compatible).
- `DEPLOY_UISP_STOP_TIMEOUT` controls how long (in seconds) the hook waits for a graceful container stop before escalating -- defaults to `30`. This matters because the UISP container runs Postgres internally (alongside Redis, RabbitMQ, and other services); `stop -t N` sends `SIGTERM` and waits N seconds before force-killing with `SIGKILL`, and killing Postgres mid-write instead of letting it shut down cleanly risks database corruption.

## Usage

    export DEPLOY_UISP_USERCERT_DIR="/path/to/uisp/usercert"
    export DEPLOY_UISP_CONTAINER="uisp"
    # optional; shown with defaults
    # export DEPLOY_UISP_CONTAINER_CMD="podman"   # or "docker"
    # export DEPLOY_UISP_STOP_TIMEOUT="30"
    acme.sh --deploy -d example.com --deploy-hook uisp

## Testing

acme.sh itself was installed and run from macOS 26.6, deploying to a real, running UISP/Podman container:
- Full deploy run completed successfully end-to-end
- Deployed certificate content confirmed byte-identical to the source cert (`diff` against acme.sh's own issued files)
- Container restart confirmed via a fresh `StartedAt` timestamp matching the deploy time
- Deleted both certificate files from the usercert directory, confirmed it empty, then reran the hook and confirmed both files were recreated with correct size/permissions at the exact deploy timestamp -- verifies the write path is a real, working copy, not a no-op against already-matching files
- Certificate served by UISP inspected via a browser TLS-inspection extension and confirmed complete (full chain present, trusted)

## Docs

Will add an entry to the [deployhooks wiki page](https://github.com/acmesh-official/acme.sh/wiki/deployhooks) once this is merged.

Please report bugs to https://github.com/acmesh-official/acme.sh/issues/7181.


neilpang:
The four uisp points from #7171 are addressed.

1. This reduces to configuration of deploy/localcopy.sh, not a new hook:

     DEPLOY_LOCALCOPY_FULLCHAIN="<dir>/custom.crt"
     DEPLOY_LOCALCOPY_CERTKEY="<dir>/custom.key"
     DEPLOY_LOCALCOPY_RELOADCMD="podman restart -t 30 uisp"

   Or with no hook at all:

     --install-cert -d example.com \
       --fullchain-file <dir>/custom.crt \
       --key-file <dir>/custom.key \
       --reloadcmd "podman restart -t 30 uisp"

   Same question as for the SSH hooks in #7171: what can localcopy.sh
   not do here?

2. The key is written with plain cp, so an already-existing custom.key
   keeps whatever mode it had -- 644 stays 644. localcopy.sh touches
   and chmod 600 before writing for exactly this reason.

3. stop + start should be one "restart -t $DEPLOY_UISP_STOP_TIMEOUT"
   (both podman and docker have it). As written, a stop that succeeds
   followed by a start that fails leaves UISP down.

4. Still no entry on the deployhooks wiki page. That has to land
   before the merge, not after.

Pablo1:
Thank you for the quick and thorough review. You are absolutely correct. Closing the PR.

Issue #7182, opened 2026-08-07
Report bugs to unifios deploy hook


Pablo1:
This is the place to report bugs in the `unifios` deploy hook.

If you experience a bug with `--deploy-hook unifios`, please report it in this issue. (Please include the `--debug` output to assist in diagnosing any problems.)

Thanks!

Pablo1:
Reopening — active PR (#7184) still in review.

Tharit:
Not a bug as such, but I would suggest to also update the docs at the beginning of unifios.sh, to be in line with the updated wiki docs, reflecting that this works on devices as well.

Pablo1:
Thank you for the report, @Tharit. @neilpang already made this change: [commit 1cbd2233](https://github.com/acmesh-official/acme.sh/commit/1cbd2233864c70d23797b6e103cfd4f5113d73b1). It landed on `dev` before your comment here.

Issue #6916, opened 2026-04-17
Problem (and solution) with unifi deploy hook.


btimby:
I wanted to deploy the certificate to my Unifi UCG Fiber (Unifi OS 5.0.16) remotely. I did not want to install the deploy hook locally and tried emulating it via SSH (`systemctl restart unifi-core`, `systemctl restart nginx, `unifi-os restart`) My certificate would always revert after about 10-15 minutes.

I inspected the communication from the web console and came up with the following script that cleanly applies the certificate using that method. Then whatever internal magic they do is performed as necessary and the cert sticks.

I am not sure how or if this can be adopted into acme.sh, but perhaps others find this useful (I saw reports about certs reverting like I was experiencing.)

```bash
#!/bin/sh
# Deploy cert to UniFi OS via API
# Usage: deploy-unifi.sh <host> <username> <password> <cert_file> <key_file>

UNIFI_HOST="$1"
UNIFI_USER="$2"
UNIFI_PASS="$3"
CERT_FILE="$4"
KEY_FILE="$5"

COOKIE_JAR=$(mktemp)

# Login and get TOKEN cookie
curl -sk -c "$COOKIE_JAR" \
  -H "Content-Type: application/json" \
  -d "{\"username\":\"$UNIFI_USER\",\"password\":\"$UNIFI_PASS\",\"rememberMe\":false}" \
  "https://$UNIFI_HOST/api/auth/login" > /dev/null

# Extract TOKEN from cookie jar
TOKEN=$(grep -m1 'TOKEN' "$COOKIE_JAR" | awk '{print $7}')
if [ -z "$TOKEN" ]; then
  echo "[ERROR] login failed — no TOKEN cookie received"
  rm -f "$COOKIE_JAR"
  exit 1
fi

# Extract CSRF token from JWT payload (base64url decode middle section)
PAYLOAD=$(echo "$TOKEN" | cut -d'.' -f2)
PAD=$(( ${#PAYLOAD} % 4 ))
[ "$PAD" -ne 0 ] && PAYLOAD="${PAYLOAD}$(printf '%0.s=' $(seq 1 $((4 - PAD))))"
CSRF=$(echo "$PAYLOAD" | tr '_-' '/+' | base64 -d 2>/dev/null | grep -o '"csrfToken":"[^"]*"' | cut -d'"' -f4)

if [ -z "$CSRF" ]; then
  echo "[ERROR] could not extract CSRF token from JWT"
[...]

PR #7184, opened 2026-08-07
Add UniFi OS Server deploy hook

Pablo1:
## Summary

This adds a deploy hook for UniFi OS Server. UniFi OS Server is Ubiquiti's self-hosted UniFi OS platform, available for macOS, Linux, and Windows with WSL2.

UniFi OS Server is a different product from the Cloud Key and UDM hardware. The existing `unifi` deploy hook already covers that hardware, including Cloud Key with UnifiOS v2.0.0 and later (see #6916). It does not cover this separate, self-hosted application.

UniFi OS Server stores certificates in its own Postgres database. It uses a REST API for this, not flat config files or a Java keystore. As a result, the `unifi` hook's method does not work here.

## How it works

UniFi OS Server exposes a REST API on its management port. The default port is 11443. This is the same API that its own web UI uses for certificate management. I found this API with reverse engineering. This method used the browser's Network tab, together with the real GUI's upload, activate, and delete actions.

The hook logs in, uploads the new certificate with a unique name, and activates it. The hook removes old entries only after it activates the new certificate. It only removes entries that it created for this domain. As a result, a failure during this process never leaves the server without a valid active certificate.

This hook uses core acme.sh helpers throughout: `_get`/`_post`, `_json_encode`, `_durl_replace_base64`, `_dbase64`, and `_egrep_o`. It does not use raw `curl -k` or an external language runtime. As a result, the wget fallback, `--debug` tracing, and `CA_BUNDLE` all work the same as they do for every other hook.

The management API's certificate can be self-signed. This is true for a fresh install. There is no reliable way to know in advance if a previous run already replaced it with a real certificate. For this reason, the hook sets `HTTPS_INSECURE=1` itself on every run. This setting is scoped to the hook's own subshell. It does not weaken TLS verification for the rest of the acme.sh run, for example the connection to the ACME CA.

`DEPLOY_UNIFIOS_HOST` defaults to `https://localhost:11443`. This is the common case, where UniFi OS Server runs on the same machine as acme.sh. You can point it at any host that the machine can reach. The Testing section below has an example.

## Usage

    export DEPLOY_UNIFIOS_USERNAME="acmeuser"
    export DEPLOY_UNIFIOS_PASSWORD="xxxxx"
    acme.sh --deploy -d example.com --deploy-hook unifios

## Testing

I installed and ran acme.sh itself from macOS 26.6. I tested it against two real UniFi OS Server instances, on different platforms, with the same unmodified code:

- **macOS, local, default `DEPLOY_UNIFIOS_HOST`**: an existing instance with a certificate already present for the domain. This tested the duplicate-detection and reuse path. The server correctly rejects a second upload of an identical certificate, by fingerprint. The hook detects this and reuses and activates the existing entry, instead of failing.
- **Linux, Ubuntu 26.04, self-hosted, remote over the network, `DEPLOY_UNIFIOS_HOST` set to the remote host's URL**: a fresh instance with zero certificates present. This tested the full create-then-activate path, end to end. It also showed that the `DEPLOY_UNIFIOS_HOST` override works for a remote target, not only localhost.

Both code paths got real test coverage between these two runs. I also compared the deployed certificate's serial number directly against the local acme.sh-issued certificate. I read the API's own `serial_number` field for both: the two matched exactly. I inspected the certificate that UniFi OS Server served. I used a browser TLS-inspection extension for this. The full chain was present and trusted.

I re-tested end to end on real hardware (macOS) after the first round of review fixes. This time, I removed the existing certificate first. This run tested the fresh create-then-activate path with the updated code, not only the duplicate-reuse branch.

After the second round of review, I re-tested again, this time against real UniFi OS Server API calls made one at a time. This tested specific claims directly against the real server, not only from code review:

- When the hook activates one entry, the server deactivates whichever other entry was active before, regardless of its name or domain.
- The `name` field has no effect on what the server serves. I uploaded a certificate under an unrelated name. The server served it correctly.
- `name` and `fingerprint` are each unique on their own: the server rejects a new entry that matches an existing entry on either field alone.

Based on these findings, I redesigned the removal step. Each upload now gets a name unique to that run: the domain plus a timestamp. The hook removes old entries only after a successful activation. It only removes entries whose name starts with the domain. I tested this against a mix of old and unrelated entries:

- 2 old entries that this hook created for the test domain: the hook removed both.
- 1 unrelated entry with a different, manual name: the hook did not touch it.
- 1 entry for a different domain that starts with the same text as the test domain. The hook did not touch this entry either. The match requires a space after the domain name, so it correctly told the two domains apart.
- The removal step also runs, and works, after a duplicate-reuse deploy (HTTP 400), not only after a new upload (HTTP 201).
- A duplicate that matches an inactive entry, with no other entry active, correctly activates that entry.
- A duplicate that matches an inactive entry, while a different entry is active, also correctly activates that entry.

Last, I ran a full production deploy of my own real domain through all seven of my personal deploy hooks:

    --deploy -d example.com --deploy-hook caddy_remote,localcopy,synology_dsm,unifios,haos,edgerouter,readynas --ecc

The UniFi OS Server UI showed that the certificate was live.

## Docs

I added an entry to the [deployhooks wiki page](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#45-deploy-the-certificate-to-unifi-os-certificate-rest-api).

Please report bugs to https://github.com/acmesh-official/acme.sh/issues/7182.

### DIFF
--- a/deploy/unifios.sh
+++ b/deploy/unifios.sh
@@ -70,7 +70,11 @@
 # Please report bugs to https://github.com/acmesh-official/acme.sh/issues/7182
 
 _uos_response_code() {
-  _egrep_o <"$HTTP_HEADER" "^HTTP[^ ]* .*$" | cut -d " " -f 2-100 | tr -d "\f\n" | _egrep_o "^[0-9]*"
+  # tr strips the trailing newline along with form feeds; re-terminate
+  # before the second _egrep_o, whose sed fallback (used wherever egrep -o
+  # is unavailable) drops an unterminated final line on some platforms.
+  _uos_code="$(_egrep_o <"$HTTP_HEADER" "^HTTP[^ ]* .*$" | cut -d " " -f 2-100 | tr -d "\f\n")"
+  printf '%s\n' "$_uos_code" | _egrep_o "^[0-9][0-9]*"
 }
 
 _uos_response_cookie() {
@@ -159,35 +163,21 @@ unifios_deploy() {
   _H2="x-csrf-token: $_uos_csrf"
   export _H2
 
-  _info "Checking for an existing '$_cdomain' certificate entry..."
-  _list_json="$(_get "$DEPLOY_UNIFIOS_HOST/api/userCertificates")"
-  _list_code="$(_uos_response_code)"
-  if [ "$_list_code" != "200" ]; then
-    _err "Failed to list existing certificates (HTTP $_list_code)."
-    _err "Response: $_list_json"
-    return 1
-  fi
-  # _normalizeJson collapses the response to one predictable line (no stray
-  # whitespace around colons, no embedded CR/LF the server might emit), then
-  # a literal embedded newline (not the two-character "\n", which GNU sed
-  # treats as a newline in the replacement but POSIX doesn't define and BSD
-  # sed emits literally) splits it one JSON object per line so grep can
-  # match a single certificate entry at a time.
-  _list_json="$(
-    echo "$_list_json" | _normalizeJson | sed 's/},{/},\
-{/g'
-  )"
-  # -F (fixed string): $_cdomain is interpolated as a literal string, not a
-  # regex. A wildcard cert's name is *.example.com -- as an unescaped BRE,
-  # the leading * would quantify the preceding character instead of
-  # matching a literal asterisk.
-  _old_ids="$(echo "$_list_json" | grep -F "\"name\":\"$_cdomain\"" | _egrep_o '"id":"[^"]*"' | cut -d '"' -f 4)"
-  _debug _old_ids "$_old_ids"
-
   _info "Uploading new certificate..."
+  # "name" is a purely cosmetic label -- the server never validates it
+  # against the certificate's actual CN/SAN, and accepts arbitrary text
+  # including spaces (confirmed: a cert for example.com served correctly
+  # after being uploaded under the unrelated name "totally unrelated label").
+  # The only constraint that matters here is uniqueness: the server rejects
+  # a second entry with a name it already has, so a bare domain name would
+  # collide with the previous deploy's entry on every renewal after the
+  # first. Appending a human-readable timestamp keeps each upload's name
+  # unique while making it obvious which entry is current when browsing the
+  # UniFi OS Server UI.
+  _uos_name="$_cdomain $(date '+%Y-%m-%d %H:%M:%S')"
   _uos_key_json="$(_json_encode <"$_ckey")"
   _uos_cert_json="$(_json_encode <"$_cfullchain")"
-  _create_body="{\"name\":\"$_cdomain\",\"key\":\"$_uos_key_json\",\"cert\":\"$_uos_cert_json\"}"
+  _create_body="{\"name\":\"$_uos_name\",\"key\":\"$_uos_key_json\",\"cert\":\"$_uos_cert_json\"}"
 
   _create_json="$(_post "$_create_body" "$DEPLOY_UNIFIOS_HOST/api/userCertificates" "" "POST" "application/json")"
   _create_code="$(_uos_response_code)"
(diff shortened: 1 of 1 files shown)


neilpang:
1. unifios.sh:156 -- sed 's/},{/}\n{/g' is a GNU extension. POSIX does
   not define \n in the s/// replacement; BSD (macOS/FreeBSD) and
   Solaris sed emit a literal n, so nothing splits. grep then matches
   the whole single-line array and _egrep_o '"id":"[^"]*"' returns
   EVERY id in the list, so the cleanup loop at :201 deletes every
   certificate on the server, not just this domain's.

     input:        [{"id":"a1","name":"foo.com"},
                    {"id":"b2","name":"example.com"},
                    {"id":"c3","name":"bar.com"}]   (one line)
     split works:  b2
     split fails:  a1 b2 c3

   The macOS run passed because that instance had a single matching
   entry, where the superset is indistinguishable from the correct
   result. Use the portable literal-newline form already used in
   dnsapi/dns_cpanel_uapi.sh:216 and dnsapi/dns_glesys.sh:142, and
   pipe the response through _normalizeJson first.

2. unifios.sh:156 -- $_cdomain goes into a grep BRE unescaped. A
   wildcard cert deploys as *.example.com, so in "name":"*.example.com"
   the * quantifies the preceding quote and every . matches any
   character. That over-broad match feeds the delete loop. Use grep -F.

3. unifios.sh:77 -- \W is a GNU grep extension, undefined in POSIX BRE
   and unsupported by Solaris /usr/bin/grep, where the lookup silently
   returns nothing and the hook aborts with "Login succeeded but no
   TOKEN cookie was returned". \W is not needed: the cookie name is
   always first after Set-Cookie:, so grep -i "^Set-Cookie: *TOKEN="
   is both portable and stricter. deploy/ruckus.sh:164 has a portable
   _response_header/_response_cookie pair worth copying
   (_uos_response_code already came from there).

4. Requiring the user to export HTTPS_INSECURE=1 turns off TLS
   verification for the whole acme.sh run, including the connection to
   the ACME CA. _deploy sources the hook in a subshell (acme.sh:6681),
   so setting it inside the hook -- as deploy/proxmoxve.sh:128 and
   deploy/fritzbox.sh:72 do -- keeps the scope to the deploy step.

5. unifios.sh:173-181 -- the duplicate branch reuses the first
   name-matching entry, not the one whose fingerprint collided. With
   two stale entries for the domain it activates the older cert and
   deletes the newer one. Match on fingerprint/serial, or re-list
   after the 400.

6. unifios.sh:155 -- the list GET response code is never checked. An
   expired session or an error response yields an empty _old_ids and
   the hook carries on silently.

7. unifios.sh:133 -- pass "base64" to _savedeployconf for the password.
   _save_conf writes KEY='value' (acme.sh:2560), so a password
   containing a single quote corrupts the domain conf. See
   deploy/synology_dsm.sh:325.

8. Issue #7182 is closed -- the "Please report bugs to" link has to
   point at an open issue. The deployhooks wiki entry needs to be in
   place before merge, not after.

neilpang:
Points 1-4 and 6-8 check out; #7182 is open and the wiki section is in
place. Three left, numbering continued from the previous comment.

9. The fingerprint match in the duplicate branch (unifios.sh:208-209)
   is untested, and the PR's own testing note says as much: the re-run
   cleared the existing certificate first, so it exercised the create
   path -- the one branch that changed is the one no longer covered.
   Two assumptions behind it are unverified:

     - that GET /api/userCertificates returns a "fingerprint" field at
       all (the PR text only ever cites serial_number);
     - that its format matches what :208 produces. openssl prints
       "sha256 Fingerprint=AB:CD:...", so after cut -d '=' -f 2 the
       value is uppercase hex WITH colons, whereas JSON APIs usually
       return lowercase hex without them.

   Under grep -F either mismatch turns a path that used to recover
   into a hard return 1, which is worse than the _head_n 1 it
   replaced. Please post a redacted list response showing the field
   and its format, plus one run that actually hits the 400.

   Also at :208, use the core helper -- _fingerprint "$_cfullchain"
   sha256 (acme.sh:1071) -- instead of calling openssl directly; it
   already strips the colons. And guard an empty _uos_fingerprint
   explicitly: 2>/dev/null swallows the openssl error today, and the
   failure surfaces as "no existing entry matching this domain and
   fingerprint" instead.

10. unifios.sh:176-179 -- _normalizeJson ends in tr -d "\r\n", so the
    sed that follows receives input with no trailing newline. Solaris
    /usr/bin/sed drops an unterminated final line, so _list_json comes
    back empty there: _old_ids is always empty (superseded entries are
    never removed) and the duplicate branch always errors out.
    Re-terminate before the split, keeping the split sed as it is:

      _list_json="$(echo "$_list_json" | _normalizeJson)"
      _list_json="$(printf '%s\n' "$_list_json" | sed ...)"

    Same shape at :73: tr -d "\f\n" | _egrep_o feeds an unterminated
    line into the sed fallback that _egrep_o uses wherever egrep -o is
    unavailable, and the status code comes back empty.

11. Nit, unifios.sh:73 -- ^[0-9]* can match the empty string, the
    pattern shape that hangs OmniOS's egrep. Use ^[0-9][0-9]*.
    (Inherited from deploy/ruckus.sh:161, but no need to carry it
    over.)

Pablo1:
Thanks for the detailed review. Points 9-11 are fixed. Testing them also found one more real problem, described below.

## 9. Fingerprint matching

I checked a real API response. The fingerprint field is SHA-1 (20 groups separated by colons), not SHA-256:

```
"fingerprint":"FC:02:50:9C:3B:3F:B7:79:9D:CA:4D:7C:AC:92:E7:D5:EA:F1:3A:29"
```

The code now uses the core helper `_fingerprint "$_cfullchain" sha1`. It also checks for an empty result before comparing.

I re-deployed the same certificate twice to trigger the duplicate path for real:

```
[...] Uploading new certificate...
[...] Certificate already present as entry 386ead06-080f-44f3-a185-6ad5af2fb82d; reusing it.
[...] Activating certificate 386ead06-080f-44f3-a185-6ad5af2fb82d...
[...] UniFi OS Server certificate deployed and activated successfully.
```

## 10 & 11. Unterminated-line handling, `^[0-9][0-9]*`

Both are fixed as you described. I could not test the original Solaris/OmniOS bug directly. `egrep -o` works fine on my hardware, so `_egrep_o`'s sed fallback never runs here. I also rebuilt `sed` from the Heirloom Toolchest (a portable rebuild of the original Sun source) to see if the bug was in the algorithm itself. It did not reproduce the bug. My guess is the issue is specific to Solaris's compiled `sed` or libc, not something I can test from here. Point 11's regex has still been exercised many times, though -- every real API call in this testing round depended on it to read the status code correctly.

## Also found: the cleanup logic could delete the wrong certificate

While testing point 9, I re-read the delete loop (was at :253-264) closely. It deleted every entry matching the domain name, not just the previous one. Worse: the duplicate-match code didn't check whether the matched entry was active. So a duplicate matching an old, inactive entry would reactivate that old cert, while the delete loop removed the certificate that was actually active. That's a real bug, not just an untested code path.

Instead of patching the matching logic, I tested the real API directly. Two facts changed the fix:

- `name` and `fingerprint` must each be unique across all certificate entries. The server rejects a new entry if either one matches an existing entry. `name` has no effect on what's served -- I uploaded a certificate under an unrelated name and it was served correctly.
- Activating one certificate always deactivates whichever one was active before, regardless of its name or domain.

Given that, the fix is simpler than patching the old logic: make every uploaded name unique (`domain + unix timestamp`). That way a new upload never collides with the previous deploy's entry, so there's nothing to delete afterward. Old entries are left in place. The UI already lets you manage them directly, and the hook's correctness doesn't depend on removing them.

There's precedent for this in the repo already: `deploy/truenas.sh:112` (and `deploy/truenas_ws.sh:267`) builds a timestamp-suffixed name for the same reason, uploading to a certificate store with the same "list of named entries" shape:

```sh
_certname="Letsencrypt_$(_utc_date | tr ' ' '_' | tr -d -- ':')"
```

I tested this against a real UniFi OS Server instance:
- a clean upload and activate
- a duplicate upload reused correctly, whether the matching entry was active or inactive
- a real renewal (new certificate content, same domain) -- creates a new entry, leaves the old one inactive
- the real `--deploy --deploy-hook unifios` command, not just the function called directly
- reusing saved credentials with no environment variables set, reading from the saved (base64-encoded) deploy config
- a full production deploy of my own real domain through all seven of my deploy hooks (`--deploy -d example.com --deploy-hook caddy_remote,localcopy,synology_dsm,unifios,haos,edgerouter,readynas --ecc`), confirmed live in the UniFi OS Server UI

neilpang:
9-11 confirmed fixed. Three on the rework.

12. Dropping the cleanup entirely goes further than the bug required,
    and the cited precedent does not support it: deploy/truenas.sh:112
    does build a timestamped name, but it also deletes the superseded
    certificate at truenas.sh:257-258. The pattern there is unique name
    AND cleanup, not unique name INSTEAD OF cleanup.

    On the default renewal cycle this leaves roughly six new entries a
    year, forever, each holding a private key in the server's Postgres
    store -- and by your own finding that the name column is
    fixed-width and does not wrap, the list gets less usable as it
    grows.

    The original bug was which entries were selected for deletion, not
    that deletion happened at all. Now that every name carries a
    hook-owned prefix, a scoped cleanup is safe: after a successful
    activation, delete the entries whose name begins with the domain
    plus a space and whose id is not _new_id. A fixed-string match on

      "name":"$_cdomain

    followed by a space only ever hits entries this hook created for
    this domain, so it cannot touch a manually uploaded or self-signed
    one. Keep delete failures non-fatal, as they were.

13. The comments no longer match the code. unifios.sh:51-57 still
    describes the removed design -- "looks up any existing certificate
    row named after the domain", "before removing the old one", "a
    stale duplicate entry is left behind for the next run". Same in
    wiki section 45 ("then remove any superseded entry for the
    domain") and in the PR body's How it works.

14. unifios.sh:179 -- use the core wrapper _time (acme.sh:1914)
    instead of date +%s. truenas.sh:112 uses _utc_date for the same
    purpose.

Pablo1:
Thank you. Point 12 is correct. This version removes old entries again. The removal is scoped, as you described.

## 12. The hook removes old entries again

You are correct. The earlier fix removed too much: it deleted the whole step that removes old entries. The bug did not require this.

The example I gave, `truenas.sh`, does not support that choice. It builds a unique name, and it also deletes the old entry. I copied only the first part of this pattern.

The new fix matches your description. After the hook activates a certificate, it removes every entry whose name starts with `"$_cdomain "`. This is a fixed-string match, not a regular expression. The hook does not remove the entry that it just activated. Every name that this hook creates has this exact form. As a result, this step can only remove entries that this hook created for this domain. It cannot affect a certificate that a user uploaded by hand. It cannot affect a self-signed certificate.

I tested this against a real UniFi OS Server. The test data had a mix of old and unrelated entries:

- 2 old entries that this hook created for the test domain. The hook removed both of them.
- 1 unrelated entry with a different, manual name. The hook did not touch it.
- 1 entry for a different domain that starts with the same text as the test domain. The trailing space in the match prevented a mistake here. The hook did not touch it.
- The removal step also runs, and works, after a duplicate-reuse deploy (HTTP 400), not only after a new upload (HTTP 201).
- 2 more scenarios: a duplicate that matches an inactive entry, with no other entry active. A second case: a duplicate that matches an inactive entry, while a different entry is active. In both cases, the hook correctly activates the matching entry.

## 13. Comments and docs updated

The file header, the wiki page, and the PR body now describe the current design. Each upload gets a unique name. The removal step only affects entries that this hook created. These texts no longer describe the old design.

## 14. `_time` instead of `date +%s`

This is done. It matches `_utc_date` in `truenas.sh`, which has the same purpose.

## Also tested

I also ran a full production deploy through all of my other deploy hooks:

    --deploy -d example.com --deploy-hook caddy_remote,localcopy,synology_dsm,unifios,haos,edgerouter,readynas --ecc

The UniFi OS Server UI showed that the certificate was live.

neilpang:
Merged.

Two non-blocking notes, if the hook gets touched again:

- An RSA and an ECC deploy of the same domain share the "$_cdomain "
  name prefix, so each run removes the other's entry. Harmless as
  things stand, since activation is exclusive server-wide and only
  one could ever be served -- but worth knowing if that changes.
- unifios.sh:234-242 and :285-289 are the same _normalizeJson +
  split block twice; worth folding into a _uos_split_json helper.

Bug reports go to #7182.

Tharit:
@neilpang @Pablo1 Great addition, thanks for getting this merged! 

However, I think its documented misleading as is. This actually works on Unifi UDM hardware as well - it's not about self hosted or hardware, I think its just about the UnifiOS version. 

I just used this hook successfully to deploy a certificate to a UDM Pro running UnifiOS 5.1.26. 

So, this actually implements whats discussed in the pending #6916 

Suggest to update the docs accordingly?

neilpang:
Confirmed, and the docs will be reframed. The endpoint set is UniFi
OS's own, not something specific to the self-hosted Server: the
original report in #6916 drives the same /api/auth/login ->
/api/userCertificates -> .../status flow against a UCG Fiber on
UniFi OS 5.0.16.

The real split is not product line but access method. deploy/unifi.sh
is file/keystore based and needs local or SSH access on the device;
deploy/unifios.sh talks to the REST API and works remotely. Both stay
-- the docs will say which one to pick.

One thing the docs miss that matters on hardware: the
https://localhost:11443 default is UniFi OS Server's management port.
On a UDM/UCG, set DEPLOY_UNIFIOS_HOST to https://<host> (443).

No version floor will be documented: 5.0.16 and 5.1.26 are the only
data points so far. If anyone has this working on an earlier UniFi
OS, say so here and it goes in.

#6916 is resolved by this hook and will be closed pointing at #7184.

Pablo1:
Thank you, @neilpang, for the wiki update and for the header comment update in unifios.sh.

@Tharit, thank you for the real hardware test and for the report about the documentation. Your report started this update.

I opened #7200 for the other two non-blocking notes: the new `_uos_split_json` helper, and a comment about the RSA/ECC name collision. That PR also fixes a wiki link in the body of this PR. The link pointed at the old wiki anchor, from before the wiki update.

PR #7081, opened 2026-07-04
dns_desec: fix advertised token variable name

Forst:
This must've been a copy-paste error from `dns_ddnss`.

Fixes: 6b7b5caf54ea ("DNS provider API: structured description")

PR #7192, opened 2026-08-12
chore: give the deSEC token a meaningful name

TheOneWithTheBraid:
The deSEC.io API implementation right now uses the option `DEDYN_TOKEN` to store the access token for their API.

I spent hours trying to figure out why it's called **DEDYN** rather than *DESEC* and couldn't find an answer but "since it always used to be called like this".

I'd propose to replace this confusing name with the more comprehensive name `DESEC_TOKEN`.

This PR renames the configuration option, adds a migration and prints a notice about the deprecation to the `_info` output.

Notably, OpenWrt currently ships a broken version of this API anyway since https://github.com/acmesh-official/acme.sh/pull/7081 is not included in their version of acmesh yet. This took ages to figure out.

Once this PR is merged, the Wiki needs to be updated with the correct configuration option : https://github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_desec

### DIFF
--- a/dnsapi/dns_desec.sh
+++ b/dnsapi/dns_desec.sh
@@ -4,7 +4,7 @@ dns_desec_info='deSEC.io
 Site: desec.readthedocs.io/en/latest/
 Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_desec
 Options:
- DEDYN_TOKEN API Token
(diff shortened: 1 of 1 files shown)


neilpang:
NO,  you can not just CHANGE the name.  there could be existing users.

PR #7199, opened 2026-08-17
fix(dns_easydns): add credential check on rm and fix wildcard record deletion

wurzelpanzer:
## Description
This PR fixes issues in the `dns_easydns` plugin:
- Added proper credential/token checks during the record removal (`_rm`) phase.
- Fixed an issue with wildcard record deletion.

## Testing
Tested successfully across the main test platforms (Ubuntu, Debian, Windows, macOS, FreeBSD, etc.). Haiku failed.

### DIFF
--- a/dnsapi/dns_easydns.sh
+++ b/dnsapi/dns_easydns.sh
@@ -75,6 +75,11 @@ dns_easydns_rm() {
   EASYDNS_Token="${EASYDNS_Token:-$(_readaccountconf_mutable EASYDNS_Token)}"
   EASYDNS_Key="${EASYDNS_Key:-$(_readaccountconf_mutable EASYDNS_Key)}"
 
+  if [ -z "$EASYDNS_Token" ] || [ -z "$EASYDNS_Key" ]; then
+    _err "You didn't specify an easydns.net token or api key. Signup at https://cp.easydns.com/manage/security/api/signup.php"
+    return 1
+  fi
+
   _debug "First detect the root zone"
   if ! _get_root "$fulldomain"; then
     _err "invalid domain"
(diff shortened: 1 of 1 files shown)


wurzelpanzer:

Done

neilpang:
(1) The multi-id loop only works where egrep -o exists. _egrep_o
    (acme.sh L958) otherwise falls back to
    sed -n 's/.*\(PAT\).*/\1/p', which on a single-line JSON returns
    exactly one match -- the last one, since .* is greedy. With a
    two-record response the egrep branch yields 12345678 12345679 and
    the sed branch yields 12345679 only. So on Solaris and similar the
    wildcard fix silently degrades to a single delete, and not the
    record the old head -n 1 picked.

(2) _rm still ignores $txtvalue and now deletes every TXT under that
    host. The convention is to match name + value and delete that id
    only (dns_cf.sh queries with ...&content=$txtvalue). Deleting all
    of them removes a concurrent run's challenge record when
    example.com and *.example.com are separate certificates.

Both points go away if the record is selected by its rdata:

    record_id=$(printf "%s\n" "$response" | tr '{' '\n' | grep "\"rdata\":\"$txtvalue\"" | _egrep_o "\"id\":\"[^\"]*\"" | cut -d : -f 2 | tr -d \")

tr '{' '\n' puts one record per line, so both _egrep_o branches
return the same single id. Treat an empty record_id as "nothing to
remove", not an error. Do not put \{ or \} in an _egrep_o pattern:
the sed fallback is BRE, where \{ opens an interval and sed aborts
with "Invalid preceding regular expression".

(3) Please restore the comments the diff drops -- the API doc URL, the
    _get_root return-value block, the section separators. They are
    unrelated to both fixes.

The credential check in _rm is correct as is. The Haiku failure is
infrastructure (pkgman install cronie -> "Failed to download package
run_parts: Resource not found"), not your change.

wurzelpanzer:
Force-pushed. The branch was based on master instead of dev, which
pulled unrelated merge commits into the diff. It is now a single commit
on top of upstream/dev.

All three review points are addressed:

(1) The multi-id loop is gone. tr '{' '\n' + grep on rdata leaves
    exactly one "id":"..." per line, so the egrep -o branch and the
    sed fallback agree. No { or } in the _egrep_o pattern.
(2) Selection is by name + value; an empty record_id returns 0 with
    _info, not an error.
(3) Restored - the API doc URL, the _get_root return-value block and
    both section separators are untouched.

PR #7200, opened 2026-08-17
unifios: extract JSON-split helper, document RSA/ECC name collision

Pablo1:
## Summary

This is a follow-up to #7184. It addresses two non-blocking notes from the merge review.

`_uos_split_json()` is a new helper function. It replaces two copies of the same code: the `_normalizeJson` step, then the split into one JSON object per line. Before this change, `unifios.sh` had this code at two places. Now it has one function, and each place calls the function.

This commit also adds a comment about a real limitation. It does not fix the limitation. An RSA deploy and an ECC deploy of the same domain share the same name prefix. As a result, the cleanup step for each deploy removes the entry that the other deploy created. Today this causes no harm, because activation is exclusive across the whole server. Only one of the two certificates can ever be active.

`deploy/haproxy.sh` and `deploy/lighttpd.sh` have the same problem for their own named certificate storage. They solve it with a `.rsa` / `.ecdsa` suffix through `_isEccKey`. If that becomes necessary later, the same method can fix this hook.

## Testing

I tested this against a real UniFi OS Server instance (nexus). Three real deploys, all through the actual `unifios_deploy()` function, cover both places that call the new helper:

1. A fresh certificate, for a new domain: the server accepted it (HTTP 201) and activated it. The cleanup step found no old entries for this domain, as expected.
2. The identical certificate again: the server rejected it as a duplicate (HTTP 400). The hook found and reused the existing entry by its fingerprint. Cleanup again found nothing to remove.
3. A different certificate, for the same domain: the server accepted it (HTTP 201). The cleanup step then found and removed the entry from step 1, through the output that the new helper produced.

Step 3 was not covered by earlier testing on #7184. It is the one case where the cleanup step actually performs a delete, not only a check.

Bugs go to #7182.

### DIFF
--- a/deploy/unifios.sh
+++ b/deploy/unifios.sh
@@ -95,6 +95,25 @@ _uos_response_cookie() {
   grep <"$HTTP_HEADER" -i "^Set-Cookie: *$1=" | _tail_n 1 | _egrep_o "$1=[^;]*" | _head_n 1
 }
 
+_uos_split_json() {
+  # $1 = raw JSON list response
+  #
+  # _normalizeJson collapses the response to one line. This removes extra
+  # space around colons. It also removes any CR or LF characters that the
+  # server can add. However, _normalizeJson also removes the newline
+  # character at the end of the line. If the line has no ending newline
+  # character, some sed programs drop the last line of input. This code
+  # adds the newline back before the split below, to prevent that problem.
+  _uos_normalized="$(echo "$1" | _normalizeJson)"
+  # A literal newline character splits the JSON into one object per line.
+  # Grep can then match a single certificate entry at a time. This is not
+  # the two-character "\n" sequence: GNU sed reads "\n" in the replacement
+  # text as a newline character. POSIX does not define this behavior, and
+  # BSD sed prints "\n" as two literal characters, not as a newline.
+  printf '%s\n' "$_uos_normalized" | sed 's/},{/},\
+{/g'
+}
+
 unifios_deploy() {
   _cdomain="$1"
   _ckey="$2"
@@ -189,6 +208,18 @@ unifios_deploy() {
   # wrap (confirmed against the real UI: a long name overlaps the Expires
   # column and makes both unreadable), so keep the suffix short instead --
   # Unix epoch seconds are still unique enough for this purpose.
+  #
+  # This name does not include the key type. As a result, an RSA
+  # deploy and an ECC deploy of the same domain share this prefix.
+  # Because of this, the cleanup step below removes the entry that the
+  # other deploy created. Today, this causes no harm, because activation is
+  # exclusive across the whole server. Only one of the two certificates
+  # can ever be active. If that changes, this needs a fix.
+  #
+  # `deploy/haproxy.sh` and `deploy/lighttpd.sh` have the same problem for their
+  # own named certificate storage. They solve it with a ".rsa" or
+  # ".ecdsa" suffix. They pick the suffix with `_isEccKey "${Le_Keylength}"`. The
+  # same method can work here.
   _uos_name="$_cdomain $(_time)"
   _uos_key_json="$(_json_encode <"$_ckey")"
   _uos_cert_json="$(_json_encode <"$_cfullchain")"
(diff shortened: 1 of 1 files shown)


neilpang:
Both grep -F in this file need to go, and this PR is the right place --
it already rewrites that exact block. Solaris /usr/bin/grep has no -F:
"grep: illegal option -- F". Its entire option set is -b -c -h -i -l
-n -q -s -v -w.

Just dropping the -F is not the fix. Verified under dash:

  - _cdomain is "*.example.com" for a wildcard cert, so the pattern
    "name":"*.example.com  reads "* as zero-or-more quotes and matches
    nothing at all -- cleanup silently stops removing anything and old
    entries pile up forever.
  - . matches any character, so "name":"x.example.com  also matches a
    name like xYexample.com and deletes another domain's entry.

case is literal by construction, same shape as your new helper:

    _uos_grep_literal() {
      # $1 = literal needle; filters stdin to lines containing it.
      while IFS= read -r _uos_line || [ -n "$_uos_line" ]; do
        case "$_uos_line" in
        *"$1"*) echo "$_uos_line" ;;
        esac
      done
    }

Use it at both call sites. The fingerprint needle is hex and colons
only, so plain grep would be safe there, but one helper for both is
simpler.

Second point: the RSA/ECC note is accurate, but it is 10 lines for
something that causes no harm, in a file that is already mostly
comments. Cut it to two lines -- or actually fix it, by putting the
_isEccKey suffix into _uos_name and matching that suffix in the
cleanup pattern.

The extraction itself is fine, behavior is identical.

Pablo1:
Both fixed.

`_uos_grep_literal` replaces both `grep -F` calls with a case-based literal match. This works without `-F`, and it does not read `*` or `.` as a regex.

`_uos_name` now includes the key type (`rsa`/`ecdsa`, via `_isEccKey`). The cleanup match uses it too.

I tested both changes against a real UniFi OS Server. I deployed an RSA certificate and an ECC certificate for the same domain, then redeployed each with new content. Each cleanup step removed only the old entry with the same key type. It never removed the entry for the other key type. I also tested an ECC duplicate deploy: it correctly reused and reactivated the existing entry.

The `_uos_split_json` extraction needed no changes.

neilpang:
Drop these three lines:

    if [ -z "${Le_Keylength}" ]; then
      Le_Keylength=""
    fi

Assigning empty to something already empty is a no-op, and as a nounset
guard it would not work either -- [ -z "${Le_Keylength}" ] trips set -u
by itself, that needs ${Le_Keylength:-}. It is also unnecessary:
_isEccKey returns 1 on an empty argument, which is why haproxy.sh and
lighttpd.sh call it bare. And a hook should not assign to a core Le_*
variable.

In _uos_grep_literal, use printf '%s\n' "$_uos_line" rather than echo
"$_uos_line" -- my snippet was wrong there. The list response carries
\n escapes inside the cert field and dash's echo expands them, so one
JSON object comes out as five lines:

    {"id":"a1","name":"example.com rsa 1700000000","cert":"-----BEGIN CERTIFICATE-----
    MIIC
    QQ==
    -----END CERTIFICATE-----
    ","fingerprint":"AA:BB"}

grep -F passed those bytes through untouched. Extraction still works
because "id" is a separate token, but that is luck rather than design --
_uos_split_json and _uos_response_code already use printf '%s\n' for
exactly this reason.

Last one needs no action, noting it so it does not come back as a
separate PR: the rename orphans entries created by the version now on
master, because "name":"example.com 1755..." no longer matches
"name":"example.com rsa ". Those entries are inactive and harmless, the
hook has only been on master since 2026-08-13 and is in no release yet,
so a migration is not worth the code.

Pablo1:
Both fixed.

`_uos_grep_literal` now uses `printf '%s\n'` instead of `echo`, matching `_uos_split_json`'s existing pattern. I wrote a failing test first, confirmed it failed the same way you described, then fixed it and confirmed it passed.

Dropped the `Le_Keylength` guard. Without it, shellcheck flags `Le_Keylength` as referenced but never assigned (SC2154) -- it's only ever set by acme.sh core, never by this hook. Suppressed that warning inline instead of assigning to a core `Le_*` variable just to quiet it.

I tested both changes against a real UniFi OS Server: a normal create-and-duplicate-reuse cycle, and a dedicated test where a new certificate for the same domain correctly found and removed the old entry through the fixed match.

No action on the orphaned-entries note.

PR #7205, opened 2026-08-21
Truenas 26 deploy fixes

BillWeiss:
I'd love some input on this as I hack on it. I _think_ I probably need to detect the version and branch a bit since there are some changes coming, but this currently applies to my box. I don't have an easy way to test against older versions but I can come up with something.

### DIFF
--- a/deploy/truenas_ws.sh
+++ b/deploy/truenas_ws.sh
@@ -43,16 +43,18 @@ _ws_call() {
   _debug "_ws_call arg1" "$1"
   _debug "_ws_call arg2" "$2"
   _debug "_ws_call arg3" "$3"
-  _ws_response=$(midclt \
-    --uri "$_ws_uri" \
-    -K "$DEPLOY_TRUENAS_APIKEY" \
-    --plain \
-    call "$@")
-  _ws_ret=$?
+
+  # TrueNAS 26.0.0-BETA3 and later need a --plain option for midclt call, detect if it is available
+  _midclt_plain=""
+  case "$(midclt --help 2>/dev/null)" in
+    *--plain*) _midclt_plain="--plain" ;;
+  esac
+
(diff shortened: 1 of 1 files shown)


BillWeiss:
(Force pushed because I started from `master` instead of `dev` and I wanted to clean that up)

neilpang:
The --plain flag only exists since 26.0.0-BETA.3 (checked
truenas/api_client tags), all older versions error with
"unrecognized arguments: --plain", so every _ws_call fails.

Don't branch on _truenas_version, the version itself comes from a
_ws_call. Just probe the flag:

    _midclt_plain=""
    case "$(midclt --help 2>/dev/null)" in
    *--plain*) _midclt_plain="--plain" ;;
    esac

    # shellcheck disable=SC2086
    _ws_response=$(midclt --uri "$_ws_uri" -K "$DEPLOY_TRUENAS_APIKEY" $_midclt_plain call "$@")

Same for jq: on 25.10 and older, ui_certificate is still an object.
Make it work for both:

    _ui_certificate_id=$(printf "%s" "$_ws_response" | jq -r '.ui_certificate | if type == "object" then .id else . end')
    _ui_certificate_name=$(printf "%s" "$_ws_response" | jq -r 'if (.ui_certificate | type) == "object" then .ui_certificate.name else .ui_certificate_name end')

Same for _changed_certid.

BillWeiss:
@neilpang thanks!

neilpang:
shfmt: case patterns go at the same indent as "case":

    case "$(midclt --help 2>/dev/null)" in
    *--plain*) _midclt_plain="--plain" ;;
    esac

or just run: shfmt -w -i 2 deploy/truenas_ws.sh

BillWeiss:
Made the shfmt fix (thanks, I should have looked at the checks), I think this is good to go? I notice that shellcheck has a fair number of complaints about other code but not mine, so I hope it's good to go? Let me know if you'd like me to spin up an older TrueNAS instance to double check, but this seems pretty safe with your suggestions.

PR #7211, opened 2026-08-26
Add DNS API for FENO (feno.no)

MrErikCodes:
Adds `dnsapi/dns_feno.sh` for FENO (https://feno.no), a Norwegian `.no` registrar with a public REST API.

**API docs:** https://github.com/mrerikcodes/feno-api/blob/main/docs/README.md
**Issues:** https://github.com/mrerikcodes/feno-api/issues

### Options

| Variable | Required | Description |
| --- | --- | --- |
| `FENO_API_KEY` | yes | Public API key (`feno_live_...`) with the `acme:write` scope, created in the FENO dashboard |
| `FENO_API_BASE` | no | API base URL, default `https://api.feno.no` |

Credentials are read/saved with `_readaccountconf_mutable` / `_saveaccountconf_mutable`.

### Implementation notes

- `_get_root` walks the labels of `fulldomain` and probes `GET /v1/domains/{candidate}` until the API answers `success:true`; `404 DOMAIN_NOT_MANAGED` means "try the parent". No label-count guessing (`.no` has both plain second-level names and delegated sub-zones).
- Auth-layer refusals (`INVALID_API_KEY`, `INSUFFICIENT_SCOPE`, `RATE_LIMITED`, ...) are detected explicitly so a bad key is not reported as "domain not found".
- `dns_feno_add` POSTs a TXT record with the zone-relative name (`_acme-challenge` / `_acme-challenge.www`), TTL 60. Records are **appended**, so wildcard + apex issuance (two TXT values at one name) works.
- `dns_feno_rm` lists the zone and deletes only the record matching name + type + **value**, so the sibling challenge record of a wildcard issuance is left alone. A record that is already gone is not an error.
- Only `_get` / `_post` with `_H1..._H3` headers; POSIX sh; string handling with `grep`/`cut`/`sed`/`tr` only (no `awk`, no `jq`).

### Checks

- `shellcheck -e SC2034,SC2154,SC2148 dnsapi/dns_feno.sh` — clean
- `shellcheck -e SC2181 -e SC2089 dnsapi/dns_feno.sh` (the exact CI invocation) — clean
- `shfmt -l -s -i 2 -d dnsapi/dns_feno.sh` and `shfmt -l -i 2 -d` (CI form) — no diff
- DNS API test workflow (`DNS.yml`) run on my fork: https://github.com/mrerikcodes/acme.sh/actions/runs/32855659984 — Docker (the issuance test: apex + wildcard on a real `.no` zone, cert issued and revoked), macOS, Windows, FreeBSD, GhostBSD, OpenBSD, NetBSD, DragonFlyBSD, MidnightBSD, Solaris, Omnios, OpenIndiana and Tribblix all green; the Haiku QEMU job timed out after ~2 h with no script output (VM killed as an orphan) and is being re-run

Note for the test workflow: the harness also adds a TXT at `acmetestXyzRandomName.<domain>`, which FENO's `acme:write` scope refuses by design (it can only touch `_acme-challenge*`). The fork run therefore uses a key with `acme:write` + `dns:read` + `dns:write` (`acme:write` for the zone lookup and the challenge records, `dns:*` for the harness's arbitrary-name TXT); real users need only `acme:write`, and the apex + wildcard issuance in the same run proves that path.

Wiki entry for `dnsapi2` is ready to add (text below / on merge).

Ref: https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide

### DIFF
--- a/dnsapi/dns_feno.sh
+++ b/dnsapi/dns_feno.sh
@@ -1,0 +1,262 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_feno_info='FENO
+Site: feno.no
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_feno
+Options:
+ FENO_API_KEY API key with the acme:write scope (feno_live_...)
+ FENO_API_BASE API base URL. Optional. Default: "https://api.feno.no".
+Issues: github.com/mrerikcodes/feno-api/issues
+Author: Erik Nilsen <eriknilsen02@hotmail.com>
+'
+
+# FENO (feno.no) is a Norwegian .no registrar. API documentation:
+# https://github.com/mrerikcodes/feno-api/blob/main/docs/README.md
+# The API key only needs the acme:write scope: it covers the zone lookup and
+# TXT records at _acme-challenge names.
+
+FENO_API_BASE_DEFAULT="https://api.feno.no"
+
+########  Public functions #####################
+
+# Usage: dns_feno_add  _acme-challenge.www.kunde.no  "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_feno_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  _info "Using FENO"
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _feno_credentials; then
+    return 1
+  fi
+
+  if ! _get_root "$fulldomain"; then
+    _err "Could not find a FENO domain covering $fulldomain."
+    _err "The domain must be registered with FENO and use FENO nameservers (ns1/ns2.feno.no)."
+    return 1
+  fi
+  _debug _domain "$_domain"
+  _debug _sub_domain "$_sub_domain"
+
+  _feno_json="{\"type\":\"TXT\",\"name\":\"$_sub_domain\",\"value\":\"$txtvalue\",\"ttl\":60}"
+  if ! _feno_rest POST "v1/domains/$_domain/dns" "$_feno_json"; then
+    return 1
+  fi
+
+  if ! _contains "$response" '"success":true'; then
+    _err "FENO refused the TXT record: $response"
+    return 1
+  fi
+
+  _info "TXT record added."
+  return 0
+}
+
+# Usage: dns_feno_rm  _acme-challenge.www.kunde.no  "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_feno_rm() {
+  fulldomain=$1
+  txtvalue=$2
+
+  _info "Using FENO"
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _feno_credentials; then
+    return 1
+  fi
+
+  if ! _get_root "$fulldomain"; then
+    _err "Could not find a FENO domain covering $fulldomain."
+    return 1
+  fi
+  _debug _domain "$_domain"
+  _debug _sub_domain "$_sub_domain"
+
+  # The API has no "delete by value" - the record id has to be looked up by listing the zone
+  # and matching name + type + value. Matching the value too is what keeps a wildcard
+  # certificate safe: kunde.no and *.kunde.no put TWO TXT records at the same name, and only
+  # the one this call created may be removed.
+  if ! _feno_rest GET "v1/domains/$_domain/dns"; then
+    return 1
+  fi
+
+  # Same guard as the zone walk: an auth failure here returns no records, which would
+  # otherwise be read as "the record is already gone" and reported as success.
+  if _feno_code=$(_feno_auth_error); then
+    _err "FENO rejected the API key while listing $_domain ($_feno_code)."
+    return 1
+  fi
+
+  _record_id=$(_feno_find_record_id "$_sub_domain" "$txtvalue")
+  if [ -z "$_record_id" ]; then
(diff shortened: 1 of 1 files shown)


MrErikCodes:
Wiki entry for `dnsapi2` (the wiki is not editable from outside the org — ready to paste on merge):

## <a name="dns_feno"/> Use FENO API

[FENO](https://feno.no) is a Norwegian `.no` registrar. Create a public API key in the FENO dashboard
(*API keys*) with the **`acme:write`** scope — that scope is enough for the plugin: it covers the zone
lookup and `TXT` records at `_acme-challenge` names, and nothing else. The domain must use FENO
nameservers (`ns1.feno.no` / `ns2.feno.no`).

```sh
export FENO_API_KEY="feno_live_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```

Optionally, to use a different API endpoint (default `https://api.feno.no`):

```sh
export FENO_API_BASE="https://api.feno.no"
```

Then issue a certificate:

```sh
acme.sh --issue --dns dns_feno -d example.no -d '*.example.no' --dnssleep 120
```

FENO's nameservers are anycast and the API returns before every edge has the record, so keep a
`--dnssleep` of at least 120 seconds (or let acme.sh's own DNS check run). Wildcard and apex
certificates work (two TXT values at the same name are supported).

The `FENO_API_KEY` and `FENO_API_BASE` values are saved in `~/.acme.sh/account.conf` and reused on renewal.

API docs: https://github.com/mrerikcodes/feno-api/blob/main/docs/README.md
Issues: https://github.com/mrerikcodes/feno-api/issues

MrErikCodes:
Fork DNS test run for this branch: https://github.com/mrerikcodes/acme.sh/actions/runs/32855659984

- Docker (real issuance: apex + wildcard on a `.no` zone, certificate issued and revoked), macOS, Windows, FreeBSD, GhostBSD, OpenBSD, NetBSD, DragonFlyBSD, MidnightBSD, Solaris, Omnios, OpenIndiana, Tribblix: **pass**.
- Haiku: fails in the VM bootstrap, before acme.sh runs — `pkgman install -y cronie` → `Refreshing repository "HaikuPorts" failed *** Failed to download package run_parts: Resource not found` (attempt 3); attempt 1 was the QEMU job timing out with no output. Nothing in `dns_feno.sh` is reached on that platform, so I have not re-run it further. Happy to re-run once HaikuPorts is healthy again.

neilpang:
1. `grep -F` / `grep -iF` in `_feno_find_record_id`: Solaris `/usr/bin/grep` has no `-F` ("illegal option -- F"), so the rm path breaks there. Match literally with a `case` loop instead, e.g.:

   ```sh
   printf "%s\n" "$response" | tr '{' '\n' | while IFS= read -r _fline; do
     case "$_fline" in
     *"\"value\":\"$_find_value\""*)
       case "$_fline" in
       *'"type":"TXT"'*)
         case "$_fline" in
         *"\"name\":\"$_find_name\""*) printf "%s\n" "$_fline" ;;
         esac
         ;;
       esac
       ;;
     esac
   done | _head_n 1
   ```

2. `_get_root`: if fulldomain equals the zone itself (DNS alias mode can point the challenge at the zone apex), the first-iteration match sets `_sub_domain` to the zone's first label, so the record is created at e.g. `kunde.kunde.no`. When the candidate at i=1 matches, set `_sub_domain=""` (the apex).

3. `Issues:` must point to a dedicated tracking issue in this repo. Please open one for FENO and reference it in the info block; the feno-api tracker can be linked from there.

4. The dnsapi2 wiki page is editable by any GitHub account - please add your entry there yourself so the `Docs:` anchor resolves.

neilpang:
The macOS job now fails before dns_feno runs: "Error loading file .../dnsapi/dns_feno.sh", i.e. the plugin no longer parses under macOS /bin/sh (bash 3.2). Run 33474510093, everything after MacOS was cancelled.

That is my fault - the snippet I posted inlines a `case` block inside `$( ... )`, and old shells (macOS bash 3.2, Solaris /bin/sh) end the command substitution at the `)` of a case pattern. The fix is to keep the `case` at statement level inside its own function and leave only the call inside `$( )`; acme.sh does exactly this in `_filter_cron_bin`.

```sh
# reads the "{"-split record lines on stdin
# $1 = TXT value (case-sensitive), $2 = lower-cased relative name
_feno_match_record() {
  _fmr_value="$1"
  _fmr_name="$2"
  while IFS= read -r _fmr_line || [ -n "$_fmr_line" ]; do
    case "$_fmr_line" in
    *"\"value\":\"$_fmr_value\""*) ;;
    *) continue ;;
    esac
    case "$_fmr_line" in
    *'"type":"TXT"'*) ;;
    *) continue ;;
    esac
    _fmr_lower=$(printf "%s" "$_fmr_line" | _lower_case)
    case "$_fmr_lower" in
    *"\"name\":\"$_fmr_name\""*)
      printf "%s\n" "$_fmr_line"
      return 0
      ;;
    esac
  done
}
```

and in `_feno_find_record_id`:

```sh
_find_name_lower=$(printf "%s" "$_find_name" | _lower_case)
_line=$(printf "%s\n" "$response" | tr '{' '\n' |
  _feno_match_record "$_find_value" "$_find_name_lower")
```

Checked under dash, bash and bash --posix: parses, and picks the right id for same-name/different-value, mixed-case name, apex (`"name":""`), and the no-match case.

Also note the nested `$( )` inside a `case` word has the same hazard on those shells, hence the `_fmr_lower` assignment.

Please re-run the fork DNS test on the fixed commit.

The dnsapi2 wiki page still has no `dns_feno` section - it is editable by any GitHub account, please add the entry yourself so the `Docs:` anchor resolves.

PR #7214, opened 2026-08-27
Update netcup DNS API to support new API

jduss:
netcup introduced a new general REST API (`https://api.netcup.com/v1`) for
domains managed by their new DNS backend. This extends `dns_netcup.sh` to
support it, keeping the legacy CCP API flow unchanged.

- The API is selected by key length: new REST API keys are 64 characters,
  legacy CCP keys are 50. With a REST key only `NC_Apikey` is needed.
- The zone is looked up via `GET /domain?fqdn=…` (walking the name, longest
  match first); challenge records are managed through the dedicated ACME
  challenge endpoints. After adding, the script waits 20 s and polls until
  the record reports `deployed`.
- Records outside `_acme-challenge.` cannot exist behind the ACME endpoints
  and are treated as a successful no-op.
- Domains with `isDnsManaged: false` fall back to the legacy CCP API when
  the optional `NC_Apikey_Legacy` (plus `NC_Apipw`/`NC_CID`) is set.

DNS-API-Test passed in my fork:
https://github.com/jduss/acme.sh/actions/runs/33067813625

The netcup section of the dnsapi wiki page has been updated. (https://github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_netcup)

### DIFF
--- a/dnsapi/dns_netcup.sh
+++ b/dnsapi/dns_netcup.sh
@@ -85,7 +85,15 @@ _nc_rest_add() {
     return
   fi
 
-  _nc_rest_get_scope "$fulldomain" "$_domain" || return 1
+  case "$fulldomain" in
+  _acme-challenge.*) ;;
+  *)
+    _info "The netcup REST API can only create _acme-challenge records. Not creating $fulldomain."
+    return 0
+    ;;
+  esac
+
+  _nc_rest_get_scope "$fulldomain" "$_domain"
   _debug _scope "$_scope"
 
   if ! _nc_rest POST "domain/$_domain_id/acme/challenge" "{\"scope\": \"$_scope\", \"value\": \"$txtvalue\"}" ||
@@ -131,7 +139,15 @@ _nc_rest_rm() {
     return
   fi
 
-  _nc_rest_get_scope "$fulldomain" "$_domain" || return 1
+  case "$fulldomain" in
+  _acme-challenge.*) ;;
+  *)
+    _info "The netcup REST API can only remove _acme-challenge records. Not removing $fulldomain."
+    return 0
+    ;;
+  esac
+
+  _nc_rest_get_scope "$fulldomain" "$_domain"
 
   if ! _nc_rest DELETE "domain/$_domain_id/acme/challenge/$_scope/$txtvalue"; then
     _err "Unable to remove the challenge record: $response"
@@ -184,15 +200,12 @@ _nc_rest_get_domain() {
 
 # fulldomain domain
 # Sets _scope to the host part of the challenge relative to the domain.
-# The REST API prepends _acme-challenge. to the scope itself.
+# The REST API prepends _acme-challenge. to the scope itself, so the
+# prefix is stripped from the record name. Records with other names
+# cannot exist behind the REST API: the callers treat them as a
+# successful no-op instead (the DNS-API-Test adds and removes such a
+# record and expects both calls to succeed).
 _nc_rest_get_scope() {
-  case "$1" in
-  _acme-challenge.*) ;;
-  *)
-    _err "The netcup REST API can only manage _acme-challenge records, $1 is not supported."
-    return 1
-    ;;
-  esac
   _scope="${1#_acme-challenge.}"
   if [ "$_scope" = "$2" ]; then
     _scope="@"
--- a/dnsapi/dns_netcup.sh
+++ b/dnsapi/dns_netcup.sh
@@ -5,30 +5,243 @@ Domains: netcup.de netcup.net
 Site: netcup.eu/
 Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_netcup
 Options:
- NC_Apikey API Key
- NC_Apipw API Password
- NC_CID Customer Number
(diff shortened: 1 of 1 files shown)


neilpang:
1. Alias mode is broken: with --domain-alias (or the "=" challenge-alias
   form) fulldomain does not start with _acme-challenge., so
   _nc_rest_add prints an info line and returns 0 without creating
   anything; issuance then fails at validation with no hint why. An
   apex alias also fails: the i=2 walk in _nc_rest_get_domain starts at
   the TLD. For a real record this must return 1 (or fall back to the
   legacy API when legacy creds are set); returning success is only
   acceptable for the synthetic DNS-API-Test record.

2. _H1 is exported and never cleared. On the isDnsManaged fallback the
   REST Bearer key is sent to ccp.netcup.net on every legacy call, and
   it stays exported for whatever hook runs next. Also blank _H2-_H5 in
   _nc_rest: in a multi-provider issuance another hook's auth headers
   (e.g. dns_cf X-Auth-Key) would ride into api.netcup.com. Set
   _H2=""... in _nc_rest and clear _H1 before the legacy calls.

3. Key-length dispatch: please confirm from netcup docs that a legacy
   CCP key can never be 64 chars and cite that in the comment. A 64-char
   legacy key would be sent as a Bearer token and end in "No DNS zone
   found"; the auth-failure path should mention the key was auto-
   detected as REST by length.

4. Drop the unconditional _sleep 20: poll the deploy status first and
   _sleep 5 between retries (same 60s cap). A 10-SAN cert currently
   pays a hard 200s even when records deploy instantly, and acme.sh
   runs its own post-add DNS wait on top.

5. _nc_rest_get_domain reports a 401 (bad key) as "No DNS zone for X
   was found", pointing users at the wrong cause. When the response has
   success:false, surface $response instead of walking on.

6. _domain_id extraction: on systems without egrep, _egrep_o falls back
   to sed which takes the LAST match, so a nested "id" (e.g. a
   nameserver object) would win over the domain id. Tighten the
   extraction so first/last cannot differ. Same for isDnsManaged.

7. [ "$_dns_managed" != "true" ] treats an empty/unparsed value as
   "not manageable" and demands legacy credentials netcup is retiring.
   Trigger the fallback only on literal false; error distinctly on
   empty.

8. _nc_is_rest_key: use [ "${#NC_Apikey}" -eq 64 ]; the wc -c | tr
   pipeline is 3 forks and depends on platform wc padding.

9. nc_apikey reads like a user-facing option but is internal state;
   rename to _nc_apikey (and prefix endrest too).

jduss:
> 1. Alias mode is broken: with --domain-alias (or the "=" challenge-alias
>    form) fulldomain does not start with _acme-challenge., so
>    _nc_rest_add prints an info line and returns 0 without creating
>    anything; issuance then fails at validation with no hint why. An
>    apex alias also fails: the i=2 walk in _nc_rest_get_domain starts at
>    the TLD. For a real record this must return 1 (or fall back to the
>    legacy API when legacy creds are set); returning success is only
>    acceptable for the synthetic DNS-API-Test record.
> 2. _H1 is exported and never cleared. On the isDnsManaged fallback the
>    REST Bearer key is sent to ccp.netcup.net on every legacy call, and
>    it stays exported for whatever hook runs next. Also blank _H2-_H5 in
>    _nc_rest: in a multi-provider issuance another hook's auth headers
>    (e.g. dns_cf X-Auth-Key) would ride into api.netcup.com. Set
>    _H2=""... in _nc_rest and clear _H1 before the legacy calls.
> 3. Key-length dispatch: please confirm from netcup docs that a legacy
>    CCP key can never be 64 chars and cite that in the comment. A 64-char
>    legacy key would be sent as a Bearer token and end in "No DNS zone
>    found"; the auth-failure path should mention the key was auto-
>    detected as REST by length.
> 4. Drop the unconditional _sleep 20: poll the deploy status first and
>    _sleep 5 between retries (same 60s cap). A 10-SAN cert currently
>    pays a hard 200s even when records deploy instantly, and acme.sh
>    runs its own post-add DNS wait on top.
> 5. _nc_rest_get_domain reports a 401 (bad key) as "No DNS zone for X
>    was found", pointing users at the wrong cause. When the response has
>    success:false, surface $response instead of walking on.
> 6. _domain_id extraction: on systems without egrep, _egrep_o falls back
>    to sed which takes the LAST match, so a nested "id" (e.g. a
>    nameserver object) would win over the domain id. Tighten the
>    extraction so first/last cannot differ. Same for isDnsManaged.
> 7. [ "$_dns_managed" != "true" ] treats an empty/unparsed value as
>    "not manageable" and demands legacy credentials netcup is retiring.
>    Trigger the fallback only on literal false; error distinctly on
>    empty.
> 8. _nc_is_rest_key: use [ "${#NC_Apikey}" -eq 64 ]; the wc -c | tr
>    pipeline is 3 forks and depends on platform wc padding.
> 9. nc_apikey reads like a user-facing option but is internal state;
>    rename to _nc_apikey (and prefix endrest too).

Changed and pipeline run again https://github.com/jduss/acme.sh/actions/runs/33485985909
The API key length is not in the documentation, but confirmed with netcup.

PR #7216, opened 2026-08-27
TrueNAS deploy script witch websocat instead of Python midclt internal command

Tok-Ra85:
TrueNAS deploy script for SCALE/CORE using websocket (websocat binary) It is recommend to use a wildcard certificate

Tested with TrueNAS SCALE 25.10 (API "wss://host/api/current", JSON-RPC 2.0). Unlike "truenas_ws" hook, this script does NOT use midclt, the truenas_api_client Python package. It only depends on:
- jq
- websocat  (a static binary you deploy)

Why: avoids installing a Python environment / TrueNAS package on remote machine (like OPNsense) just to push a certificate.

IMPORTANT: This script is written in pure POSIX sh (no coproc, no bash arrays).

This hook is a "fork" of truenas_ws

### DIFF
--- a/deploy/truenas_websocat.sh
+++ b/deploy/truenas_websocat.sh
@@ -73,7 +73,7 @@ truenas_websocat_deploy() {
   fi
   _websocat_bin=$(command -v websocat 2>/dev/null)
   if [ -z "$_websocat_bin" ]; then
-    _err "websocat binary not found in PATH. Install it using your system's package manager (e.g. 'pkg install websocat' on FreeBSD/OPNsense), or download a static binary from https://github.com/vi/websocat/releases."
+    _err "websocat binary not found in PATH. Install it using your system's package manager, or download a static binary from https://github.com/vi/websocat/releases."
     return 8
   fi
 
--- a/deploy/truenas_websocat.sh
+++ b/deploy/truenas_websocat.sh
@@ -27,8 +27,8 @@
 #
 # Optional:
 #   export DEPLOY_TRUENAS_HOSTNAME="<TRUENAS_HOSTNAME_OR_IP>				(required on first run)
-#   export DEPLOY_TRUENAS_PROTOCOL="wss"						# ws or wss		(default: wss)
-#   export DEPLOY_TRUENAS_PORT="443"						# 80, 443, 8443	(default: 443)
+#   export DEPLOY_TRUENAS_PROTOCOL="ws"						# ws or wss		(default: ws)
+#   export DEPLOY_TRUENAS_PORT="80"							# 80, 443, 8443	(default: 80)
 #   export DEPLOY_TRUENAS_UPDATE_FTP="no"					# yes or no 	(default: no) also updates the FTP certificate
 #   export DEPLOY_TRUENAS_UPDATE_APPS="no"					# yes or no		(default: no) also updates the certificate for any
 #								iX app exposing a "certificate_id" option.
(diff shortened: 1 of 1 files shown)


neilpang:
1. Insecure defaults, and the comments contradict the code: the comment
   says "default to wss ... the payload carries the API key and the
   private key in plain text" but the code defaults to ws; the comment
   says "default to 443" but the code defaults to 80. Please default to
   wss and 443, matching truenas_ws.sh.

2. _jq_bin is hardcoded to /usr/local/bin/jq (FreeBSD path). On TrueNAS
   SCALE (Debian) jq is /usr/bin/jq. Locate it with "command -v jq",
   the same way you locate websocat. The "pkg install websocat" error
   hint is also FreeBSD-only, please generalize it.

3. Deploy hooks are sourced into the same shell as acme.sh and all
   other hooks, so every private function and variable needs a hook
   prefix to avoid collisions: _rpc_call, _rpc_has_error,
   _wait_for_job, _ws_cleanup, _method, _params, _payload, _response,
   _ws_uri (already defined by truenas_ws.sh, and multiple deploy
   hooks can run in one invocation) -> _truenas_websocat_rpc_call etc.

4. _wait_for_job loops forever if the job never leaves RUNNING. Please
   add a timeout.

5. CI enforces 2-space indent (shfmt -i 2); the file is tab-indented
   and has trailing whitespace, so the shfmt check will fail. Run:
   shfmt -w -i 2 deploy/truenas_websocat.sh

6. Please add a usage entry for this hook on the wiki "deployhooks"
   page.

Tok-Ra85 (commented):
Insecure defaults, and the comments contradict the code: the comment
says "default to wss ... the payload carries the API key and the
private key in plain text" but the code defaults to ws; the comment
says "default to 443" but the code defaults to 80. Please default to
wss and 443, matching truenas_ws.sh.

-- Done, but still in port 80 because TrueNAS run, out-of-the-box on the 80 and 443 is not activated

_jq_bin is hardcoded to /usr/local/bin/jq (FreeBSD path). On TrueNAS
SCALE (Debian) jq is /usr/bin/jq. Locate it with "command -v jq",
the same way you locate websocat. The "pkg install websocat" error
hint is also FreeBSD-only, please generalize it.

-- Good catch, done.

Deploy hooks are sourced into the same shell as acme.sh and all
other hooks, so every private function and variable needs a hook
prefix to avoid collisions: _rpc_call, _rpc_has_error,
_wait_for_job, _ws_cleanup, _method, _params, _payload, _response,
_ws_uri (already defined by truenas_ws.sh, and multiple deploy
hooks can run in one invocation) -> _truenas_websocat_rpc_call etc.

-- Since I taken ws hook, I did not change all var and fonction name, my bad, fixed.

_wait_for_job loops forever if the job never leaves RUNNING. Please
add a timeout.

-- Correct, fixed.

CI enforces 2-space indent (shfmt -i 2); the file is tab-indented
and has trailing whitespace, so the shfmt check will fail. Run:
shfmt -w -i 2 deploy/truenas_websocat.sh

-- Fixed.

Please add a usage entry for this hook on the wiki "deployhooks"
page.

-- Already did, entry N°48

Tok-Ra85 (commented):
removed exemple of FreeBSD, not enough generic.

neilpang:
ShellCheck fails on SC2016: jq expressions like '{name: $name}' in
single quotes are correct jq usage, so it is a false positive here.
Silence it with a directive, either one file-wide on line 2:

  # shellcheck disable=SC2016

or one above each of the three jq calls (the
_truenas_websocat_payload=, _key_params= and _create_params=
assignments).

Tok-Ra85:
> ShellCheck fails on SC2016: jq expressions like '{name: $name}' in single quotes are correct jq usage, so it is a false positive here. Silence it with a directive, either one file-wide on line 2:
> 
> # shellcheck disable=SC2016
> or one above each of the three jq calls (the _truenas_websocat_payload=, _key_params= and _create_params= assignments).

Did it on line 2.
Thanks.

Tok-Ra85 (commented):
Typo, forget the # ....

Tok-Ra85:
Test pass.
Ready to deploy !
Thank you for your help @neilpang !!

Next step will be to propose and implement that into OPNsense !

invario:
@Tok-Ra85 Just wanted to add a successful testing data point for you.  I have `acme.sh` running as non-root as a custom Docker compose app on TrueNAS Scale 25.10.6.

I have replaced my previous config (which used [deploy-freenas](https://github.com/danb35/deploy-freenas)) with your deploy-hook and it is working with no problems. It deploys to the TrueNAS host server without an issue and it also deploys to a remotely located TrueNAS Scale install that I have connected with a VPN tunnel.  These deployments are done using your hook via the `multideploy` hook also.

As a result of your hook, my custom Docker image build no longer needs to install the TrueNAS API or python. Thank you!!

Tok-Ra85:
Hi @invario,
Thank you very much for this feedback !
It is very usefull.

My hook was added into the dev branch, as soon is it in the master branch, I will start to write the addon for OPNsense implementation.

Have a great day.

PR #7219, opened 2026-08-29
Add AK & SK based Huawei Cloud DNS API

mashirozx:
## Summary

This PR adds `dns_hw.sh`, a Huawei Cloud DNS API integration authenticated with IAM Access Key (AK) and Secret Access Key (SK).

The existing `dns_huaweicloud.sh` plugin remains unchanged to avoid disrupting users who already rely on username/password authentication. Its Huawei Cloud documentation link is no longer available, while the new implementation follows the current AK/SK request-signing authentication model.

## API Comparison

Both plugins use the same Huawei Cloud DNS v2 API endpoints for DNS record management:

- `GET /v2/zones`
- `GET /v2/zones/{zone_id}/recordsets`
- `POST /v2/zones/{zone_id}/recordsets`
- `PUT /v2/zones/{zone_id}/recordsets/{recordset_id}`
- `DELETE /v2/zones/{zone_id}/recordsets/{recordset_id}`

The difference is the authentication mechanism:

| Plugin | Credentials | Authentication |
| --- | --- | --- |
| dns_huaweicloud.sh | Username, password, and account domain name | Obtains an IAM `X-Auth-Token` through `/v3/auth/tokens` |
| dns_hw.sh | IAM AK and SK | Signs each DNS request with `SDK-HMAC-SHA256` |

## Why AK/SK Authentication

AK/SK authentication is better suited for automated certificate issuance and renewal:

- It supports least-privilege IAM policies without exposing an account password.
- Credentials can be independently rotated, disabled, or replaced.
- It avoids an interactive login/token acquisition flow.
- It uses Huawei Cloud's standard request-signing mechanism for API access.

## Backward Compatibility

The old plugin is intentionally retained. Existing users may have stored credentials and renewal jobs configured with `HUAWEICLOUD_Username`, `HUAWEICLOUD_Password`, and `HUAWEICLOUD_DomainName`; replacing it would break those deployments.

Although both authentication methods could technically be placed in one plugin, doing so would require maintaining two credential models, authentication flows, configuration precedence rules, and test paths. The plugins also have different defaults and record-cleanup behavior, which would make a combined implementation harder to understand and maintain.

Keeping the integrations separate provides a clean migration path:

- Existing users can continue using dns_huaweicloud.sh without changes.
- New users can use dns_hw.sh with the recommended AK/SK authentication method.


mashirozx:
Wiki content (https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_hw) follows:

***

# Huawei Cloud DNS API (Access Key Based)

Use `dns_hw` to complete DNS-01 validation with Huawei Cloud DNS. This hook
uses an IAM Access Key (AK) and Secret Access Key (SK); it does not use the
legacy username, password, or domain name settings.

## Prerequisites

1. Create or select an IAM user that has access to the Huawei Cloud DNS zone.
2. Create an [access key](https://support.huaweicloud.com/intl/en-us/usermanual-ca/ca_01_0003.html)
  for that IAM user in the Huawei Cloud console and keep its AK and SK secure.
3. Grant the IAM user permission to list DNS zones and create, update, and
   delete DNS record sets. Scope the permission to only the required zone when
   possible.
4. Ensure the zone already exists in Huawei Cloud DNS and its authoritative
   name servers are active for the domain.

## Configure Credentials

Export the credentials before the first issuance. Use the DNS service region
that contains the zone; `HW_Region` is optional and defaults to
`cn-north-4`.

```sh
export HW_AK="your_access_key"
export HW_SK="your_secret_access_key"
export HW_Region="cn-north-4"
```

## Issue a Certificate

```sh
./acme.sh --issue --dns dns_hw -d example.com -d '*.example.com'
```

`HW_AK`, `HW_SK`, and, when set, `HW_Region` are saved in
`~/.acme.sh/account.conf` after a successful issuance and reused during
renewals. Do not place these credentials in shell history or source control.

## Notes

- The hook searches for the most specific existing Huawei Cloud public DNS zone
  for each requested domain.
- It adds the ACME challenge to the matching TXT record set and preserves other
  TXT values in that set. Cleanup removes only the challenge value it added.
- `dns_huaweicloud` remains available for existing users and uses the separate
  `HUAWEICLOUD_Username`, `HUAWEICLOUD_Password`, and
  `HUAWEICLOUD_DomainName` credentials. These credentials are not used by
  `dns_hw`.

PR #7220, opened 2026-08-29
Add AK & SK based Huawei Cloud DNS API

mashirozx:
## Summary

This PR adds `dns_hw.sh`, a Huawei Cloud DNS API integration authenticated with [IAM Access Key (AK) and Secret Access Key (SK)](https://support.huaweicloud.com/intl/en-us/usermanual-ca/ca_01_0003.html).

The existing `dns_huaweicloud.sh` plugin remains unchanged to avoid disrupting users who already rely on username/password authentication. Its Huawei Cloud documentation link is no longer available, while the new implementation follows the current AK/SK request-signing authentication model.

## API Comparison

Both plugins use the same Huawei Cloud DNS v2 API endpoints for DNS record management:

- `GET /v2/zones`
- `GET /v2/zones/{zone_id}/recordsets`
- `POST /v2/zones/{zone_id}/recordsets`
- `PUT /v2/zones/{zone_id}/recordsets/{recordset_id}`
- `DELETE /v2/zones/{zone_id}/recordsets/{recordset_id}`

The difference is the authentication mechanism:

| Plugin | Credentials | Authentication |
| --- | --- | --- |
| dns_huaweicloud.sh | Username, password, and account domain name | Obtains an IAM `X-Auth-Token` through `/v3/auth/tokens` |
| dns_hw.sh | IAM AK and SK | Signs each DNS request with `SDK-HMAC-SHA256` |

## Why AK/SK Authentication

AK/SK authentication is better suited for automated certificate issuance and renewal:

- It supports least-privilege IAM policies without exposing an account password.
- Credentials can be independently rotated, disabled, or replaced.
- It avoids an interactive login/token acquisition flow.
- It uses Huawei Cloud's standard request-signing mechanism for API access.

Relevant Huawei Cloud documentation:

- [DNS API Reference](https://support.huaweicloud.com/intl/en-us/api-dns/dns_api_60000.html)

## Backward Compatibility

The old plugin is intentionally retained. Existing users may have stored credentials and renewal jobs configured with `HUAWEICLOUD_Username`, `HUAWEICLOUD_Password`, and `HUAWEICLOUD_DomainName`; replacing it would break those deployments.

Although both authentication methods could technically be placed in one plugin, doing so would require maintaining two credential models, authentication flows, configuration precedence rules, and test paths. The plugins also have different defaults and record-cleanup behavior, which would make a combined implementation harder to understand and maintain.

Keeping the integrations separate provides a clean migration path:

- Existing users can continue using dns_huaweicloud.sh without changes.
- New users can use dns_hw.sh with the recommended AK/SK authentication method.

## Pipeline

Mostly passed. One test failed due to an internal pipeline issue, which is unrelated to this PR.

https://github.com/moezx/acme.sh/actions/runs/33258854426

## Wiki Content Update

https://github.com/acmesh-official/acme.sh/pull/7220#issuecomment-5462506920

### DIFF
--- a/dnsapi/dns_hw.sh
+++ b/dnsapi/dns_hw.sh
@@ -1,0 +1,226 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_hw_info='Huawei Cloud DNS
+Site: HuaweiCloud.com
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_hw
+Options:
+ HW_AK Access Key
+ HW_SK Secret Access Key
+ HW_Region Region. E.g. "cn-north-4". Optional, defaults to "cn-north-4".
+Issues: github.com/acmesh-official/acme.sh/issues/7221
+Author: mashirozx
+'
+
+dns_hw_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  if ! _hw_init; then
+    return 1
+  fi
+
+  if ! _hw_get_zoneid "$fulldomain"; then
+    return 1
+  fi
+  if ! _hw_get_recordset "$fulldomain" "$_hw_zoneid"; then
+    return 1
+  fi
+
+  # Huawei Cloud stores each TXT value with its required inner quotes.
+  _hw_txt_record="\"\\\"${txtvalue}\\\"\""
+  case "$_hw_records" in
+  *"$txtvalue"*)
+    _debug "TXT record already exists"
+    ;;
+  *)
+    if [ -z "$_hw_recordid" ]; then
+      _hw_body="{\"name\":\"${fulldomain}.\",\"type\":\"TXT\",\"ttl\":300,\"records\":[${_hw_txt_record}]}"
+      # Create a TXT record set: https://support.huaweicloud.com/api-dns/dns_api_64001.html
+      _hw_rest "POST" "/v2/zones/${_hw_zoneid}/recordsets" "" "$_hw_body" || return 1
+    else
+      _hw_body="{\"name\":\"${fulldomain}.\",\"type\":\"TXT\",\"ttl\":${_hw_recordttl},\"records\":[${_hw_records},${_hw_txt_record}]}"
+      # Update the existing TXT record set: https://support.huaweicloud.com/api-dns/UpdateRecordSets.html
+      _hw_rest "PUT" "/v2/zones/${_hw_zoneid}/recordsets/${_hw_recordid}" "" "$_hw_body" || return 1
+    fi
+    ;;
+  esac
+
+  _saveaccountconf_mutable HW_AK "$HW_AK"
+  _saveaccountconf_mutable HW_SK "$HW_SK"
+  if [ -n "$HW_Region" ]; then
+    _saveaccountconf_mutable HW_Region "$HW_Region"
+  fi
+}
+
+dns_hw_rm() {
+  fulldomain=$1
+  txtvalue=$2
+
+  if ! _hw_init; then
+    return 1
+  fi
+
+  if ! _hw_get_zoneid "$fulldomain" || ! _hw_get_recordset "$fulldomain" "$_hw_zoneid"; then
+    return 1
+  fi
+  if [ -z "$_hw_recordid" ]; then
+    _debug "TXT record not found"
+    return 0
+  fi
+
+  # Keep unrelated TXT values that share this record set.
+  _hw_txt_record="\"\\\"${txtvalue}\\\"\""
+  case "$_hw_records" in
+  *"$txtvalue"*) ;;
+  *)
+    _debug "TXT record value not found"
+    return 0
+    ;;
+  esac
+  _hw_sed_txt_record=$(echo "$_hw_txt_record" | sed 's/\\/\\\\/g')
+  _hw_new_records=$(echo "$_hw_records" | sed "s/${_hw_sed_txt_record},//; s/,${_hw_sed_txt_record}//; s/${_hw_sed_txt_record}//")
+  if [ -z "$_hw_new_records" ]; then
+    # Delete an empty TXT record set: https://support.huaweicloud.com/api-dns/dns_api_64005.html
+    _hw_rest "DELETE" "/v2/zones/${_hw_zoneid}/recordsets/${_hw_recordid}" "" "" || return 1
+  else
+    _hw_body="{\"name\":\"${fulldomain}.\",\"type\":\"TXT\",\"ttl\":${_hw_recordttl},\"records\":[${_hw_new_records}]}"
+    # Update the record set after removing this challenge value: https://support.huaweicloud.com/api-dns/UpdateRecordSets.html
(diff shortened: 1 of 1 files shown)


mashirozx:
Wiki content (https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_hw) follows:

***

# Huawei Cloud DNS API (Access Key Based)

Use `dns_hw` to complete DNS-01 validation with Huawei Cloud DNS. This hook
uses an IAM Access Key (AK) and Secret Access Key (SK); it does not use the
legacy username, password, or domain name settings.

## Prerequisites

1. Create or select an IAM user that has access to the Huawei Cloud DNS zone.
2. Create an [access key](https://support.huaweicloud.com/intl/en-us/usermanual-ca/ca_01_0003.html)
  for that IAM user in the Huawei Cloud console and keep its AK and SK secure.
3. Grant the IAM user permission to list DNS zones and create, update, and
   delete DNS record sets. Scope the permission to only the required zone when
   possible.
4. Ensure the zone already exists in Huawei Cloud DNS and its authoritative
   name servers are active for the domain.

## Configure Credentials

Export the credentials before the first issuance. Use the DNS service region
that contains the zone; `HW_Region` is optional and defaults to
`cn-north-4`.

```sh
export HW_AK="your_access_key"
export HW_SK="your_secret_access_key"
export HW_Region="cn-north-4"
```

## Issue a Certificate

```sh
./acme.sh --issue --dns dns_hw -d example.com -d '*.example.com'
```

`HW_AK`, `HW_SK`, and, when set, `HW_Region` are saved in
`~/.acme.sh/account.conf` after a successful issuance and reused during
renewals. Do not place these credentials in shell history or source control.

## Notes

- The hook searches for the most specific existing Huawei Cloud public DNS zone
  for each requested domain.
- It adds the ACME challenge to the matching TXT record set and preserves other
  TXT values in that set. Cleanup removes only the challenge value it added.
- `dns_huaweicloud` remains available for existing users and uses the separate
  `HUAWEICLOUD_Username`, `HUAWEICLOUD_Password`, and
  `HUAWEICLOUD_DomainName` credentials. These credentials are not used by
  `dns_hw`.

neilpang:
Thanks. A few blockers:

1. _hw_get_recordset: the name filter is fuzzy by default (doc
   dns_api_64004: "A fuzzy search will be used by default"). Add
   search_mode=equal and verify the returned name equals
   "${fulldomain}." like _hw_get_zoneid already does. Without it,
   a recordset whose name merely contains the challenge name can be
   returned, and the hook PUTs/DELETEs the wrong recordset.

2. Same function: _hw_recordid takes the FIRST "id" in the response
   but the greedy sed (s/.*"records"...) takes the LAST records
   array. With 2+ recordsets returned, the PUT rewrites recordset A
   with recordset B's values. search_mode=equal (single result)
   fixes this too.

3. Line 142: tr -d '\r\n' | sed feeds sed a line with no trailing
   newline. Solaris /usr/bin/sed silently drops such a line, so
   _hw_records is always empty there (add builds records:[,...] ->
   400; rm never removes). Run sed before tr, or feed the input via
   printf "%s\n".

4. _hw_rest sets _H1.._H4 but never clears _H5. Hooks share one
   shell; in a multi-provider renewal a previous hook's _H5 (e.g.
   an auth header) is sent to the Huawei endpoint. Clear _H1.._H5
   like dns_volcengine.sh does.

Minor:
- Check _post's return code and truncate "$HTTP_HEADER" before the
  request; on a transport failure the file still holds the previous
  request's 200 and _hw_rest reports success.
- The PUT bodies hardcode ttl:300, silently rewriting the TTL of a
  pre-existing recordset on the same name.
- The credential-load/endpoint block is duplicated in add and rm;
  move it to a shared _hw_init (see _hetznercloud_init).
- Line 162: _hw_date=$(echo "$(_utc_date)" | ...) -> SC2005; one
  pipeline does it: _hw_date="$(_utc_date | sed 's/ /T/' |
  tr -d ':-')Z".

Process (required for a new dnsapi):
- File a dedicated bug-report issue for this provider and reference
  it in the info block's Issues: field.
- Add the usage docs to the wiki dnsapi2 page (the Docs: anchor is
  currently dead); the wiki is publicly editable.
- Re-run DNS.yml in your fork: the MidnightBSD VM flake skipped the
  whole Solaris/illumos chain, so those jobs never ran.

---

1. _hw_get_recordset: name 过滤默认是模糊匹配 (文档 dns_api_64004:
   "A fuzzy search will be used by default")。请加 search_mode=equal
   并像 _hw_get_zoneid 那样校验返回的 name 等于 "${fulldomain}."。
   否则任何 name 包含 challenge 域名的 recordset 都可能被返回,
   hook 会 PUT/DELETE 错误的 recordset。

2. 同一函数: _hw_recordid 取响应中第一个 "id", 但贪婪 sed
   (s/.*"records"...) 取的是最后一个 records 数组。返回 2 个以上
   recordset 时, PUT 会用 B 的值改写 A。search_mode=equal (单结果)
   也能一并解决。

3. 第 142 行: tr -d '\r\n' | sed 给 sed 的输入行没有结尾换行。
   Solaris /usr/bin/sed 会静默丢弃这样的行, _hw_records 在 Solaris
   上恒为空 (add 拼出 records:[,...] -> 400; rm 永远删不掉)。
   先 sed 后 tr, 或用 printf "%s\n" 喂入。

4. _hw_rest 设置了 _H1.._H4 但没清 _H5。hooks 共享同一个 shell;
   多提供商续期时上一个 hook 的 _H5 (例如认证头) 会被发到华为
   endpoint。请像 dns_volcengine.sh 那样清空 _H1.._H5。

次要:
- 检查 _post 的返回码, 并在请求前清空 "$HTTP_HEADER"; 传输层失败时
  文件里还是上一次请求的 200, _hw_rest 会误报成功。
- PUT body 硬编码 ttl:300, 会悄悄改写同名已有 recordset 的 TTL。
- 凭据加载/endpoint 初始化块在 add 和 rm 里重复; 建议提取共用的
  _hw_init (参考 _hetznercloud_init)。
- 第 162 行: _hw_date=$(echo "$(_utc_date)" | ...) -> SC2005;
  一条管道即可: _hw_date="$(_utc_date | sed 's/ /T/' |
  tr -d ':-')Z"。

流程 (新 dnsapi 必需):
- 为该 provider 建一个专门的 bug-report issue, 并写进 info block 的
  Issues: 字段。
- 把使用文档加到 wiki dnsapi2 页面 (目前 Docs: 锚点是死链); wiki
  可以直接编辑。
- 在你的 fork 重跑 DNS.yml: MidnightBSD VM flake 把 Solaris/illumos
  整条链 skip 了, 那些 job 没跑过。

mashirozx:
@neilpang Fixed. 

DNS Pipeline: https://github.com/moezx/acme.sh/actions/runs/33304660119

Issue: #7221

Wiki: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_hw

mashirozx:
@neilpang DNS pipeline succeed https://github.com/moezx/acme.sh/actions/runs/33304660119/job/99262118234

PR #7232, opened 2026-09-03
fix: poweradmin dnsapi leaves hanging txt records

InvisibleDuck:
Improved string and json handling to prevent certain acme challenge txt records to be left hanging.
Changes have been tested in our production environment for 3 weeks, with no leftover records.
ci run: https://github.com/InvisibleDuck/acme.sh/actions/runs/32261080269


InvisibleDuck:
Test is passed

neilpang:
1. grep -o is not portable: Solaris /usr/bin/grep has no -o (its
   option set is -b -c -h -i -l -n -q -s -v -w). The Solaris job is
   green only because CI puts /usr/gnu/bin first. Use the core
   wrapper, and escape the braces -- under ERE an unescaped leading
   { is not matched literally by GNU egrep, so the output loses the
   opening brace and gains stray lines like ]} and ,"message":"ok"}:

       _egrep_o '\{[^{}]*\}'

2. grep -F is a hard error on the same Solaris grep (illegal option
   -- F). Since this pipeline is being rewritten anyway, drop the
   three grep -F and match with case, which is literal by
   construction (a . in $full stays a dot). Capture first and feed a
   here-doc; a case inside $(...) breaks bash 3.2 and Solaris sh:

       _records="$(printf '%s\n' "$response" | tr -d '\r\n' | _egrep_o '\{[^{}]*\}')"
       _txt_record_obj=""
       while IFS= read -r _rec || [ -n "$_rec" ]; do
         case "$_rec" in
         *"\"name\":\"$full\""*) ;;
         *) continue ;;
         esac
         case "$_rec" in
         *"\"type\":\"TXT\""*) ;;
         *) continue ;;
         esac
         case "$_rec" in
         *"\"content\":\"$txtvalue\""*)
           _txt_record_obj="$_rec"
           break
           ;;
         esac
       done <<EOF
       $_records
       EOF

   Verified under dash and bash: matches the right object, no match
   on a mangled name or unknown value, empty output on an empty
   response. _get_root has the same grep -F; same fix, optional
   here.

3. _json_escape: private helpers need the provider prefix
   (_poweradmin_*), hooks share one shell namespace. But fulldomain
   is a hostname and txtvalue is base64url, neither can contain "
   or \, so the escaping is a no-op. Simplest is to drop it and
   keep the original data= line.

Re-run DNS.yml in your fork after the change.

Issue #7221, opened 2026-08-30
Add AK & SK based Huawei Cloud DNS API


mashirozx:
## Summary

[This PR](https://github.com/acmesh-official/acme.sh/pull/7220) adds `dns_hw.sh`, a Huawei Cloud DNS API integration authenticated with [IAM Access Key (AK) and Secret Access Key (SK)](https://support.huaweicloud.com/intl/en-us/usermanual-ca/ca_01_0003.html).

The existing `dns_huaweicloud.sh` plugin remains unchanged to avoid disrupting users who already rely on username/password authentication. Its Huawei Cloud documentation link is no longer available, while the new implementation follows the current AK/SK request-signing authentication model.

## API Comparison

Both plugins use the same Huawei Cloud DNS v2 API endpoints for DNS record management:

- `GET /v2/zones`
- `GET /v2/zones/{zone_id}/recordsets`
- `POST /v2/zones/{zone_id}/recordsets`
- `PUT /v2/zones/{zone_id}/recordsets/{recordset_id}`
- `DELETE /v2/zones/{zone_id}/recordsets/{recordset_id}`

The difference is the authentication mechanism:

| Plugin | Credentials | Authentication |
| --- | --- | --- |
| dns_huaweicloud.sh | Username, password, and account domain name | Obtains an IAM `X-Auth-Token` through `/v3/auth/tokens` |
| dns_hw.sh | IAM AK and SK | Signs each DNS request with `SDK-HMAC-SHA256` |

## Why AK/SK Authentication

AK/SK authentication is better suited for automated certificate issuance and renewal:

- It supports least-privilege IAM policies without exposing an account password.
- Credentials can be independently rotated, disabled, or replaced.
- It avoids an interactive login/token acquisition flow.
- It uses Huawei Cloud's standard request-signing mechanism for API access.

Relevant Huawei Cloud documentation:

- [DNS API Reference](https://support.huaweicloud.com/intl/en-us/api-dns/dns_api_60000.html)

## Backward Compatibility

The old plugin is intentionally retained. Existing users may have stored credentials and renewal jobs configured with `HUAWEICLOUD_Username`, `HUAWEICLOUD_Password`, and `HUAWEICLOUD_DomainName`; replacing it would break those deployments.

[...]

PR #7236, opened 2026-09-06
Fix dns_hw: prevent record-set lookup from hanging

mashirozx:
## Summary

Fixes a hang in the Huawei Cloud DNS (`dns_hw`) hook while looking up an existing TXT record set.

`_lower_case()` is a stdin-based acme.sh helper. The record-set name comparison incorrectly passed DNS names as positional arguments, so the helper waited for terminal input instead of receiving data to transform.

## Changes

- Pipe DNS names into `_lower_case()` before comparing them.
- Compare record-set names case-insensitively, matching DNS semantics and Huawei Cloud's lowercase normalization.
- Add `limit=1` to the exact TXT record-set query.

The query already filters by record name, `search_mode=equal`, and `type=TXT`; therefore at most one record set is needed. Limiting the response avoids processing the API's default large record-set page through portable shell parsing.

## DNS Pipeline

https://github.com/moezx/acme.sh/actions/runs/34039268031

### DIFF
--- a/dnsapi/dns_hw.sh
+++ b/dnsapi/dns_hw.sh
@@ -134,7 +134,7 @@ _hw_get_recordset() {
   _hw_recordid=""
   _hw_records=""
   _hw_recordttl=""
-  _hw_query="name=$(printf "%s" "$_hw_domain" | _url_encode upper-hex)&search_mode=equal&type=TXT"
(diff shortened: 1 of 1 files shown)


mashirozx:
ref: #7221

PR #7238, opened 2026-09-07
Add DNSMint DNS API

kxbnb:
Adds `dnsapi/dns_dnsmint.sh` for [DNSMint](https://dnsmint.com).

DNSMint mints hostnames on domains it operates and serves from its own authoritative nameservers, so a customer has no zone of their own and the DNS-01 challenge is published through its API. Users reach it today through `dns_acmedns.sh` and a per-hostname acme-dns credential; this replaces that with one API key.

**One option, `DNSMINT_API_KEY`.** Saved with `_saveaccountconf_mutable` on first use, so renewals need no environment.

**No root zone detection.** The hostname is derived server-side from the challenge name, so there is nothing to look up and no record id to track — the value published is the value removed. `dns_acmedns.sh` skips `_get_root` for the same reason.

**Wildcards work.** The API keeps the two newest values for a name, so the apex and wildcard pair both resolve, per the note in the dev guide about appending rather than replacing.

**Tested** against Let's Encrypt staging for a wildcard plus its apex, on a hostname minted for the purpose:

```console
$ export DNSMINT_API_KEY=...
$ acme.sh --issue --server letsencrypt_test --dns dns_dnsmint \
    -d lego.dnsmint-108bd2.top -d '*.lego.dnsmint-108bd2.top'
[...] Your cert is in: .../lego.dnsmint-108bd2.top_ecc/lego.dnsmint-108bd2.top.cer

$ openssl x509 -in fullchain.cer -noout -subject -issuer -text | grep -E 'subject=|issuer=|DNS:'
subject= /CN=lego.dnsmint-108bd2.top
issuer= /C=US/O=Let's Encrypt/CN=(STAGING) Artificial Amaranth YE1
    DNS:*.lego.dnsmint-108bd2.top, DNS:lego.dnsmint-108bd2.top
```

`shellcheck -s sh` and `shfmt -d -i 2 -ci` are both clean. Branched from `dev`.

### DIFF
--- a/dnsapi/dns_dnsmint.sh
+++ b/dnsapi/dns_dnsmint.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_dnsmint_info='DNSMint.com
+ DNSMint mints hostnames on domains it operates and serves from its own
+ authoritative nameservers, so the challenge is published through its API
+ rather than a zone you run.
+Site: dnsmint.com
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi#dns_dnsmint
+Options:
+ DNSMINT_API_KEY API key carrying the dns01:write scope
+Issues: github.com/dnsmint/acme.sh
+Author: DNSMint
+'
+
+DNSMint_Api="https://dnsmint.com/api/httpreq"
+
+########  Public functions #####################
+
+#Usage: dns_dnsmint_add _acme-challenge.q7k4m2.example.dev "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_dnsmint_add() {
+  fulldomain=$1
+  txtvalue=$2
+
+  _info "Using DNSMint"
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _dnsmint_key; then
+    return 1
+  fi
+
+  if _dnsmint_rest present "$fulldomain" "$txtvalue"; then
+    _info "Added, OK"
+    return 0
+  fi
+  return 1
+}
+
+#Usage: dns_dnsmint_rm _acme-challenge.q7k4m2.example.dev "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_dnsmint_rm() {
+  fulldomain=$1
+  txtvalue=$2
+
+  _info "Using DNSMint"
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _dnsmint_key; then
+    return 1
+  fi
+
+  if _dnsmint_rest cleanup "$fulldomain" "$txtvalue"; then
+    _info "Removed, OK"
+    return 0
+  fi
+  return 1
+}
+
+####################  Private functions below ##################################
+
+_dnsmint_key() {
+  DNSMINT_API_KEY="${DNSMINT_API_KEY:-$(_readaccountconf_mutable DNSMINT_API_KEY)}"
+  if [ -z "$DNSMINT_API_KEY" ]; then
+    DNSMINT_API_KEY=""
+    _err "You did not specify DNSMINT_API_KEY yet."
+    _err "Create a key with the dns01:write scope at https://dnsmint.com/dashboard"
+    _err "e.g."
+    _err "export DNSMINT_API_KEY=dnsm_xxxxxxxxxxxx_xxxxxxxx"
+    return 1
+  fi
+  _saveaccountconf_mutable DNSMINT_API_KEY "$DNSMINT_API_KEY"
+  return 0
+}
+
+# DNSMint derives the hostname from the challenge name, so there is no zone to
+# look up and no record id to track: the value published is the value removed.
+_dnsmint_rest() {
+  action="$1"
+  fqdn="$2"
+  value="$3"
+
+  export _H1="Authorization: Bearer $DNSMINT_API_KEY"
+  export _H2="Accept: application/json"
+  export _H3="Content-Type: application/json"
+
+  data="{\"fqdn\":\"$fqdn\",\"value\":\"$value\"}"
+  _secure_debug2 data "$data"
+
+  response="$(_post "$data" "$DNSMint_Api/$action" "" "POST")"
+  _ret="$?"
+  _code="$(grep "^HTTP" "$HTTP_HEADER" | _tail_n 1 | cut -d " " -f 2 | tr -d "\\r\\n")"
+  _debug "http response code $_code"
+  _debug2 response "$response"
+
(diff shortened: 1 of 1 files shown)


kxbnb:
Read the [DNS API Dev Guide](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide) and the [DNS-API-Test](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test), and this targets `dev`.

**The DNS API test passes on every platform**: https://github.com/dnsmint/acme.sh/actions/runs/34075386104

Docker, MacOS, Windows, FreeBSD, GhostBSD, OpenBSD, NetBSD, DragonFlyBSD, MidnightBSD, Solaris, Omnios, OpenIndiana, Tribblix, Haiku, Hurd, OpenEuler, HardenedBSD, OPNsense — all green. It ran on `dnsmint/acme.sh`; both that and `kxbnb/acme.sh` (this PR's head) are ours, and I can re-run it on the head fork if you would rather see it there.

A note on one thing the test caught, in case it matters to review. DNSMint's DNS-01 endpoint only publishes `_acme-challenge` records — that label is reserved on our side precisely so a customer cannot break their own issuance. The first run failed the case commented `#add a txt record to a subdomain not "_acme-challenge"`, because the plugin only implemented challenge publishing, which is narrower than what a `dns_` plugin promises.

So the add and rm functions now branch on the label: a challenge goes to the DNS-01 endpoint, and any other TXT name goes to the records endpoint as an ordinary record under the hostname. Both paths are real writes, verified against the live API. Issuing certificates needs only the `dns01:write` scope; the records path needs `hostnames:read` and `hostnames:write`, which is why the `Options:` block documents the narrower one.

`shellcheck -s sh` and `shfmt -d -i 2 -ci` are clean.

Happy to update the [dnsapi2 wiki page](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2) with the usage — let me know if I should, or if you would rather add it when merging.

kxbnb:
Following up on the two things the welcome bot asks for.

The DNS-API-Test passes on all eighteen platforms in the matrix: https://github.com/dnsmint/acme.sh/actions/runs/34075386104. No workflow has run on this PR itself; happy to re-run it here if a maintainer approves the workflow.

**The dnsapi2 wiki entry** I cannot add — the wiki is not writable by non-collaborators (`403` on push). Here it is ready to paste as entry 226, in the format the surrounding entries use:

---

## 226. Use DNSMint API
<a name="dns_dnsmint"/>

[DNSMint](https://dnsmint.com) issues hostnames on domains it operates, for machines that need a name and a certificate. Create an API key in the dashboard; the `dns01:write` scope is enough to issue certificates, and it can be narrowed to a single hostname so the key on a web server cannot reach the rest of your account.

```sh
export DNSMINT_API_KEY="your-api-key"
```

`DNSMint_Api` is optional and overrides the API base url, which defaults to `https://dnsmint.com/api`.

To issue a cert:

```sh
./acme.sh --issue --dns dns_dnsmint -d q7k4m2.example.dev -d *.q7k4m2.example.dev
```

The key is saved on the first issue and reused on renewal, so an unattended renewal needs nothing in its environment.

Report any bugs or issues [here](https://github.com/acmesh-official/acme.sh/issues/7238)

kxbnb:
Correction to my previous comment, and a commit fixing what it exposed.

The wiki text I posted said `DNSMint_Api` overrides the API base url. **It did not.** The assignment was unconditional:

```sh
DNSMint_Api="https://dnsmint.com/api"
```

so exporting it was overwritten every time the plugin was sourced. Every neighbouring plugin with an `_Api` variable treats it as an override; this one only looked like it did. Pushed `4f361c8`, which makes it one:

```sh
DNSMint_Api="${DNSMint_Api:-https://dnsmint.com/api}"
```

The wiki entry above is now accurate as written. `DNSMINT_API_KEY` persistence was already correct — the plugin reads it with `_readaccountconf_mutable` and saves it with `_saveaccountconf_mutable`, so an unattended renewal needs nothing in its environment.

kxbnb:
Wiki entry added as 227, and #7251 is the tracking issue step 9 asks for.

shfmt wanted the two `case` blocks a level shallower; fixed in `6bed53d0`, so the red check above is stale.

neilpang:
1. dnsapi/dns_dnsmint.sh:207: grep -F. Solaris /usr/bin/grep has no -F
   ("illegal option -- F"). Match the value with a case pattern on each
   line instead (case "$_line" in *"\"$_value\""*), and keep the case
   statement outside $(...) -- old shells choke on the ")" there.

2. Same line: sed 's/},{/}\n{/g'. "\n" in the replacement is a GNU
   extension; BSD/macOS sed inserts a literal "n", so the response stays
   on one line, the whole line matches the value, and _head_n 1 returns
   the id of the FIRST record under the hostname -- rm deletes the wrong
   record. Use a literal newline in the replacement (see dns_glesys.sh).
   The DNS test cannot catch either: a failed lookup falls into "Record
   already gone" and returns 0.

3. Issues: -> github.com/acmesh-official/acme.sh/issues/7251 (not your
   fork).

4. Docs: -> github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_dnsmint
   (the entry is on dnsapi2).

5. Re-run the DNS test after the fixes.

kxbnb:
Fixed in `61ef816c`.

`grep -F` is gone; the value is matched with a `case` pattern per line, and the `case` is outside the command substitution. The replacement now carries a literal newline as `dns_glesys.sh` does.

The second one is worse than a portability nit. Without the split the whole reply stays on one line, so the match succeeds whatever the value is and the id taken is the first record under the hostname — removal deletes somebody else's TXT record rather than the challenge:

```
old, sed that did not split:  rec_AAA
new:                          rec_BBB
the record holding the value: rec_BBB
```

`Docs:` is now `dnsapi2#dns_dnsmint` and `Issues:` is `acmesh-official/acme.sh/issues/7251`.

DNS test on the fixed commit: https://github.com/dnsmint/acme.sh/actions/runs/34742177450

neilpang:
1. _dnsmint_record_rm: printf "%s" "$response" | sed ... feeds sed a
   final line with no newline. Oracle Solaris /usr/bin/sed drops such a
   line, so the last record is never seen and rm silently reports
   "already gone". Use echo "$response" like the rest of the file.

2. The id extraction lost the | _head_n 1. If a record object ever
   carries a nested "id" under "data", _rid becomes two lines and the
   DELETE URL breaks. Keep _head_n 1.

kxbnb:
Both fixed in 2019d0d9.

1. `echo "$response"`, matching lines 154 and 172 in the same file. `printf "%s"` left the reply with no trailing newline, and that one line is the whole reply, so dropping it empties the loop and every removal reports the record already gone.

2. `_head_n 1` is back on the id extraction.

`shellcheck` and `shfmt -i 2` are clean. Removal logic exercised under sh, dash, bash and ksh for four cases: target is the last record, target is the first, a record object carrying a nested `"id"` under `"data"`, and the value absent.

Issue #7241, opened 2026-09-09
Report bugs to Optidata Cloud DNS API (dns_optidata)


EduardoLangner:
This is the place to report bugs in the Optidata Cloud DNS API (`dns_optidata`).

If you experience a bug, please report it in this issue.

Thanks!

PR #7243, opened 2026-09-09
Add Optidata Cloud DNS API (dns_optidata)

EduardoLangner:
## Summary

This PR adds `dns_optidata.sh`, an integration with the [Optidata Cloud](https://console.optidata.com) DNS API, authenticated with a DNS-scoped API key sent in the `x-api-key` header.

The hook is a port of `dns_cf.sh`. Every route lives under `$OPTIDATA_Api/api/v1`:

- `GET dns-zones?name=<record fqdn>` — resolve the zone containing the name
- `GET dns-zones/<zone_id>` — read one zone, when `OPTIDATA_Zone_ID` pins it
- `POST dns-zones/<zone_id>/recordsets` — create/upsert the TXT record set
- `DELETE dns-zones/<zone_id>/recordsets?name&type&value` — remove one TXT value

## Options

| Variable | Required | Purpose |
| --- | --- | --- |
| `OPTIDATA_Token` | yes | DNS API key created in the Optidata Console (API Keys); starts with `ocs_` |
| `OPTIDATA_Api` | no | API base url; defaults to `https://console.optidata.com` |
| `OPTIDATA_Location` | no | Location code or UUID of the zone; only needed when the zone lives outside the account default location |
| `OPTIDATA_Zone_ID` | no | Pins the zone and skips the zone lookup, so a zone-restricted key can be used for a single certificate |

Following `dns_cf.sh`, a pinned zone saves the key in the domain config; otherwise it goes to the account config. Both are reused on renewal, so an unattended renewal needs nothing in its environment.

## Zone resolution

The zone is verified locally instead of trusting the API's `name` filter: `_optidata_pick_zone` selects the longest zone name that is the fqdn itself or one of its parents, and `_get_root` then re-checks that the selected zone is a suffix of the record name. A zone whose name merely contains the challenge name cannot be selected, and an API that ignored the filter still resolves to the right zone.

## Record handling

A record set is unique per `(name, type)` and the apex and wildcard challenges of one certificate share the challenge name, so `add` merges the second value into the existing set (`upsert`) and `rm` deletes a single value, leaving the other challenge in place. `require_active_zone` makes the API answer 409 when the zone is not delegated yet, instead of queueing a record that would only be published long after the ACME server gave up. HTTP 429/502/503/504 are retried up to three times, honouring `Retry-After` (capped at 60s).

## Portability

- `_H1` to `_H5` are all assigned, since hooks share one shell and `_get`/`_post` always send every slot — a previous provider's auth header must not reach this endpoint.
- `$HTTP_HEADER` is truncated before each attempt, so a transport failure cannot be read as the previous request's status code.
- Every `sed` and `_egrep_o` pipeline is fed with a trailing newline (`printf "%s\n"`), because Solaris `/usr/bin/sed` silently drops a line without one and `_egrep_o` falls back to `sed` where `egrep -o` is missing. `_url_encode` inputs keep `printf "%s"` on purpose, so no `%0A` is appended.

## Pipeline

DNS API test passing in my fork, all platforms:

https://github.com/optidatacloud/acme.sh/actions/runs/34360015350

## Process

- Bug report tracking issue: #7241, referenced in the info block's `Issues:` field
- Wiki usage: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_optidata
- `shellcheck -e SC2181 -e SC2089` and `shfmt -i 2` (v3.1.2) are clean

## Wiki Content Update

https://github.com/acmesh-official/acme.sh/pull/7243#issuecomment-5605505721

### DIFF
--- a/dnsapi/dns_optidata.sh
+++ b/dnsapi/dns_optidata.sh
@@ -1,0 +1,440 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_optidata_info='Optidata Cloud
+Site: console.optidata.com
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_optidata
+Options:
+ OPTIDATA_Token DNS API key. Create a key of kind DNS in the Optidata Console (API Keys); it starts with "ocs_".
+ OPTIDATA_Api API base URL. Default "https://console.optidata.com".
+ OPTIDATA_Location Location code or UUID of the zone. Only needed when the zone lives outside the account default location and the lookup cannot tell. Optional.
+ OPTIDATA_Zone_ID Zone ID. Pins the zone and skips the zone lookup. Optional.
+Issues: github.com/acmesh-official/acme.sh/issues/7241
+Author: Eduardo Langner <https://github.com/optidatacloud>
+'
+
+# Port of dnsapi/dns_cf.sh (CloudFlare) to the Optidata Cloud DNS API served by
+# ocs-backend. Routes used, all under $OPTIDATA_Api/api/v1 and authenticated
+# with the x-api-key header:
+#
+#   GET    dns-zones?name=<record fqdn>            resolve the zone containing the name
+#   GET    dns-zones/<zone_id>                     read one zone (OPTIDATA_Zone_ID)
+#   POST   dns-zones/<zone_id>/recordsets          create / upsert the TXT record set
+#   DELETE dns-zones/<zone_id>/recordsets?name&type&value   remove one TXT value
+#
+# Every response is wrapped in {"success":true,"data":...}; errors are flat
+# {"status_code":<n>,"message":"...","error":"..."}.
+
+OPTIDATA_DEFAULT_API="https://console.optidata.com"
+OPTIDATA_TTL=120
+OPTIDATA_MAX_ATTEMPTS=3
+
+########  Public functions #####################
+
+#Usage: dns_optidata_add  _acme-challenge.www.domain.com  "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_optidata_add() {
+  fulldomain=$1
+  txtvalue=$2
+  _info "Using Optidata Cloud DNS API"
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _optidata_load_config; then
+    return 1
+  fi
+  _optidata_save_config
+
+  _debug "First detect the root zone"
+  if ! _get_root "$fulldomain"; then
+    _err "invalid domain"
+    return 1
+  fi
+  _debug _domain_id "$_domain_id"
+  _debug _sub_domain "$_sub_domain"
+  _debug _domain "$_domain"
+
+  _info "Adding TXT record $fulldomain"
+  # A record set is unique per (name, type) and the apex and wildcard
+  # challenges share the same name, so the second value has to be merged into
+  # the existing set: that is what upsert does. require_active_zone makes the
+  # API fail now (409) instead of queueing a record that would only be
+  # published after delegation, long after the ACME server gave up.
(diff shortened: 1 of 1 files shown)


EduardoLangner:
Read the [DNS API Dev Guide](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide) and the [DNS-API-Test](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Test). This PR targets `dev`.

- DNS API test passing in my fork, all platforms: https://github.com/optidatacloud/acme.sh/actions/runs/34360015350
- Bug report tracking issue: #7241, referenced in the info block's `Issues:` field

Wiki content (https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_optidata) follows:

***

## 225. Use Optidata Cloud DNS API
<a name="dns_optidata"/>

Create an API key of kind **DNS** in the [Optidata Console](https://console.optidata.com) under *API Keys*. The key starts with `ocs_` and needs the `dns_zones` scope with the `zones_read`, `records_create`, `records_update` and `records_delete` permissions, plus access to the zone you are validating.

```sh
export OPTIDATA_Token="ocs_xxxxxxxxxxxxxxxx"
```

Optional settings:

- `OPTIDATA_Api` overrides the API base url, which defaults to `https://console.optidata.com`.
- `OPTIDATA_Location` is the location code or UUID of the zone. Only needed when the zone lives outside the account default location and the zone lookup cannot tell.
- `OPTIDATA_Zone_ID` pins the zone and skips the zone lookup, so a zone-restricted key can be used for a single certificate.

To issue a cert:

```sh
./acme.sh --issue --dns dns_optidata -d example.com -d '*.example.com'
```

The settings are saved on the first issuance and reused on renewal, so an unattended renewal needs nothing in its environment.

The zone must already be delegated to the Optidata name servers and be `ACTIVE`; otherwise the API rejects the challenge record with HTTP 409 instead of queueing a record the ACME server would never see.

Report any bugs or issues [here](https://github.com/acmesh-official/acme.sh/issues/7241)

neilpang:
1. dnsapi/dns_optidata.sh line 64: grep -F is not portable. Solaris
   /usr/bin/grep has no -F (and no --): "grep: illegal option -- F".
   The DNS test only passed there because CI puts /usr/gnu/bin first
   in PATH. txtvalue is base64url ([A-Za-z0-9_-]), so the JSON escape
   is a no-op; match it with a case pattern instead:

     case "$response" in
     *"$txtvalue"*)
       _info "Added, OK"
       return 0
       ;;
     esac
     _err "The API accepted the record but the value is missing from the record set: $response"
     return 1

2. _get_root: what is the "internal" special case in
   `! _contains "$_domain" "internal"`? It is not documented anywhere;
   either document it in the info block or drop it.

EduardoLangner:
Thanks for the review. Both points are addressed in 1544e366.

**1. `grep -F` portability** — applied exactly as you suggested. You are right that the escape was a no-op there, so the match no longer goes through `_optidata_json_escape` at all:

```sh
    # The API echoes the whole record set back. The value is base64url
    # ([A-Za-z0-9_-]), so it needs no JSON escaping and a case pattern matches
    # it without depending on a grep that supports -F (Solaris grep does not).
    case "$response" in
    *"$txtvalue"*)
      _info "Added, OK"
      return 0
      ;;
    esac
    _err "The API accepted the record but the value is missing from the record set: $response"
    return 1
```

I also swept the rest of the file: no other `grep -F` or `--` is left, the remaining `grep` calls use only POSIX flags (`-i`) and plain BREs, so nothing else depends on `/usr/gnu/bin` being ahead in `PATH`.

**2. The `internal` special case** — dropped. It only silenced the `_info` notice for zones whose name contained `internal`, and it does not deserve to be documented either: a dns-01 challenge has to be publicly resolvable, so a private/internal zone can never validate a certificate through this hook. There was no real case for it to protect, and as written it was a loose substring test (`myinternalzone.com` matched too). The status notice is now unconditional:

```sh
  if [ "$_zone_status" ] && [ "$_zone_status" != "ACTIVE" ]; then
```

Verified locally before pushing:

- `sh -n dnsapi/dns_optidata.sh` — clean
- `shellcheck -e SC2181 -e SC2089` (same flags as the Shellcheck workflow) — no warnings
- `shfmt -i 2` (v3.1.2, same as CI) — no diff

PR #7244, opened 2026-09-09
Add Opteamax DNS API

opteamax:
Adds a DNS API for [Opteamax](https://www.opteamax.de), an ISP/hosting operator
whose customers manage their zones through the Opteamax customer API.

Usage:

```sh
export OPTEAMAX_Token="oxt_..."      # created in the customer panel under "API-Tokens"
acme.sh --issue --dns dns_opteamax -d example.com -d '*.example.com'
```

`OPTEAMAX_Api` optionally overrides the endpoint (default
`https://api.opteam.ax/api/v2`). Both are saved with
`_saveaccountconf_mutable` so renewals reuse them.

Notes on the implementation:

- The root zone is detected by listing the account's zones once and walking
  the name up against them, so subzones, multi-label TLDs and DNS alias mode
  all resolve to the right zone (longest match wins).
- `dns_opteamax_rm` deletes only the record carrying the value it was handed,
  so the two TXT records of a wildcard order do not remove each other.
- Only `_get`/`_post` and the acme.sh string helpers are used; no `awk`, no
  direct `curl`/`wget`.

Tested against Let's Encrypt staging for `acme-plugin-test.opteamax.de` and
`*.acme-plugin-test.opteamax.de`: both challenge records were added, validated
and removed again. `shellcheck -e SC2181 -e SC2089` and `shfmt -i 2` are clean.

I will add the wiki page under `dnsapi2` and open the tracking issue for the
`Issues:` line once this is reviewed.

### DIFF
--- a/dnsapi/dns_opteamax.sh
+++ b/dnsapi/dns_opteamax.sh
@@ -1,0 +1,229 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC2034
+dns_opteamax_info='Opteamax.de
+Site: opteamax.de
+Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_opteamax
+Options:
+ OPTEAMAX_Token API token. Create it in the customer panel under "API-Tokens"; it starts with "oxt_".
+ OPTEAMAX_Api API endpoint. Default "https://api.opteam.ax/api/v2". Optional.
+Issues: github.com/acmesh-official/acme.sh/issues/7245
+Author: Jens Ott <jo@opteamax.de>
+'
+
+OPTEAMAX_Api_Default="https://api.opteam.ax/api/v2"
+
+######## Public functions #####################
+
+#Usage: dns_opteamax_add _acme-challenge.www.domain.com "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_opteamax_add() {
+  fulldomain=$1
+  txtvalue=$2
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _opteamax_init; then
+    return 1
+  fi
+
+  _debug "First detect the root zone"
+  if ! _get_root "$fulldomain"; then
+    return 1
+  fi
+  _debug _domain "$_domain"
+  _debug _domain_id "$_domain_id"
+
+  _info "Adding the TXT record"
+  _opteamax_body="{\"type\":\"TXT\",\"name\":\"$fulldomain.\",\"content\":\"$txtvalue\",\"ttl\":300}"
+  if ! _opteamax_rest POST "/dns/domains/$_domain_id/records/" "$_opteamax_body"; then
+    return 1
+  fi
+  if ! _contains "$response" "data_id"; then
+    _err "Could not add the TXT record: $response"
+    return 1
+  fi
+
+  _info "TXT record added"
+  return 0
+}
+
+#Usage: dns_opteamax_rm _acme-challenge.www.domain.com "XKrxpRBosdIKFzxW_CT3KLZNf6q0HG9i01zxXp5CPBs"
+dns_opteamax_rm() {
+  fulldomain=$1
+  txtvalue=$2
+  _debug fulldomain "$fulldomain"
+  _debug txtvalue "$txtvalue"
+
+  if ! _opteamax_init; then
+    return 1
+  fi
+
+  _debug "First detect the root zone"
+  if ! _get_root "$fulldomain"; then
+    return 1
+  fi
+  _debug _domain "$_domain"
+  _debug _domain_id "$_domain_id"
+
+  _debug "Getting the record id"
+  if ! _opteamax_rest GET "/dns/domains/$_domain_id/records/"; then
+    return 1
+  fi
+
+  # Only this challenge's record may go: a wildcard certificate puts two TXT
+  # values on the same name, and acme.sh removes them one call at a time.
+  # PowerDNS hands TXT content back in wire format, so the value arrives inside
+  # escaped quotes ("content": "\"<value>\""); matching from the field name up
+  # to the next field keeps the value pinned to the content field without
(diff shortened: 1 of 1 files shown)


opteamax:
Replying as asked. Everything on the checklist is done:

- **Branch**: this PR targets `dev`.
- **DNS API Dev Guide**: read and followed — `_get_root` walks the name up against the account's zone list (so subzones, multi-label TLDs and DNS alias mode all work), only `_get`/`_post` and the acme.sh helpers are used, no `awk` or direct `curl`, and `_rm` deletes only the record carrying the value it was handed, so a wildcard's two TXT records survive each other.
- **DNS-API-Test**: passed on every platform — https://github.com/opteamax/acme.sh/actions/runs/34411637670 (Docker, macOS, Windows, FreeBSD, GhostBSD, OpenBSD, NetBSD, DragonFlyBSD, MidnightBSD, HardenedBSD, OPNsense, Solaris, Omnios, OpenIndiana, Tribblix, Haiku, Hurd, OpenEuler).
- **Bug tracking issue**: #7245, and its link is in the `Issues:` line of `dns_opteamax_info`.
- **Structural info description**: present at the top of the script.
- `shellcheck -e SC2181 -e SC2089` and `shfmt -i 2` are clean.

The one item I cannot do myself is the wiki: pushing to `acme.sh.wiki` is refused for my account, so a maintainer would need to add this (or grant wiki access and I will add it). It belongs right before "Use custom API", numbered after the last entry:

```markdown
## 226. Use Opteamax DNS API
<a name="dns_opteamax"/>

Create an API token in the [Opteamax panel](https://panel.opteam.ax) under *API-Tokens*. The token starts with `oxt_`, and the user it belongs to needs DNS access to the zone you are validating. A token can be restricted to individual zones, which is enough for issuing a certificate.

```sh
export OPTEAMAX_Token="oxt_xxxxxxxxxxxxxxxxxxxxxxxx"
```

Optional settings:

- `OPTEAMAX_Api` overrides the API base url, which defaults to `https://api.opteam.ax/api/v2`.

To issue a cert:

```sh
./acme.sh --issue --dns dns_opteamax -d example.com -d '*.example.com'
```

The settings are saved on the first issuance and reused on renewal, so an unattended renewal needs nothing in its environment.

The zone is detected from the account's own zone list, so a delegated subzone is used in preference to its parent, and DNS alias mode works as well.

Report any bugs or issues [here](https://github.com/acmesh-official/acme.sh/issues/7245)
```

**Comment by neilpang** [comment]
1. The wiki entry is added by the PR author, Use
   the Edit button on
   https://github.com/acmesh-official/acme.sh/wiki/dnsapi2 (the web
   editor works without git push access) and add the section you
   pasted above. The PR is merged once
   https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_opteamax
   resolves.
2. dns_opteamax_rm, `grep -- "$txtvalue"`: `--` is not portable
   (Solaris grep). Use `grep "\"content\": *\"$txtvalue\""` instead;
   the pattern then starts with a quote, so no `--` is needed, and it
   only matches the content field.
3. `_record_id` needs `| _head_n 1`. `add` does not check for an
   existing record, so a leftover from an aborted run gives two ids
   and a DELETE URL with a newline in it.
4. `_get_root` reads `/dns/domains/` once. Is that endpoint paginated
   (DRF style `count`/`next`/`results`)? If so, accounts with more
   zones than one page will not find their zone.
5. Lines 161 and 203: replace the em-dash in the comments with `--`
   (keep the file ASCII).
6. Line 108 `OPTEAMAX_Token=""` inside the `-z` branch is a no-op,
   drop it.

**Comment by opteamax** [comment]
Thanks — all six addressed, force-pushed as one commit.

**2. `grep -- "$txtvalue"`** — fixed, though not with the exact pattern you suggested: PowerDNS hands TXT content back in wire format, so the value arrives inside escaped quotes and the field reads `"content": "\"<value>\""`. `"\"content\": *\"$txtvalue\""` therefore never matches. I used

```sh
grep "\"content\":[^,]*$txtvalue"
```

which starts with a quote (so no `--`), stays inside the content field, and does not need the backslashes spelled out.

**3. `_head_n 1`** — added. You are right that `add` does not check first; I also pinned it with a test that seeds the same value twice and fails without the `_head_n 1` (two ids produced a DELETE path with a newline and the records stayed).

**4. Pagination** — no. `GET /dns/domains/` returns a bare JSON array, not a DRF-style envelope; there is no `count`/`next`/`results` and no page limit, so one request always carries the account's whole zone list:

```
[{"domain_id": 895, "domain": "opteam.ax", "is_ptr": false}, {"domain_id": 186, "domain": "opteamax.de", ...
```

**5. Em-dashes** — gone; `LC_ALL=C grep '[^ -~]'` now reports nothing for the whole file.

**6. `OPTEAMAX_Token=""`** — dropped.

**1. Wiki** — understood, thanks for the pointer about the web editor; the section is being added there now and I will confirm here once
https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_opteamax resolves.

`shellcheck -e SC2181 -e SC2089` and `shfmt -i 2` stay clean, and the DNS-API-Test is running again on the updated branch: https://github.com/opteamax/acme.sh/actions/runs/34495998160

opteamax:
The wiki entry is in: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_opteamax resolves (section 226, listed in the page's table of contents, immediately before "Use custom API").

The DNS-API-Test re-run on the reviewed branch is working through the platforms — Docker, macOS and the token check are green so far: https://github.com/opteamax/acme.sh/actions/runs/34495998160 — I will report here if anything turns red.

opteamax:
The re-run finished green on every platform: https://github.com/opteamax/acme.sh/actions/runs/34495998160 (Docker, macOS, Windows, FreeBSD, GhostBSD, OpenBSD, NetBSD, DragonFlyBSD, MidnightBSD, HardenedBSD, OPNsense, Solaris, Omnios, OpenIndiana, Tribblix, Haiku, Hurd, OpenEuler). Nothing turned red, so the review changes are covered by the test as well.

PR #7250, opened 2026-09-11
fix: honor account email precedence and registration semantics

go-goll:
## Honor explicit account email and keep the per-CA email authoritative

  When `account.conf` contains `ACCOUNT_EMAIL`, `--update-account -m new@example.com` submits the old
  address and still prints `Account update success`. Argument parsing sets `ACCOUNT_EMAIL`, but
  `_initpath` then sources `account.conf` and overwrites it. The `-m`, `--email` and `--accountemail`
  aliases share this path, and `--register-account` is affected too.

  Changes:

  - `_getAccountEmail`: use the command-line value, then the per-CA `CA_EMAIL`, then the global
    `ACCOUNT_EMAIL`. Without the per-CA step, the address saved by a successful update was shadowed
    by `account.conf`, and the next `--update-account` without `-m` pushed the old one back.
  - `_clearCA` resets `CA_EMAIL`, so a renewal that switches CA does not reuse the previous CA's address.
  - `_regAccount` saves `CA_EMAIL` only when a new account is created (201). A `newAccount` request for
    an existing key returns 200 and ignores the contact (RFC 8555 7.3.1). In that case it prints a hint to
    use `--update-account -m` instead.

  Behavior change: once a CA has `CA_EMAIL` saved, editing `ACCOUNT_EMAIL` in `account.conf` or re-running
  `--install -m` no longer changes that CA's contact; use `--update-account -m`. CAs without `CA_EMAIL`
  still fall back to the global address. The mail notify hook still reads the global `ACCOUNT_EMAIL`.

  Validation (at efb4c9bf):

  - Pebble v2.10.1, with `ACCOUNT_EMAIL=old` in `account.conf`: `--update-account -m new` now sets the
    contact to new (before: old); a following `--update-account` without `-m` keeps new; `--register-account -m`
    on an existing account returns 200, leaves the contact unchanged, and prints the hint.
  - Stubbed cases under sh/dash/bash/zsh: cross-CA renewal, a stray `ACCOUNT_EMAIL` in `ca.conf`,
    an address containing `'`, and `CA_EMAIL` passed through the environment.
  - `sh -n`, `dash -n` and `git diff --check` pass. Not tested against Let's Encrypt staging;
    ShellCheck and shfmt were not run locally.

  Target branch: `dev`.

### DIFF
--- a/acme.sh
+++ b/acme.sh
@@ -3373,6 +3373,7 @@ _initAPI() {
 
 _clearCA() {
   export CA_CONF=
+  CA_EMAIL=
   export ACCOUNT_KEY_PATH=
   export ACCOUNT_JSON_PATH=
 }
@@ -4283,15 +4284,22 @@ _mailto_contacts() {
 }
 
 _getAccountEmail() {
-  if [ "$ACCOUNT_EMAIL" ]; then
-    echo "$ACCOUNT_EMAIL"
+  # Config loading can overwrite ACCOUNT_EMAIL after command-line parsing.
+  if [ "$_accountemail" ]; then
+    echo "$_accountemail"
     return 0
   fi
+  # The per-CA email must win over the global one, or an email saved by
+  # --update-account is shadowed by account.conf on the next run.
   if [ -z "$CA_EMAIL" ]; then
     CA_EMAIL="$(_readcaconf CA_EMAIL)"
   fi
   if [ "$CA_EMAIL" ]; then
     echo "$CA_EMAIL"
+    return 0
+  fi
+  if [ "$ACCOUNT_EMAIL" ]; then
+    echo "$ACCOUNT_EMAIL"
     return 0
   fi
   _readaccountconf "ACCOUNT_EMAIL"
@@ -4327,9 +4335,6 @@ _regAccount() {
   _secure_debug3 _eab_kid "$_eab_kid"
   _secure_debug3 _eab_hmac_key "$_eab_hmac_key"
   _email="$(_getAccountEmail)"
-  if [ "$_email" ]; then
-    _savecaconf "CA_EMAIL" "$_email"
-  fi
 
   if [ "$ACME_DIRECTORY" = "$CA_ZEROSSL" ]; then
     if [ -z "$_eab_kid" ] || [ -z "$_eab_hmac_key" ]; then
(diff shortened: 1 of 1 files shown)


neilpang:
1. With the save moved under 201, an address given only via -m is lost
   whenever nothing is stored yet:
   - first ZeroSSL --issue -m fails at the EAB fetch / newAccount: the
     domain conf is already written, so every cron retry runs with no
     email and stops at "Please update your account with an email
     address first" until -m is passed again by hand (dev saved
     CA_EMAIL before the request and recovered by itself);
   - legacy account, ca.conf without CA_EMAIL: --register-account -m x
     (200) then a plain --update-account sends {"contact": []} and
     clears the contact; the hint also repeats on every run.

   Keep a first-contact save and the rest of the logic stands:

   ```sh
   _email="$(_getAccountEmail)"
   if [ "$_email" ] && [ -z "$(_readcaconf CA_EMAIL)" ]; then
     _savecaconf "CA_EMAIL" "$_email"
   fi
   ```

2. Help text for -m still says "only valid for the '--install' and
   '--update-account' command"; both hints now point at
   --register-account -m.

Precedence change itself is fine; reproduced the old behaviour and the
fix with a stubbed _send_signed_request under dash and bash.

go-goll:
> 1. With the save moved under 201, an address given only via -m is lost
>    whenever nothing is stored yet:
>    
>    * first ZeroSSL --issue -m fails at the EAB fetch / newAccount: the
>      domain conf is already written, so every cron retry runs with no
>      email and stops at "Please update your account with an email
>      address first" until -m is passed again by hand (dev saved
>      CA_EMAIL before the request and recovered by itself);
>    * legacy account, ca.conf without CA_EMAIL: --register-account -m x
>      (200) then a plain --update-account sends {"contact": []} and
>      clears the contact; the hint also repeats on every run.
>    
>    Keep a first-contact save and the rest of the logic stands:
>    ```shell
>    _email="$(_getAccountEmail)"
>    if [ "$_email" ] && [ -z "$(_readcaconf CA_EMAIL)" ]; then
>      _savecaconf "CA_EMAIL" "$_email"
>    fi
>    ```
> 2. Help text for -m still says "only valid for the '--install' and
>    '--update-account' command"; both hints now point at
>    --register-account -m.
> 
> Precedence change itself is fine; reproduced the old behaviour and the fix with a stubbed _send_signed_request under dash and bash.

  Both reproduced (my earlier check ran the two commands in one shell, where the exported
  `ACCOUNT_EMAIL` from `-m` leaked into the second one and hid it):

  - legacy account without `CA_EMAIL`: a plain `--update-account` after `--register-account -m x`
    sent `{"contact": []}`;
  - ZeroSSL: after a failed EAB fetch, the next run stopped at "Please update your account with an
    email address first".

  Applied your first-contact save and kept the 201 save, so re-registering with a different address
  still updates the stored one. The legacy case now sends `x`, the hint no longer repeats, and the
  ZeroSSL retry fails only on the EAB fetch as on dev. Help text for `-m` now mentions
  `--register-account`.

  Re-verified against Pebble v2.10.1 plus the stubbed cases under dash, bash and zsh.

PR #7254, opened 2026-09-12
Add JetKVM SSH deploy hook

daemonhorn:
## Summary

Adds `deploy/jetkvm.sh`, a deploy hook to push an ACME-issued certificate
to a [JetKVM](https://jetkvm.com) KVM-over-IP device over plain SSH:

```sh
export DEPLOY_JETKVM_HOST="jetkvm.example.com"
acme.sh --deploy -d jetkvm.example.com --deploy-hook jetkvm
```

## Why this approach

JetKVM only supports key-based SSH authentication (`root@<device>`,
password logins disabled), enabled via "Developer Mode" and a pasted
public key in the device's web UI (Settings > Advanced). SSH keys must
already be exchanged before using this hook, same expectation as
`deploy/ssh.sh` and `deploy/routeros.sh`.

JetKVM's minimal userspace does not ship an `scp` binary or SFTP server,
so unlike `deploy/ssh.sh` this hook has no "use scp" option. Instead it
opens a plain SSH exec session and pipes a small POSIX shell script to
the remote `sh` via stdin (only depends on `sh`/`cat`/`chmod`/`mkdir`/`mv`/`rm`
on the device side). The remote path, filenames, and file permissions are
fixed firmware constants on this single-purpose, single-root appliance,
not user-configurable -- only `USER`/`HOST`/`PORT`/`SSH_CMD`/`RESTART_CMD`/
`REQUIRE_CUSTOM_MODE` are.

JetKVM's "Custom" TLS mode (device web UI: Settings > Network > HTTPS
Mode) reads the certificate/key from a fixed path and does not
hot-reload -- the device's own certificate-apply script does a full
reboot to pick up a new "Custom" certificate. This hook defaults its
post-upload command to `reboot` for that reason, since it's meant to run
unattended from cron-driven renewals. A blank `DEPLOY_JETKVM_RESTART_CMD`
is deliberately *not* treated as "skip" -- it falls back to `reboot` --
since a renewed certificate that never actually gets applied defeats the
point of automating this; set it to the literal value `none` to opt out.
The restart command is run detached on the device (`nohup ... &`) so this
hook's ssh call can return as soon as it's launched, before the reboot
itself lands, rather than racing the connection teardown. The tradeoff:
since it then runs as an unwaited background job, this hook cannot tell
whether it actually succeeded -- only an outright SSH connection failure
(unreachable host, auth failure, etc.) is caught; a bad restart command,
or even a missing `nohup`, still reports success once launched.

**HTTPS Mode precondition check.** Uploading a certificate that mode
won't even serve was otherwise a silent no-op. JetKVM's own
`getTLSState`/`setTLSState` JSON-RPC calls require an authenticated
WebRTC session (see
[jetkvm/kvm#1240](https://github.com/jetkvm/kvm/issues/1240) and the
still-open [jetkvm/kvm#1515](https://github.com/jetkvm/kvm/pull/1515)),
so there's no documented/headless way to query this yet. Its firmware
does persist the mode as a plain JSON field, `tls_mode` (values `""`,
`"self-signed"`, or `"custom"`), in `/userdata/kvm_config.json` -- this
hook checks that file is readable and greps it before writing anything,
reporting the config-file-missing and mode-not-custom cases as distinct,
specific errors (`DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE=no` opts out of the
check entirely, since this file/field is exactly as undocumented as
everything else this hook depends on and could change in a future
firmware version).

The certificate and key are staged under fixed temporary filenames on
the device and only `mv`-ed into their final names (an atomic rename)
once both have been fully written and `chmod`-ed. This keeps a dropped
connection or a failed write from ever leaving the device with a
truncated or mismatched cert/key pair for its own HTTPS listener, and a
`trap ... EXIT` in the generated script removes any leftover staged file
however that script exits.

## Background

This factors out the SSH upload logic originally proposed as an OPNsense
ACME Client plugin automation in
[opnsense/plugins#5621](https://github.com/opnsense/plugins/pull/5621),
per [maintainer feedback there](https://github.com/opnsense/plugins/pull/5621#issuecomment-5570647257):

> This PR suggests to add all the code/logic to the Acme Client plugin. A
> better approach would be to add a new deploy hook to acme.sh. Then this
> new deploy hook could be utilized by the Acme Client plugin with
> minimal code. This would lower the (code) maintenance burden.

Once this is merged, a much smaller follow-up PR will be opened against
`opnsense/plugins` (tracked in
[opnsense/plugins#5622](https://github.com/opnsense/plugins/issues/5622))
that just wires up a thin automation to invoke this hook, rather than
duplicating the SSH/remote-script logic in plugin-specific PHP.

None of the defaults here (storage path, filenames, permissions, the
`kvm_config.json` path/field, reboot-to-apply behavior) are part of
JetKVM's stable/documented API -- they come from a mix of hardware
testing (see opnsense/plugins#5621's description) and reading the
open-source firmware source (`web_tls.go`/`config.go` in
[jetkvm/kvm](https://github.com/jetkvm/kvm)), and are worth a spot-check
after a JetKVM firmware upgrade.

## Review

@neilpang reviewed the first version of this hook and flagged 12 points,
all now addressed. Ten in code, as of the latest commit:
POSIX portability (dropped `[[:space:]]`/`-E`, not read consistently
across `grep`/`sed` implementations), the dead `date +%s || echo 0`
fallback (replaced with the core `_time()` helper, matching every other
hook/dnsapi script that needs a timestamp), local temp files holding the
private key (removed entirely -- the upload script is now piped directly
into `ssh ... sh`, same pattern as `deploy/windows_rdp.sh`), no input
validation for a `--signcsr`-only run (now refused up front), `echo`
instead of `printf` for generated script lines (fixed throughout --
dash's `echo` interprets backslash escapes), unsafe `_savedeployconf` of
values that can contain shell metacharacters (now saved with the
`base64` flag, as `deploy/docker.sh` does), the config-file-missing case
being misreported as "HTTPS Mode not custom" (now a distinct error),
unnecessary `DEPLOY_JETKVM_*` knobs for firmware constants (removed,
see "Why this approach" above), a restart-command race that could report
a successful deploy without the device ever rebooting (fixed via the
detached-restart rewrite above), and staged files stranded by an aborted
remote script (fixed with fixed filenames + `trap ... EXIT`).

The other two: the header now links to the
[deployhooks wiki page](https://github.com/acmesh-official/acme.sh/wiki/deployhooks),
which now has a JetKVM section.

**Second round** (commit `146a83b0`): two more points --
`DEPLOY_JETKVM_RESTART_CMD` was interpolated unescaped inside the
detached command's own `sh -c '...'` wrapper, so a value containing a
single quote (e.g. `sh -c 'sync; reboot'`) broke that quoting and only
part of the command ran, un-detached; now escaped before nesting
(verified against `sh` and `dash`, plus a smoke-test case that actually
executes the generated command and checks both halves ran). And the
"only a failure to launch it at all is caught" claim (both here and in
"Why this approach" above) was simply wrong -- measured 0 for both a
nonexistent restart command and a missing `nohup` binary, since the
remote shell doesn't wait on the backgrounded job; reworded to describe
what's actually caught instead.

## Testing done

- `shellcheck deploy/jetkvm.sh` -- clean, no exclusions needed.
- `shfmt -d -i 2 deploy/jetkvm.sh` -- no reformatting needed.
- Local smoke test (stubbed acme.sh core helpers, a fake `ssh` that
  redirects the hardcoded device paths into a scratch directory instead
  of a real device) covering: a clean deploy with byte-exact
  content/permissions and no leftover staged files; `RESTART_CMD=none`;
  HTTPS Mode being `"self-signed"` (certificate correctly not written);
  the config file missing entirely (reported distinctly, not blamed on
  HTTPS Mode); `REQUIRE_CUSTOM_MODE=no` skipping the check; an empty key
  file (`--signcsr` case) rejected before any ssh call; a failed
  restart-command *launch* reported as an error; and (added for the
  second review round) a restart command containing single quotes
  running both halves intact through the actual generated detached
  command, not just a string comparison.
- **Real JetKVM hardware**, run directly (not through acme.sh's own
  `--deploy`/`--issue` machinery), against the pre-review version of
  this hook (upload path unchanged since):
  - A full run with a throwaway self-signed test cert succeeded end to
    end -- upload, atomic rename, correct `0644`/`0600` permissions, a
    reboot triggered, and the device's HTTPS listener confirmed
    (post-reboot) serving that exact test certificate.
  - The `tls_mode` config-file check confirmed matching on the real
    device's busybox `grep`.
- **Re-verified against real hardware after the review remediation**,
  through a real OPNsense ACME Client pipeline using locally-built test
  packages:
  - The happy path (upload -> detached restart -> reboot -> certificate
    served) succeeded end to end.
  - HTTPS Mode not set to `"custom"` correctly fails with the expected,
    specific error message.
  - The config-file-missing case (exit code 4) is **not** re-verified on
    real hardware -- intentionally, to avoid renaming a live config file
    on the device. It's only covered by the local smoke test above.

### DIFF
--- a/deploy/jetkvm.sh
+++ b/deploy/jetkvm.sh
@@ -45,6 +45,22 @@
 # JetKVM hardware, but is worth a spot-check after a JetKVM firmware
 # upgrade.
 #
+# Before writing anything, this hook also checks that the device's HTTPS
+# Mode is already "Custom" -- uploading a certificate that mode won't
+# even serve would otherwise fail silently. There is currently no
+# documented/headless way to read this back (JetKVM's own JSON-RPC
+# getTLSState/setTLSState calls require an authenticated WebRTC session,
+# see https://github.com/jetkvm/kvm/issues/1240 and the still-open
+# https://github.com/jetkvm/kvm/pull/1515), so this greps the device's
+# own config file instead: JetKVM's firmware (see web_tls.go / config.go
+# in https://github.com/jetkvm/kvm) persists the mode as the plain-JSON
+# field "tls_mode" (values "", "self-signed", or "custom") in
+# /userdata/kvm_config.json. Like the rest of this hook's defaults, this
+# file path/field is undocumented/internal, not a stable public API, and
+# worth a spot-check after a firmware upgrade; set
+# DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE=no to skip the check entirely if a
+# future firmware version changes this format out from under it.
+#
 # The following variables exported from environment will be used. If not
 # set then values previously saved in the domain.conf file are used. All
 # of them are optional.
@@ -59,6 +75,8 @@
 # export DEPLOY_JETKVM_CHMOD_CERT="0644"                  # defaults to 0644
 # export DEPLOY_JETKVM_CHMOD_KEY="0600"                   # defaults to 0600
 # export DEPLOY_JETKVM_RESTART_CMD="reboot"               # defaults to "reboot" (JetKVM has no hot-reload); set to "none" to skip it
+# export DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE="yes"          # defaults to "yes" (verify tls_mode=custom before upload); set to "no" to skip
+# export DEPLOY_JETKVM_CONFIG_FILE="/userdata/kvm_config.json" # defaults to JetKVM's confirmed config file location
 #
 # Example:
 # ```sh
@@ -145,6 +163,18 @@ jetkvm_deploy() {
   fi
   _savedeployconf DEPLOY_JETKVM_RESTART_CMD "$DEPLOY_JETKVM_RESTART_CMD"
 
+  _getdeployconf DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE
+  if [ -z "$DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE" ]; then
+    DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE="yes"
+  fi
+  _savedeployconf DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE "$DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE"
+
+  _getdeployconf DEPLOY_JETKVM_CONFIG_FILE
+  if [ -z "$DEPLOY_JETKVM_CONFIG_FILE" ]; then
+    DEPLOY_JETKVM_CONFIG_FILE="/userdata/kvm_config.json"
+  fi
+  _savedeployconf DEPLOY_JETKVM_CONFIG_FILE "$DEPLOY_JETKVM_CONFIG_FILE"
+
   _info "Deploying certificate to JetKVM device $DEPLOY_JETKVM_USER@$DEPLOY_JETKVM_HOST:$DEPLOY_JETKVM_PORT"
 
   _jetkvm_remote_path="${DEPLOY_JETKVM_REMOTE_PATH%/}"
@@ -156,6 +186,7 @@ jetkvm_deploy() {
   _jetkvm_key_tmp="$_jetkvm_remote_path/.$DEPLOY_JETKVM_KEY_NAME.tmp.$_jetkvm_run_id"
   _jetkvm_cert_target="$_jetkvm_remote_path/$DEPLOY_JETKVM_CERT_NAME"
   _jetkvm_key_target="$_jetkvm_remote_path/$DEPLOY_JETKVM_KEY_NAME"
+  _jetkvm_mode_exitcode=3
 
   # Command substitution strips all trailing newlines, so the printf below
   # always emits the content with exactly one trailing newline before the
@@ -170,6 +201,16 @@ jetkvm_deploy() {
     echo "#!/bin/sh"
     echo "set -e"
     echo "umask 077"
+    if [ "$DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE" != "no" ]; then
+      # Uploading a certificate that HTTPS Mode won't even serve would
+      # otherwise fail silently -- see the header comment for why this
+      # greps the device's own config file rather than querying it
+      # through a documented API (there isn't one for reading this
+      # headlessly yet).
+      echo "if ! grep -Eq '\"tls_mode\"[[:space:]]*:[[:space:]]*\"custom\"' '$DEPLOY_JETKVM_CONFIG_FILE' 2>/dev/null; then"
+      echo "  exit $_jetkvm_mode_exitcode"
+      echo "fi"
+    fi
     echo "mkdir -p '$_jetkvm_remote_path'"
     echo "cat > '$_jetkvm_cert_tmp' <<'$_jetkvm_cert_marker'"
     printf '%s\n' "$_jetkvm_cert_content"
@@ -185,17 +226,24 @@ jetkvm_deploy() {
 
   _secure_debug "Generated upload script" "$(cat "$_jetkvm_upload_script")"
 
-  _info "Uploading certificate and key to $_jetkvm_remote_path on the device"
+  _info "Connecting to JetKVM device $DEPLOY_JETKVM_USER@$DEPLOY_JETKVM_HOST:$DEPLOY_JETKVM_PORT to deploy certificate"
   # shellcheck disable=SC2086
   $DEPLOY_JETKVM_SSH_CMD -p "$DEPLOY_JETKVM_PORT" "$DEPLOY_JETKVM_USER@$DEPLOY_JETKVM_HOST" sh <"$_jetkvm_upload_script"
   _ret=$?
 
   rm -f "$_jetkvm_upload_script"
 
+  if [ "$_ret" = "$_jetkvm_mode_exitcode" ]; then
+    _err "JetKVM HTTPS Mode is not set to \"Custom\" (checked \"tls_mode\" in $DEPLOY_JETKVM_CONFIG_FILE on the device). Set it in the device's web UI (Settings > Network > HTTPS Mode) before this hook can take effect. Certificate was NOT uploaded. Set DEPLOY_JETKVM_REQUIRE_CUSTOM_MODE=no to skip this check."
+    return $_ret
+  fi
+
   if [ "$_ret" != "0" ]; then
     _err "Error code $_ret returned uploading certificate to JetKVM device"
     return $_ret
   fi
+
+  _info "Certificate and key uploaded to $_jetkvm_remote_path on the device"
 
   if [ "$DEPLOY_JETKVM_RESTART_CMD" = "none" ]; then
     _info "Certificate successfully deployed to JetKVM device. DEPLOY_JETKVM_RESTART_CMD=none, skipping restart command."
--- a/deploy/jetkvm.sh
+++ b/deploy/jetkvm.sh
@@ -22,6 +22,17 @@
 # certificate. This hook defaults its restart command to "reboot" for that
 # reason, since it's meant to run unattended from cron-driven renewals
 # (typically overnight, when an active KVM-over-IP session is unlikely).
+# A renewed certificate that never actually gets applied without a human
+# rebooting the device defeats the point of automating this, so a blank
+# DEPLOY_JETKVM_RESTART_CMD is treated the same as unset (falls back to
+# "reboot") rather than silently skipping the restart -- set it to the
+# literal value "none" to opt out and apply/verify manually instead.
+#
+# Because the restart command normally tears the device down (and this
+# SSH connection along with it), it is run as a *second*, separate SSH
+# call after the upload has already completed and been confirmed -- see
+# the comments in jetkvm_deploy() for why the exit code of that second
+# call alone cannot be trusted to tell success from failure.
 #
 # The certificate and key are staged under temporary names on the device
 # and only renamed into their final names (an atomic "mv", on the same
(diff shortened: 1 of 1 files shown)


daemonhorn:
Update: this hook has now been validated end-to-end through a real OPNsense ACME Client pipeline, not just standalone against the device.

Built local test packages (`os-acme-client` + `acme.sh`, each adding only the JetKVM-related file(s) on top of the real published packages) and installed them on a real OPNsense 26.7 box. Used a new minimal ACME Client automation ("Upload certificate to JetKVM (SSH)", a thin wrapper that just invokes `acme.sh --deploy --deploy-hook jetkvm` -- the follow-up plugin change requested on opnsense/plugins#5621) to deploy a real Let's Encrypt certificate, issued by OPNsense itself, to the JetKVM device. Confirmed working end-to-end: issuance → this deploy hook → upload → HTTPS-mode check → reboot → certificate live on the device.

No changes needed to this PR as a result -- just closing the loop on validation now that both halves (this hook, and the plugin automation that will follow once this merges) have been exercised together for real.

neilpang:
1. L210: no POSIX character classes anywhere in this repo, remote script
   included (Solaris sed/grep read [[:space:]] as a literal bracket set).
   kvm_config.json is written by Go's json encoder, so
     grep -q '"tls_mode" *: *"custom"'
   covers both the indented and the compact form; -E is not needed.

2. L181: use the core wrapper: _jetkvm_run_id="$$.$(_time)". The
   "|| echo 0" branch is dead (a date without %s prints "%s", exit 0).

3. L195-234, L263-278: drop the local temp files and pipe the generated
   script straight into ssh (see deploy/windows_rdp.sh). _mktemp's
   fallback only echoes a /tmp name without creating it, so the chmod
   600 at L199 fails and L225 then creates the file holding the private
   key with the default umask. Piping keeps the key off local disk, $?
   stays ssh's exit code, and the rm -f goes away. The restart phase
   needs no file at all: pass "echo MARKER; $DEPLOY_JETKVM_RESTART_CMD"
   as the ssh command argument.

4. L195-196: guard the inputs first, e.g.
     if [ ! -s "$_ckey" ] || [ ! -s "$_cfullchain" ]; then _err ...; return 1; fi
   With --signcsr there is no key file; today that becomes an empty
   user-defined.key on the device followed by a reboot.

5. L268 and every generated line that embeds a config value: printf
   '%s\n', not echo. dash's echo interprets backslash escapes, so the
   script the device receives depends on which /bin/sh runs acme.sh.

6. L128, L164: save DEPLOY_JETKVM_SSH_CMD and DEPLOY_JETKVM_RESTART_CMD
   with the base64 flag (_savedeployconf X "$X" "base64", as
   deploy/docker.sh does). A value containing a single quote otherwise
   breaks the conf line and is eval'd on the host at the next renewal.

7. L210/L237: grep exit 2 (config file missing/unreadable) takes the
   same exit-3 path as "not custom", so the error blames the HTTPS Mode
   setting when the file has moved; test -r it separately. And do not
   advertise REQUIRE_CUSTOM_MODE=no in that error: following it makes
   every deploy a silent no-op (the firmware only serves the store when
   tls_mode is custom) and the "no" gets persisted per cert.

8. Too many knobs. CHMOD_CERT/CHMOD_KEY, CERT_NAME/KEY_NAME and
   CONFIG_FILE are firmware constants on a single-root appliance;
   hardcode them. If the firmware moves them that is a hook update, not
   a user setting -- and _savedeployconf pins today's defaults into
   every domain.conf, so existing users would never pick up a corrected
   default anyway. USER/HOST/PORT/SSH_CMD/RESTART_CMD (+
   REQUIRE_CUSTOM_MODE) is enough.

9. L253-311, optional: a self-detaching default such as
     nohup sh -c 'sleep 2; reboot' >/dev/null 2>&1 &
   lets the session exit 0 before the reboot lands, so the exit code can
   be trusted and the marker / "0|255" logic goes away. As written, a
   connection drop after the marker but before reboot runs is reported
   as success and never retried.

10. L214-224: staged files are not removed when the remote script
    aborts, and the per-run suffix means every failed attempt leaves
    another key file on the device. Fixed .tmp names plus a trap.

11. Header: trim to what a user needs (purpose, env vars, the two
    device caveats); the 0|255 rationale appears three times. Also
    "typically overnight" is wrong -- the default cron runs four times
    a day, the reboot lands whenever a renewal crosses the threshold.

12. Add a JetKVM section to the deployhooks wiki page (wiki Edit
    button) and reference it from the header.

daemonhorn:
Thanks for the thorough review — all 12 points addressed in the latest commit (`e300dba8`):

1. Dropped `[[:space:]]`/`-E` from the `tls_mode` grep; a plain space with `*` covers both the compact and indented JSON forms.
2. Swapped the `date +%s || echo 0` dead-fallback for the core `_time()` helper.
3. Dropped the local temp files for both SSH calls entirely — the upload script is now built in a variable and piped straight into `ssh ... sh`, same pattern as `deploy/windows_rdp.sh`. The restart command is now passed as a direct ssh argument (see #9 below), so no file there either.
4. Added a guard for an empty key/fullchain (`--signcsr`-only run) before doing anything else.
5. Every generated script line now uses `printf '%s\n'` instead of `echo`.
6. `DEPLOY_JETKVM_SSH_CMD` and `DEPLOY_JETKVM_RESTART_CMD` are now saved with `_savedeployconf`'s `"base64"` flag, matching `deploy/docker.sh`'s reload-command handling.
7. Split "config file missing/unreadable" from "HTTPS Mode not custom" into distinct exit codes and error messages, and dropped the `REQUIRE_CUSTOM_MODE=no` suggestion from the mode-not-custom error text.
8. Hardcoded remote path, filenames, chmod values, and the config file path as constants. Only `USER`/`HOST`/`PORT`/`SSH_CMD`/`RESTART_CMD`/`REQUIRE_CUSTOM_MODE` remain as env vars.
9. Restart is now launched detached (`nohup sh -c 'sleep 2; $CMD' >/dev/null 2>&1 &`) so the ssh call returns as soon as it's launched, before the reboot lands — this also removes the marker/`0|255` case logic entirely, and with it the bug where a connection dropping after the marker printed but before the restart command actually ran was reported as success. Tradeoff called out in the PR description: a restart command failing *after* launch can no longer be detected, only a failure to launch it at all.
10. Staged files now use fixed names plus a `trap ... EXIT` in the generated script, so any abort (mode check, a write failure under `set -e`) cleans up instead of leaving another stray file.
11. Trimmed the header (the marker/`0|255` rationale is gone along with #9's old logic) and corrected the "typically overnight" claim.
12. Added a JetKVM section to the [deployhooks wiki](https://github.com/acmesh-official/acme.sh/wiki/deployhooks) and linked it from the header.

Verification: `shellcheck` is now clean with no exclusions needed (previously excluded SC2181/SC2089), `shfmt -i 2` clean, and I wrote an 11-case local smoke test (stubbed acme.sh core, a fake `ssh` redirecting the hardcoded device paths into a scratch directory) covering the happy path, `RESTART_CMD=none`, HTTPS Mode not custom, the config file missing entirely, `REQUIRE_CUSTOM_MODE=no`, the `--signcsr` guard, and a failed restart-launch — all passing. I haven't re-run this against real JetKVM hardware since this remediation (the earlier hardware validation predates it); PR description updated to reflect that honestly rather than imply otherwise.

One thing I didn't do: verify `_time()` is actually the convention other *deploy hooks* use specifically, versus just core acme.sh — checked `grep -rn '_time' deploy/ dnsapi/ notify/` and it's used throughout `dnsapi/` and in `deploy/unifios.sh` for exactly this kind of run-id generation, so it's the right call.

daemonhorn:
Update: re-verified the review remediation (`e300dba8`) against real JetKVM hardware, through the same OPNsense ACME Client pipeline as before.

Confirmed:
- The happy path -- upload, the new detached restart (`nohup sh -c 'sleep 2; reboot' &`), and the device's HTTPS listener serving the newly issued certificate after reboot.
- HTTPS Mode not set to `"custom"` still fails cleanly with the expected, specific error message.

Not re-verified on real hardware: the config-file-missing case (exit code 4) -- didn't want to rename a live config file on the device I'm testing against. Still only covered by the local smoke test. PR description updated to reflect this precisely.

neilpang:
1. L229-232: "only a failure to launch it at all is caught below" is not
   accurate. `nohup ... >/dev/null 2>&1 &` is an async list, so the
   remote sh returns 0 regardless -- measured 0 both for a nonexistent
   restart command and for nohup itself missing. Only a transport error
   fails the ssh call. Reword so nobody later relies on that check; the
   tradeoff itself is fine.

2. L235: the command is interpolated inside `sh -c '...'`, so a single
   quote in DEPLOY_JETKVM_RESTART_CMD breaks the quoting -- and point
   6's base64 saving now lets such values reach the device. With
   RESTART_CMD="sh -c 'sync; reboot'":

     nohup sh -c 'sleep 2; sh -c 'sync; reboot'' >/dev/null 2>&1 &

   splits into `nohup sh -c "sleep 2; sh -c sync"` plus a separate,
   un-nohup'd `reboot >/dev/null 2>&1 &` -- the user's `sync` never
   runs. Either document that the command must not contain a single
   quote, or avoid the nesting.

Otherwise verified: exit 3/4 survive the EXIT trap with distinct
messages, the base64 round-trip holds, no staged files left on any
failure path, wiki section is up.

daemonhorn:
Both fixed in `146a83b0`:

1. Confirmed by testing directly (`sh -n`/`sh -c`/`dash`) — `DEPLOY_JETKVM_RESTART_CMD` is now escaped (the standard `'\''` substitution) before being nested inside the detached command's `sh -c '...'` wrapper. Added a smoke-test case that runs the actual generated detached command with a restart command containing a single quote and confirms both halves execute — it caught a timing bug in my first attempt at that test (the "sleep 2" is inside the detached job, so checking immediately after the ssh call returns doesn't work), good reminder to actually exercise generated output rather than just eyeball the escaping logic.

2. You're right, and I appreciate you actually measuring it rather than taking my comment at face value. Reworded both the code comment and the PR description to say what's actually caught (an outright SSH connection failure) instead of implying detection this design doesn't provide.

Issue #7261, opened 2026-09-16
Report bugs to bergdns DNS API


kekropp:
Tracking issue for the `dns_bergdns` DNS API (added in #PR), as asked for by
step 9 of the [DNS API Dev Guide](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Dev-Guide).

Please report bugs and questions about the bergdns DNS API here.

Usage:

```sh
export BERGDNS_API_KEY="bgd_..."     # created in the account UI, a "Challenges only" key is enough
acme.sh --issue --dns dns_bergdns -d example.com -d '*.example.com'
```

Optional settings:

- `BERGDNS_API_URL` overrides the API base url, default `https://bergdns.at/v1`.
- `BERGDNS_TTL` is the TTL of the challenge record, default `60`.
- `BERGDNS_PROPAGATION_TIMEOUT` is how many seconds to wait for every bergdns nameserver to serve the record, default `120`.

Docs: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_bergdns

When reporting a problem, please include the output of `acme.sh --issue --dns dns_bergdns -d example.com --debug 2`
with the API key removed.

PR #7263, opened 2026-09-16
 Add bergdns.at DNS API

kekropp:
Adds `dns_bergdns.sh`, solving dns-01 through the bergdns.at API. Auth is an API
key from the account UI, which is can be restricted to `_acme-challenge.*` TXT
records in chosen zones.

The zone is found by walking the name up against the account's zone list, so
subzones and DNS alias mode both work. `rm` deletes only the value it was given,
so the two TXT records of an apex plus its wildcard survive each other (#1261),
and removing the last value drops the RRset.

After writing, the plugin polls the zone's propagation endpoint until
bergdns's own authoritative servers answer with the record, up to
`BERGDNS_PROPAGATION_TIMEOUT` seconds. A timeout warns and continues rather than
failing, since this does not imply the record is not written, but rather the
propagation checking has failed (which is non-critical).

Usage:

```sh
export BERGDNS_API_KEY="bgd_..."
acme.sh --issue --dns dns_bergdns -d example.com -d '*.example.com'
```

`BERGDNS_API_URL`, `BERGDNS_TTL` and `BERGDNS_PROPAGATION_TIMEOUT` are optional
and documented in the wiki.

- Support issue: #7261
- Wiki: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_bergdns
- DNS API test on the fork: https://github.com/kekropp/acme.sh/actions/runs/35064016967


kekropp:
Read the guide. The PR targets `dev`, the usage is on the
[dnsapi2 wiki page](https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_bergdns),
and the support issue is #7261.

The DNS API test passes on my fork across all platforms:
https://github.com/kekropp/acme.sh/actions/runs/35064016967

neilpang:
1) dnsapi/dns_bergdns.sh:140 and :173 use grep -F. Solaris /usr/bin/grep
   has no -F at all -- it errors with "grep: illegal option -- F". The
   green Solaris job in your fork run does not clear this: DNS.yml line
   599 does export PATH=/usr/gnu/bin:$PATH, so that job ran GNU grep.
   Use a shell case, which is literal by construction and needs no
   external command:

     while IFS= read -r _bergdns_line || [ -n "$_bergdns_line" ]; do
       case "$_bergdns_line" in
       *'"name":"'"$_bergdns_cand"'",'*) ... ;;
       esac
     done

   Do not reach for _contains / _startswith instead: they take the
   needle as a regex.

2) _bergdns_rest (:240-250) decides success or failure by looking for
   "status": and "code":" in the body rather than at the HTTP status.
   Any error response that is not problem+json is then read as success.
   bergdns.at sits behind a reverse proxy (via: 1.1 Caddy), so a 502 or
   504 HTML page is a realistic case: _bergdns_rest returns 0, the id
   match yields an empty _bergdns_rrset_id, _bergdns_wait returns at
   once because that id is empty, and dns_bergdns_add returns 0 with no
   TXT record written. Issuance then fails later at the CA with an
   unrelated authorization error.

   The status code is already there: _inithttp runs curl with
   --dump-header $HTTP_HEADER and the wget branch redirects headers to
   the same file; around twenty dnsapi hooks read it. Branch on that,
   and keep the problem+json parsing only for the message and the code.

3) :140 and :173 require the fields to be adjacent and comma-terminated
   ("name":"X",  and  "name":"X","type":"TXT",). A server-side key
   reorder silently breaks the lookup and surfaces as "no zone in this
   account holds ...". Match name and type independently.

4) :114-117 persist the defaults into account.conf. After the first run
   BERGDNS_API_URL, BERGDNS_TTL and BERGDNS_PROPAGATION_TIMEOUT are
   pinned to today's values, and a later change to the shipped defaults
   -- an API base move in particular -- never reaches those installs.
   Save only what the user actually set.

5) :35 interpolates the TTL unquoted: "ttl":$BERGDNS_TTL. A non-numeric
   value from the environment or from a stale account.conf produces
   malformed JSON and an opaque 400. Validate it in _bergdns_init.

Minor:

  - _fqdn, _line, _candidate, _zone_lines, _waited, _method, _endpoint,
    _body, _url and _ret are unprefixed. The hook runs in a subshell so
    nothing leaks back to the caller, but _url, _body and _ret are core
    names; prefix them _bergdns_ as the functions are.

  - :246 the greedy .* makes sed take the last "detail" in the body,
    not the first.

  - :82 rm tolerates only rrset_not_found, while the header comment
    says removing a value that does not exist is not an error. If the
    API has a separate record_not_found, cleanup fails.

  - _bergdns_wait stacks on top of _check_dns_entries, which acme.sh
    runs after the add loop. Polling the authoritative servers is not
    redundant, but is 120s the right default on top of that?

kekropp:
Hi @neilpang!

Thank you very much for taking the time reviewing this.
I am sorry, i am not very good at writing multi-platform bash scripts.

I hope I have resolved all of the issues you found.
Thank you for your amazing work keeping this running.

Best Regards

neilpang:
1) _bergdns_init:51-53 -- when a value equals the default, nothing is
   saved, but nothing is cleared either. An install that once saved a
   non-default BERGDNS_TTL keeps it: exporting 60 again skips the save,
   the old value stays in account.conf, and the next unattended renewal
   reads it back. So the environment cannot put a setting back to its
   default. Call _clearaccountconf_mutable in the equal-to-default
   branch.

2) The default BERGDNS_PROPAGATION_TIMEOUT is 60 now, but the wiki
   entry and the body of #7261 still say 120.

Everything else checks out, including on wget-only hosts: _get and
_post both normalise the header file, so the "^HTTP" match holds there
too. With those two it is good to merge.

kekropp:
Hi again!

Thank you for your time again.
These issues should now be resolved.

Best Regards

PR #7267, opened 2026-09-19
Add deploy hook for nginx proxy manager and NPM plus

invario:
Add a new deploy hook to deploy to nginx proxy manager or nginx proxy manager plus. Configuration via the following, all of which are required.
```
# DEPLOY_NPM_HOST       - host/IP address of NPM admin panel (example.com, 192.168.1.1, etc.)
# DEPLOY_NPM_PORT       - port number of NPM admin panel (80, 443, 8443, etc.)
# DEPLOY_NPM_PROTOCOL   - protocol to connect to NPM admin panel (http/https)
#                         (https is STRONGLY recommended)
# DEPLOY_NPM_CERTNUM    - custom certificate number to overwrite (1, 2, 3, etc.)
#                         (look it up in NPM admin panel, must already exist)
# DEPLOY_NPM_USER       - authorized user for NPM (adminuser, admin etc.)
# DEPLOY_NPM_PASSWORD   - authorized user's password for NPM (MySecurePassword, etc.)
```

1. Connect to the NPM REST API using the configured credentials
2. Request an access token
3. Uses the token to upload the fullchain and key to replace an existing custom certificate

Full disclosure: I've only been able to test it successfully on [npmplus](https://github.com/ZoeyVid/NPMplus) but I believe it should work on the original NPM as well since NPMPlus is a fork of the original NPM.

Wiki entry added:
https://github.com/acmesh-official/acme.sh/wiki/deployhooks#50-deploy-to-nginx-proxy-manager--npmplus


neilpang:
1. Raw `curl` is not allowed in hooks -- use `_post`. Raw calls bypass the proxy / CA bundle / timeout settings and break on wget-only systems. Multipart bodies are built by hand and passed as the content type (`deploy/kong.sh`, `deploy/synology_dsm.sh`, `deploy/ruckus.sh`); Set-Cookie is read from `$HTTP_HEADER` (`deploy/unifios.sh`, `deploy/peplink.sh`, `deploy/ikuai.sh`).

2. `curl -k` disables certificate verification unconditionally, so the admin credentials are posted to an unverified endpoint. Core skips verification only when the user passes `--insecure`. Moving to `_post` fixes this.

3. It cannot work on upstream NPM. In jc21/nginx-proxy-manager (v2.12.6 and main), `backend/lib/express/jwt-decode.js` reads only `Authorization: Bearer` and `POST /api/tokens` returns the token in the JSON body -- there is no `__Host-Http-token` cookie. That cookie is NPMplus-only (`ZoeyVid/NPMplus` reads `req.signedCookies["__Host-Http-token"]`). Either handle both auth styles, or drop NPM from the title, the header comment and the wiki entry.

4. `[ "$_npm_error" != "{}" ]` is NPMplus-specific and leaks the key elsewhere: NPMplus `upload()` returns `_.omit(meta, allowedSslFiles)` = `{}`, upstream returns `_.pick(meta, allowedSslFiles)` = the certificate and the private key, which this then prints via `_err`. Check the HTTP status from `$HTTP_HEADER` instead.

5. `_debug _npm_token_raw "$_npm_token_raw"` prints the Set-Cookie with the full session token at `--debug 1` -- the level users paste into issues. Use `_secure_debug`.

6. The password is interpolated into the JSON body unescaped; a `"` or `\` in it produces an invalid request. Pipe it through `_json_encode`.

7. `curl -f` discards the error body and its exit status is never checked, so every failure reports "check your credentials" -- including a 2FA account, where NPMplus sets `__Host-Http-challenge_token` instead.

8. `--deploy-hook npm` reads as Node's npm; `npmplus` or `nginx_proxy_manager` is clearer.

invario:
Thanks @neilpang, I will make the necessary changes/fixes and update the PR.

invario:
@neilpang Ready for review whenever you are available.  I have a question though: 
`_post` outputs the body/payload of a post request when `--debug` is specified. When this contains the private certificate key, is there a way to prevent this from happening?

